PR #35, opened 2019-09-03
Protocol registry and CLI tool `aea`

### LINKED ISSUE #24: Add the main entry point for the `aea` package.
In order to manage the installation of extras for the `aea` package, it would be nice to have a main entry point script for the `aea` package.

Then, eventually, after `pip install aea`, one might be able to set up an agent by just a bunch of  commands from the terminal:
```
$ aea create my_agent
$ aea add <skill-name>
$ aea run
```

Using the `setuptools` feature `entry_points` to install the `aea` command-line tool in the system path.

None.

**Additional context**
- [Automatic Script Creation](https://setuptools.readthedocs.io/en/latest/setuptools.html#automatic-script-creation)
- [Python Packaging: Command Line Scripts](https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html)

marcofavorito:
## Proposed changes

- Introduce the command-line tool `aea`, installable with `pip install -e .[cli]`
- Enable dynamic loading of protocols and protocols handlers.

## Fixes

Fixes #24 

## Types of changes

## Checklist

## Further comments

None.


aea/channel/local.py
```diff
@@ -128,17 +124,6 @@ def handle_agent_message(self, envelope: Envelope) -> None:
         else:
             self._send(envelope)
 
-    def register_agent(self, public_key: str, agent_description: Description) -> None:
```
DavidMinarsch: Why are the register and unregister method being removed?

aea/protocols/base/abstract_handler.py
```diff
@@ -0,0 +1,54 @@
...
+    """This is an interface for a message handler."""
+
+    # TODO: should we have the agent be accessible by the handlers?
+    # def __init__(self, agent):
+    #     """
+    #     Initialize the handler.
+    #
+    #     :param agent: the agent.
+    #     """
+    #     self.agent = agent
```
DavidMinarsch: It would probably be saver not to. If you recall the principle of object capability then  we don't want the handler to control the entire agent. At most we want it to be able to change (parts of) the agent's state.

aea/protocols/base/protocol.py
```diff
@@ -0,0 +1,70 @@
...
+    def serializer(self) -> Serializer:
+        """Get the serializer."""
+        return self._serializer
+
+    @property
+    def handler(self) -> AbstractHandler:
+        """Get the handler."""
+        return self._handler
+
+    # @abstractmethod
```
DavidMinarsch: I think should be uncommented.

aea/protocols/oef/message.py
```diff
@@ -53,12 +51,10 @@ class OEFErrorOperation(Enum):
 
         REGISTER_SERVICE = 0
         UNREGISTER_SERVICE = 1
-        REGISTER_DESCRIPTION = 2
```
DavidMinarsch: Are we removing REGISTER_DESCRIPTION and UNREGISTER_DESCRIPTION because they are duplicating the SERVICE_ equivalents?

aea/aea.py
```diff
@@ -0,0 +1,290 @@
...
+        """
+
+        protocols_spec = importlib.util.find_spec(".".join([directory, "protocols"]))
+        protocols_packages = list(filter(lambda x: not x.startswith("__"), protocols_spec.loader.contents()))
+        logger.debug("Processing the following protocol packages: {}".format(protocols_packages))
+        for protocol_name in protocols_packages:
+            try:
+                self._add_protocol(directory, protocol_name)
+            except Exception:
+                logger.exception("Not able to add protocol {}.".format(protocol_name))
```
DavidMinarsch: After adding we should probably quickly test whether each protocol implements all required methods?

aea/aea.py
```diff
@@ -0,0 +1,290 @@
...
+        return self._protocols.get(envelope.protocol_id, None)
+
+    def fetch_handler(self, envelope: Envelope) -> Optional[AbstractHandler]:
+        """
+        Fetch the handler for the envelope.
+
+        :pass envelope: the envelope
+        :return: the protocol id or None if the protocol is not registered
+        """
+        return self._protocols.get(envelope.protocol_id, None)
```
DavidMinarsch: Not sure this is correct, it returns the same value as the previous function as it has the same return line...

aea/aea.py
```diff
@@ -0,0 +1,290 @@
...
+
+        logger.debug("Found serializer class {serializer_class} for protocol {protocol_name}"
+                     .format(serializer_class=serializer_class, protocol_name=protocol_name))
+        serializer = serializer_class()
+
+        # get the handler
+        handler_module = importlib.import_module(".".join([directory, "protocols", protocol_name, "handlers"]))
+        classes = inspect.getmembers(handler_module, inspect.isclass)
+        handler_classes = list(filter(lambda x: re.match("Handler", x[0]), classes))
+        handler_class = handler_classes[0][1]
```
DavidMinarsch: I think the handler should be separate from the protocol. Consider this: I add tac protocol but rather than handling it with the default tac handler I do it with some other handler I have written myself. Note, that I consider 'handlers' to do deal with 'reactions'.

aea/channel/local.py
marcofavorito: `register_agent` and `unregister_agent` methods are not supported anymore:
https://github.com/fetchai/oef-sdk/blob/feat/python-sdk2/oef/src/python/agents.py#L521-L546

aea/protocols/base/abstract_handler.py
marcofavorito: agreed.

aea/protocols/oef/message.py
marcofavorito: No, because they are related to the `register_agent` and `unregister_agent`, which have been removed.

aea/aea.py
marcofavorito: yeah, that's wrong. That should read the `.handler` property of the protocol object, if any.

aea/aea.py
marcofavorito: agreed.

aea/channel/local.py
DavidMinarsch: Ok, thanks, I saw this thread on Slack now.

aea/protocols/oef/message.py
DavidMinarsch: yes

PR #51, opened 2019-09-07
Add mypy for static type checking, fix mypy issues, fixes #25

### LINKED ISSUE #25: Add static type checker to tests
We would like to have a high consistency in our code.

Add static type checker mypy: https://mypy.readthedocs.io/en/latest/

None.

**Additional context**
None.

DavidMinarsch:
## Proposed changes

This PR adds mypy to the repo for static type checking. It amends the code base to resolve issues raised by mypy. It adds mypy to the testing pipeline (Jenkins).

## Fixes

It fixes #25 .

## Types of changes

## Checklist

## Further comments

Mypy can be used in a stricter fashion than currently configured. More details here: https://mypy.readthedocs.io/en/latest/common_issues.html#unexpected-errors-about-none-and-or-optional-types

### DIFF
--- a/aea/aea.py
+++ b/aea/aea.py
@@ -20,11 +20,12 @@
 """This module contains the implementation of an Autonomous Economic Agent."""
 import logging
 from pathlib import Path
-from typing import Optional
+from typing import Optional, cast
 
 from aea.agent import Agent
 from aea.mail.base import Envelope
 from aea.skills.base import Resources, Context
+from aea.skills.default.handler import DefaultHandler
 
 logger = logging.getLogger(__name__)
 
@@ -105,28 +106,29 @@ class AEA(Agent):
 
         # fetch the handler of the "default" protocol for error handling. TODO: change with the handler of "error" protocol.
         default_handler = self.resources.handler_registry.fetch("default")
+        default_handler = cast(DefaultHandler, default_handler)
 
         if protocol is None:
             if default_handler is not None:
-                default_handler.send_unsupported_protocol(envelope)  # type: ignore
+                default_handler.send_unsupported_protocol(envelope)
             return
 
         try:
             msg = protocol.serializer.decode(envelope.message)
         except Exception:
             if default_handler is not None:
-                default_handler.send_decoding_error(envelope)  # type: ignore
+                default_handler.send_decoding_error(envelope)
             return
 
         if not protocol.check(msg):
             if default_handler is not None:
-                default_handler.send_invalid_message(envelope)  # type: ignore
+                default_handler.send_invalid_message(envelope)
             return
 
         handler = self.resources.handler_registry.fetch(protocol.id)
         if handler is None:
             if default_handler is not None:
-                default_handler.send_unsupported_skill(envelope, protocol)  # type: ignore
+                default_handler.send_unsupported_skill(envelope, protocol)
             return
 
         handler.handle_envelope(envelope)
--- a/aea/aea.py
+++ b/aea/aea.py
@@ -101,32 +101,32 @@ class AEA(Agent):
         :param envelope: the envelope to handle.
         :return: None
         """
-        protocol = self.resources.protocol_registry.fetch_protocol(envelope.protocol_id)
+        protocol = self.resources.protocol_registry.fetch(envelope.protocol_id)
 
         # fetch the handler of the "default" protocol for error handling. TODO: change with the handler of "error" protocol.
         default_handler = self.resources.handler_registry.fetch("default")
 
         if protocol is None:
             if default_handler is not None:
-                default_handler.send_unsupported_protocol(envelope)
+                default_handler.send_unsupported_protocol(envelope)  # type: ignore
             return
 
         try:
             msg = protocol.serializer.decode(envelope.message)
         except Exception:
             if default_handler is not None:
-                default_handler.send_decoding_error(envelope)
+                default_handler.send_decoding_error(envelope)  # type: ignore
             return
 
         if not protocol.check(msg):
             if default_handler is not None:
-                default_handler.send_invalid_message(envelope)
+                default_handler.send_invalid_message(envelope)  # type: ignore
             return
 
-        handler = self.resources.handler_registry.fetch_handler(protocol.id)
+        handler = self.resources.handler_registry.fetch(protocol.id)
         if handler is None:
             if default_handler is not None:
-                default_handler.send_unsupported_skill(envelope, protocol)
+                default_handler.send_unsupported_skill(envelope, protocol)  # type: ignore
             return
 
         handler.handle_envelope(envelope)
--- a/aea/channel/gym.py
+++ b/aea/channel/gym.py
@@ -60,7 +60,7 @@ class GymChannel:
             return None
 
         assert len(self._queues.keys()) <= 1, "Only one public key can register to a gym."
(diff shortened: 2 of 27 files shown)


examples/gym/agent.py
```diff
@@ -169,7 +170,8 @@ def react(self) -> None:
             envelope = self.inbox.get_nowait()  # type: Optional[Envelope]
             if envelope is not None:
                 if envelope.protocol_id == 'gym':
-                    gym_msg = GymSerializer().decode(envelope.message)
+                    gym_msg = GymSerializer().decode(envelope.message)  # type: Message
+                    gym_msg = cast(GymMessage, gym_msg)
```
DavidMinarsch: Note, @marcofavorito I expect this to be an issue across all skills so we will likely want to move the type casting down to the message object before the return statement.

aea/protocols/tac/serialization.py
```diff
@@ -68,28 +70,28 @@ def encode(self, msg: Message) -> bytes:
 
         if tac_type == TACMessage.Type.REGISTER:
             agent_name = msg.get("agent_name")
-            tac_msg = tac_pb2.TACAgent.Register()
+            tac_msg = tac_pb2.TACAgent.Register()  # type: ignore
```
DavidMinarsch: we ignore these as stubbing them makes little sense

aea/mail/base.py
```diff
@@ -242,6 +242,8 @@ def put_message(self, to: Optional[Address] = None, sender: Optional[Address] =
 class Connection:
     """Abstract definition of a connection."""
 
+    bridge: Optional[Any]
```
DavidMinarsch: This is an annoying requirement of mypy. It always checks against the base class and throws an error if an instance var is not referenced there.

aea/crypto/base.py
```diff
@@ -117,7 +117,7 @@ def _compute_pbk(self) -> object:
 
         :return: public key object
         """
-        public_key = self._private_key.public_key()
+        public_key = self._private_key.public_key()  # type: ignore
```
DavidMinarsch: Rather than stubbing we just ignore external libraries for now.

aea/aea.py
```diff
@@ -101,32 +101,32 @@ def handle(self, envelope: Envelope) -> None:
...
-        protocol = self.resources.protocol_registry.fetch_protocol(envelope.protocol_id)
+        protocol = self.resources.protocol_registry.fetch(envelope.protocol_id)
 
         # fetch the handler of the "default" protocol for error handling. TODO: change with the handler of "error" protocol.
         default_handler = self.resources.handler_registry.fetch("default")
 
         if protocol is None:
             if default_handler is not None:
-                default_handler.send_unsupported_protocol(envelope)
+                default_handler.send_unsupported_protocol(envelope)  # type: ignore
```
DavidMinarsch: We have to ignore here as the base class 'Handler' does not know about these attributes.

marcofavorito: I see. Shall we apply `cast`?

something like:
```
from aea.skills.default.handler import DefaultHandler
default_handler = cast(DefaultHandler, default_handler) 
```

marcofavorito: That would let us to remove the `# type: ignore` statements.

aea/channel/gym.py
```diff
@@ -60,7 +60,7 @@ def connect(self, public_key: str) -> Optional[Queue]:
             return None
 
         assert len(self._queues.keys()) <= 1, "Only one public key can register to a gym."
-        q = Queue()
+        q = Queue()  # type: Optional[Queue]
```
marcofavorito: Why `Optional[Queue]` and not simply `Queue`?

DavidMinarsch: Oops. My bad, will fix.

aea/aea.py
DavidMinarsch: Yeah good point. Will do.

aea/mail/base.py
marcofavorito: Removing that line, I get only the following error:
```
aea/channel/oef.py:526: error: "Connection" has no attribute "bridge"
```

That corresponds to the `return` statement of the following snippet:
```
    @property
    def mail_stats(self) -> MailStats:
        """Get the mail stats object."""
        return self._connection.bridge.mail_stats
```

Putting the usual `# type: ignore` will make the linter to skip the problem.

I'm wondering if this could be a better solution. The advantage would be that we remove the concept of `bridge` in the abstract `Connection` class. In fact, the `bridge` attribute in the `OEFConnection` is an "implementation detail", and doesn't correspond to an abstract concept. However, we could surely design and develop an abstract `class Bridge` in order to solve a more general problem in the connection implementation.

aea/protocols/fipa/serialization.py
```diff
@@ -40,31 +40,31 @@ def encode(self, msg: Message) -> bytes:
 
         performative_id = msg.get("performative").value
         if performative_id == "cfp":
-            performative = fipa_pb2.FIPAMessage.CFP()
+            performative = fipa_pb2.FIPAMessage.CFP()  # type: ignore
```
marcofavorito: This is fine for the moment. However, I'd like to share this project: [MyPy Protobuf](https://github.com/dropbox/mypy-protobuf)

 `mypy-protobuf` is an open-source tool to generate MyPy stubs from Protobuf files. I haven't investigated in deep, but I believe that might help for all the Protobuf-based protocol serializations we have developed, and future ones.

DavidMinarsch: Good point!

aea/mail/base.py
DavidMinarsch: Having looked at the code once more we should probably consider renaming `bridge` to `channel`. I then fixed all the channel logic across the 3 examples. It now is consistent. Let me know what you think.

aea/channel/oef.py
```diff
@@ -523,4 +523,4 @@ def __init__(self, public_key: str, oef_addr: str, oef_port: int = 10000):
     @property
     def mail_stats(self) -> MailStats:
         """Get the mail stats object."""
-        return self._connection.bridge.mail_stats
+        return self._connection.channel.mail_stats  # type: ignore
```
DavidMinarsch: `mail_stats` is implementation specific to OEFChannel

aea/channel/local.py
```diff
@@ -228,6 +228,45 @@ def disconnect(self, public_key: str) -> None:
...
+        return self.local_node.disconnect(self.public_key)
+
+    def send(self, envelope: Envelope) -> None:
+        """
+        Send an envelope.
+
+        :param envelope: the envelope to send.
+        :return: None.
+        """
+        return self.local_node.send(envelope)
```
DavidMinarsch: this just wraps the local node methods to ensure consistent interface across channels

aea/channel/gym.py
```diff
@@ -39,32 +39,32 @@
...
 
-        :param public_key: the public key of the agent.
         :return: an asynchronous queue, that constitutes the communication channel.
         """
-        if public_key in self._queues:
+        if self.public_key in self._queues:
             return None
 
-        assert len(self._queues.keys()) <= 1, "Only one public key can register to a gym."
+        assert len(self._queues.keys()) == 0, "Only one public key can register to a gym."
```
DavidMinarsch: could remove some of this logic but no harm in having it to mirror the others

PR #61, opened 2019-09-11
Fix/tests

marcofavorito:
## Proposed changes

- fix how the OEF channel sends messages of the FIPA protocol. In particular, how the query of a CFP is passed to the `send_cfp` method (handling the case when it's None) and how the proposal is encoded (use only the `bytes` propose type).
- fix the decoding of the Default protocol: not use only the body keyword in the constructor, but also the 'type' keyword argument. Otherwise, in the parent `Message` constructor, the values in the keyword of the child constructor overwrite the values in the body (in particular, `type=None` used to overwrite `body["type"])`, hence losing information about the default message type.
- fix default message test, to include the (now mandatory) error_data field in the Error message type.
- add the `mypy` tox environment, and run linters by default.

## Fixes

None.

## Types of changes

## Checklist

## Further comments

None.

### DIFF
--- a/aea/protocols/default/serialization.py
+++ b/aea/protocols/default/serialization.py
@@ -68,4 +68,4 @@ class DefaultSerializer(Serializer):
         else:
             raise ValueError("Type not recognized.")
 
-        return DefaultMessage(body=body)
+        return DefaultMessage(type=msg_type, body=body)
--- a/tox.ini
+++ b/tox.ini
@@ -24,4 +24,4 @@ commands = flake8 aea examples scripts tests --exclude=.md,aea/*_pb2.py,aea/__in
 [testenv:mypy]
 basepython = python3.7
 deps = mypy
-commands = mypy aea examples tests
+commands = mypy aea examples tests scripts
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py37, py36
+envlist = flake8, mypy, py37, py36
 skipsdist = True
 ignore_basepython_conflict = True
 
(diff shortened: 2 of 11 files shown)


aea/protocols/default/serialization.py
```diff
@@ -68,4 +68,4 @@ def decode(self, obj: bytes) -> Message:
         else:
             raise ValueError("Type not recognized.")
 
-        return DefaultMessage(body=body)
+        return DefaultMessage(type=msg_type, body=body)
```
DavidMinarsch: do we need to change this on the other protocols too?

tox.ini
```diff
@@ -19,4 +19,9 @@ basepython = python3.7
...
        flake8-docstrings
        pydocstyle==3.0.0
-commands = flake8 aea examples scripts tests --exclude=.md,aea/*_pb2.py,aea/__init__.py,aea/cli/__init__.py,tests/common//oef_search_pluto_scripts --ignore=E501,E701
\ No newline at end of file
+commands = flake8 aea examples scripts tests --exclude=.md,aea/*_pb2.py,aea/__init__.py,aea/cli/__init__.py,tests/common//oef_search_pluto_scripts --ignore=E501,E701
+
+[testenv:mypy]
+basepython = python3.7
+deps = mypy
+commands = mypy aea examples
```
DavidMinarsch: Should be: mypy aea examples scripts tests

aea/protocols/default/serialization.py
marcofavorito: yes! good point.

tox.ini
marcofavorito: thanks, will fix.

PR #83, opened 2019-09-18
Add and remove a connection from command line.

### LINKED ISSUE #71: Simplify adding connections via cli
Currently, adding a connection is a manual process.

- Allow adding connection via cli for local, oef and gym.
- Validate agent config file for correct connection specs

-

**Additional context**
-

marcofavorito:
## Proposed changes

Let the CLI tool to add/remove connections.

to add:
```
aea add connection <connection_name> <connection_type> --config key1 value1 --config key2 value2 
```

## Fixes

Fixes #71 

## Types of changes

## Checklist

## Further comments

None.


aea/cli/add.py
```diff
@@ -41,6 +41,20 @@ def add(ctx: Context):
     _try_to_load_agent_config(ctx)
 
 
+@add.command()
+@click.argument('connection_name', type=str, required=True)
```
DavidMinarsch: Why do we distinguish here between type and name? Aren't they the same?

aea/cli/add.py
```diff
@@ -41,6 +41,20 @@ def add(ctx: Context):
...
+@click.argument('connection_type', type=click.Choice(['oef', 'local', 'gym']), required=True)
+@click.option('--config', nargs=2, multiple=True, metavar='KEY VALUE', help='Overrides a config key/value pair.')
```
DavidMinarsch: I think I said before. Not a fan of this at all. We should have it all via YAML configs in my opinion to keep the CLI neat and clean. I don't see the use case for this once you have the yaml config.

examples/gym_skill/README.md
```diff
@@ -22,12 +22,8 @@ A guide to create an AEA with the gym_skill.
...
 
 - Add a gym connection to the `aea-config.yaml`:
-
-      - connection:
-        config:
-          env: gyms.env.BanditNArmedRandom
-        name: gym
-        type: gym
+        
+      aea add connection gym gym --config env gyms.env.BanditNArmedRandom
```
DavidMinarsch: I think it should be `aea add connection gym` and then the rest is specified in config yaml. That way it mirrors the approach to skills and protocols.

aea/cli/add.py
DavidMinarsch: I think it is because at the moment you do this differently than for the skills approach (i.e. the connections don't get copied in). I think they should though. Happy to discuss.

PR #104, opened 2019-09-23
Test/mail

Totoual:
Created a test to test the aea/mail/base.py script


tests/test_mail.py
```diff
@@ -0,0 +1,67 @@
...
+	to_adr = []
+	msg = Message(content="hello")
+	message_bytes = ProtobufSerializer().encode(msg)
+	envelope = Envelope(to=to_adr, sender="Agent0", protocol_id="my_own_protocol", message=message_bytes)
+	assert (envelope, "Passing an empty receiver. Testing try/except by passing a list as sender")
+
+def test_inbox_empty():
+	my_queue = Queue() 
+	_inbox = InBox(my_queue)
+	assert(_inbox.empty(), "Must return an empty inbox")
```
DavidMinarsch: Comment should be relevant in the fail case, i.e. "Does not return an empty inbox."

tests/test_mail.py
```diff
@@ -0,0 +1,67 @@
...
+from queue import Queue
+import pytest
+
+
+def test_envelope_initialisaztion() : 
+	msg = Message(content='hello')
+	message_bytes = ProtobufSerializer().encode(msg)
+	envelope = Envelope(to="Agent1", sender="Agent0", protocol_id="my_own_protocol", message=message_bytes)
+	
+	envelope.to = "Agent1"
```
DavidMinarsch: If you want to test the setters than maybe do:
assert envelope.to = "Agent1", "Cannot set to value on Envelope."

Similarly, for the others.

tests/test_mail.py
```diff
@@ -0,0 +1,67 @@
...
+	envelope.message = b"hello"
+	
+	assert (envelope,"Cannot generate a new envelope")
+
+def test_envelope_empty_receiver() : 
+	to_adr = []
+	msg = Message(content="hello")
+	message_bytes = ProtobufSerializer().encode(msg)
+	envelope = Envelope(to=to_adr, sender="Agent0", protocol_id="my_own_protocol", message=message_bytes)
+	assert (envelope, "Passing an empty receiver. Testing try/except by passing a list as sender")
```
DavidMinarsch: Same, this won't fail..

tests/test_mail.py
```diff
@@ -0,0 +1,67 @@
...
+	my_queue = Queue()
+	my_queue.put(message_bytes)
+	_inbox = InBox(my_queue)
+	assert(_inbox.get_nowait(), "Check for a message on the in queue and wait for no time.")
+
+#Test thet the outbox queue is empty
+def test_outbox_empty() :
+	my_queue = Queue()
+	_outbox = OutBox(my_queue)
+	assert(_outbox.empty(), "Checking if the outbox is empty")
```
DavidMinarsch: Write message for fail case, i.e." "Outbox is not empty."

aea/mail/base.py
```diff
@@ -50,13 +50,12 @@ def __init__(self, to: Address,
...
-        assert type(self._to) == str or self._to is None
-        try:
+        try : 
+            assert type(self._to) == str or self._to is None
             if self._to is not None and type(self._to) == str:
                 self._to.encode('utf-8')
-        except Exception:
-            assert False
-
+        except AssertionError as error :
```
marcofavorito: the handling of the failing assertion is a good idea. However, we should decide if we tolerate this bad state or not. if we raise another exception or we just log a warning/error/exception message.

tests/test_mail.py
```diff
@@ -0,0 +1,67 @@
...
+	msg = Message(content='hello')
+	message_bytes = ProtobufSerializer().encode(msg)
+	envelope = Envelope(to="Agent1", sender="Agent0", protocol_id="my_own_protocol", message=message_bytes)
+	
+	envelope.to = "Agent1"
+	envelope.sender = "Agent0"
+	envelope.protocol_id = "my_own_protocol"
+	envelope.message = b"hello"
+	
+	assert (envelope,"Cannot generate a new envelope")
```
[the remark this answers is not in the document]
Totoual: Removed the parenthesis in order not to be always true.

tests/test_mail.py
```diff
@@ -0,0 +1,66 @@
...
+
+#Test thet the outbox queue is empty
+def test_outbox_empty() :
+	my_queue = Queue()
+	_outbox = OutBox(my_queue)
+	assert _outbox.empty(), "Checking if the outbox is empty"
+
+
+''' Testing the mailBox()'''
+#It creates a new thread for the connection ( So interupt after passing all tests)
```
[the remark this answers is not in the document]
marcofavorito: @Totoual the same fix for the other functions.

tests/test_mail.py
```diff
@@ -0,0 +1,90 @@
...
+    assert envelope.sender == "ChangedSender", "Cannot set sender value on Envelope"
+    assert envelope.protocol_id == "my_changed_protocol", "Cannot set protocol_id on Envelope "
+    assert envelope.message == b"HelloWorld", "Cannot set message on Envelope"
+
+    assert envelope, "Cannot generate a new envelope"
+
+
+def test_envelope_empty_receiver():
+    """Tests what happens when we are passing wrong type of receiver."""
+    to_adr = []
```
DavidMinarsch: This should be = None

tests/test_mail.py
```diff
@@ -0,0 +1,90 @@
...
+    msg = Message(content="hello")
+    message_bytes = ProtobufSerializer().encode(msg)
+    envelope = Envelope(to=to_adr, sender="Agent0",
```
DavidMinarsch: then here assert it throws an error

PR #124, opened 2019-09-25
Tests/aea test

DavidMinarsch:
## Proposed changes

Provides test for aea.

## Fixes

-

## Types of changes

## Checklist

## Further comments

- opened this on behalf of Aris


aea/skills/error/handler.py
```diff
@@ -63,10 +63,11 @@ def send_unsupported_protocol(self, envelope: Envelope) -> None:
         """
         logger.warning("Unsupported protocol: {}".format(envelope.protocol_id))
         reply = DefaultMessage(type=DefaultMessage.Type.ERROR,
-                               error_code=DefaultMessage.ErrorCode.UNSUPPORTED_PROTOCOL,
+                               error_code=DefaultMessage.ErrorCode.UNSUPPORTED_PROTOCOL.value,
                                error_msg="Unsupported protocol.",
                                error_data={"protocol_id": envelope.protocol_id})
-        self.context.outbox.put_message(to=envelope.sender, sender=self.context.agent_name, protocol_id=DefaultMessage.protocol_id,
+        self.context.outbox.put_message(to=envelope.sender, sender=self.context.agent_public_key,
+                                        protocol_id="error",
```
DavidMinarsch: I am not sure why this protocol_id has been replaced. @marcofavorito I think it should stay. Thoughts?

aea/registries/base.py
```diff
@@ -229,6 +229,20 @@ def fetch(self, protocol_id: ProtocolId) -> Optional[List[Handler]]:
             # TODO: introduce a controller class which intelligently selects the appropriate handler.
             return list(result.values())
 
+    def fetch_by_skill(self, protocol_id: ProtocolId, skill_id: SkillId) -> Optional[Handler]:
+        """
+        Fetch the handler for the protocol_id and skill id.
+
+        :param protocol_id: the protocol id
+        :param skill_id: the skill id
+        :return: the list of handlers registered for the protocol_id
```
DavidMinarsch: change to: ':return: the handler registered for the protocol_id and skill_id'

tests/data/dummy_aea/connections/local/connection.py
```diff
@@ -0,0 +1,365 @@
+# -*- coding: utf-8 -*-
```
DavidMinarsch: Ideally we would copy these in at runtime of the tests and remove again at the end. Otherwise they can get out of sync with the actual implementations. @marcofavorito  thoughts?

aea/skills/error/handler.py
marcofavorito: Yes, it should stay.

tests/data/dummy_aea/connections/local/connection.py
marcofavorito: Agreed. Even better, they could be symbolic links to the modules in `aea`.

DavidMinarsch: I tried symbolic links but then our import functionality doesn;t play with it..

marcofavorito:
Great work! The only lines in `aea/aea.py` not covered by the tests are in `AEA.handle`.

aea-test-coverage

To cover them, you might draw inspiration from [this](https://github.com/fetchai/agents-aea/blob/d327f988b08ce7f9d0e685e3d17a9013396e4767/tests/test_aea.py#L78-L87).

To generate a "decoding error", the agent should receive an envelope belonging to a known protocol but with `envelope.message` field not encoded properly, e.g. an envelope with dummy bytes in the `message`field.

To generate an "unsupported skill error", the agent should receive a message belonging to a known protocol (e.g. `"fipa"`) but no skill is registered to handle that protocol.

The `if not protocol.check(msg):` bit can be skipped by appending to every line the comment `# pragma: no cover` in this way:
```
if not protocol.check(msg):  # pragma: no cover
    ...  # pragma: no cover
    ...  # pragma: no cover
```

Issue #103, opened 2019-09-20
v1 concept and prototype of decision maker


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
This is a missing pease of our architecture.

**Describe the solution you'd like**
See architecture slides.

**Describe alternatives you've considered**
-

**Additional context**
-

Issue #97, opened 2019-09-20
Introduce multiple protocols to one skill


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
Currently a skill can handle one protocol.

**Describe the solution you'd like**
Adapt CLI and registries so multiple protocols can be added to a skill.

**Describe alternatives you've considered**
-

**Additional context**
-

Issue #49, opened 2019-09-07
Add FIPA negotiation skill


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
Currently, we have no proper skill.

**Describe the solution you'd like**
Wrap the fipa negotiation of the participant agent into a `fipa_negotiation` skill.

**Describe alternatives you've considered**
None

**Additional context**
This will help transitioning to a skills based TAC architecture and allow us to demonstrate the skills concept.

PR #142, opened 2019-09-30
First step towards writing a proper skill: fipa negotiation

DavidMinarsch:
## Proposed changes

This PR adds fipa_negotiation and tac skills and introduces necessary additional concepts to the framework (e.g. decision_maker, shared classes)

## Fixes

It addresses #49 and #103 and #156 and #97.

## Types of changes

## Checklist

## Further comments

The skills are WIP.


aea/aea.py
```diff
@@ -61,9 +62,20 @@ def __init__(self, name: str,
         self._directory = directory if directory else str(Path(".").absolute())
 
         self.mailbox = mailbox
-        self._context = AgentContext(self.name, self.crypto.public_key, self.outbox)
+        self._decision_maker = DecisionMaker(self.max_reactions, self.outbox)
+        self._context = AgentContext(self.name,
+                                     self.crypto.public_key,
+                                     self.outbox,
+                                     self.decision_maker.message_queue,
+                                     self.decision_maker.ownership_state,
```
DavidMinarsch: So we distinguish now between preferences and ownership spate. That is a relevant separation but might not be generic enough. We need to see.

aea/aea.py
```diff
@@ -114,8 +126,10 @@ def handle(self, envelope: Envelope) -> None:
         """
         protocol = self.resources.protocol_registry.fetch(envelope.protocol_id)
 
-        error_handler = self.resources.handler_registry.fetch_by_skill("default", "error")
-        assert error_handler is not None, "ErrorHandler not initialized"
+        handlers = self.resources.handler_registry.fetch_by_skill("default", "error")
```
DavidMinarsch: I am not sure why you amended this @marcofavorito . It should not be possible for more than one to be returned (we only allow unique handlers per protocol in a skill and only unique skills).

aea/protocols/state_update/message.py
```diff
@@ -0,0 +1,82 @@
...
+TransactionId = str
+Address = str
+
+CurrencyEndowment = Dict[str, float]  # a map from identifier to quantity
+GoodEndowment = Dict[str, int]   # a map from identifier to quantity
+UtilityParams = Dict[str, float]   # a map from identifier to quantity
+ExchangeParams = Dict[str, float]   # a map from identifier to quantity
+
+
+class StateUpdateMessage(Message):
```
DavidMinarsch: Currently, I am thinking of using the same protocols internally for comms with the DM. Not sure if that makes sense though. Of course no serialization is needed

aea/protocols/transaction/message.py
```diff
@@ -0,0 +1,139 @@
...
+                         sender=sender,
+                         counterparty=counterparty,
+                         is_sender_buyer=is_sender_buyer,
+                         currency_pbk=currency,
+                         sender_tx_fee=sender_tx_fee,
+                         counterparty_tx_fee=counterparty_tx_fee,
+                         amount=amount,
+                         quantities_by_good_pbk=quantities_by_good_pbk,
+                         **kwargs)
+        assert self.check_consistency(), "FIPAMessage initialization inconsistent."
```
DavidMinarsch: This could be simplified to be:
buyer_pbk
seller_pbk
currency_pbk
seller_tx_fee (share of)
buyer_tx_fee (share of)
amount
quantities_by_good_pbk

packages/skills/fipa_negotiation/__init__.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# ------------------------------------------------------------------------------
```
DavidMinarsch: The core thing to notice with the skills is that they are written so that the fipa_negotiation works completely independent of tac and vice versa. That is we can run tac with other protocols/skills/markets and similarly reuse them in settings outside tac.

aea/aea.py
marcofavorito: My change is wrong, thanks for spotting it.

DavidMinarsch:
Merging this so we are no longer blocked!

PR #159, opened 2019-10-02
Feature/aea docker

chr15murray:
* Updated the deploy-image to use the git repo at runtime instead of build time.

* added entryscript.sh with logic transplanted from dockerfile and augmented slightly to run until we have sample agents to test with. 

* Updated readme accoringly


chr15murray:
Thanks for merging

PR #177, opened 2019-10-04
Clean up weather examples

DavidMinarsch:
## Proposed changes

This PR addresses issues in the weather station examples.

## Fixes

It fixes problems in the examples.

## Types of changes

## Checklist

## Further comments

WIP


packages/skills/weather_station/dummy_weather_station_data.py
```diff
@@ -1,19 +1,42 @@
...
+import sqlite3
+import time
+from typing import Dict, Union
+
+DB_SOURCE = 'dummy_weather_station_data.db'
 
 
-#  Checking if the database exists#
-con = sqlite3.connect('weather_fake.db')
+# Checking if the database exists
```
DavidMinarsch: @Totoual you should handle errors here with try/except

packages/skills/weather_station/dummy_weather_station_data.py
```diff
@@ -87,11 +112,10 @@ def main(self):
             dict_of_data['wind_ave'] = random.randrange(0, 10, 1)
             dict_of_data['wind_dir'] = random.randrange(0, 14, 1)
             dict_of_data['wind_gust'] = random.randrange(1, 7, 1)
-            print(dict_of_data)
-            self.addData(dict_of_data)
+            self.add_data(dict_of_data)
             time.sleep(5)
 
 
 if __name__ == '__main__':
```
DavidMinarsch: @Totoual rather than creating these manually upfront we could create these in a task as part of the skill i.e. each call to task `execute()` causes a call to `generate` (with the while loop removed)

packages/skills/weather_station/dummy_weather_station_data.py
```diff
@@ -38,44 +60,47 @@
 
 
 class Forecast():
+    """Represents a whether forecast."""
+
+    def add_data(self, tagged_data: Dict[str, Union[int, datetime.datetime]]) -> None:
```
marcofavorito: We could think of using some ORM for Python like [SQLAlchemy](https://www.sqlalchemy.org/)

PR #192, opened 2019-10-08
Ledger Entity Early Stage

### LINKED ISSUE #180: Fetch Ledger Integration
We cannot transact on the Fetch ledger at the moment.

Integrate fetch ledger api into AEA framework.

None

**Additional context**
None

Totoual:
## Proposed changes

Early stage of Ledger Integration. Entity Generation and dictionary of public keys. (Kind of hacky)

## Fixes
#180 
## Types of changes

## Checklist

## Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...


aea/context/base.py
```diff
@@ -29,7 +29,7 @@ class AgentContext:
     """Provide read access to relevant data of the agent for the skills."""
 
     def __init__(self, agent_name: str,
-                 public_key: str,
+                 public_key: {},
```
DavidMinarsch: I would be more specific and use typing: Dict[str, str]

aea/crypto/base.py
```diff
@@ -29,6 +29,7 @@
 from cryptography.hazmat.primitives import hashes, serialization
 from cryptography.hazmat.primitives.asymmetric import ec, utils
 from cryptography.hazmat.primitives.serialization import load_pem_private_key
+from .fetchai_base import FetchCrypto
```
DavidMinarsch: please provide full path for imports where possible, so `aea.crypto.fetchai_base`

aea/crypto/base.py
```diff
@@ -67,7 +68,11 @@ def public_key(self) -> str:
 
         :return: a public key string in base58 format
         """
-        return self._public_key_b58
+        fetch_crypt = FetchCrypto()
+        public_keys = {}
```
DavidMinarsch: Would use typing: `public_keys = {}  # type: Dict[str, str]`.

aea/crypto/base.py
```diff
@@ -67,7 +68,11 @@ def public_key(self) -> str:
 
         :return: a public key string in base58 format
         """
-        return self._public_key_b58
+        fetch_crypt = FetchCrypto()
```
DavidMinarsch: `fetch_crypt` > `fetch_crypto`

DavidMinarsch: However, I would initialize this and the entire public keys dictionary in the `__init__` method and just call the initialized dict here

aea/crypto/base.py
```diff
@@ -67,7 +68,11 @@ def public_key(self) -> str:
```
DavidMinarsch: note, the function signature needs updating such that `public_keys(self) -> Dict[str, str]:`

aea/crypto/fetchai_base.py
```diff
@@ -0,0 +1,98 @@
...
+    def __init__(self, private_key_path: Optional[str] = None):
+        """Instantiate a crypto object."""
+        self._entity = Entity() if private_key_path is None else self._load_private_key_from_path(private_key_path)
+        self._public_key_obj = self._entity.public_key
+        self._public_key_bytes = self._entity.public_key_bytes
+        self._public_key_hex = self._entity.public_key_hex
+        self._display_address = Address(Identity.from_hex(self._public_key_hex))
+        self._private_key = self._entity.private_key
+        self._private_key_hex = self._entity.private_key_hex
+        self._private_key_bytes = self._entity.private_key_bytes
```
DavidMinarsch: Are you using all these private variables anywhere? Maybe just instantiate those vars which you actually use.

aea/crypto/base.py
Totoual: Removed from this file.

aea/crypto/fetchai_base.py
Totoual: I am not sure yet which of them I am going to use :D

aea/agent.py
```diff
@@ -71,7 +71,7 @@ def __init__(self, name: str,
         :return: None
         """
         self._name = name
-        self._crypto = Crypto(private_key_pem_path=private_key_pem_path)
+        self._crypto = Wallet(private_key_pem_path=private_key_pem_path)
```
DavidMinarsch: I am tempted to use `wallet` throughout in place of `crypto`. Any thoughts on this @marcofavorito ?

aea/cli/run.py
```diff
@@ -85,17 +84,12 @@ def run(ctx: Context, connection_name: str):
...
     agent_name = cast(str, ctx.agent_config.agent_name)
-    private_key_pem_path = cast(str, ctx.agent_config.private_key_pem_path)
-    if private_key_pem_path == "":
-        private_key_pem_path = _create_temporary_private_key_pem_path()
-    else:
-        _try_validate_private_key_pem_path(private_key_pem_path)
-    crypto = Crypto(private_key_pem_path=private_key_pem_path)
-    public_key = crypto.public_key
+    wallet = Wallet(private_key_pem_path=ctx.agent_config.private_key_pem_path)
+    public_key = wallet.public_keys
```
DavidMinarsch: Please call it `public_keys` rather than `public_key`.

aea/context/base.py
```diff
@@ -60,7 +61,7 @@ def agent_name(self) -> str:
         return self._agent_name
 
     @property
-    def public_key(self) -> str:
+    def public_key(self) -> Dict[str, str]:
```
DavidMinarsch: Please call it `public_keys`. Then also create the old `public_key` which we will make the default one, so:
def public_key(self) -> str:
    " Get the default public key"
     return self._public_keys['default']

That way you don't break the interface and are backwards compatible.

aea/skills/base.py
```diff
@@ -55,7 +55,7 @@ def agent_name(self) -> str:
         return self._agent_context.agent_name
 
     @property
-    def agent_public_key(self) -> str:
+    def agent_public_key(self) -> Dict[str, str]:
```
DavidMinarsch: Use this for the `default` public key so the interface is stable but add the whole dictionary as `agent_public_keys`.

aea/cli/run.py
Totoual: I was waiting for your review. I used it like that because in the framework we are using public_key and didn't want to modify every script at the moment.

aea/crypto/ethereum_base.py
```diff
@@ -0,0 +1,105 @@
...
+                with open(path, "r") as key:
+                    data = key.read()
+                    account = Account.from_key(data)
+            else:
+                account = self._generate_private_key()
+            return account
+        except IOError as e:
+            logger.exception(str(e))
+
+    def sign_transaction(self, message: str) -> bytes:
```
DavidMinarsch: shouldn't the message be bytes?

aea/crypto/fetchai_base.py
```diff
@@ -0,0 +1,117 @@
...
+
+            else:
+                entity = self._generate_private_key()
+
+            return entity
+        except IOError as e:
+            logger.exception(str(e))
+
+    def _generate_private_key(self) -> Entity:
+        path = Path("pk.txt")
```
DavidMinarsch: `fet_pk.txt` for consistency?

aea/crypto/fetchai_base.py
```diff
@@ -0,0 +1,117 @@
...
+        print(path)
```
DavidMinarsch: remove print

aea/crypto/fetchai_base.py
```diff
@@ -0,0 +1,117 @@
...
+        entity = Entity()
+        with open(path, "w+") as file:
```
DavidMinarsch: What if a private key file already exists?

aea/crypto/fetchai_base.py
DavidMinarsch: I would remove all but the entity, then expose those you want via @property.

aea/crypto/initialiser.py
```diff
@@ -0,0 +1,79 @@
...
+            "fetchai": self.crypto_objects['fetchai'].public_key,
+            "ethereum": self.crypto_objects['ethereum'].public_key
+        }
+
+        self.private_key_pem_path = ""
+
+    def _setup_crypto(self, private_key_pem_path: Optional[str] = None):
+        """Create the crypto object."""
+        private_key_pem_path = cast(str, private_key_pem_path)
+        if private_key_pem_path == "":
```
DavidMinarsch: we could move that logic into crypto to mirror the below functions for ethereum and fetchai

aea/crypto/initialiser.py
```diff
@@ -0,0 +1,79 @@
...
+        fetch_crypto = FetchCrypto("pk.txt")
+        return fetch_crypto
+
+    def _setup_ethereum_crypto(self):
+        """Create an ethereum account."""
+        ethereum_crypto = EthCrypto("eth_pk.txt")
+        return ethereum_crypto
+
+    @property
+    def public_key(self):
```
DavidMinarsch: public_keys

aea/crypto/ethereum_base.py
```diff
@@ -0,0 +1,105 @@
...
+        m_message = encode_defunct(text=message)
+        signature = self.account.sign_message(m_message)
+        return signature
+
+    def _generate_private_key(self) -> Account:
+        """Generate a key pair for ethereum network."""
+        path = Path("eth_pk.txt")
+        print(path)
+        pk = Account.create()
+        with open(path, "w+") as file:
```
DavidMinarsch: what if a private key already exists?

aea/crypto/ethereum_base.py
Totoual: No because it serialise on the background.

aea/crypto/ethereum_base.py
Totoual: You just load it and you create the account.

aea/crypto/fetchai_base.py
Totoual: You just load it and you create the Entity.

aea/crypto/initialiser.py
Totoual: I would prefer to have the logic of setting up the cryptos somewhere else rather than wrap everything inside crypto. It will make it easier to add more cryptos.

aea/crypto/ethereum_base.py
DavidMinarsch: Hm, but what is the `message` then here?

aea/crypto/fetchai_base.py
DavidMinarsch: Yes, but your code will overwrite it. So basically, what I am suggesting with this question is that you need to adjust the code to check whether a private key exists, if so load it, otherwise, generate it.

aea/crypto/initialiser.py
DavidMinarsch: I am not saying to move the `_setup_crypto` method out. I am saying to make it look similar to the other `_setup_x` methods by moving the first lines of the method into crypto (in fact, it might already be there).

aea/crypto/ethereum_base.py
DavidMinarsch: See above. Overwrite issue.

aea/crypto/initialiser.py
Totoual: Yes, sorry I get it now. This code doesn't exist in the crypto. Since it is extracted from the run.py. But I'll move it inside the crypto to match the other two.

aea/agent.py
```diff
@@ -57,7 +57,7 @@ class Agent(ABC):
     """This class implements a template agent."""
 
     def __init__(self, name: str,
-                 private_key_pem_path: Optional[str] = None,
+                 private_key_paths: Dict[str, str],
```
DavidMinarsch: Can you change it to `Optional[Dict[str, str]] = None`?

aea/agent.py
```diff
@@ -71,7 +71,7 @@ def __init__(self, name: str,
         :return: None
         """
         self._name = name
-        self._crypto = Crypto(private_key_pem_path=private_key_pem_path)
+        self._crypto = Wallet()
```
DavidMinarsch: I think this needs to be passed the private keys?

aea/cli/run.py
```diff
@@ -85,17 +83,13 @@ def run(ctx: Context, connection_name: str):
...
     _try_to_load_agent_config(ctx)
     agent_name = cast(str, ctx.agent_config.agent_name)
-    private_key_pem_path = cast(str, ctx.agent_config.private_key_pem_path)
-    if private_key_pem_path == "":
-        private_key_pem_path = _create_temporary_private_key_pem_path()
-    else:
-        _try_validate_private_key_pem_path(private_key_pem_path)
-    crypto = Crypto(private_key_pem_path=private_key_pem_path)
-    public_key = crypto.public_key
+    wallet = Wallet()
```
DavidMinarsch: I think you need to pass the private key paths?

Totoual: Not anymore. I am loading them from the aea-config.yaml

aea/crypto/ethereum_base.py
Totoual: It is any message that will encoded to EIP-101_messages. Based on the eth_account comment on the library. They also have a specific sign_transaction.

Issue #182, opened 2019-10-05
Investigate addressable peer to peer message routing similar to Ledger


DavidMinarsch:
This is a scoping exercise. Nothing needs to be built.

Issue #171, opened 2019-10-04
Add a Stub connector for enqueing and dequeueing


DavidMinarsch:

marcofavorito:
Fixed by #211

PR #209, opened 2019-10-11
WIP: TCP connection client/server side

marcofavorito:
## Proposed changes

This PR adds a new type of connection relying on `asyncio` sockets. It introduces:
- `TCPConnection` and `TCPChannel` abstractions
- `TCPServerConnection` and `TCPServerChannel` to implement the connection server-side
- `TCPClientConnection` and `TCPClientChannel` to implement the connection client-side

It can be used both in networked version and UNIX version (only the latter has been tested). The behaviour can be switched by using the `unix` flag in the channel constructor.

## Fixes

This implementation allows the fix for  #171. It addresses #182

## Types of changes

## Checklist

## Further comments

These two new connections are building blocks for P2P communication. The `P2PConnection` will use those two abstractions in the following way:
- every `P2PConnection` uses a `TCPServerConnection` to listen incoming requests
- whenever the `P2PConnection` wants to send an envelope to a certain `address`, it'll instantiate a `TCPClientConnection` that will handle the communication.
 -then, the fetch and receiving loop of `P2PConnection` will just send the envelopes through the right `TCPClientConnection` and receive from `TCPServerConnection`, respectively.

At the moment, `TCPClientConnection` support the connection to only one address. 

Should it be called `SocketConnection`? Since we support also the UNIX socket.


aea/connections/p2p/base.py
```diff
@@ -0,0 +1,173 @@
...
+
+        self.public_key = public_key
+        self._channel = channel
+        self._stopped = True
+
+    def _fetch(self) -> None:
+        """Fetch the envelopes from the outqueue and send them."""
+        while not self._stopped:
+            try:
+                msg = self.out_queue.get(block=True, timeout=2.0)
```
DavidMinarsch: Call it `envelope` rather than `msg` for clarity?

aea/connections/p2p/base.py
```diff
@@ -0,0 +1,173 @@
...
+                if msg is not None:
+                    self.send(msg)
+            except queue.Empty:
+                pass
+
+    def _receive_loop(self):
+        """Receive envelopes."""
+        while not self._stopped:
+            try:
+                data = self._channel.recv(block=True, timeout=2.0)
```
DavidMinarsch: Call it `envelope` as well?

aea/connections/p2p/base.py
```diff
@@ -0,0 +1,173 @@
...
+        nbytes = struct.pack("I", len(data))
+        logger.debug("#bytes: {}".format(nbytes))
+        try:
+            writer.write(nbytes)
+            writer.write(data)
+            await writer.drain()
+        except CancelledError:
+            return None
+
+    async def _recv_loop(self, reader) -> None:
```
DavidMinarsch: `reader: StreamReader`

aea/connections/p2p/tcp_server.py
```diff
@@ -0,0 +1,139 @@
...
+        future = asyncio.run_coroutine_threadsafe(self._send(writer, envelope.encode()), loop=self._loop)
+        future.result()
+
+
+class TCPServerConnection(TCPConnection):
+    """Implementation of a TCP server connection."""
+
+    @classmethod
+    def from_config(cls, public_key: str, connection_configuration: ConnectionConfig) -> 'Connection':
+        """Get the Local OEF connection from the connection configuration.
```
DavidMinarsch: So we are again connecting to a central server?

DavidMinarsch:
Thanks a lot for this work @marcofavorito. At this point I suggest to not add more features as this is a good starting point for further discussions with ledger team etc. How complicated would a test of the networked version be?

aea/connections/tcp/tcp_client.py
```diff
@@ -0,0 +1,93 @@
...
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+# ------------------------------------------------------------------------------
+
+"""Implementation of the TCP server."""
```
DavidMinarsch: Please update docstring.

aea/connections/tcp/tcp_client.py
```diff
@@ -0,0 +1,93 @@
...
+            self._fetch_task.result()
+            self._reader.feed_eof()
+            self._read_task.cancel()
+        except CancelledError:
+            pass
+        self._writer.close()
+
+    def select_writer_from_envelope(self, envelope: Envelope) -> Optional[StreamWriter]:
+        """Select the destination, given the envelope."""
+        return self._writer
```
DavidMinarsch: Could add `TODO` here to clarify that this is unfinished (one connection only -> envelope `to` not used).

PR #227, opened 2019-10-14
Example transaction

Totoual:
## Proposed changes

An example that demonstrates the based usage of the Decision maker. Uses the weather_station example as a base and takes this example a step ahead by completing a transaction through the Fetch Ledger Api  

## Fixes

None

## Types of changes

## Checklist

## Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...


aea/decision_maker/base.py
```diff
@@ -301,11 +305,36 @@ def _handle_tx_message(self, tx_message: TransactionMessage) -> None:
         :param tx_message: the transaction message
         :return: None
         """
-        score_diff = self.preferences.get_score_diff_from_transaction(self.ownership_state, tx_message)
-        if score_diff >= 0.0:
-            pass  # TODO: implement tx flow
-            # envelope = Envelope()
-            # self.outbox.put(envelope)
+        amount = tx_message.get("amount")
+        m_address = self.generate_address_from_public_key(self._wallet.public_keys['fetchai'])
```
DavidMinarsch: Two suggestions here:
- `m_address` is not a good name for a variable as its not explicit
- this variable should be instantiated in wallet and exposed via it (I can imagine many use cases for the address)

aea/decision_maker/base.py
```diff
@@ -301,11 +305,36 @@ def _handle_tx_message(self, tx_message: TransactionMessage) -> None:
...
+        api = LedgerApi("127.0.0.1", 8100)
```
DavidMinarsch: Generation of wealth should not happen in the agent. Wealth generation is a special case for test nets. So there should be a standalone script which accepts as `argument` the `m_address` and then generates some wealth for it.

aea/decision_maker/base.py
```diff
@@ -301,11 +305,36 @@ def _handle_tx_message(self, tx_message: TransactionMessage) -> None:
...
+        self._generate_wealth()
+        print(api.tokens.balance(m_address))
```
DavidMinarsch: You often add `print` statements. Please adapt the habit to remove them for PRs if they are only for you during development. If you want to log stuff then please use the logger :)

aea/decision_maker/base.py
```diff
@@ -301,11 +305,36 @@ def _handle_tx_message(self, tx_message: TransactionMessage) -> None:
...
+        if amount <= api.tokens.balance(m_address):
+            m_entity = Entity.from_hex(self._wallet.crypto_objects['fetchai'].private_key)
+            to = self.generate_address_from_public_key(tx_message.get("counterparty"))
+            api.sync(api.tokens.transfer(m_entity, to, 100, 20))
+            print(api.tokens.balance(m_address))
+
+    def _generate_wealth(self) -> None:
```
DavidMinarsch: This method should not be part of the decision maker. See above.

aea/decision_maker/base.py
```diff
@@ -301,11 +305,36 @@ def _handle_tx_message(self, tx_message: TransactionMessage) -> None:
...
+        Generate tokens to be able to make a transaction.
+
+        :return:
+        """
+        m_entity = Entity.from_hex(self._wallet.crypto_objects['fetchai'].private_key)
+        print(m_entity.public_key)
+        api = LedgerApi("127.0.0.1", 8100)
+        api.tokens.wealth(m_entity, 1000)
+
+    def generate_address_from_public_key(self, public_key) -> Address:
```
DavidMinarsch: This should not be needed here, see above.

aea/decision_maker/base.py
```diff
@@ -301,11 +305,36 @@ def _handle_tx_message(self, tx_message: TransactionMessage) -> None:
...
-            # envelope = Envelope()
-            # self.outbox.put(envelope)
+        amount = tx_message.get("amount")
+        m_address = self.generate_address_from_public_key(self._wallet.public_keys['fetchai'])
+        api = LedgerApi("127.0.0.1", 8100)
+        self._generate_wealth()
+        print(api.tokens.balance(m_address))
+        if amount <= api.tokens.balance(m_address):
+            m_entity = Entity.from_hex(self._wallet.crypto_objects['fetchai'].private_key)
+            to = self.generate_address_from_public_key(tx_message.get("counterparty"))
```
DavidMinarsch: We should already pass the `counterpart` address into the tx_message (since the counterpart will have to send it to us)

aea/decision_maker/base.py
DavidMinarsch: If you want such a helper method then I would move it into the crypto module.

packages/skills/weather_client_ledger/handlers.py
```diff
@@ -0,0 +1,230 @@
...
+        proposals = cast(List[Description], message.get("proposal"))
+        message_id = cast(int, message.get("message_id"))
+        dialogue_id = cast(int, message.get("dialogue_id"))
+        if msg_performative == FIPAMessage.Performative.PROPOSE:
+            if proposals is not []:
+                for item in proposals:
+                    logger.info("[{}]: received proposal={} in dialogue={}".format(self.context.agent_name, item.values,
+                                                                                   dialogue_id))
+                    if "Price" in item.values.keys():
+                        if item.values["Price"] < self.max_price:
```
DavidMinarsch: The client should also have an account on ledger and check whether it can actually afford the transaction given it's wealth. So it will only accept if the tx price < max_price and if it has enough wealth to afford price.

packages/skills/weather_client_ledger/handlers.py
```diff
@@ -0,0 +1,230 @@
...
+        :param sender: the sender
+        :return: None
+        """
+        json_data = json.loads(message.get("content").decode())
+        if json_data is not None:
+            if "Command" in json_data.keys():
+                if json_data['Command'] == 'success':
+                    logger.info("[{}]: receiving data ...".format(self.context.agent_name))
+                    logger.info("[{}]: this is the data I got: {}".format(self.context.agent_name, json_data))
+                elif json_data['Command'] == "address":
```
DavidMinarsch: Can you add a check in both weather clients to check that they only send the transaction if they previously received the data?

Totoual: I can do that but I don't agree. If you think an online transaction, you first pay and then you wait to receive the actual product. What will happen if the client receives the data and then disconnect ?

aea/decision_maker/base.py
Totoual: The counterpart sends the public_key and not the actual address. We had this discussion and we said that since we can generate the address from the public key there is no need to change how we store things in the wallet.

aea/decision_maker/base.py
Totoual: @DavidMinarsch  I could add this in the Wallet as a property. But is it something that we want? I mean there is the functionality in the fetch_crypto_base but we want something that specific in the wallet (It is only for the fetch public_key)?

aea/decision_maker/base.py
DavidMinarsch: Ok, so in this case you can just create a helper method which is available from aea.crypto.helpers and lets you recover the address from the public key

aea/decision_maker/base.py
```diff
@@ -301,11 +306,15 @@ def _handle_tx_message(self, tx_message: TransactionMessage) -> None:
         :param tx_message: the transaction message
         :return: None
         """
-        score_diff = self.preferences.get_score_diff_from_transaction(self.ownership_state, tx_message)
-        if score_diff >= 0.0:
-            pass  # TODO: implement tx flow
-            # envelope = Envelope()
-            # self.outbox.put(envelope)
+        amount = 0
+        m_address = _generate_address_from_public_key(self._wallet.public_keys['fetchai'])
```
DavidMinarsch: I think I mentioned in a previous review. The funding off the account of an agent has to happen outside of the agent run. Here, you either have money in the account and accept or you don't and reject.

aea/decision_maker/base.py
```diff
@@ -301,11 +306,15 @@ def _handle_tx_message(self, tx_message: TransactionMessage) -> None:
...
+        api = LedgerApi("127.0.0.1", 8100)
+        if amount <= api.tokens.balance(m_address) and amount <= api.tokens.balance(m_address):
```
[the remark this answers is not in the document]
Totoual: Fair sorry.

aea/crypto/wealth_generation.py
```diff
@@ -0,0 +1,49 @@
...
+import sys
+from typing import Optional
+
+from fetchai.ledger.api import LedgerApi        # type: ignore
+from fetchai.ledger.crypto import Entity, Address, Identity  # type: ignore
+
+logger = logging.getLogger(__name__)
+
+
+def _generate_wealth(amount, private_key: Optional[str] = None) -> None:
```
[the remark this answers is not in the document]
Totoual: No specific reason. Removed it.

aea/decision_maker/base.py
Totoual: I should move the check (if we have money in the wallet and if it is less than the amount we are willing to spend ) to the handler before accepting or declining the proposal. Though I don't want to create a new connection to the ledger to ask for the balance from the handler and that's why I am going to check in the decision-maker for now. Is that ok? (When we have the messaging system complete we can query the decision-maker for that kind of stuff and the decision-maker could pass the balance for example to the handler and then answer ACCEPT or DECLINE)

aea/decision_maker/base.py
```diff
@@ -301,11 +306,14 @@ def _handle_tx_message(self, tx_message: TransactionMessage) -> None:
         :param tx_message: the transaction message
         :return: None
         """
-        score_diff = self.preferences.get_score_diff_from_transaction(self.ownership_state, tx_message)
-        if score_diff >= 0.0:
-            pass  # TODO: implement tx flow
-            # envelope = Envelope()
-            # self.outbox.put(envelope)
+        m_address = generate_address_from_public_key(self._wallet.public_keys['fetchai'])
```
DavidMinarsch: we have an address dict right, so just take it from there?

aea/decision_maker/base.py
```diff
@@ -301,11 +306,14 @@ def _handle_tx_message(self, tx_message: TransactionMessage) -> None:
...
+        api = LedgerApi("127.0.0.1", 8000)
+        if tx_message.get("amount") <= api.tokens.balance(m_address):
+            m_entity = Entity.from_hex(self._wallet.crypto_objects['fetchai'].private_key)
```
DavidMinarsch: we should make the entity a property of the wallet I think, like public keys, addresses etc.

aea/decision_maker/base.py
```diff
@@ -301,11 +306,14 @@ def _handle_tx_message(self, tx_message: TransactionMessage) -> None:
...
+            to = generate_address_from_public_key(tx_message.get("counterparty"))
+            api.sync(api.tokens.transfer(m_entity, to, 1, 1))
```
DavidMinarsch: what do 1, 1 stand for? please create constants :)

packages/skills/weather_client_ledger/handlers.py
```diff
@@ -0,0 +1,245 @@
...
+
+class FIPAHandler(Handler):
+    """This class scaffolds a handler."""
+
+    SUPPORTED_PROTOCOL = FIPAMessage.protocol_id  # type: Optional[ProtocolId]
+
+    def __init__(self, **kwargs):
+        """Initiliase the handler."""
+        super().__init__(**kwargs)
+        self.max_price = kwargs['max_price'] if 'max_price' in kwargs.keys() else MAX_PRICE
```
DavidMinarsch: These are preferences. So try to see if you can represent them in the current preference module :)

packages/skills/weather_client_ledger/handlers.py
```diff
@@ -0,0 +1,245 @@
...
+        proposals = cast(List[Description], message.get("proposal"))
+        message_id = cast(int, message.get("message_id"))
+        dialogue_id = cast(int, message.get("dialogue_id"))
+        if msg_performative == FIPAMessage.Performative.PROPOSE:
+            if proposals is not []:
+                for item in proposals:
+                    logger.info("[{}]: received proposal={} in dialogue={}".format(self.context.agent_name, item.values,
+                                                                                   dialogue_id))
+                    if "Price" in item.values.keys():
+                        # TODO: Add  if tx_message.get("amount") <= api.tokens.balance(m_address)
```
DavidMinarsch: We can expose the wealth of an agent via the AgentContext.

packages/skills/weather_station_ledger/handlers.py
```diff
@@ -0,0 +1,262 @@
...
+        """
+        Implement the reaction to an message.
+
+        :param message: the message
+        :param sender: the sender
+        :return: None
+        """
+        fipa_msg = cast(FIPAMessage, message)
+        msg_performative = FIPAMessage.Performative(fipa_msg.get('performative'))
+        message_id = cast(int, fipa_msg.get('id'))
```
DavidMinarsch: get('message_id')

packages/skills/weather_station_ledger/handlers.py
```diff
@@ -0,0 +1,262 @@
...
+        :param dialogue_id: the dialogue id
+        :return: None
+        """
+        new_message_id = 1
+        new_target = message_id
+        fetched_data = self.db.get_data_for_specific_dates(DATE_ONE, DATE_TWO)
+
+        if len(fetched_data) >= 1:
+            self.fetched_data = fetched_data
+            total_price = self.fet_price * len(fetched_data)
```
DavidMinarsch: the fet_price should also be expressed via preferences

scripts/wealth_generation.py
```diff
@@ -0,0 +1,59 @@
...
+    api.sync(api.tokens.wealth(entity_to_generate_wealth, arguments.amount))
+    address = Address(entity_to_generate_wealth)
+    logger.info('The new balance of the address {} is : {} FET'.format(address, api.tokens.balance(address)))
+
+
+def parse_arguments():
+    """Arguments parsing."""
+    parser = argparse.ArgumentParser("wealth_creation")
+    parser.add_argument("--amount", type=int, default=10, help="The amount we want to generate to the address")
+    parser.add_argument("--private-key", type=str, default="dd50e375684f4a9013129483883fff2d3402335d847547819a45c11198b627db",
```
DavidMinarsch: I would not do that as a fixed PK. Just generate a default PK dynamically? Otherwise, successive people doing this demo will wonder why the wealth isn't aligned with their amount.

packages/skills/weather_client_ledger/handlers.py
DavidMinarsch: Have you had a look at the Dialogue, Dialogues and DialogueLabel classes in fipa_negotiation? It would probably be good to implement something like this. Otherwise, you can short cut the dialogue etc.

aea/decision_maker/base.py
Totoual: No, no. Because through the entity you can access the private key.

aea/decision_maker/base.py
Totoual: Nah, I just forgot to change them back to tx_message.get("amount") and tx_message.get("sender_tx_fee"). These were changes I made before we fix the ledger.

packages/skills/weather_client_ledger/handlers.py
Totoual: I moved the max_price to the decision_maker/base.Preferences. And I have access through self.context.agent_preferences.

packages/skills/weather_station_ledger/handlers.py
Totoual: It can be expressed via preferences though, at the moment the decision-maker expresses the client and not the weather_station.

packages/skills/weather_client_ledger/handlers.py
Totoual: And why not a query to the decision-maker? It makes sense if you think that it will be the only thing that will use the ledger connection for queries and that kind of stuff right ?

aea/cli/run.py
```diff
@@ -93,9 +93,12 @@ def run(click_context, connection_name: str, env_file: str, install_deps: bool):
     _try_to_load_agent_config(ctx)
     _load_env_file(env_file)
     agent_name = cast(str, ctx.agent_config.agent_name)
+
     _verify_or_create_private_keys(ctx)
     wallet = Wallet(dict([(identifier, config.path) for identifier, config in ctx.agent_config.private_key_paths.read_all()]))
 
+    _verify_ledger_apis_access(ctx)
```
DavidMinarsch: Here we verify that we can actually connect to the ledgers specified via configs

aea/crypto/ethereum.py
```diff
@@ -94,10 +92,41 @@ def sign_transaction(self, message: str) -> bytes:
...
+        signature = self._account.sign_message(m_message)
         return signature
 
     def _generate_private_key(self) -> Account:
         """Generate a key pair for ethereum network."""
         account = Account.create()
         return account
+
+    @property
+    def token_balance(self) -> float:
```
DavidMinarsch: for now crypto objects bundle both the entity and the ledger access. later we can refactor and separate it into two

aea/decision_maker/base.py
```diff
@@ -301,11 +306,37 @@ def _handle_tx_message(self, tx_message: TransactionMessage) -> None:
...
         """
-        score_diff = self.preferences.get_score_diff_from_transaction(self.ownership_state, tx_message)
-        if score_diff >= 0.0:
-            pass  # TODO: implement tx flow
-            # envelope = Envelope()
-            # self.outbox.put(envelope)
+        crypto_identifier = CURRENCY_TO_ID_MAP.get(cast(str, tx_message.get("currency")))
+        amount = cast(float, tx_message.get("amount"))
+        crypto_object = self._wallet.crypto_objects.get(crypto_identifier)
+        if self._is_acceptable_tx(crypto_object, amount):
```
DavidMinarsch: for now we accept any transaction which is affordable with no regard for preferences.

PR #240, opened 2019-10-15
Exposed generated addresses to the agent Context

### LINKED ISSUE #196: Start using Address rather than Public Key
No

Start using address rather than the public key in messages and the oef. 

-

**Additional context**
-

Totoual:
## Proposed changes

Storing the address of each crypto in a dictionary. The addresses are generated inside the wallet and we can have access to them through the agent_context.

## Fixes
#196 

## Types of changes

## Checklist

- [x ] I am making a pull request against the `develop` branch (left side). Also you should start your branch off our `develop`.
- [ x] Lint and unit tests pass locally with my changes

## Further comments
-


aea/configurations/base.py
```diff
@@ -435,6 +435,7 @@ def __init__(self,
         self.url = url
         self.registry_path = registry_path
         self.private_key_paths = CRUDCollection[PrivateKeyPathConfig]()
+        self.addresses = {}  # type: Dict[str, str]
```
DavidMinarsch: I don't think we need this here.

aea/context/base.py
```diff
@@ -66,6 +68,11 @@ def public_keys(self) -> Dict[str, str]:
         """Get public keys."""
         return self._public_keys
 
+    @property
+    def addresses(self) -> Dict[str, str]:
+        """Get addresses."""
+        return self._addresses
```
DavidMinarsch: let's similarly to below (`public_key`), create also an `address` getter which uses the one associated with the default crypto.

aea/crypto/ethereum_base.py
```diff
@@ -58,7 +58,7 @@ def public_key(self) -> str:
         return self._public_key
 
     @property
-    def display_address(self) -> str:
+    def address(self) -> str:
         """
         Return the display_address for the key pair.
```
DavidMinarsch: please update doc string too `display_address` is out dated

aea/crypto/fetchai_base.py
```diff
@@ -60,7 +60,7 @@ def private_key(self) -> str:
         return self._entity.private_key_hex
 
     @property
-    def display_address(self) -> str:
+    def address(self) -> str:
         """
         Return the display_address for the key pair.
```
DavidMinarsch: Same here, doc string is out of date.

aea/crypto/wallet.py
```diff
@@ -66,3 +69,8 @@ def public_keys(self):
     def crypto_objects(self):
         """Get the crypto objects (key pair)."""
         return self._crypto_objects
+
+    @property
+    def crypto_addresses(self):
```
DavidMinarsch: can we drop the `crypto_`?

aea/skills/base.py
```diff
@@ -65,6 +65,11 @@ def agent_public_keys(self) -> Dict[str, str]:
         """Get public keys."""
         return self._agent_context.public_keys
 
+    @property
```
DavidMinarsch: And similar, create the matching `agent_address`.

aea/aea.py
```diff
@@ -67,6 +67,7 @@ def __init__(self, name: str,
         self._decision_maker = DecisionMaker(self.max_reactions, self.outbox)
         self._context = AgentContext(self.name,
                                      self.wallet.public_keys,
+                                     self.wallet.crypto_addresses,
```
DavidMinarsch: see below re renaming `crypto_addresses` to simply `addresses`

aea/crypto/wallet.py
```diff
@@ -66,3 +69,8 @@ def public_keys(self):
     def crypto_objects(self):
         """Get the crypto objects (key pair)."""
         return self._crypto_objects
+
+    @property
+    def addresses(self):
+        """Get the crypt addresses."""
```
marcofavorito: fix typo

aea/crypto/wallet.py
```diff
@@ -66,3 +69,8 @@ def public_keys(self):
     def crypto_objects(self):
         """Get the crypto objects (key pair)."""
         return self._crypto_objects
+
+    @property
+    def addresses(self):
```
marcofavorito: add type annotation for the return value

aea/context/base.py
```diff
@@ -31,6 +31,7 @@ class AgentContext:
 
     def __init__(self, agent_name: str,
                  public_keys: Dict[str, str],
+                 addresses: Dict[str, str],
```
marcofavorito: worth considering having a type like
```
Address = str
```

so we can then easily extend with other types if needed:
```
Address = Union[str, AnotherType]
```

marcofavorito: This comment applies to all the other places where `str` appears when referring to an address

aea/context/base.py
```diff
@@ -66,6 +68,16 @@ def public_keys(self) -> Dict[str, str]:
...
 
+    @property
+    def addresses(self) -> Dict[str, str]:
+        """Get addresses."""
+        return self._addresses
+
+    @property
+    def address(self) -> str:
+        """Get the defualt address."""
+        return self._addresses['default']
```
marcofavorito: Why this returns only the default address?

DavidMinarsch: @marcofavorito for now we want to stay backwards compatible in all the skills to avoid a rewrite, hence `public_key` points to the default one. So to mirror that `address` does the same.

DavidMinarsch: In future, we can give the developer more control over what address exposed here.

aea/context/base.py
DavidMinarsch: Agreed, though I'd prefer if we open a separate issue for that and deal with it another day. Not a priority atm.

PR #277, opened 2019-10-18
GUI bug fix

dishmop:
## Proposed changes

GUI now lists all skills, connections and protocols from aea in virtual env as well as from packge registry dir. Previously it just listed from package dir. 

## Fixes

-

## Types of changes

## Checklist

- [ x] I am making a pull request against the `develop` branch (left side). Also you should start your branch off our `develop`.
- [x ] I have added necessary documentation (if appropriate)
- [x ] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/aea/cli_gui/__init__.py
+++ b/aea/cli_gui/__init__.py
@@ -101,19 +101,22 @@ def get_agents():
     return agent_list
 
 
-def get_registered_items(item_type):
-    """Return list of all protocols, connections or skills in the registry."""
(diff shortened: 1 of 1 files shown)


dishmop:
I agree that aea search would be better (I only realised they existed recently) - though search doesn't return the description, so I couldn't use it out of the box.

The fix I have here seems to work ok.

I need to go now so if you want this fix you will need to merge it yourself (sorry)

PR #308, opened 2019-10-23
Almost 100% on aea/helpers

Totoual:
## Fixes

Added tests for the aea/helpers.

## Types of changes

- [x] Added tests
expected)

## Checklist

### DIFF
--- a/aea/connections/local/connection.py
+++ b/aea/connections/local/connection.py
@@ -44,7 +44,7 @@ class LocalNode:
 
     def __init__(self):
         """Initialize a local (i.e. non-networked) implementation of an OEF Node."""
-        self.agents = dict()  # type: Dict[str, Description]
+        self.agents = defaultdict(lambda: [])  # type: Dict[str, List[Description]]
         self.services = defaultdict(lambda: [])  # type: Dict[str, List[Description]]
         self._lock = threading.Lock()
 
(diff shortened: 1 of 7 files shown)


aea/connections/local/connection.py
```diff
@@ -138,6 +140,17 @@ def register_service(self, public_key: str, service_description: Description):
         with self._lock:
             self.services[public_key].append(service_description)
 
+    def register_agent(self, public_key: str, agent_description: Description):
```
marcofavorito: this is not used anywhere - shall we remove it?

aea/connections/local/connection.py
```diff
@@ -178,9 +191,10 @@ def search_agents(self, public_key: str, search_id: int, query: Query) -> None:
         if query.model is None:
             result = list(set(self.services.keys()))
         else:
-            for agent_public_key, description in self.agents.items():
-                if query.model == description.data_model:
-                    result.append(agent_public_key)
+            for agent_public_key, descriptions in self.agents.items():
+                for description in descriptions:
+                    if query.model == description.data_model:
+                        result.append(agent_public_key)
```
marcofavorito: notice, this was the old version. Here we should call the `check` method of the `query` object, which is now implemented.

aea/connections/local/connection.py
Totoual: It is pulled from the feature PR. I messed up with the PRs a little bit.

aea/connections/local/connection.py
Totoual: it is based on the service functionality.

marcofavorito: Oh ok, I think we can remove also `search_agents` since it is not used (right?)

Issue #296, opened 2019-10-19
Address TODOs in weather skill


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
Self explanatory.

**Describe the solution you'd like**
-
**Describe alternatives you've considered**
-
**Additional context**
-

PR #324, opened 2019-10-27
Feature/weather skills

DavidMinarsch:
## Proposed changes

This PR introduces the necessary changes to allow for the weather skills to work with ledger integration and usage of the decision maker.

## Fixes

It addresses #296 

## Types of changes

## Checklist

## Further comments

-


aea/aea.py
```diff
@@ -175,6 +177,15 @@ def update(self) -> None:
         for task in self.resources.task_registry.fetch_all():
             task.execute()
         self.decision_maker.execute()
+        counter = 0
```
DavidMinarsch: I have added this temporarily. I don't think we should keep this for long. Once we have the `Filter` next week I propose that we pipe the messages straight to the `Filter` and the `Filter` can then allocate them to the correct `Handler` (see comments below as well).

aea/crypto/ledger_apis.py
```diff
@@ -99,6 +100,31 @@ def transfer(self, identifier: str, entity: Entity, destination_address: str, am
...
+        assert identifier in self.apis.keys(), "Unsupported ledger identifier."
+        is_successful = False
+        try:
+            api = self.apis[identifier]
+            logger.info("Checking the transaction ...")
+            # tx_status = cast(TxStatus, api.tx.status(tx_digest))
+            tx_status = cast(str, api.tx.status(tx_digest))
+            SUCCESSFUL_TERMINAL_STATES = ('Executed', 'Submitted')
+            if tx_status in SUCCESSFUL_TERMINAL_STATES:
+                # TODO: check the amount of the transaction is correct
```
DavidMinarsch: Currently, the ledger api does not allow for this check. Ed says it will be available in a few days :).

packages/skills/weather_client_ledger/behaviours.py
```diff
@@ -62,6 +76,63 @@ def act(self) -> None:
...
+        """
+        Implement the task teardown.
+
+        :return: None
+        """
+        balance = self.context.ledger_apis.token_balance('fetchai', cast(str, self.context.agent_addresses.get('fetchai')))
+        logger.info("[{}]: ending balance on fetchai ledger={}.".format(self.context.agent_name, balance))
+
+
+class MyTransactionBehaviour(Behaviour):
```
DavidMinarsch: Currently, transaction message handling from decision maker is implemented as a behaviour. That is conceptually wrong and I would rather implement it as a handler. However, we first need the `Filter` so we can keep a separation between external and internal message filtering. I will work on the `Filter` tomorrow. Then we can make this a `Handler` which listens to `SUPPORTED_PROTOCOL = TransactionMessage.protocol_id`.

aea/decision_maker/base.py
```diff
@@ -319,36 +323,49 @@ def _handle_tx_message(self, tx_message: TransactionMessage) -> None:
...
+        counterparty_address = cast(str, tx_message.get("counterparty"))
+
+        # adjust payment amount to reflect transaction fee split
+        amount -= counterparty_tx_fee
+        tx_fee = counterparty_tx_fee + sender_tx_fee
+        payable = amount + tx_fee
+
+        # check if the transaction is acceptable and process it accordingly
+        if self._is_acceptable_tx(crypto_object, payable):
+            tx_digest = self._settle_tx(crypto_object, counterparty_address, amount, tx_fee)
```
DavidMinarsch: This is currently blocking. We can keep it that way for now I think. We should address this in the context of the actor model/async discussions.

packages/skills/weather_client_ledger/handlers.py
```diff
@@ -246,15 +231,15 @@ def _handle_inform(self, msg: FIPAMessage, sender: str, message_id: int, dialogu
         :param dialogue: the dialogue object
         :return: None
         """
-        logger.info("[{}]: received INFORM from sender={}".format(self.context.agent_name, sender))
+        logger.info("[{}]: received INFORM from sender={}".format(self.context.agent_name, sender[-5:]))
         json_data = cast(dict, msg.get("json_data"))
         if 'weather_data' in json_data.keys():
             weather_data = json_data['weather_data']
             logger.info("[{}]: received the following weather data={}".format(self.context.agent_name, weather_data))
```
marcofavorito: what about using `pprint.pformat(weather_data)` instead of just `weather_data`?

aea/crypto/default.py
```diff
@@ -280,3 +280,13 @@ def _pvk_obj_to_pem(self, pvk: object) -> bytes:
...
+
+    @staticmethod
+    def get_address_from_public_key(self, public_key: str) -> str:
+        """
+        Get the address from the public key.
+
+        :param public_key: the public key
+        :return: str
+        """
+        pass
```
marcofavorito: shouldn't do it raise `NotImplementedError`?

aea/crypto/ethereum.py
```diff
@@ -104,3 +104,13 @@ def _generate_private_key(self) -> Account:
...
+
+    @staticmethod
+    def get_address_from_public_key(self, public_key: str) -> str:
+        """
+        Get the address from the public key.
+
+        :param public_key: the public key
+        :return: str
+        """
+        pass
```
marcofavorito: as above

aea/crypto/fetchai.py
```diff
@@ -46,7 +46,7 @@ def __init__(self, private_key_path: Optional[str] = None):
         :param ledger_api_config: the ledger api config
```
marcofavorito: remove this line from the docstring.

aea/crypto/fetchai.py
```diff
@@ -46,7 +46,7 @@ def __init__(self, private_key_path: Optional[str] = None):
...
         """
         self._entity = self._generate_private_key() if private_key_path is None else self._load_private_key_from_path(private_key_path)
-        self._address = str(Address(Identity.from_hex(self.public_key)))
+        self._address = Address(Identity.from_hex(self.public_key)).__str__()
```
marcofavorito: I believe this change can be reverted.

aea/crypto/ledger_apis.py
```diff
@@ -99,6 +100,31 @@ def transfer(self, identifier: str, entity: Entity, destination_address: str, am
             tx_digest = None
         return tx_digest
 
+    def is_tx_settled(self, identifier: str, tx_digest: str, amount: float) -> bool:
```
marcofavorito: why `amount: float` and not `amount: int`?

aea/helpers/dialogue/base.py
```diff
@@ -102,6 +121,16 @@ def is_self_initiated(self) -> bool:
         """Check whether the agent initiated the dialogue."""
         return self._is_self_initiated
 
+    @property
+    def last_incoming_message(self) -> Optional[Message]:
+        """Get the last incoming message."""
+        return self._incoming_messages[-1]
```
marcofavorito: what if this is called when `self._incoming_messages` is an empty list? The property should check this corner case.

aea/helpers/dialogue/base.py
```diff
@@ -102,6 +121,16 @@ def is_self_initiated(self) -> bool:
...
+
+    @property
+    def last_outgoing_message(self) -> Optional[Message]:
+        """Get the last incoming message."""
+        return self._outgoing_messages[-1]
```
marcofavorito: same as above

examples/weather_skills/README.md
```diff
@@ -53,10 +62,16 @@ ledger_apis:
     addr: alpha.fetch-ai.com
     port: 80
 ```
+- Generate the private key for the weather client:
```
marcofavorito: there must be a better way to generate private keys... `aea generate_keys [fetch|ethereum|all]`?

examples/weather_skills/README.md
```diff
@@ -53,10 +62,16 @@ ledger_apis:
...
+
+     aea run
```
marcofavorito: mention the `-v DEBUG` flag at some point? So we remind the user (or make him aware) of this option.

examples/weather_skills/README.md
DavidMinarsch: Yes, great point!

examples/weather_skills/README.md
DavidMinarsch: No I think we have this in the main docs. Let's keep things simple and brief.

examples/weather_skills/README.md
Totoual: We had a discussion on that. Why don't we generate the keys on the create. Don't remember the reason why we didn't. Moreover, is it possible to support more than one private/public key per agent ?

Issue #295, opened 2019-10-19
Develop Filter


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
A skill should be able to deregister itself from being called by the Filter. Then let the decision maker terminate the liveness of the agent when there is no non-default skills registered anymore.

**Describe the solution you'd like**
Develop the Filter with the features needed.

**Describe alternatives you've considered**
-

**Additional context**
-

PR #341, opened 2019-10-29
Initial implementation of filter concept

DavidMinarsch:
## Proposed changes

This PR introduces the concept of the `Filter`. The filter is responsible for keeping track of inactive skill components and for routing internal messages.

## Fixes

It addresses #295 

## Types of changes

## Checklist

## Further comments

-

### DIFF
--- a/aea/aea.py
+++ b/aea/aea.py
@@ -27,9 +27,8 @@ from aea.context.base import AgentContext
 from aea.crypto.ledger_apis import LedgerApis
 from aea.crypto.wallet import Wallet
 from aea.decision_maker.base import DecisionMaker
-from aea.decision_maker.messages.transaction import TransactionMessage
 from aea.mail.base import Envelope, MailBox
-from aea.registries.base import Resources
+from aea.registries.base import Filter, Resources
 from aea.skills.error.handlers import ErrorHandler
 
 
@@ -83,6 +82,7 @@ class AEA(Agent):
                                      self.decision_maker.preferences,
                                      self.decision_maker.is_ready_to_pursuit_goals)
         self._resources = None  # type: Optional[Resources]
+        self._filter = None  # type: Optional[Filter]
 
     @property
     def decision_maker(self) -> DecisionMaker:
@@ -100,6 +100,12 @@ class AEA(Agent):
         assert self._resources is not None, "No resources initialized. Call setup."
         return self._resources
 
+    @property
+    def filter(self) -> Filter:
+        """Get filter."""
+        assert self._filter is not None, "No filter initialized. Call setup."
+        return self._filter
+
     def setup(self) -> None:
         """
         Set up the agent.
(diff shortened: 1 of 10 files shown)


aea/decision_maker/base.py
```diff
@@ -41,6 +41,7 @@
 ExchangeParams = Dict[str, float]   # a map from identifier to quantity
 
 QUANTITY_SHIFT = 100
+INTERNAL_PROTOCOL_ID = 'internal'
```
DavidMinarsch: Not sure where to place this constant yet, hence duplicated throughout.

aea/aea.py
```diff
@@ -160,33 +167,25 @@ def handle(self, envelope: Envelope) -> None:
             error_handler.send_invalid_message(envelope)    # pragma: no cover
             return                                          # pragma: no cover
 
-        handlers = self.resources.handler_registry.fetch(protocol.id)
+        handlers = self.filter.get_active_handlers(protocol.id)
         if handlers is None:
             if error_handler is not None:
                 error_handler.send_unsupported_skill(envelope)
```
marcofavorito: should it really be an "unsupported skill" error? Should discern between "inactive" handler and "unsupported" handler?

aea/decision_maker/base.py
```diff
@@ -299,11 +300,9 @@ def execute(self) -> None:
...
         :return: None
         """
-        counter = 0
-        while not self.message_in_queue.empty() and counter < self._max_reactions:
-            counter += 1
+        while not self.message_in_queue.empty():
             message = self.message_in_queue.get_nowait()  # type: Optional[Message]
-            if message is not None:
+            if message is not None and message.protocol_id == INTERNAL_PROTOCOL_ID:
                 self.handle(message)
```
marcofavorito: As a suggestion, I'd put the following code
```python
else:
    logger.warning("Message received by the decision maker is not labelled as 'internal'")
```

or something like that. Might be useful to have for debugging.

aea/protocols/base.py
```diff
@@ -33,6 +33,8 @@
 class Message:
     """This class implements a message."""
 
+    protocol_id = 'base'
```
marcofavorito: does that suggest we could also support the serialization for the `base` protocol? 
E.g. we could recommend using `JSONSerializer` to serialize objects of type `Message`.

aea/aea.py
DavidMinarsch: I am not sure. Basically, this is just telling the other agent: "I cannot deal with this". We probably should keep private to the agent the info why that is (i.e. unsupported or not active skill). So the point I would make is that it should maybe be renamed.

aea/protocols/base.py
DavidMinarsch: No, we don't support that at the moment. But we need to provide a class variable in the base class for typing consistency.

aea/registries/base.py
```diff
@@ -533,3 +546,76 @@ def teardown(self):
...
+    def __init__(self, resources: Resources, decision_maker_out_queue: Queue):
+        """
+        Instantiate the filter.
+
+        :param resources: the resources
+        :param decision_maker_out_queue: the decision maker queue
+        """
+        self._resources = resources
+        self._decision_maker_out_queue = decision_maker_out_queue
+        # TODO: self._inactive_handlers = {}  # type: Dict[SkillId, List[HandlerId]]
```
DavidMinarsch: Note, @marcofavorito how this demonstrates the need for `HandlerId`, `BehaviourId` and `TaskId`

PR #363, opened 2019-10-31
Feature/carpark skills

dishmop:
## Proposed changes

Carpark skills working to the extent that we did the live demo at pydata Meetup

## Fixes

-

## Types of changes

## Checklist

- [x ] I have read the [CONTRIBUTING](../master/CONTRIBUTING.rst) doc
- [ x] I am making a pull request against the `develop` branch (left side). Also you should start your branch off our `develop`.
- [ x] Lint and unit tests pass locally with my changes
- [ x] Any dependent changes have been merged and published in downstream modules


packages/skills/carpark_client/dialogues.py
```diff
@@ -178,7 +179,8 @@ def _next_dialogue_id(self) -> int:
 
         :return: the next id
         """
-        self._dialogue_id += 1
+        self._dialogue_id = random.randint(0, 1000000)
```
DavidMinarsch: I propose we revert this. We will develop a dialogue error handler next week to address the issue properly.

packages/skills/carpark_detection/strategy.py
```diff
@@ -60,10 +61,19 @@ def __init__(self, **kwargs) -> None:
...
+
+        if not os.path.isdir(db_dir):
+            print("WARNING - DATABASE dir does not exist")
+
+        self.db = DetectionDatabase(db_dir, False)
         self.lat = 43
         self.lon = 42
+        self.record_balance(balance)
+
+    def record_balance(self, balance):
```
DavidMinarsch: Why are you keeping track of balance and not just query it from the ledger when needed?

packages/skills/carpark_client/handlers.py
```diff
@@ -218,7 +220,8 @@ def _handle_match_accept(self, msg: FIPAMessage, sender: str, message_id: int, d
                                     sender_tx_fee=0,
                                     counterparty_tx_fee=0,
                                     quantities_by_good_pbk={},
-                                    dialogue_label=dialogue.dialogue_label.json)
+                                    dialogue_label=dialogue.dialogue_label.json,
+                                    ledger_id='fetchai')
```
DavidMinarsch: Have a look at my PR to see how I move all constants to configs or load from aea.

packages/skills/carpark_client/dialogues.py
dishmop: I don't think I agree.

I agree that we should have a "duplicate dialogue id" error handler. However, what do you envisage being the correct process to select a dialogue new id? 

I think the simplest (from a user perspective) is that each time you start a dialogue, you generate a unique id. Ideally this would be a GUID, but a random number will approximate this for the moment.

Then yes, we should have an error handler in case someone creates two dialogues with the same id (which is perhaps indistinguishable from some simply violating the message ordering protocol). The error handler should obviously be tested by the automated test routines (in case you are worried about a bug in it being hidden).

packages/skills/carpark_detection/strategy.py
dishmop: This is not really to keep track of it, but to put it in the database so it can be read and displayed by the GUI.

packages/skills/carpark_client/handlers.py
dishmop: Do you mean things like this (from your latest PR):
    sender_tx_fee=strategy.max_buyer_tx_fee,

And do you envisage this being applied to:
    sender_tx_fee
    counterparty_tx_fee
    ledger_id

and specifically this function or were you talking more generally in the code?

DavidMinarsch: I was referring to `fetchai` constant which I think should be provided via an instance variable in the strategy (or some other class of your choice) and loaded from config. That way it is trivial to also run the carpark agent on other ledgers.

DavidMinarsch: It might indeed help to apply it to `sender_tx_fee`, `counterpartt_tx_fee`, `ledger_id` and `currency_pbk` :)

packages/skills/carpark_client/dialogues.py
DavidMinarsch: The issue is we cannot use a GUID as it is an `int` field on the OEF SDK. We can eventually do the GUID like approach plus dialogue id error handler. The correct process is iteration of the id in my opinion as this gives a straightforward recovery when an error is detected > jump to the `dialogue_id` proposed by the opponent (i.e. the last one he knows from you). Note, the dialogue is not uniquely identified by the `dialogue_id`, for unique identification we have the `dialogue_label`!

DavidMinarsch: Note, Ethereum does something similar with the `nonce` of transactions. You increment the `nonce` for each new transaction.

dishmop: OK - if we are limited by the OEF implementation then I am persuaded by your argument.

packages/skills/carpark_client/handlers.py
dishmop: Hi,

I am not going to have time to fix this prior to 3:30 which is when I need to get things ready by.

Issue #262, opened 2019-10-16
Scaffold does not rename resource in the configuration file.


marcofavorito:
**Is your feature request related to a problem? Please describe.**

When doing `aea scaffold <resource-type> <resource-name>`, the field `name` in the `<resource>.yaml` configuration file is not equal to `<resource-name>` but the default one is used.

E.g. for `<resource_type>` equal to `protocol` and `<resource-name>` equal to `myprotocol`, `protocol.yaml` in `myagent/protocols/myprotocol/` contains the field `name` equal to `scaffold_protocol`.

**Describe the solution you'd like**

Instead of the default name, I'd like the name given as a parameter for the `aea scaffold` command.

In the example, `name` in `protocol.yaml` should be `myprotocol` instead of `scaffold_protocol`.

**Describe alternatives you've considered**

None.

**Additional context**

None.

Issue #320, opened 2019-10-25
'aea list <resource-type>' includes invalid resources.


marcofavorito:
**Describe the bug**

`aea list <resource-type>` (e.g. `aea list protocol`) includes resources that are invalid. E.g. if a protocol folder is invalid because does not contain the `protocol.yaml` file, or the configuration file is misformatted, or other things are missing, it should not be displayed in the list.

It seems that `aea list` checks only the folders inside `<agent-folder>/protocols/`. Probably, what should happen is that the folder gets ignored.

**To Reproduce**
1. `aea create myagent`
2. `cd myagent`
3. remove `./protocols/default/protocol.yaml`
4. `aea list protocols`

**Expected behaviour**

expected output: No output, or maybe a warning message on the invalidity of the `default` protocol.

**Actual Behaviour**

the `aea list protocols` gives:
```
default
```

**Desktop (please complete the following information):**
 - OS: `Ubuntu 18.04.3 LTS`
 - Browser: N/A
 - `aea` Version: `0.1.10`

**Smartphone (please complete the following information):**
N/A

**Additional context**
None.

PR #409, opened 2019-11-12
cli returns descriptions

### LINKED ISSUE #406: Get descriptions of components from CLI
Currently you can get a list of skills, connections and protocols, but cannot get their descriptions.
The GUI needs to be able to display descriptions, so currently it reads the files directly from disc which breaks the encapsulation of the CLI somewhat and makes the system rather brittle.

Here is my proposal:

```
(agents-aea) bash-3.2$ aea list connections 
gym [The gym connection wraps an OpenAI gym.]
oef [The oef connection provides a wrapper around the OEF sdk.]
```

Though could also keep current functionality but have an additional argument to get the descriptions:
```
(agents-aea) bash-3.2$ aea list connections 
gym
oef
```

```
(agents-aea) bash-3.2$ aea list connections -d
gym [The gym connection wraps an OpenAI gym.]
oef [The oef connection provides a wrapper around the OEF sdk.]
```

**Additional context**
I'd like to get this sorted quite quickly as I don't want to write tests for the gui which covers all the directory canning logic if it is just going to be removed. I'm happy to implement the functionality in the current CLI tool - though obvs. Oleg will need to do it in the package manager version.

dishmop:
CLI now returns descriptions as well as the names when you either list of search for protocols, connections or skills. It does it in the format that Oleg has implemented. I have also used the "mock" thing to decouple (most of) the guy code from the CLI. Running an agent still does it for real as I want to test the extracting of stdout and stderr text (it could be mocked, but its non trivial). 

This is now merged with Oleg's code. 

It also fixed a few of existing bugs where:
1. Search was not working when aea framework was installed with  `pip install .`, it only worked with `pip install -e .`
2. Any components with a name containing a 'p' or 'y' was not being displayed in a search (so p2p connection was not showing up'
3. Scaffolding didn't rename the name in the yaml file - its does now

## Proposed changes
This was originally done in order to clean up what CLI GUI code - which we previously parsing the directories and yams files itself to extract the description information. This was the wrong way to do it and rather than write a load of tests for wrong code I thought it best to fix it. 

## Fixes

#406
#262 

## Types of changes

## Checklist


tests/test_cli/test_search.py
```diff
@@ -111,43 +131,37 @@ def setup_class(cls):
         cls.cwd = os.getcwd()
         cls.runner = CliRunner()
 
+    def _generated_expected_output(self):
```
DavidMinarsch: You're redefining the same function here.

tests/test_cli/test_search.py
```diff
@@ -78,19 +88,29 @@ def setup_class(cls):
         cls.cwd = os.getcwd()
         cls.runner = CliRunner()
 
+
+    def _generated_expected_output(self):
```
DavidMinarsch: Also redefining here

aea/cli_gui/__init__.py
```diff
@@ -203,13 +203,16 @@ def _call_aea_async(param_list, dir):
     return ret
 
 
-def start_oef_node(dummy):
+def start_oef_node():
     """Start an OEF node running."""
     _kill_running_oef_nodes()
+    temp_dir = os.getcwd()
+
+    print("cwd = {}".format(temp_dir))
```
DavidMinarsch: Please remove the print statement.

tests/test_cli/test_search.py
dishmop: These are in different functions

aea/cli/common.py
```diff
@@ -140,5 +141,80 @@ def _load_env_file(env_file: str):
     load_dotenv(dotenv_path=Path(env_file), override=False)
 
 
+def format_items_dc(items):
+    """
+    Format information about a protocol, connection or skill in a human readable form.
+
+    This is a placeholder until Oleg's function is merged in.
```
DavidMinarsch: @dishmop please update given that Oleg's work in merged now.

aea/cli/common.py
```diff
@@ -140,5 +141,80 @@ def _load_env_file(env_file: str):
...
+    """Return description of a protocol, skill or connection."""
+    try:
+        config = loader.load(open(str(config_filepath)))
+        if config.name != name:
+            logger.error("Mismatch of config name and folder name. config.name = {}, name = {}".format(config.name, name))
+        assert config.name == name
+        return {"name": config.name, "description": config.description, "version": config.version}
+    except ValidationError as e:
+        logger.error(str(e))
+        return {"name": name, "description": None, "version": None}
```
DavidMinarsch: What case does this cover and why do we not just want to raise the error? If it is a temporary work around for an open issue then please add a # TODO comment so we can find and return to this.

aea/cli/scaffold.py
```diff
@@ -39,135 +39,92 @@ def scaffold(ctx: Context):
...
+        shutil.rmtree(os.path.join(item_type_plural, item_name), ignore_errors=True)
         sys.exit(1)
 
 
+@scaffold.command()
+@click.argument('connection_name', type=str, required=True)
+@pass_ctx
+def connection(ctx: Context, connection_name: str) -> None:
+    """Add a connection scaffolding to the configuration file and agent."""
+    _scaffold_item(
```
DavidMinarsch: Minor comment, you pass all of ctx and its associated attributes. You could also drop the second and third to last arguments and use the `getattr` function to retrieve the correct method (https://docs.python.org/3/library/functions.html#getattr, https://stackoverflow.com/questions/3061/calling-a-function-of-a-module-by-using-its-name-a-string) using the `ctx` and `item_name` as an argument to it.

aea/cli/search.py
```diff
@@ -41,11 +42,28 @@ def search(ctx: Context, registry):
     if registry:
         ctx.set_config("is_registry", True)
     else:
-        registry = os.path.join(AEA_DIR, DEFAULT_REGISTRY_PATH)
+        registry = os.path.join(ctx.cwd, DEFAULT_REGISTRY_PATH)
         ctx.set_config("registry", registry)
         logger.debug("Using registry {}".format(registry))
 
 
+def _is_invalid_item(name, dir_path, config_path):
```
DavidMinarsch: This seems to be addressing #320.

aea/cli/search.py
```diff
@@ -41,11 +42,28 @@ def search(ctx: Context, registry):
...
+    """Return true if this protocol, connection or skill should not be returned in the list."""
+    return ".py" in name or "__" in name or name == "scaffold" or os.path.isfile(dir_path) or not os.path.isfile(config_path)
```
DavidMinarsch: We should also exclude `.DS_Store`, so really any item starting with `.` as we don't consider that valid naming of an item

aea/cli_gui/__init__.py
```diff
@@ -157,19 +151,12 @@ def remove_local_item(agent_id, item_type, item_id):
 
 def get_local_items(agent_id, item_type):
```
DavidMinarsch: @dishmop here and in other places you're missing the typing annotations. Please add them for consistency and readability. Note, in the `cli` we cannot use them as this messes with the `click` module. But here I suspect it is ok, unless you think otherwise?

tests/test_gui/test_run_oef.py
```diff
@@ -0,0 +1,55 @@
...
+"""This test module contains the tests for the `aea gui` sub-commands."""
+# import json
+# import time
+# from .test_base import create_app
+
+
+def test_create_and_run_oef():
+    """Test for running oef, reading TTY and errors."""
+    pass
+    # """Test for running oef, reading TTY and errors."""
```
DavidMinarsch: Why is this being excluded?

aea/cli/common.py
dishmop: done

aea/cli/scaffold.py
dishmop: yes your approach is much nicer (I didn't realise you could do this). I've changed the code accordingly,

aea/cli/search.py
dishmop: Yes true .

aea/cli/search.py
dishmop: .DS_Store is already invalid because it doesn't contains a config file.

aea/cli_gui/__init__.py
dishmop: OK I'll put those in

dishmop: hmm actually I've now put then in the arguments list - but not on the return types as the functions often do return different structures depending on whether they pass or fail.

tests/test_gui/test_run_oef.py
dishmop: I just haven't written it properly yet

PR #434, opened 2019-11-23
Make use of `supported_protocols` field in connection configuration.

### LINKED ISSUE #249: Supported protocols in connections not used
Currently, `supported_protocols` in connections config is not used.

We should use this to only let messages into the connection which satisfy the list.

This could be treated in a wider context by introducing a router between the mailbox layer and the connection layer. This would then prepare the stage for more than one connection to be serving an aea at runtime.
 
**Additional context**
-

marcofavorito:
## Proposed changes

This PR give meaning to `supported_protocols` in the connection configuration files `connection.yaml`.

## Fixes

fixes #249 

## Types of changes

## Checklist

### DIFF
--- a/aea/connections/base.py
+++ b/aea/connections/base.py
@@ -44,27 +44,24 @@ class ConnectionStatus(object):
 class Connection(ABC):
     """Abstract definition of a connection."""
 
-    def __init__(self, connection_id: Optional[str] = None, supported_protocols: Optional[Set[str]] = None):
+    def __init__(self, connection_id: str, restricted_to_protocols: Optional[Set[str]] = None):
         """
         Initialize the connection.
 
         :param connection_id: the connection identifier.
-        :param supported_protocols: the set of protocol ids supported by the connection.
+        :param restricted_to_protocols: the set of protocols ids of the only supported protocols for this connection.
         """
-        self._connection_id = connection_id if connection_id is not None else type(self).__name__
-        self._supported_protocols = self._get_supported_protocols(supported_protocols)
+        self._connection_id = connection_id
+        self._restricted_to_protocols = self._get_restricted_to_protocols(restricted_to_protocols)
 
         self._loop = None  # type: Optional[AbstractEventLoop]
         self._connection_status = ConnectionStatus()
 
-        if len(self.supported_protocols) == 0:
-            logger.warning("No supported protocol for connection '{}'".format(connection_id))
-
-    def _get_supported_protocols(self, supported_protocols: Optional[Set[str]] = None) -> Set[str]:
-        if supported_protocols is not None:
-            return supported_protocols
-        elif hasattr(type(self), "supported_protocols") and isinstance(getattr(type(self), "supported_protocols"), set):
-            return getattr(type(self), "supported_protocols")
+    def _get_restricted_to_protocols(self, restricted_to_protocols: Optional[Set[str]] = None) -> Set[str]:
+        if restricted_to_protocols is not None:
+            return restricted_to_protocols
+        elif hasattr(type(self), "restricted_to_protocols") and isinstance(getattr(type(self), "restricted_to_protocols"), set):
+            return getattr(type(self), "restricted_to_protocols")
         else:
             return set()
 
@@ -90,9 +87,9 @@ class Connection(ABC):
         return self._connection_id
 
     @property
-    def supported_protocols(self) -> Set[str]:
-        """Get the supported protocols."""
-        return self._supported_protocols
+    def restricted_to_protocols(self) -> Set[str]:
+        """Get the restricted to protocols.."""
+        return self._restricted_to_protocols
 
     @property
     def connection_status(self) -> ConnectionStatus:
--- a/aea/connections/base.py
+++ b/aea/connections/base.py
@@ -19,14 +19,18 @@
 # ------------------------------------------------------------------------------
 
 """The base connection package."""
+import logging
 from abc import abstractmethod, ABC
 from asyncio import AbstractEventLoop
-from typing import TYPE_CHECKING, Optional
+from typing import TYPE_CHECKING, Optional, Set
 
 from aea.configurations.base import ConnectionConfig
 
 if TYPE_CHECKING:
     from aea.mail.base import Envelope  # pragma: no cover
+
+
+logger = logging.getLogger(__name__)
 
 
 class ConnectionStatus(object):
@@ -40,16 +44,29 @@ class ConnectionStatus(object):
 class Connection(ABC):
     """Abstract definition of a connection."""
 
-    def __init__(self, connection_id: str):
+    def __init__(self, connection_id: Optional[str] = None, supported_protocols: Optional[Set[str]] = None):
         """
         Initialize the connection.
 
         :param connection_id: the connection identifier.
+        :param supported_protocols: the set of protocol ids supported by the connection.
         """
-        self._connection_id = connection_id
+        self._connection_id = connection_id if connection_id is not None else type(self).__name__
+        self._supported_protocols = self._get_supported_protocols(supported_protocols)
 
         self._loop = None  # type: Optional[AbstractEventLoop]
         self._connection_status = ConnectionStatus()
+
+        if len(self.supported_protocols) == 0:
+            logger.warning("No supported protocol for connection '{}'".format(connection_id))
+
+    def _get_supported_protocols(self, supported_protocols: Optional[Set[str]] = None) -> Set[str]:
+        if supported_protocols is not None:
+            return supported_protocols
+        elif hasattr(type(self), "supported_protocols") and isinstance(getattr(type(self), "supported_protocols"), set):
+            return getattr(type(self), "supported_protocols")
+        else:
+            return set()
 
     @property
     def loop(self) -> Optional[AbstractEventLoop]:
@@ -71,6 +88,11 @@ class Connection(ABC):
     def connection_id(self) -> str:
         """Get the id of the connection."""
         return self._connection_id
+
+    @property
+    def supported_protocols(self) -> Set[str]:
+        """Get the supported protocols."""
+        return self._supported_protocols
 
     @property
     def connection_status(self) -> ConnectionStatus:
--- a/aea/connections/local/connection.py
+++ b/aea/connections/local/connection.py
@@ -53,7 +53,7 @@ class LocalNode:
         self._loop = loop if loop is not None else asyncio.new_event_loop()
         self._thread = Thread(target=self._run_loop)
 
-        self._in_queue = asyncio.Queue(loop=self._loop)  # type: Optional[asyncio.Queue]
+        self._in_queue = asyncio.Queue(loop=self._loop)  # type: asyncio.Queue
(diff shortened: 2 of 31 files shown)


aea/connections/base.py
```diff
@@ -40,17 +44,30 @@ def __init__(self):
 class Connection(ABC):
     """Abstract definition of a connection."""
 
-    def __init__(self, connection_id: str):
+    def __init__(self, connection_id: Optional[str] = None, supported_protocols: Optional[Set[str]] = None):
```
marcofavorito: now the base class `Connection` has an optional parameter `supported_protocols`.
That means both at the code level and configuration level the user can change the behaviour of the outgoing messages by ignoring the envelopes belonging to some protocols.

aea/connections/base.py
```diff
@@ -40,17 +44,30 @@ def __init__(self):
...
+        self._connection_id = connection_id if connection_id is not None else type(self).__name__
+        self._supported_protocols = self._get_supported_protocols(supported_protocols)
 
         self._loop = None  # type: Optional[AbstractEventLoop]
         self._connection_status = ConnectionStatus()
 
+        if len(self.supported_protocols) == 0:
+            logger.warning("No supported protocol for connection '{}'".format(connection_id))
+
+    def _get_supported_protocols(self, supported_protocols: Optional[Set[str]] = None) -> Set[str]:
```
marcofavorito: It works as follows:
- if the user specifies `supported_protocols` in the constructor, set the supported protocols to that value
- else, if the connection class has an attribute `supported_protocols` (some examples below) and it is a `set`, then use it
- else, return the empty set. A warning is logged to notify the user of this corner case.

aea/connections/local/connection.py
```diff
@@ -53,7 +53,7 @@ def __init__(self, loop: AbstractEventLoop = None):
         self._loop = loop if loop is not None else asyncio.new_event_loop()
         self._thread = Thread(target=self._run_loop)
 
-        self._in_queue = asyncio.Queue(loop=self._loop)  # type: asyncio.Queue
+        self._in_queue = None  # type: Optional[asyncio.Queue]
```
marcofavorito: This change is due to the deprecation of `loop` parameter of `asyncio.Queue` from Python 3.8:
https://docs.python.org/3/library/asyncio-queue.html#asyncio.Queue

that implies the asyncio queue must be initialized when the right loop has been set - in this context, in `_run_loop`

aea/connections/local/connection.py
```diff
@@ -322,14 +323,17 @@ class OEFLocalConnection(Connection):
     It is useful for local testing.
     """
 
-    def __init__(self, public_key: str, local_node: LocalNode, connection_id: str = "local"):
+    supported_protocols = {"default", "oef", "fipa", "tac"}
```
marcofavorito: I feel this approach has a problem of "code scalability": let's say we have hundreds of protocols, and that the connections can support them. We should update constantly this variable (the same holds for its default configuration file).

aea/connections/local/connection.py
```diff
@@ -344,7 +348,7 @@ def public_key(self) -> str:
     async def connect(self) -> None:
         """Connect to the local OEF Node."""
         if not self.connection_status.is_connected:
-            self._reader = Queue(loop=self._loop)
+            self._reader = Queue()
```
marcofavorito: see above

aea/connections/base.py
marcofavorito: also, worth noticing that this adds a burden to the developer of a new connection.

aea/mail/base.py
```diff
@@ -239,7 +239,7 @@ def __init__(self, connections: List['Connection'],
         self._thread = Thread(target=self._run_loop)
 
         self._in_queue = queue.Queue()  # type: queue.Queue
-        self._out_queue = asyncio.Queue(loop=self._loop)  # type: asyncio.Queue
+        self._out_queue = None  # type: Optional[asyncio.Queue]
```
marcofavorito: see above

aea/mail/base.py
```diff
@@ -252,6 +252,7 @@ def in_queue(self) -> queue.Queue:
     @property
     def out_queue(self) -> asyncio.Queue:
         """Get the out queue."""
+        assert self._out_queue is not None, "Accessing out queue before loop is started."
```
marcofavorito: this is needed because we cannot set the queue in the constructor.

aea/mail/base.py
```diff
@@ -472,6 +474,11 @@ def _stop(self):
             logger.debug("Using default connection: {}".format(self.default_connection))
             connection = self.default_connection
 
+        if envelope.protocol_id not in connection.supported_protocols:
+            logger.warning("Connection {} does not support protocol {}. Cannot send the message."
+                           .format(connection.connection_id, envelope.protocol_id))
+            return
```
marcofavorito: Observe that, in case the protocol id of the outgoing envelope is not supported  by the chosen connection, we just log a warning message. That might not be always ideal.

tests/test_aea.py
```diff
@@ -166,7 +166,7 @@ def test_react():
             dummy_handler = dummy_skill.handlers[0]
 
             expected_envelope = envelope
-            agent.outbox.put(expected_envelope)
+            agent._multiplexer.in_queue.put(expected_envelope)
```
marcofavorito: to make the test succeed, in order to handle an envelope with a not supported protocol, we can't rely on `outbox.put` because that would be filtered out by the multiplexer.

tests/test_connections/test_local/test_misc.py
```diff
@@ -151,11 +151,11 @@ def test_communication():
 @pytest.mark.asyncio
 async def test_connecting_to_node_with_same_key():
     """Test that connecting twice with the same key works correctly."""
-    node = LocalNode()
-    public_key = "my_public_key"
-    my_queue = asyncio.Queue()
+    with LocalNode() as node:
```
marcofavorito: now we require the local node started for every operation (including `connect`)

tests/test_decision_maker/test_base.py
```diff
@@ -265,6 +266,7 @@ def setup_class(cls):
         cls.preferences = Preferences()
         cls.decision_maker = DecisionMaker(agent_name=cls.agent_name, max_reactions=MAX_REACTIONS, outbox=cls.outbox,
                                            wallet=cls.wallet, ledger_apis=cls.ledger_apis)
+        cls.multiplexer.connect()
```
marcofavorito: again, this is needed because now the `out_queue` (called below by `outbox.empty()`) is not initialized unless the multiplexer set up the connections.

tests/test_multiplexer.py
```diff
@@ -276,3 +279,68 @@ def test_get_from_multiplexer_when_empty():
 
     with pytest.raises(aea.mail.base.Empty):
         multiplexer.get()
+
+
+def test_multiple_connection():
```
marcofavorito: that is not strictly relevant to the PR

aea/connections/local/connection.py
DavidMinarsch: Yes good point. An alternative could be to support by default any protocol (i.e. when `{}` is passed to supported_protocols) and then only if supported_protocols is specified different from `{}` will the filtering kick in.

tests/test_multiplexer.py
DavidMinarsch: nice test!

tests/test_decision_maker/test_base.py
DavidMinarsch: hmm ok

tests/test_aea.py
DavidMinarsch: Ok so why don't you add the correct list of supported protocols?

aea/mail/base.py
DavidMinarsch: Ok, for now that is fine.

DavidMinarsch: Alternative is to raise an error?

aea/mail/base.py
DavidMinarsch: OK

aea/mail/base.py
DavidMinarsch: ok

aea/connections/local/connection.py
DavidMinarsch: Maybe this is the right approach anyway, so relabel `supported_protocols` to `restricted_to_protocols` and this is an optional param. If I pass nothing then no filtering takes place.

aea/connections/local/connection.py
DavidMinarsch: Ok. How did this not cause issues earlier?

aea/connections/base.py
DavidMinarsch: See the comments below re `restricted_to_protocols`

aea/connections/base.py
DavidMinarsch: See comments below. Valid points and I think I have an answer. Re making `connection_id` optional I am not sure. I could see this causing more confusion than being helpful.

aea/connections/local/connection.py
marcofavorito: it was just a warning when using Python 3.8 (which has been ignored), it does not affect functionalities. But it will when the `loop` parameter will be removed in Python 3.10.

aea/mail/base.py
marcofavorito: An alternative is to let the user specifying a callback to handle this case.

tests/test_aea.py
marcofavorito: I should add `"unknown_protocol"` to the list of supported protocols of `OEFLocalConnection` which is not super-neat.
Anyway, with the change to `restricted_to_protocols` this problem will be solved.

DavidMinarsch: @marcofavorito can you solve this now?

tests/data/dummy_aea/connections/local/connection.py
```diff
@@ -53,7 +53,7 @@ def __init__(self, loop: AbstractEventLoop = None):
         self._loop = loop if loop is not None else asyncio.new_event_loop()
         self._thread = Thread(target=self._run_loop)
 
-        self._in_queue = asyncio.Queue(loop=self._loop)  # type: asyncio.Queue
+        self._in_queue = asyncio.Queue(loop=self._loop)  # type: Optional[asyncio.Queue]
```
DavidMinarsch: @marcofavorito did you intend it this way? it is not optional now ..

aea/connections/local/connection.py
```diff
@@ -53,7 +53,7 @@ def __init__(self, loop: AbstractEventLoop = None):
         self._loop = loop if loop is not None else asyncio.new_event_loop()
         self._thread = Thread(target=self._run_loop)
 
-        self._in_queue = asyncio.Queue(loop=self._loop)  # type: asyncio.Queue
+        self._in_queue = asyncio.Queue(loop=self._loop)  # type: Optional[asyncio.Queue]
```
DavidMinarsch: @marcofavorito did you intend it this way? it is not optional now ..

PR #438, opened 2019-11-25
[WiP] Feature/cli registry add

panasevychol:
## Proposed changes

Add items from Registry via CLI implemented

## Types of changes

## Checklist


aea/cli/add.py
```diff
@@ -87,6 +77,40 @@ def connection(click_context, connection_name):
         logger.error(str(e))
         sys.exit(1)
 
+
+@add.command()
+@click.argument('connection_name', type=str, required=False)
```
DavidMinarsch: @panasevychol can we remove the `--id` arg and instead assume that it is passed as `connection_name`. So if the registry flag is used the `connection_name` must be a public identifier (otherwise abort) and otherwise it can be as is. This simplifies the cli and avoids us having to change the api in the future.

aea/cli/registry/utils.py
```diff
@@ -36,7 +38,116 @@ def request_api(method, path, params=None):
...
+        raise Exception('Unknown file type: {}'.format(source))
+
+    os.remove(source)
+
+    folder_name = os.path.basename(source).replace('.tar.gz', '')
+    extracted_folder = os.path.join(target, folder_name)
+    return extracted_folder
+
+
+def fetch(obj_type, public_id, cwd):
```
DavidMinarsch: would rename to `fetch_package`

aea/cli/registry/utils.py
```diff
@@ -36,7 +38,116 @@ def request_api(method, path, params=None):
...
+    """Fetch connection/protocol/skill from Registry.
+
+    Parameters
```
DavidMinarsch: please stick to our docstring convention

aea/cli/add.py
panasevychol: The idea of having it separately was for understanding that those 2 params are completely different. I can unify that to one but that would require a bit of new logic in function. Cause it uses name in both cases (for existing connection check).

DavidMinarsch: Well, eventually we will just have the public identifier. So it makes sense to merge them now. The logic will be minimal ..if registry, apply the string split, otherwise use what is given..

aea/cli/registry/utils.py
```diff
@@ -36,7 +38,78 @@ def request_api(method, path, params=None):
...
+        raise Exception('Unknown file type: {}'.format(source))
+
+    os.remove(source)
+
+    folder_name = os.path.basename(source).replace('.tar.gz', '')
+    extracted_folder = os.path.join(target, folder_name)
+    return extracted_folder
+
+
+def fetch_package(obj_type, public_id, cwd):
```
DavidMinarsch: Alle these methods are missing docstrings listing the params and the params have no type annotations.

Issue #417, opened 2019-11-18
Test coverage CLI 100%


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
Some files in CLI have no coverage

**Describe the solution you'd like**
Add tests so we cover 100%

PR #448, opened 2019-11-27
Feature/cli tests

dishmop:
## Proposed changes

All cli tests now at 100% test coverage (excpet the registry ones which I am assuming Oleg will do).

## Fixes

- #417

## Types of changes

## Checklist


aea/cli/common.py
```diff
@@ -183,8 +183,6 @@ def format_skills(items):
 def retrieve_details(name: str, loader: ConfigLoader, config_filepath: str):
     """Return description of a protocol, skill or connection."""
     config = loader.load(open(str(config_filepath)))
-    if config.name != name:
```
DavidMinarsch: Would this check plus a `raise AssertionError` not be preferable to the assertion? Then the user knows what's going on.

tests/test_cli/test_run.py
```diff
@@ -86,6 +87,442 @@ def test_run(pytestconfig):
...
+
+    whole_file = whole_file.replace(find_text, replace_text)
+
+    # close the file
+    file.close()
+
+    with open("aea-config.yaml", 'w') as f:
+        f.write(whole_file)
+
+    with open("fet_private_key.txt", 'w') as f:
```
DavidMinarsch: Why are you creating the private keys? If it is to test the fact that they should already exist when `run` is called then please add a comment to make clear this is the case.

tests/test_cli/test_run.py
```diff
@@ -86,6 +87,442 @@ def test_run(pytestconfig):
...
```
DavidMinarsch: Same as below

aea/cli/common.py
dishmop: I remove the if because I didn't want to write a test for that path. Also I don't feel this is the place to report an error like this back to "user land". We should have a "validity checker" for skills/connections/protocols which is run on skills as they are added to the registry and when the agent is run and report back to the user there.

tests/test_cli/test_run.py
dishmop: done - and I could remove some of them

PR #459, opened 2019-11-29
Remove dependency on OEF FIPA API, introduce dialogue reference inste…

### LINKED ISSUE #342: Duplicate Dialogue ID crash
**Describe the bug**
The weather station crashes if you start a dialogue with it and then start another dialog with the same dialogue ID. This is likely to also happen with other agents ad probably requires a system-wide solution.

**To Reproduce**
Run the weather station Ledger agent
Run the weather station client ledger agent (until it requests some data)
Shut down the client and re-run it

The weather station agent will crash.

**Expected behavior**
Weather station should return some kind of error "You have already used this ID". This should be part of the dialogue management.

DavidMinarsch:
…ad of dialogue id

## Proposed changes

This PR:
- removes the dependency on the OEF SDK's FIPA API. The OEF channel will henceforth only accept CFPs from that API and drop all other messages. Messages are no longer sent via the FIPA API, rather via the standard `send_message` API.
- introduces a `dialogue_reference` which replaces the `dialogue_id` in the `FIPAMessage` and comes in the form of a `Tuple[str,str]`. The first slot is the `dialogue_starter_reference`, the second slot is the `dialogue_responder_reference`. A dialogue is initiated with `dialogue_reference=('any_string', '')`. The responder then populates the second field, returning a message with `dialogue_reference=('any_string', 'any_other_string'). This ensures that the dialogue can be uniquely identified by each participant and clashes can be avoided.
- all the changes necessary regarding the above.

## Fixes

It fixes #342 

## Types of changes

## Checklist

## Further comments

-


aea/connections/oef/connection.py
```diff
@@ -377,58 +353,18 @@ def send(self, envelope: Envelope) -> None:
         :param envelope: the message.
         :return: None
         """
-        if envelope.protocol_id == "default":
-            self.send_default_message(envelope)
-        elif envelope.protocol_id == "fipa":
-            self.send_fipa_message(envelope)
+        if envelope.protocol_id == "gym":
```
DavidMinarsch: @marcofavorito maybe the connections should also have an `excluded_protocols` array in the configs?

aea/protocols/fipa/dialogues.py
```diff
@@ -201,16 +216,26 @@ def is_belonging_to_registered_dialogue(self, fipa_msg: Message, sender: Address
         :return: boolean indicating whether the message belongs to a registered dialogue
         """
         fipa_msg = cast(FIPAMessage, fipa_msg)
-        dialogue_id = cast(int, fipa_msg.get("dialogue_id"))
-        self_initiated_dialogue_label = DialogueLabel(dialogue_id, sender, agent_pbk)
-        other_initiated_dialogue_label = DialogueLabel(dialogue_id, sender, sender)
+        dialogue_reference = cast(Tuple[str, str], fipa_msg.get("dialogue_reference"))
```
DavidMinarsch: This method could be cleaned up eventually!

aea/connections/oef/connection.py
marcofavorito: yes that makes sense! Do you mind opening an issue?

aea/protocols/fipa/dialogues.py
marcofavorito: Yes, for the moment maybe some comments in the docstring or the code can help

aea/protocols/fipa/dialogues.py
```diff
@@ -263,7 +290,9 @@ def create_opponent_initiated(self, dialogue_opponent_pbk: Address, dialogue_id:
 
         :return: the created dialogue
         """
-        dialogue_label = DialogueLabel(dialogue_id, dialogue_opponent_pbk, dialogue_opponent_pbk)
+        assert dialogue_reference[0] != '' and dialogue_reference[1] == '', "Cannot initiate dialogue with preassigned dialogue_responder_reference!"
```
marcofavorito: That is (one of the) potentially harmful assertion, because a malicious opponent can easily make the skill fail.

Alternative is raise an exception that must be handled (or ignored) by the caller function.

marcofavorito: It is OK to leave as it is for now. We could add a `TODO` to don't forget that.

DavidMinarsch: Good point, the fix is a bit more involved so I will merge now and open an issue!

PR #482, opened 2019-12-02
Docs now have walkthrough of building an AEA with code api

dishmop:
## Proposed changes

Done a walkthrough - also fixed an issue where stub connection failed to read message if it was on the last line of the file without a carriage return

## Fixes

-

## Types of changes

## Checklist

### DIFF
--- a/docs/hacking-an-agent.md
+++ b/docs/hacking-an-agent.md
@@ -1,14 +1,21 @@
 ## Preliminaries
 
-These instructions detail the Python code you need for running an AEA outside the `cli` tool.
+These instructions detail the Python code you need for running an AEA outside the `cli` tool, using the code interface. 
 
-<div class="admonition note">
-  <p class="admonition-title">Note</p>
-  <p>You have already coded up your agent. See the <a href="../aea/skill-guide/" target=_blank>build your own skill guide</a> for a reminder.</p>
-</div>
+  
+This guide assumes you have already followed the [quick start](quickstart.md) guide and so have the framework installed and the packages and scripts directory downloaded into the directory you are working in.
 
+
+## Create private key
+Before we make our agent we make a private key to be used in the wallet. Open a terminal and type:
+    
+``` bash
+python scripts/generate_private_key.py my_key.txt
+```
 
 ## Imports
+
+Now we can start to write our python script. 
 
 First, import the necessary common Python libraries and classes.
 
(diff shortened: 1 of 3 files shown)


dishmop:
I haven't managed to run the tests locally as they keep stalling.

docs/hacking-an-agent.md
```diff
@@ -1,99 +1,222 @@
...
 
-These instructions detail the Python code you need for running an AEA outside the `cli` tool.
+These instructions detail the Python code you need for running an AEA outside the `cli` tool, using the code interface. 
 
-<div class="admonition note">
-  <p class="admonition-title">Note</p>
-  <p>You have already coded up your agent. See the <a href="../aea/skill-guide/" target=_blank>build your own skill guide</a> for a reminder.</p>
-</div>
+  
+This guide assumes you have already followed the [quick start](quickstart.md) guide and so have the framework installed and the packages and scripts directory downloaded into the directory you are working in.
```
DavidMinarsch: I would say they need to specifically follow the `Preliminaries` and `Installation` sections of the quickstart.

docs/hacking-an-agent.md
```diff
@@ -1,99 +1,222 @@
...
+Set up a variable pointing to where the packages directory is located - this should be our current directory. 
+``` python
+root_dir = "./"
+```
 
+## Clearing the input and output files
+We will use the stub connection to pass envelopes in and out of the agent. Ensure that any input and output text files are removed before we start.
+``` python
+# Ensure the input and output files do not exist initially
+input_filename = "input.txt"
```
DavidMinarsch: maybe it is easier to do that via command line, like `rm input.txt && touch input.txt` etc?

docs/hacking-an-agent.md
```diff
@@ -1,99 +1,222 @@
...
+t = Thread(target=my_agent.start)
+t.start()
+
+# Wait for everything to start up
+time.sleep(4)
 ```
 
-Create the resources pointing to the working directory.
+## Send and receive an envelope
+We use the input and output text files to send an envelope to our agent and recieve a response (from the echo skill)
```
DavidMinarsch: `receive`

docs/hacking-an-agent.md
```diff
@@ -1,99 +1,222 @@
...
 ``` python
-my_agent = AEA("my_agent", stub_connection, wallet, ledger_apis, resources)
+# Shut down the agent
+my_agent.stop()
+t.join()
 ```
 
-## Add skills and protocols
+## Running the agent
+If you now run this pythong script file, you should see this output:
```
DavidMinarsch: `Python`

docs/hacking-an-agent.md
```diff
@@ -1,99 +1,222 @@
...
 
-We can add the echo skill as follows...
+    No protocol found.
```
DavidMinarsch: Hmm..what causes these log messages? I see it is caused by line 150 in `aea.registries.base`. We should not raise this warning in this case. @marcofavorito should we just change the logger level or come up with a more elaborate solution?

docs/hacking-an-agent.md
dishmop: Yes, this might be better though I would like to leave this as is so I can wrap this up (your proposed change requires, me to do code changes and re-testing for development and stranded installs etc and I think the current setup is basically fine).

docs/hacking-an-agent.md
dishmop: agreed

docs/hacking-an-agent.md
dishmop: agreed

docs/hacking-an-agent.md
dishmop: agreed

PR #483, opened 2019-12-03
Added ability to specify host when starting GUI

8ball030:
## Proposed changes

Added in a arg for the cli tool to serve from a specified host.

This allows the gui to be run within a docker container.

(flask when dockerised, default will NOT allow connections from outside the docker unless you specify a host of "0.0.0.0" which is why i have made this pull request :)
## Types of changes

## Checklist


aea/cli_gui/__init__.py
```diff
@@ -446,12 +446,12 @@ def favicon():
     return app
 
 
-def run(port: int):
```
DavidMinarsch: @8ball030 please set a default arg on host `127.0.0.1` for backwards compatibility

8ball030: This should do it! :)

PR #490, opened 2019-12-08
Minor test additions

DavidMinarsch:
## Proposed changes

Adds missing tests.

## Types of changes

## Checklist

### DIFF
--- a/aea/cli/common.py
+++ b/aea/cli/common.py
@@ -218,5 +218,5 @@ class ConnectionsOption(click.Option):
 
             connection_names = set(arg_strip(s) for s in value.split(",") if arg_strip(s) != "")
             return list(connection_names)
-        except Exception:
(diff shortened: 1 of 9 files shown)


aea/cli/common.py
```diff
@@ -218,5 +218,5 @@ def arg_strip(s):
 
             connection_names = set(arg_strip(s) for s in value.split(",") if arg_strip(s) != "")
             return list(connection_names)
-        except Exception:
-            raise click.BadParameter(value)
+        except Exception:  # pragma: no cover
+            raise click.BadParameter(value)  # pragma: no cover
```
Totoual: If you use only #pragma: no cover on the except you don't need to add it in the next row. (It will help to keep the files cleaner I think :D )

DavidMinarsch: Nice, thanks :)

Issue #214, opened 2019-10-13
Make TAC work end to end on AEA


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
Currently, we have `fipa_negotiation` and `tac` skill. However, they don't work yet with the TAC setup.

**Describe the solution you'd like**
Finish the TAC example end to end.

**Describe alternatives you've considered**
-

**Additional context**
-

PR #491, opened 2019-12-09
Refactor the fipa protocol to include inform component

DavidMinarsch:
## Proposed changes

This PR:
- refactors fipa to include `W_INFORM` rather than `W_ADDRESS`. This makes it more versatile and is needed for TAC.

## Fixes

Related to #214 

## Types of changes

## Checklist

## Further comments

-

### DIFF
--- a/aea/protocols/fipa/message.py
+++ b/aea/protocols/fipa/message.py
@@ -43,8 +43,8 @@ class FIPAMessage(Message):
         MATCH_ACCEPT = "match_accept"
         DECLINE = "decline"
         INFORM = "inform"
-        ACCEPT_W_ADDRESS = "accept_w_address"
-        MATCH_ACCEPT_W_ADDRESS = "match_accept_w_address"
+        ACCEPT_W_INFORM = "accept_w_inform"
+        MATCH_ACCEPT_W_INFORM = "match_accept_w_inform"
 
         def __str__(self):
             """Get string representation."""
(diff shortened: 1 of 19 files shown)


aea/protocols/fipa/message.py
```diff
@@ -43,8 +43,8 @@ class Performative(Enum):
         MATCH_ACCEPT = "match_accept"
         DECLINE = "decline"
         INFORM = "inform"
-        ACCEPT_W_ADDRESS = "accept_w_address"
-        MATCH_ACCEPT_W_ADDRESS = "match_accept_w_address"
+        ACCEPT_W_INFORM = "accept_w_inform"
+        MATCH_ACCEPT_W_INFORM = "match_accept_w_inform"
```
marcofavorito: Why an `ACCEPT` or `MATCH_ACCEPT` message can't have an optional `inform` instead of creating a new performative for that?

DavidMinarsch: @marcofavorito it can but it makes the definition of `ACCEPT` less tight. I think it is good to avoid optional fields were possible with performatives as it in practice means more conditional checks. I would prefer to even get rid of the `info` dict and have it fully specified by performative, but that would be excessive.

tests/test_connections/test_oef/test_communication.py
```diff
@@ -407,8 +407,8 @@ def test_accept_w_address(self):
         accept_w_address = FIPAMessage(message_id=0,
                                        dialogue_reference=(str(0), ''),
                                        target=0,
-                                       performative=FIPAMessage.Performative.ACCEPT_W_ADDRESS,
```
marcofavorito: please change name of the test method above.

PR #501, opened 2019-12-10
Feature/composite behaviours

DavidMinarsch:
## Proposed changes

Composite behaviours added

## Fixes

-

## Types of changes

## Checklist

## Further comments

-

### DIFF
--- a/aea/skills/behaviours.py
+++ b/aea/skills/behaviours.py
@@ -176,7 +176,7 @@ class State(SimpleBehaviour, ABC):
         self.next_state = state_name
 
 
-class FMSBehaviour(CompositeBehaviour):
+class FSMBehaviour(CompositeBehaviour):
     """This class implements a finite-state machine behaviour."""
 
     def __init__(self, **kwargs):
@@ -205,6 +205,7 @@ class FMSBehaviour(CompositeBehaviour):
         self.name2state[name] = state
         if initial:
             self._initial_state = name
+            self.current = self._initial_state
 
     @property
     def initial_state(self) -> Optional[str]:
@@ -218,9 +219,9 @@ class FMSBehaviour(CompositeBehaviour):
             raise ValueError("Name is not registered as state.")
         self._initial_state = value
 
-    def get_state(self, name) -> State:
+    def get_state(self, name) -> Optional[State]:
         """Get a state from its name."""
-        return self.name2state[name]
+        return self.name2state.get(name, None)
 
     def reset(self):
         """Reset the behaviour to its initial conditions."""
@@ -229,13 +230,16 @@ class FMSBehaviour(CompositeBehaviour):
     def act(self):
         """Implement the behaviour."""
         if self.current is None:
-            if self.initial_state is None:
-                return
-            else:
-                self.current = self.initial_state
+            return
 
         current_state = self.get_state(self.current)
+        if current_state is None:
+            return
         current_state.act_wrapper()
 
         if current_state.done():
             self.current = current_state.next_state
+
+    def done(self) -> bool:
+        """Return True if the behaviour is terminated, False otherwise."""
+        return self.current is None
--- a/aea/skills/behaviours.py
+++ b/aea/skills/behaviours.py
@@ -19,7 +19,7 @@
 """This module contains the classes for specific behaviours."""
 import datetime
 from abc import ABC
-from typing import Optional
+from typing import Optional, List, Dict
 
 from aea.skills.base import Behaviour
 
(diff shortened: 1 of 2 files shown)


aea/skills/behaviours.py
```diff
@@ -42,11 +42,14 @@ def __init__(self, **kwargs):
 
     def act_wrapper(self) -> None:
         """Wrap the call of the action. This method must be called only by the framework."""
-        self.act()
-        self._number_of_executions += 1
+        if self.done():
```
DavidMinarsch: You mean `if not self.done()`?

aea/skills/behaviours.py
```diff
@@ -115,3 +118,128 @@ def is_time_to_act(self) -> bool:
...
+        if not self.done() and not self.current_behaviour.done():
+            self.current_behaviour.act_wrapper()
+
+    def done(self) -> bool:
+        """Return True if the behaviour is terminated, False otherwise."""
+        return self._index >= len(self._behaviour_sequence)
+
+
+class State(SimpleBehaviour, ABC):
+    """A state of a FSMBehaviour is a OneShotBehaviour"""
```
DavidMinarsch: So should it inherit from OneShotBehaviour then?

aea/skills/behaviours.py
```diff
@@ -115,3 +118,128 @@ def is_time_to_act(self) -> bool:
...
+
+
+class FSMBehaviour(CompositeBehaviour):
+    """This class implements a finite-state machine behaviour."""
+
+    def __init__(self, **kwargs):
+        """Initialize the finite-state machine behaviour."""
+        super().__init__(**kwargs)
+
+        self.name2state = {}  # type: Dict[str, State]
```
DavidMinarsch: we need to adopt some conventions on dict naming, so far we use `state_by_name`.. not sure thats ideal either

aea/skills/behaviours.py
```diff
@@ -115,3 +118,128 @@ def is_time_to_act(self) -> bool:
...
+    def initial_state(self) -> Optional[str]:
+        """Get the initial state name."""
+        return self._initial_state
+
+    @initial_state.setter
+    def initial_state(self, value: str):
+        """Set the initial state."""
+        if value not in self.name2state:
+            raise ValueError("Name is not registered as state.")
+        self._initial_state = value
```
DavidMinarsch: call it `name` rather than `value`?
also, should it have line `208` as well?

aea/skills/behaviours.py
```diff
@@ -115,3 +118,128 @@ def is_time_to_act(self) -> bool:
...
+
+
+class State(SimpleBehaviour, ABC):
+    """A state of a FSMBehaviour is a OneShotBehaviour"""
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.next_state = None
+
+    def set_next_state(self, state_name: str):
```
DavidMinarsch: hmm what if the next state is dependent on behaviour execution? I guess then the user needs to extend `set_next_state`?

tests/test_skills/test_behaviours.py
```diff
@@ -0,0 +1,118 @@
...
+        def setup(self) -> None:
+            pass
+
+        def teardown(self) -> None:
+            pass
+
+        def act(self) -> None:
+            outputs.append(self.id)
+
+    # TODO let the initialization of a behaviour action from constructor
```
DavidMinarsch: what do you mean by this sentence?

tests/test_skills/test_behaviours.py
```diff
@@ -0,0 +1,118 @@
...
+
+        def setup(self) -> None:
+            pass
+
+        def teardown(self) -> None:
+            pass
+
+        def act(self) -> None:
+            outputs.append(self.id)
+            self.set_next_state(str(self.id + 1))
```
DavidMinarsch: I think a transition table approach like done in JADE might be better as it aids clarity and allows for implementation of complex conditional transition logic more easily.

aea/skills/behaviours.py
marcofavorito: yes, thank you!

aea/skills/behaviours.py
marcofavorito: Not necessarily. For instance, it might be a `CyclicBehaviour`. The docstring is not correct. However, it is a good idea to restrict it to `OneShotBehaviour` as you suggested.

aea/skills/behaviours.py
marcofavorito: True.  Probably `X_to_Y` is a good compromise. Using the `2` instead of `to` might not be clear enough.

aea/skills/behaviours.py
marcofavorito: `set_next_state` is supposed to be called by the user at some point. For instance. from the other methods of the behaviour like `act` or `setup`.

tests/test_skills/test_behaviours.py
marcofavorito: The fact that I had to create a class ad-hoc suggest that we can support a constructor that takes the `setup`, `act` and `teardown` callables. In other words, having the ['Command'](https://en.wikipedia.org/wiki/Command_pattern) pattern (i.e. our behaviour) not only specifiable by _inheritance_but also by _composition_.

tests/test_skills/test_behaviours.py
marcofavorito: Agreed. Or, automatically generated by a protocol specification.

Issue #494, opened 2019-12-09
Global context for skills to access


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
Currently, skills cannot share information other than via the decision maker.

**Describe the solution you'd like**
Introduce a global context so different skills can write to it and read from it to share info. It goes a bit against the design principle of actor like setup but then we have the global (read only) context (agent context) anyway.

This needs to be thread-safe and a simple key value store suffices.

Issue #471, opened 2019-12-01
Finish tac skills 


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
The TAC skills are not fully finished

**Describe the solution you'd like**
Finish them

PR #524, opened 2019-12-17
AEA-223 AEA-192 Feature/tac

### LINKED ISSUE #468: Callback for transaction request

Currently, to process a response from the decision-maker, it is necessary the implementation of a `TransactionHandler`.

A more lightweight approach. E.g. an implicit transaction handler and/or a way to register callbacks to be called when the decision-maker sends back the response.

e.g.:
```python

def on_transaction_response(tx_message):
    # handle transaction response

tx_request = TransactionMessage(...)
decision_maker.submit(tx_request, callback=on_transaction_response)
```

**Additional context**

DavidMinarsch:
## Proposed changes

This PR:
- improves naming conventions around currency > `currency_id` is used throughout
- refactors the `TransactionMessage` and the `decision_maker.py` to be compatible for both message signing and ledger transfers
- fixes miscellaneous issues noticed when doing the above

## Fixes

#468 
#471 
#214
#494

## Types of changes

## Checklist

## Further comments

-


aea/decision_maker/base.py
```diff
@@ -110,42 +111,45 @@ def check_transaction_is_consistent(self, tx_message: TransactionMessage) -> boo
...
-            # check if we have the money to cover amount and tx fee.
-            result = self.amount_by_currency[currency_id] >= tx_message.amount + tx_message.sender_tx_fee
-        else:
-            # check if we have the goods.
-            result = True
-            for good_id, quantity in tx_message.quantities_by_good_id.items():
-                result = result and (self.quantities_by_good_id[good_id] >= quantity)
-            # check if we have the money to cover tx fee.
-            result = self.amount_by_currency[currency_id] + tx_message.amount >= tx_message.sender_tx_fee
+        result = len(tx_message.tx_amount_by_currency_id) == 1
```
marcofavorito: what does that mean? Do we only support one currency?

If yes, I think it's going to be a temporary state of the software. Maybe adding a comment, or a `# TODO` would help. Worth opening an issue as well.

aea/decision_maker/base.py
```diff
@@ -110,42 +111,45 @@ def check_transaction_is_consistent(self, tx_message: TransactionMessage) -> boo
...
+        for currency_id, amount in tx_message.tx_amount_by_currency_id.items():
+            if amount <= 0 and all(quantity >= 0 for quantity in tx_message.tx_quantities_by_good_id.values()):
+                # check if the agent has the money to cover amount and tx fee (the agent is the buyer).
+                transfer_amount = -amount - tx_message.tx_counterparty_fee
+                result = result and (transfer_amount >= 0)
+                max_tx_fee = tx_message.tx_sender_fee + tx_message.tx_counterparty_fee
+                max_payable = transfer_amount + max_tx_fee
+                result = result and (self.amount_by_currency_id[currency_id] >= max_payable)
+            elif amount >= 0 and all(quantity <= 0 for quantity in tx_message.tx_quantities_by_good_id.values()):
```
marcofavorito: I noticed that both in the buyer and in the seller cases, the comparisons are `<=` and `>=` (equality included), respectively.
What happens if `amount = 0` and for all the quantities,  `quantity = 0`?

aea/decision_maker/base.py
```diff
@@ -165,22 +169,20 @@ def update(self, tx_message: TransactionMessage) -> None:
...
         """
-        if tx_message.is_sender_buyer:
-            diff = tx_message.amount + tx_message.sender_tx_fee
-            self._amount_by_currency[tx_message.currency_id] -= diff
-        else:
-            diff = tx_message.amount - tx_message.sender_tx_fee
-            self._amount_by_currency[tx_message.currency_id] += diff
+        assert self.check_transaction_is_consistent(tx_message), "Inconsistent transaction."
+        for currency_id, amount in tx_message.tx_amount_by_currency_id.items():
+            # we apply the worst case where all the tx_sender_fee is used up in the transaction
```
marcofavorito: Can it be different? That is, for some transactions the agent might not be supposed to pay (because it's not the sender)? Or because the fee was too high than actually needed?

aea/decision_maker/messages/transaction.py
```diff
@@ -39,55 +39,52 @@ class TransactionMessage(InternalMessage):
...
-        SIGN = "sign"
-        ACCEPT = "accept"
-        REJECT = "reject"
+        PROPOSE_FOR_SETTLEMENT = "propose_for_settlement"
+        SUCCESSFUL_SETTLEMENT = "successful_settlement"
+        FAILED_SETTLEMENT = "failed_settlement"
+        REJECTED_SETTLEMENT = "rejected_settlement"
+        PROPOSE_FOR_SIGNING = "propose_for_signing"
+        SUCCESSFUL_SIGNING = "successful_signing"
+        REJECTED_SIGNING = "rejected_signing"
```
marcofavorito: Would be useful at some point to have documentation and illustration (e.g. with sequence diagrams) about the possible usages of the different performatives.

aea/decision_maker/messages/transaction.py
```diff
@@ -221,27 +216,77 @@ def check_consistency(self) -> bool:
...
+                                        tx_sender_addr=other.tx_sender_addr,
+                                        tx_counterparty_addr=other.tx_counterparty_addr,
+                                        tx_amount_by_currency_id=other.tx_amount_by_currency_id,
+                                        tx_sender_fee=other.tx_sender_fee,
+                                        tx_counterparty_fee=other.tx_counterparty_fee,
+                                        tx_quantities_by_good_id=other.tx_quantities_by_good_id,
+                                        ledger_id=other.ledger_id,
+                                        info=other.info,
+                                        tx_digest=tx_digest)
+        return tx_msg
```
marcofavorito: couldn't that simplified with:
```python
tx_msg = TransactionMessage(...)
if tx_digest is None:
    tx_msg.set("tx_digest", tx_digest)
return tx_msg
```

aea/decision_maker/messages/transaction.py
```diff
@@ -221,27 +216,77 @@ def check_consistency(self) -> bool:
...
+                                        tx_sender_addr=other.tx_sender_addr,
+                                        tx_counterparty_addr=other.tx_counterparty_addr,
+                                        tx_amount_by_currency_id=other.tx_amount_by_currency_id,
+                                        tx_sender_fee=other.tx_sender_fee,
+                                        tx_counterparty_fee=other.tx_counterparty_fee,
+                                        tx_quantities_by_good_id=other.tx_quantities_by_good_id,
+                                        ledger_id=other.ledger_id,
+                                        info=other.info,
+                                        signing_payload=other.signing_payload,
+                                        tx_signature=tx_signature)
```
marcofavorito: see above

aea/decision_maker/base.py
DavidMinarsch: Well, a transaction should for the forseeable future just reference one currency. But I will add an issue to do multi currency transactions :)

aea/decision_maker/messages/transaction.py
DavidMinarsch: Yes agreed. Opening an issue.

aea/decision_maker/messages/transaction.py
DavidMinarsch: Well it could, but then the consistency check is not applied as this is only run upon the init. We can clean this up in the future to have consistency run also on setters!

aea/decision_maker/messages/transaction.py
DavidMinarsch: see above

aea/decision_maker/base.py
DavidMinarsch: There is a wider issue around transaction fees. The first is that the split of the `max_tx_fee` needs to be part of the negotiation. The actual `tx_fee` is only known at the time of transacting. The update() method here is actually a bit of a placeholder. Currently, it is used for forward looking calculations by the AEA, i.e. what happens if I transact like this to my overall utility. Actual state is updated via the `apply_state_update` method, which assumes that tx fees are roled into the `amount_by_currency_id` diff. This only gets used by the TAC skills atm. There is stilll a lot of work to be done to align the state the agent holds with the ledger state and come up with the right synchronisation.

aea/decision_maker/base.py
DavidMinarsch: Good point, I changed this now to reject the transaction for the special, all `0` case.

aea/crypto/ethereum.py
```diff
@@ -103,6 +103,16 @@ def sign_message(self, message: str) -> bytes:
         signature = self._account.sign_message(m_message)
         return signature
```
Totoual: I think that we can remove the sign_message from the ethereum.py. If we are trying to make the ethereum.py and fetchai.py similar.

aea/decision_maker/base.py
```diff
@@ -42,6 +42,7 @@
 SENDER_TX_SHARE = 0.5
 QUANTITY_SHIFT = 100
 INTERNAL_PROTOCOL_ID = 'internal'
+OFF_CHAIN_SETTLEMENT_DIGEST = cast(Optional[str], 'off_chain_settlement')
```
Totoual: Are these something that we could move inside the aea-config.yaml?

Issue #475, opened 2019-12-01
Expose handlers, behaviours by name on context


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
Currently, the way to access handlers works like this: `self.context.handlers[0]`

**Describe the solution you'd like**
We would like to use this like shared classes: `self.context.fipa_handler`

PR #525, opened 2019-12-17
Change skill configuration resources types from lists to dictionaries.

### LINKED ISSUE #205: Use id as key for resources list in skill.yaml

skill.yaml file format is not very handy when listing different components, e.g.

```
...
handlers:
    handler:
        - field1
        - field2
    handler:
        - field1
        - field2
... # the same for other component types
```

Instead of repeating `handler:`, we could use its identifier, e.g.

```
...
handlers:
    handler1:
        - field1
        - field1
        - field2
    handler2:
        - field1
        - field2
... # the same for other component types
```

None.

**Additional context**

None

marcofavorito:
## Proposed changes

Before this PR, a skill.yaml would have the following format:
```yaml
...
behaviours:
- behaviour:
  class_name: {..}
  args: {...}
tasks: [...] # analogous structure as 'behaviours'
handlers: [...] # analogous structure as 'behaviours'
shared_classes: [...] # analogous structure as 'behaviours'
```

That is, the components are listed as 'lists'. Instead, change with the following format:
```yaml

...
behaviours:
  behaviour_1_id:
    class_name: {...}
    args: {...}
    ...
...
```
The same for the other component types. That means the actual components, e.g. a single behaviour, can be retrived by using its identifier, instead of having an indexing based on the position.
## Fixes

Fixes #205 

## Types of changes

## Checklist

### DIFF
--- a/aea/skills/scaffold/skill.yaml
+++ b/aea/skills/scaffold/skill.yaml
@@ -5,23 +5,23 @@ license: Apache 2.0
 url: ""
 description: "The scaffold skill is a scaffold for your own skill implementation."
 behaviours:
-  - behaviour:
-      class_name: MyScaffoldBehaviour
-      args:
-        foo: bar
+  scaffold:
+    class_name: MyScaffoldBehaviour
+    args:
+      foo: bar
 handlers:
-  - handler:
-      class_name: MyScaffoldHandler
-      args:
-        foo: bar
+  scaffold:
+    class_name: MyScaffoldHandler
+    args:
+      foo: bar
 tasks:
-  - task:
-      class_name: MyScaffoldTask
-      args:
-        foo: bar
+  scaffold:
+    class_name: MyScaffoldTask
+    args:
+      foo: bar
 shared_classes:
-  - shared_class:
-      class_name: MySharedClass
-      args:
-        foo: bar
+  scaffold:
+    class_name: MySharedClass
+    args:
+      foo: bar
 protocols: []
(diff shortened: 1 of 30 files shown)


aea/skills/scaffold/skill.yaml
```diff
@@ -5,23 +5,23 @@ license: Apache 2.0
 url: ""
 description: "The scaffold skill is a scaffold for your own skill implementation."
 behaviours:
-  - behaviour:
-      class_name: MyScaffoldBehaviour
-      args:
-        foo: bar
+  scaffold:
```
DavidMinarsch: maybe call them `my_scaffold_behaviour` etc.

marcofavorito:
> Looks good, thanks for this!
> 
> Can you address #475 in this PR as well. I know it is mixing concerns but the implementation might be informed by this issue. In particular, are we thinking of calling a behaviour in the context like `self.context.my_behaviour` where `my_behaviour` is the key value in the skill.yaml?

Sure!

tests/test_aea.py
```diff
@@ -167,7 +167,7 @@ def test_react():
             t.start()
             time.sleep(2.0)
             dummy_skill = agent.resources.get_skill("dummy")
-            dummy_handler = dummy_skill.handlers[0]
+            dummy_handler = dummy_skill.handlers["dummy"]
```
DavidMinarsch: maybe use the new method: `dummy_skill.handlers.dummy` as well and add an assert so we show both are possible

Issue #513, opened 2019-12-15
Create unique string to identify/reference a skill/protocol/connection


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
Skills, protocols and connections are uniquely identified by `user_name:package_name/version_id` (the public identifier). These are available as three fields in the config. For referencing a package it makes sense to create a unique, fixed length identifier from these three.

*Describe the solution you'd like**
Use a hashing function to create the unique id and add it to the config file as an additional field.
 
The uid can then be used in the envelopes to reference protocols and in the aea to reference packages (to avoid clashes between similarly names packages).

DavidMinarsch:
We will hash the package itself (think IPFS) and then use this as a unique reference for protocols

Issue #155, opened 2019-10-01
Unique names for skills/protocols/connections


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
Currently, no naming conventions are applied.

**Describe the solution you'd like**
USER_HANDLER/RESOURCE_NAME:VERSION as a unique id/public id. This should be added to the config file of a skill/connection/protocol/agent and needs to be present for add & push. The agent can run without it though..

**Describe alternatives you've considered**
-

**Additional context**
-

DavidMinarsch:
Alias for local names: with the package public ids it means you can have two similarly named packages `echo` on two different ids `user_1:echo/v1` and `user_2:echo/v1` say, but you cannot not have them both as part of the same agent. I was thinking, we could add an optional alias field in the aea-config.yaml so that you can do so. If the `alias` field is present we use that for loading the package locally, so you can do `echo_one` and `echo_two` say. This has now been agreed by @marcofavorito and myself for implementation.

PR #526, opened 2019-12-18
AEA-219  Identify the packages by unique_id

Totoual:
## Proposed changes

Generate a uuid for each connection/protocol/skill so it can be unique.

## Fixes

Fixes: #513 and #155 

## Types of changes

## Checklist

- [] Lint and unit tests pass locally with my changes


aea/configurations/base.py
```diff
@@ -197,6 +198,7 @@ def __init__(self,
         self.authors = authors
         self.version = version
         self.license = license
+        self.uuid = generate_uuid(author=self.authors, package_name=self.name, version=self.version)
```
DavidMinarsch: @marcofavorito and @Totoual
- are we happy calling it uuid? It makes sense, but maybe not quite in line with crypto lingo.
- we should allow for a fourth, optional field in the `generate_uuid` function called `nonce`. Then we can by default set the nonce to None. If the user wants to use two packages with different configs then they can achieve that by feeding a `nonce` to the `generate_uuid` function. Note, we do not allow for the submission of the same package with two different nonces to the package registry, hence we do not need to keep track of the nonce in the config (it is only for local usage).
- I think we need to add a `user_name` field and replace the authors with the `user_name`
- All packages should have `user_name = fetchai`
- All packages should have their uuid in the config (only do this once we have finished everything else int his PR)
- Once this PR is finished, we need another PR which applies the usage of uuid to how packages are identified within the AEA framework, but we cannot do this for 2.0, hence please open an issue.

aea/helpers/base.py
```diff
@@ -68,3 +70,13 @@ def locate(path):
...
+
+
+def generate_public_id(user_name: str, package_name: str, version: str, nonce: Optional[int] = None) -> str:
+    """Generate a unique id for the package."""
+    import hashlib
+    string_for_hash = "".join([user_name, package_name, version, str(nonce)])
+    m_hash = hashlib.sha3_256()
+    m_hash.update(string_for_hash.encode())
+    encoded_str = m_hash.digest().hex()
+    return encoded_str
```
panasevychol: So what do we have on the output? Hashed string?

aea/configurations/base.py
```diff
@@ -182,7 +183,7 @@ class ConnectionConfig(Configuration):
 
     def __init__(self,
                  name: str = "",
-                 authors: str = "",
+                 user_name: str = "",
```
panasevychol: I would prefer `owner` or `owner_name` or smth. like that. To explain which user_name is that.

aea/helpers/base.py
Totoual: yeap

aea/configurations/base.py
Totoual: We will switch back to author to avoid any conflict.

aea/connections/scaffold/connection.yaml
```diff
@@ -1,7 +1,8 @@
 name: scaffold_connection
```
DavidMinarsch: This should be called `scaffold` only

aea/protocols/scaffold/protocol.yaml
```diff
@@ -1,6 +1,7 @@
 name: 'scaffold_protocol'
```
DavidMinarsch: This should be `scaffold` only

aea/skills/scaffold/skill.yaml
```diff
@@ -1,7 +1,8 @@
 name: scaffold_skill
```
DavidMinarsch: This should be `scaffold` only

tests/data/aea-config.example.yaml
```diff
@@ -1,6 +1,6 @@
 aea_version: 0.1.1
 agent_name: myagent
-authors: Fetch.AI Limited
```
DavidMinarsch: missing fingerprint field

tests/data/aea-config.example_w_keys.yaml
```diff
@@ -1,6 +1,6 @@
 aea_version: 0.1.1
 agent_name: myagent
-authors: Fetch.AI Limited
```
DavidMinarsch: missing fingerprint field

tests/data/dummy_aea/aea-config.yaml
```diff
@@ -1,6 +1,6 @@
 aea_version: 0.1.7
 agent_name: Agent0
-authors: Fetch.AI Limited
```
DavidMinarsch: missing fingerprint field

aea/helpers/base.py
```diff
@@ -68,3 +70,13 @@ def locate(path):
         except AttributeError:
             return None
     return object
+
+
+def generate_fingerprint(author: str, package_name: str, version: str, nonce: Optional[int] = None) -> str:
+    """Generate a unique id for the package."""
+    import hashlib
+    string_for_hash = "".join([author, package_name, version, str(nonce)])
```
DavidMinarsch: If the nonce = None, we should not convert it to the string, rather just have the first three

tests/data/dummy_aea/aea-config.yaml
Totoual: I don't think that we want fingerprint here.

tests/test_packages/test_skills/data/connection.yaml
```diff
@@ -1,5 +1,5 @@
 name: gym
-authors: Fetch.AI Limited
+author: fetchai
 version: 0.1.0
```
DavidMinarsch: add fingerprint

PR #529, opened 2019-12-18
Fix minor issues in CLI GUI tests.

marcofavorito:
## Proposed changes

Fix minor issues in the tests relative to the CLI GUI component.

- the main reason of this PR is about some tests that started failing for Python 3.8. Namely:
    - `tests/test_gui/test_search.py::test_search_protocols`
    - `tests/test_gui/test_search.py::test_search_connections`
    - `tests/test_gui/test_search.py::test_list_skills`

Each of them raised the following error:
```
  File ".../agents-aea/.tox/py38/lib/python3.8/site-packages/connexion/apis/abstract.py", line 325, in _response_from_handler
    raise TypeError(
TypeError: The view function did not return a valid response tuple. The tuple must have the form (body), (body, status, headers), (body, status), or (body, headers).
```
Indeed, the `aea/cli_gui/__init__.py::search_registered_items` returned 4 values instead of the allowed number between 1 and 3. It is probably due to a change in the `connexion` dependency. 
That change also implied a change in the `home.js` file in the front-end of the app.

- To call the `aea` command,  `["sys.executable", "-m", "aea.cli", ...]` is used, replacing `["aea", ...]`. That would probably give more robustness since we are sure to use the same executable that is being used by the calling process.

## Types of changes

## Checklist

### DIFF
--- a/aea/cli/common.py
+++ b/aea/cli/common.py
@@ -31,6 +31,7 @@ import click
 import jsonschema  # type: ignore
 from dotenv import load_dotenv
 
+import aea
 from aea.cli.loggers import default_logging_config
 from aea.configurations.base import DEFAULT_AEA_CONFIG_FILE, AgentConfig, SkillConfig, ConnectionConfig, ProtocolConfig, \
     DEFAULT_PROTOCOL_CONFIG_FILE, DEFAULT_CONNECTION_CONFIG_FILE, DEFAULT_SKILL_CONFIG_FILE, Dependencies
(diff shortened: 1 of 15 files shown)


aea/cli/common.py
```diff
@@ -39,7 +40,7 @@
 logger = logging.getLogger("aea")
 logger = default_logging_config(logger)
 
-DEFAULT_REGISTRY_PATH = "../packages"
+DEFAULT_REGISTRY_PATH = str(Path(aea.AEA_DIR, "..", "packages").resolve())
```
DavidMinarsch: Are we sure this is sensible? For instance, in line 70 of aea/cli/create.py we now save the entire path to the config..

PR #533, opened 2019-12-19
fix issues in aea config

marcofavorito:
## Proposed changes

`aea config set` if the value is not valid against the json schema it will overwrite the target yaml file.

## Types of changes

## Checklist

### DIFF
--- a/aea/configurations/base.py
+++ b/aea/configurations/base.py
@@ -243,7 +243,7 @@ class ConnectionConfig(Configuration):
             restricted_to_protocols=cast(Set[str], restricted_to_protocols),
             excluded_protocols=cast(Set[str], excluded_protocols),
             dependencies=dependencies,
-            description=cast(str, obj.get("description")),
+            description=cast(str, obj.get("description", "")),
             **cast(dict, obj.get("config"))
         )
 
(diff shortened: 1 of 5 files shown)


aea/configurations/base.py
```diff
@@ -243,7 +243,7 @@ def from_json(cls, obj: Dict):
             restricted_to_protocols=cast(Set[str], restricted_to_protocols),
             excluded_protocols=cast(Set[str], excluded_protocols),
             dependencies=dependencies,
-            description=cast(str, obj.get("description")),
+            description=cast(str, obj.get("description", "")),
```
marcofavorito: without the default value `""`, the default was `None` and this caused issues when validating the configuration in json format, since it expeted a `"string"` as per JSON schema.

tests/test_gui/test_search.py
```diff
@@ -138,29 +138,23 @@ def test_real_search():
...
     i += 1
-
     assert data[i]['id'] == 'stub'
     assert data[i]['description'] == 'The stub connection implements a connection stub which reads/writes messages from/to file.'
     i += 1
-
     assert data[i]['id'] == 'tcp'
-    assert data[i]['description'] == 'None'
-    i += 1
+    assert data[i]['description'] == ''
```
marcofavorito: see above.

Issue #455, opened 2019-11-28
Import quirks


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
Currently, some packages are part of the distribution whilst others are not. Those packages which are part of the distribution can be referenced directly, `aea.protocols.fipa.*` for instance. However, they are also copied into the agent project on skills creation, for instance. This is somewhat confusing as the developer might think they can amend the files in the agent project when in fact this will not be the place from which they are called at run time!

**Describe the solution you'd like**
One solution is to no longer copy packages which are part of the distribution into the agent project. This reduces the amount of code the developer is confronted with (a plus!).

**Describe alternatives you've considered**
I am keen to hear alternatives.

DavidMinarsch:
We decided to remove all packages apart from the default protocol, error skill and stub connection from the aea package and move it to the packages dir instead.

PR #534, opened 2019-12-19
AEA-222 - move non-default packages into packages dir

DavidMinarsch:
## Proposed changes

This PR moves non-default packages into the `packages` directory.

## Fixes

It addresses #455 

## Types of changes

## Checklist


aea/cli/add.py
```diff
@@ -77,6 +78,12 @@ def _find_connection_locally(ctx, connection_name):
         logger.error(str(e))
         sys.exit(1)
```
DavidMinarsch: Note, we need to recursively load protocol dependencies and therefore also need a protocol dependency list in the connection.yaml.

aea/helpers/search/models.py
```diff
@@ -27,28 +27,6 @@
 ATTRIBUTE_TYPES = Union[float, str, bool, int]
```
DavidMinarsch: OEF Models now live here, the reason being that they are different from protocols (we want to have control over them from the framework side). Them being in OEF protocol also introduces many cross dependencies.

tests/test_cli/test_add/test_protocol.py
```diff
@@ -44,7 +44,7 @@ def setup_class(cls):
         cls.agent_name = "myagent"
         cls.cwd = os.getcwd()
         cls.t = tempfile.mkdtemp()
-        cls.protocol_name = "oef"
+        cls.protocol_name = "gym"
```
DavidMinarsch: Note, due to the dependencies now being installed for the connections automatically (and since the default connection is oef with protocol oef & fipa dependency) this had to be changed to a protocol different from oef, default and fipa

tests/test_cli/test_create.py
```diff
@@ -128,7 +128,7 @@ def test_license_field_is_empty_string(self):
     def test_protocols_field_is_not_empty_list(self):
         """Check that the 'protocols' field is a list with the 'default' protocol."""
         agent_config_instance = self._load_config_file()
-        assert agent_config_instance["protocols"] == ["default"]
+        assert agent_config_instance["protocols"] == ["default", "fipa", "oef"]
```
DavidMinarsch: since oef is currently the default connection these end up being installed

Issue #500, opened 2019-12-09
Skill config: add required_protocols identified with public_ids


panasevychol:
Now required protocols described in Skill config as `protocols: ["name"]`.
This can not be used when pushing a Skill to a remote Registry because we identify items in Registry with public_id. Now we have just names. It's possible to generate public id from owner_name/item_name:version.

There're 2 options I see:
- replace current protocol_names with public_ids. Protocol name can be grabbed from public id for local usage
- add an extra field (i.e. we have `protocols: [...]`, we can have an extra option `registry_required_protocols: ['public_id', ...]` to keep em both for now)

PR #550, opened 2019-12-23
Set pattern for `author` and `version` field in configuration schema.

marcofavorito:
## Proposed changes

This PR enforces a proper format for the `author` and `version` field in the configuration schemas.

## Fixes

Preliminary work for #155 and #500 

## Types of changes

## Checklist


aea/configurations/schemas/aea-config_schema.json
```diff
@@ -103,47 +103,5 @@
     "description": {
       "type": "string"
     }
-  },
```
marcofavorito: we are not going to copy `definitions` for every schema file. It is not maintainable. 
We introduced this workaround in PR #289. However, that is not a long-term solution, so if the problem still occurs on Windows we have to find another solution.

aea/configurations/schemas/definitions.json
```diff
@@ -34,6 +35,54 @@
...
+        }
+      }
+    },
+    "author": {
+      "type": "string",
+      "pattern": "[a-zA-Z0-9_]*"
+    },
+    "version": {
+      "type": "string",
+      "pattern": "^(0|[1-9]\\d*)\\.(0|[1-9]\\d*)\\.(0|[1-9]\\d*)(?:-((?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\\.(?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\\+([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?$"
```
marcofavorito: We use semantic versioning for version strings (not backward compatible):
[https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string](https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string)

aea/configurations/schemas/definitions.json
```diff
@@ -34,6 +35,54 @@
...
+    },
+    "public_id": {
+      "type": "string",
+      "pattern": "[a-zA-Z_][a-zA-Z0-9_]*"
```
marcofavorito: This is not used atm, to be removed.

aea/configurations/schemas/definitions.json
```diff
@@ -34,6 +35,54 @@
...
+          "type": "string"
+        },
+        "port": {
+          "type": "number"
+        }
+      }
+    },
+    "author": {
+      "type": "string",
+      "pattern": "[a-zA-Z0-9_]*"
```
marcofavorito: Author `""` is allowed in the configuration file for backward compatibility. However, it is not going to work when interacting with the registry. 

That suggests `aea scaffold` sets the `author` field according to the session kept by the CLI tool (if any, eventually).

aea/cli/create.py
```diff
@@ -75,7 +75,7 @@ def create(click_context, agent_name):
         # create a config file inside it
         logger.info("Creating config file {}".format(DEFAULT_AEA_CONFIG_FILE))
         config_file = open(os.path.join(agent_name, DEFAULT_AEA_CONFIG_FILE), "w")
-        agent_config = AgentConfig(agent_name=agent_name, aea_version=aea.__version__, author="", version="v1", license="", fingerprint="", url="", registry_path=DEFAULT_REGISTRY_PATH, description="")
+        agent_config = AgentConfig(agent_name=agent_name, aea_version=aea.__version__, author="", version="1.0.0", license="", fingerprint="", url="", registry_path=DEFAULT_REGISTRY_PATH, description="")
```
DavidMinarsch: I think the default version should be `0.1.0`. We should not suggest that a default agent project is somehow a v1.

aea/configurations/schemas/definitions.json
DavidMinarsch: Ok

aea/configurations/schemas/definitions.json
DavidMinarsch: Sounds good

aea/configurations/schemas/definitions.json
DavidMinarsch: Ok so remove..?

aea/cli/create.py
marcofavorito: done

PR #553, opened 2019-12-23
Use public ids

### LINKED ISSUE #500: Skill config: add required_protocols identified with public_ids

marcofavorito:
## Fixes

Fixes #500 and #155

## Types of changes

## Checklist


aea/configurations/base.py
```diff
@@ -544,17 +642,20 @@ def __init__(self,
     def default_connection(self) -> str:
         """Get the default connection."""
         assert self._default_connection is not None, "Default connection not set yet."
-        return self._default_connection
+        return str(self._default_connection)
 
     @default_connection.setter
-    def default_connection(self, connection_name: str):
+    def default_connection(self, connection_name: Optional[str]):
```
DavidMinarsch: Not sure why this is needed as Optional?

aea/cli/common.py
```diff
@@ -117,13 +117,13 @@ def _try_to_load_protocols(ctx: Context):
...
         try:
-            ctx.protocol_loader.load(open(os.path.join("protocols", protocol_name, DEFAULT_PROTOCOL_CONFIG_FILE)))
+            ctx.protocol_loader.load(open(os.path.join("protocols", protocol_name.package_name, DEFAULT_PROTOCOL_CONFIG_FILE)))
         except FileNotFoundError:
             logger.error("Protocol configuration file for protocol {} not found.".format(protocol_name))
             sys.exit(1)
 
         try:
-            protocol_spec = importlib.util.spec_from_file_location(protocol_name, os.path.join("protocols", protocol_name, "__init__.py"))
+            protocol_spec = importlib.util.spec_from_file_location(protocol_name, os.path.join("protocols", protocol_name.package_name, "__init__.py"))
```
DavidMinarsch: should it be protocol_id?

marcofavorito: The namings are going to change.
However, at the moment, to be precise, `protocol_name` is an instance of `ProtocolId`, and `package_name` is the name of the package, e.g. in the public id of the protocol `fetchai/oef:0.1.0`, `package_name` is `oef`.

Since we store the components in directories whose names are the names of the packages that they contain, we have to use `package_name` to point to them.

In the example, for the protocol `oef`, we have to load `protocols/oef/__init__.py`
```
protocols/
    oef/
        __init__.py
        protocol.yaml
        ...
```

aea/configurations/base.py
marcofavorito: the type of `self._default_connection` (after the changes) is `Optional[PublicId]`, but the setter didn't accept `None` values.

(the same holds for the previous code, where the type was simply `str`)

PR #559, opened 2019-12-24
Local "push" item implemented, handling ConnectionError added

panasevychol:
Local "push" item implemented, handling ConnectionError while request added

## Proposed changes

Local "push" item implemented, handling ConnectionError while request added

## Types of changes

## Checklist

### DIFF
--- a/aea/cli/registry/push.py
+++ b/aea/cli/registry/push.py
@@ -23,6 +23,8 @@ import os
 import tarfile
 import shutil
 
+from distutils.dir_util import copy_tree
+
 from aea.cli.common import logger
 from aea.cli.registry.utils import request_api, load_yaml, clean_tarfiles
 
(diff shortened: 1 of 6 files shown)


aea/cli/registry/push.py
```diff
@@ -93,3 +95,59 @@ def push_item(item_type: str, item_name: str) -> None:
...
+
+def get_packages_path() -> str:
+    """
+    Get path to packages folder.
+
+    Should not be called from outside the project dir.
+
+    :return: str path to packages dir.
+    """
+    project_parent_dir = os.path.abspath('__file__').split(
```
DavidMinarsch: We have a path already defined as a constant: `DEFAULT_REGISTRY_PATH`. Use this instead.

tests/test_cli/test_registry/test_push.py
```diff
@@ -101,3 +108,97 @@ def test_remove_pycache_no_pycache(self, path_exists_mock, rmtree_mock):
         source_dir = 'somedir'
         _remove_pycache(source_dir)
         rmtree_mock.assert_not_called()
+
+
+@mock.patch(
+    'aea.cli.registry.push.os.path.abspath',
+    return_value=('somepath/agents-aea/somefile')
```
DavidMinarsch: see above, should reference DEFAULT_REGISTRY_PATH not agents-aea

aea/cli/registry/push.py
panasevychol: I've tried. But it always leads to "packages" one level up. It won't work with any nested folder structure.
Sure I can use that. But don't you think it's better to have a possibility to use it with nested structure?

DavidMinarsch: well you neet to construct the correct path I suppose. `packages` is assumed to be on the root dir. So if you build cwd + DEFAULT_REGISTRY+PATH then cwd would have to be the agent project dir. You can figure out whether you're in agent project dir by checking if the aea-config file is present.

PR #606, opened 2020-01-06
fix import of agent components.

marcofavorito:
## Proposed changes

This PR fixes some issues of the import system for agent components.
In particular:
- uses the right dotted path (e.g. `packages.protocols.fipa`) for imports.
- make the framework to import the whole connection package, and not only the `connection.py` module.

## Fixes

None

## Types of changes

## Checklist

### DIFF
--- a/aea/helpers/base.py
+++ b/aea/helpers/base.py
@@ -96,24 +96,25 @@ def add_module_to_sys_modules(dotted_path: str, module_obj) -> None:
     :param dotted_path: the dotted path to be used in the imports.
     :param module_obj: the module object. It is assumed it has been already executed.
     :return: None
-    :raises ValueError: if the dotted path is already used.
+    :raises ValueError: if the dotted path is already used. # TODO this cannot be done atm.
     """
-    if dotted_path in sys.modules:
-        raise ValueError("The dotted path {} is already used.".format(dotted_path))
+    # if dotted_path in sys.modules:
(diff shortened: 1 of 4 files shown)


aea/helpers/base.py
```diff
@@ -72,27 +73,63 @@ def locate(path):
...
+def add_module_to_sys_modules(dotted_path: str, module_obj) -> None:
+    """
+    Add module to sys.modules.
+
+    :param dotted_path: the dotted path to be used in the imports.
+    :param module_obj: the module object. It is assumed it has been already executed.
+    :return: None
+    :raises ValueError: if the dotted path is already used. # TODO this cannot be done atm.
+    """
+    # if dotted_path in sys.modules:
```
DavidMinarsch: Why is this commented out?

marcofavorito: because I wanted feedback :-) 

Those two line should be there for "safety" purposes. That is, to avoid removing previously loaded modules, that could break the code working properly.
In practice, for a (single) AEA, this function shouldn't be a big problem since the namespace of the modules is very narrow (e.g. unlikely that there exist `packages.package_type.package_name`). Still, name conflicts may occur.

A practical consequence of those two lines is that tests fail because each test is not independent in that sense - later test try to load a package which has been already loaded by previous tests, causing the commented `ValueError`  to be raised.

PR #612, opened 2020-01-09
Local push and publish commands re-implemented to use context generat…

panasevychol:
…ed value of packages path. Tests fixed.

## Fixes

The variable `DEFAULT_REGISTRY_PATH` was changed not to be used as a path to `packages` folder anymore so the push and publish functionality that was using it was broken. By request it's re-implemented "exactly like in search command". Tests updated.

WARNING!
This fix seem to be acting wrongly. Context generated value `registry_path` doesn't lead us to packages folder. Instead of it it points to `aea` inside of the current folder.

## Types of changes

## Checklist

### DIFF
--- a/aea/cli/fetch.py
+++ b/aea/cli/fetch.py
@@ -19,9 +19,12 @@
 
 """Implementation of the 'aea fetch' subcommand."""
 import click
+from distutils.dir_util import copy_tree
+import os
 
-from aea.cli.common import PublicIdParameter
-from aea.cli.registry.fetch import fetch_agent, fetch_agent_locally
+from aea.cli.common import Context, pass_ctx, PublicIdParameter, DEFAULT_REGISTRY_PATH, try_get_item_source_path
+from aea.cli.registry.fetch import fetch_agent
+from aea.configurations.base import PublicId
 
 
 @click.command(name='fetch')
@@ -29,9 +32,31 @@ from aea.cli.registry.fetch import fetch_agent, fetch_agent_locally
     '--registry', is_flag=True, help="For fetching agent from Registry."
 )
 @click.argument('public-id', type=PublicIdParameter(), required=True)
-def fetch(public_id, registry):
+@pass_ctx
+def fetch(ctx: Context, public_id, registry):
     """Fetch Agent from Registry."""
     if not registry:
-        fetch_agent_locally(public_id)
+        _fetch_agent_locally(ctx, public_id)
     else:
-        fetch_agent(public_id)
+        fetch_agent(ctx, public_id)
+
+
+def _fetch_agent_locally(ctx: Context, public_id: PublicId) -> None:
+    """
+    Fetch Agent from local packages.
+
+    :param ctx: Context
+    :param public_id: public ID of agent to be fetched.
+
+    :return: None
+    """
+    packages_path = os.path.basename(DEFAULT_REGISTRY_PATH)
+    source_path = try_get_item_source_path(packages_path, 'agents', public_id.name)
+    target_path = os.path.join(ctx.cwd, public_id.name)
+    if os.path.exists(target_path):
+        raise click.ClickException(
+            'Item "{}" already exists in target folder.'.format(public_id.name)
+        )
+    copy_tree(source_path, target_path)
+    click.echo('Agent {} successfully fetched.'.format(public_id.name))
+    # TODO: install all dependencies recursively
--- a/aea/cli/publish.py
+++ b/aea/cli/publish.py
@@ -22,7 +22,7 @@ import click
 import os
 from shutil import copyfile
 
-from aea.cli.common import pass_ctx, Context, _try_to_load_agent_config, try_get_item_target_path, DEFAULT_AEA_CONFIG_FILE
+from aea.cli.common import pass_ctx, Context, try_to_load_agent_config, try_get_item_target_path, DEFAULT_AEA_CONFIG_FILE
 from aea.cli.registry.publish import publish_agent
 
 
@@ -33,7 +33,7 @@ from aea.cli.registry.publish import publish_agent
 @pass_ctx
 def publish(ctx: Context, registry):
     """Publish Agent to Registry."""
-    _try_to_load_agent_config(ctx)
+    try_to_load_agent_config(ctx)
     if not registry:
         # TODO: check agent dependencies are available in local packages dir.
         _save_agent_locally(ctx)
--- a/aea/cli/publish.py
+++ b/aea/cli/publish.py
@@ -19,10 +19,11 @@
 
 """Implementation of the 'aea publish' subcommand."""
 import click
+import os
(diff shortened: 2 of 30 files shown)


aea/cli/registry/utils.py
```diff
@@ -306,18 +309,39 @@ def get_item_source_path(
...
+def get_default_registry_path(click_ctx: Context) -> str:
+    """
+    Get agent's default registry path (packages path).
+
+    :param click_ctx context object of click command.
+
+    :return: str agent's default registry path.
+    """
+    try:
+        with open(DEFAULT_AEA_CONFIG_FILE, mode="r", encoding="utf-8") as f:
```
DavidMinarsch: you can simply do `click_ctx.agent_config.registry_path` I believe

aea/cli/registry/utils.py
```diff
@@ -306,18 +309,39 @@ def get_item_source_path(
...
-    packages_path = DEFAULT_REGISTRY_PATH
     target_path = os.path.join(packages_path, item_type_plural, item_name)
     if os.path.exists(target_path):
         raise click.ClickException(
             'Item "{}" already exists in packages folder.'.format(item_name)
         )
     return target_path
+
+
+def get_default_registry_path(click_ctx: Context) -> str:
```
DavidMinarsch: We prefer to use `ctx` for the `click_context.obj` and click_context otherwise

aea/cli/push.py
```diff
@@ -33,14 +34,19 @@ def push(ctx: Context, registry):
     """Push item to Registry or save it in local packages."""
     ctx.set_config("registry", registry)
 
+    packages_path = get_default_registry_path(ctx)
```
DavidMinarsch: Not a fan of this. The path is already available on `ctx` via `ctx.agent_config.registry_path` so no need to set it here, just pass `ctx` as an argument

aea/cli/push.py
```diff
@@ -33,14 +34,19 @@ def push(ctx: Context, registry):
...
 
 @push.command(name='connection')
 @click.argument('connection-id', type=PublicIdParameter(), required=True)
 @pass_ctx
 def connection(ctx: Context, connection_id):
     """Push connection to Registry or save it in local packages."""
     if not ctx.config.get("registry"):
-        save_item_locally('connection', connection_id)
+        save_item_locally(
+            'connection', connection_id, ctx.config.get("packages_path")
```
DavidMinarsch: You can do `ctx.agent_config.registry_path

aea/cli/push.py
```diff
@@ -51,7 +57,9 @@ def connection(ctx: Context, connection_id):
 def protocol(ctx: Context, protocol_id):
     """Push protocol to Registry or save it in local packages."""
     if not ctx.config.get("registry"):
-        save_item_locally('protocol', protocol_id)
+        save_item_locally(
+            'protocol', protocol_id, ctx.config.get("packages_path")
```
DavidMinarsch: see above

aea/cli/push.py
```diff
@@ -62,6 +70,8 @@ def protocol(ctx: Context, protocol_id):
 def skill(ctx: Context, skill_id):
     """Push skill to Registry or save it in local packages."""
     if not ctx.config.get("registry"):
-        save_item_locally('skill', skill_id)
+        save_item_locally(
+            'skill', skill_id, ctx.config.get("packages_path")
```
DavidMinarsch: see above

aea/cli/registry/fetch.py
```diff
@@ -17,38 +17,32 @@
...
-from aea.cli.common import DEFAULT_REGISTRY_PATH
+from aea.cli.common import Context
 from aea.cli.registry.utils import (
     request_api, download_file, extract, fetch_package
 )
 from aea.configurations.base import PublicId
 
 
-def fetch_agent(public_id: Union[PublicId, str]) -> None:
+def fetch_agent(ctx: Context, public_id: PublicId) -> None:
```
panasevychol: We don't need the whole context object here. It's enough to pass `ctx.cwd` as `cwd` argument.
And by the way what a difference between `os.getcwd()` and `ctx.cwd`?

aea/cli/registry/publish.py
```diff
@@ -49,13 +47,12 @@ def _load_agent_config(agent_config_path: str) -> Dict:
 
 
 @clean_tarfiles
-def publish_agent():
+def publish_agent(ctx: Context):
```
panasevychol: Same as above. Only cwd needed.

aea/cli/registry/push.py
```diff
@@ -49,7 +44,7 @@ def _compress_dir(output_filename: str, source_dir: str):
 
 
 @clean_tarfiles
-def push_item(item_type: str, item_name: str) -> None:
+def push_item(ctx: Context, item_type: str, item_id: PublicIdParameter) -> None:
```
panasevychol: Same as above.

aea/cli/fetch.py
```diff
@@ -19,19 +19,44 @@
...
     """Fetch Agent from Registry."""
     if not registry:
-        fetch_agent_locally(public_id)
+        _fetch_agent_locally(ctx, public_id)
     else:
-        fetch_agent(public_id)
+        fetch_agent(ctx, public_id)
+
+
+def _fetch_agent_locally(ctx: Context, public_id: PublicId) -> None:
```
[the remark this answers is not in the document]
DavidMinarsch: Because we value consistency more. All the other commands implement the local methods here so this should too

aea/cli/publish.py
```diff
@@ -19,17 +19,43 @@
 
 """Implementation of the 'aea publish' subcommand."""
 import click
+import os
+from shutil import copyfile
 
-from aea.cli.registry.publish import publish_agent, save_agent_locally
+from aea.cli.common import pass_ctx, Context, _try_to_load_agent_config, try_get_item_target_path, DEFAULT_AEA_CONFIG_FILE
```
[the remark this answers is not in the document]
DavidMinarsch: Good point. So we should rename it.

aea/cli/registry/fetch.py
DavidMinarsch: I prefer passing the whole object as it makes it easier to read the code and extend later in case you need more from context or need to write to context

tests/test_cli/tools_for_testing.py
```diff
@@ -0,0 +1,42 @@
...
+    """An object to mock Click Context."""
+
+    cwd = 'cwd'
+    agent_config = AgentConfigMock()
+
+
+class PublicIdMock:
+    """An object to mock PublicId."""
+
+    def __init__(self, name='name', author='author', version='1.0.0'):
```
DavidMinarsch: please everywhere where you use verison, import DEFAULT_VERSION from configurations base

PR #621, opened 2020-01-11
AEA-263 Fix/author

DavidMinarsch:
## Proposed changes

This reopnes #618 to simplify comparison

This will avoid clashes between packages with the same name and type
but from different author.

E.g. to import a module of a package of the agent, in agent context:

```
import packages.author_name.package_type_plural.package_name
```
For instance:
```
import packages.fetchai.protocols.fipa.serialization
```

## Types of changes

## Checklist


aea/cli/add.py
```diff
@@ -48,16 +48,18 @@ def add(ctx: Context, registry):
 
 def _find_connection_locally(ctx, connection_public_id, click_context):
     # check that the provided path points to a proper connection directory -> look for connection.yaml file.
-    # first check in aea dir
+    # first check in registry
     connection_name = connection_public_id.name
     registry_path = ctx.agent_config.registry_path
-    connection_configuration_filepath = Path(os.path.join(registry_path, "connections", connection_name, DEFAULT_CONNECTION_CONFIG_FILE))
+    package_path = Path(registry_path, connection_public_id.author, "connections", connection_name)
```
DavidMinarsch: can we make `connections` a constant? Similar for the other pluralised package types.

aea/cli/add.py
```diff
@@ -48,16 +48,18 @@ def add(ctx: Context, registry):
...
+    connection_configuration_filepath = package_path / DEFAULT_CONNECTION_CONFIG_FILE
```
DavidMinarsch: Is this recommended style, never seen it before?

tests/test_helpers/test_base.py
```diff
@@ -33,18 +32,16 @@ def test_locate(self):
         """Test the locate function to locate modules."""
         cwd = os.getcwd()
         os.chdir(os.path.join(CUR_PATH, ".."))
-        sys.modules["gym_connection"] = locate("packages.connections.gym")
-        assert sys.modules['gym_connection'] is not None
-        sys.modules["gym_connection"] = locate("packages.connections.weather")
-        assert sys.modules['gym_connection'] is None
+        assert locate("packages.fetchai.connections.gym") is not None
+        assert locate("packages.fetchai.connections.weather") is None
```
DavidMinarsch: We don't have a weather connection... so I'd add a comment that we are testing absence of a connection to make it easier to read

tests/data/dummy_aea/connections/local/connection.py
```diff
@@ -30,8 +30,8 @@
 from aea.connections.base import Connection
 from aea.helpers.search.models import Description, Query
 from aea.mail.base import Envelope, AEAConnectionError, Address
-from packages.protocols.oef.message import OEFMessage
```
DavidMinarsch: The paths have been accidentally shortened, fix please.

aea/cli/add.py
marcofavorito: yes good idea, will do

aea/cli/add.py
marcofavorito: yes good idea, will do

aea/cli/add.py
marcofavorito: I don't know if it is recommended, but I love that syntactic sugar :-) 

of course it requires that the left operand is an instance of `pathlib.Path`, and the right operand can be both `Path`, `str` or any other type accepted by the constructor of `Path`.

You can see it equivalent as the following expression:
`Path(package_path, DEFAULT_CONNECTION_CONFIG_FILE)`

DavidMinarsch: Yeah, it looks neat. But when I first saw it I thought it's some maths operation :D

tox.ini
```diff
@@ -21,7 +21,7 @@ commands =
     pip install git+https://github.com/pytoolz/cytoolz.git#egg=cytoolz==0.10.1.dev0
     pip install -e .[all]
     pip install -i https://test.pypi.org/simple/ fetch-p2p-api==0.0.1
-    pytest --doctest-modules aea packages tests/ --cov-report=html --cov-report=term --cov=aea --cov=packages/protocols --cov=packages/connections {posargs}
+    pytest --doctest-modules aea packages/fetchai/protocols packages/fetchai/connections tests/ --cov-report=html --cov-report=term --cov=aea --cov=packages/fetchai/protocols --cov=packages/fetchai/connections {posargs}
```
DavidMinarsch: @marcofavorito since you also imported the skills code into coverage checks we needed the `scikit-image` dependency. But we don't want to run coverage on skill code atm as we have no tests other than the end-to-end tests.

tests/test_cli/test_generate/test_protocols.py
```diff
@@ -72,13 +72,13 @@ def test_exit_code_equal_to_0(self):
...
         p = Path(self.t, self.agent_name, "protocols", "two_party_negotiation", "message.py")
-        original = Path(CUR_PATH, "data", "generator", "two_party_negotiation", "message.py")
+        original = Path(CUR_PATH, "..", "examples", "protocol_specification_ex", "output", "two_party_negotiation", "message.py")
         assert filecmp.cmp(p, original)
 
     def test_resource_folder_contains_module_protocol(self):
         """Test that the protocol folder contains serialization.py module."""
         p = Path(self.t, self.agent_name, "protocols", "two_party_negotiation", "serialization.py")
-        original = Path(CUR_PATH, "data", "generator", "two_party_negotiation", "serialization.py")
+        original = Path(CUR_PATH, "..", "examples", "protocol_specification_ex", "output", "two_party_negotiation", "serialization.py")
```
DavidMinarsch: I moved the generated protocols out of the `tests/` directory. Otherwise coverage loads the files and then throws an error because it cannot find the modules.

tox.ini
marcofavorito: I see. thanks

PR #626, opened 2020-01-13
AEA-268 Properties for message fields; Check content names;

5A11:
Added properties per message fields (i.e. message_id, target, performative, contents)
Better error messages
Added check that content names matches content definition

## Types of changes

## Checklist


aea/protocols/generator.py
```diff
@@ -161,56 +161,67 @@ def _message_class_str(self) -> str:
...
+        cls_str += '    @property\n'
+        cls_str += '    def performative(self) -> str:\n'
+        cls_str += '        \"\"\"Get the performative of the message.\"\"\"\n'
+        cls_str += '        assert self.is_set("performative"), "performative is not set"\n'
+        cls_str += '        return cast(str, self.get("performative"))\n\n'
+
+        cls_str += '    @property\n'
+        cls_str += '    def contents(self) -> Dict:\n'
+        cls_str += '        \"\"\"Get the contents of the message.\"\"\"\n'
+        cls_str += '        assert self.is_set("contents"), "contents is not set"\n'
```
DavidMinarsch: We need to dynamically generate a property for each one of the content `names` to expose their `value`. So if the contents = {'price':1, 'size':2} then there need to be properties for `price` and `size`

5A11: You mean for all contents of all allowed performatives?

What if there are some performatives with similarly named contents? Say "request(resource = r)" and "promise(resource r)".

5A11: Actually currently, "cfp" and "propose" both have a content named "query" of type "DataModel".

DavidMinarsch: Well yes, but we should have a restricting which says that a name must be of the same type for all performatives. I don't see why we would want to support a scenario where `my_var = int` for CFP and `my_var = str` for Propose. The developer should use different variable names for different things. Any given message instantiation has one performative only so if the above condition (unique type per var name across performatives) is satisfied there is no problem. The current implementation means I still have to extract from contents and then cast to a type. The same issue applies in this case that we should not allow the above (`my_var = int` for CFP and `my_var = str` for Propose) and the UX is much worse.

DavidMinarsch: And note, `query = DataModel` in more than one performative is completely compatible with the approach we have implemented in the protocols and I am suggesting.

examples/protocol_specification_ex/default.yaml
```diff
@@ -0,0 +1,12 @@
...
+version: 0.1.0
+license: Apache 2.0
+description: 'A protocol for exchanging any bytes message.'
+speech_acts:
+  bytes:
+    content: bytes
+  error:
+    error_code: ErrorCode
+    error_msg: str
+    error_data: dict
```
DavidMinarsch: @5A11 for version control its good to always add a blank line at the end of each file

examples/protocol_specification_ex/fipa.yaml
```diff
@@ -0,0 +1,25 @@
...
+  match_accept_w_inform:
+    info: dict
+  inform:
+    info: dict
+  accept: {}
+  decline: {}
+  match_accept: {}
+
+  # Differences:
+  # propose -> proposal is a list of Descriptions. Declare each d in proposal is of type description.
```
DavidMinarsch: Why not include a Proposal special data type?

examples/protocol_specification_ex/fipa.yaml
```diff
@@ -0,0 +1,25 @@
...
+  # dialogue_reference is missing
```
DavidMinarsch: Please add it, it is required.

DavidMinarsch: Note, this points at a limitation in the current spec, if I want to add a field in each performative I need to repeat myself...

examples/protocol_specification_ex/gym.yaml
```diff
@@ -0,0 +1,20 @@
...
+  percept:
+    step_id: int
+    observation: Observation
+    reward: float
+    done: bool
+    info: dict
+  reset: {}
+  close: {}
+
+  # Should we allow Any as a type and recognise it (e.g. for action: Action or observation: Observation)?
```
DavidMinarsch: I think we cannot allow Any as we would not know how to serialise it?

examples/protocol_specification_ex/fipa.yaml
```diff
@@ -0,0 +1,25 @@
...
+license: Apache 2.0
+description: 'A protocol for FIPA ACL.'
+speech_acts:
+  cfp:
+    query: Query
+  propose:
+    proposal: list
+    price: float
+  accept_w_inform:
+    info: dict
```
DavidMinarsch: I am concerned that dict is not further specified. I think we should make it strongly typed.

examples/protocol_specification_ex/ml_trade.yaml
```diff
@@ -0,0 +1,20 @@
...
+  cfp:
+    query: Query
+  terms:
+    terms: Descrition
+  accept:
+    terms: Descrition
+    tx_digest: str
+  data:
+    terms: Descrition
+    data: Tuple
```
DavidMinarsch: I think it needs to be strongly typed, like `Tuple[Array, Array]` or whatever you want to have there...

examples/protocol_specification_ex/oef.yaml
```diff
@@ -0,0 +1,31 @@
...
+    service_id: str
+  unregister_agent:
+    agent_description: Description
+    agent_id: str
+  search_services:
+    query: Query
+  search_agents:
+    query: Query
+  search_result:
+    agents: list
```
DavidMinarsch: I think it needs to be List[str]

examples/protocol_specification_ex/output/two_party_negotiation/message.py
```diff
@@ -43,47 +43,61 @@ def __init__(self, message_id: int, target: int, performative: str, contents: Di
...
+        return cast(str, self.get("performative"))
+
+    @property
+    def contents(self) -> Dict:
+        """Get the contents of the message."""
+        assert self.is_set("contents"), "contents is not set"
+        return cast(Dict, self.get("contents"))
+
+    @property
+    def performatives_definition(self) -> set:
```
DavidMinarsch: Why not call it `permitted_performatives`? Also, if `permitted_performatives` is only used in the message class I would make it 'private': `_permitted_performatives` and not a property

examples/protocol_specification_ex/tac.yaml
```diff
@@ -0,0 +1,45 @@
...
+description: 'A protocol for participating in a TAC.'
+speech_acts:
+  register:
+    agent_name: str
+  unregister: {}
+  transaction:
+    tx_id: str
+    tx_sender_addr: str
+    tx_counterparty_addr: str
+    amount_by_currency_id: dict
```
DavidMinarsch: `Dict[str, int]` etc.. for all needs to be specified

examples/protocol_specification_ex/tac.yaml
```diff
@@ -0,0 +1,45 @@
...
+    tx_id: str
+    amount_by_currency_id: dict
+    quantities_by_good_id: dict
+  tac_error:
+    error_code: ErrorCode
+    info: dict
+
+# transaction -> amount_by_currency_id is a dictionary. Declare each key, value in 'amount_by_currency_id' is of type str int respectively.
+# similarly for quantities_by_good_id, amount_by_currency_id, exchange_params_by_currency_id, quantities_by_good_id, utility_params_by_good_id, good_id_to_name,
+# currently you check whether keys in the above are unique (line 225). Is that necessary give it is dictionary?
```
DavidMinarsch: No keys are by design unique, so no need to check.

examples/protocol_specification_ex/tac.yaml
```diff
@@ -0,0 +1,45 @@
...
+# declare tx_sender_fee>=0 and tx_counterparty_fee>=0. Declare amount_by_currency_id.keys() == exchange_params_by_currency_id.keys(). Declare quantities_by_good_id.keys == utility_params_by_good_id.keys
```
DavidMinarsch: should we have signed and unsigned integers?

examples/protocol_specification_ex/tac.yaml
```diff
@@ -0,0 +1,45 @@
...
+# game_data -> agent_addr_to_nameand and good_id_to_name are defined to be either dict or defaultdict. Why? do we allow for either T1 or T2 or ... for types?
```
DavidMinarsch: I think dict is fine.

aea/protocols/generator.py

examples/protocol_specification_ex/default.yaml
5A11: Done

examples/protocol_specification_ex/output/two_party_negotiation/message.py
5A11: At some point I decided to differentiate the fields of a message being validated vs the correct definitions by appending the laters' variable names with "_definition", e.g. "contents": the contents of the message, and "contents_definition": its definition which comes from the specification file.

Happy to change for all instances to "valid_..." or "allowed_...", "permitted_...", "defined_...".

examples/protocol_specification_ex/tac.yaml
5A11: You mean I remove lines 225, 233, 269, 273 (where the checks for uniqueness happen) from message.py of tac protocol?

aea/protocols/generator.py
```diff
@@ -161,56 +168,71 @@ def _message_class_str(self) -> str:
...
-        # variables
-        # cls_str += str.format('        self.performatives = {}\n', self._performatives_set())
-        cls_str += str.format('        self.speech_acts = {}\n\n', self._speech_acts_str())
+        cls_str += str.format('        self.speech_acts_definition = {}\n\n', self._speech_acts_str())
         cls_str += '        assert self.check_consistency()\n\n'
 
+        # Properties
         cls_str += '    @property\n'
-        cls_str += '    def performatives(self) -> set:\n'
+        cls_str += '    def performatives_definition(self) -> set:\n'
```
DavidMinarsch: Please relabel this to `allowed_performatives` or `permitted_performatives` as the variable name is confusing and the docstring clear.

aea/protocols/generator.py
```diff
@@ -161,56 +168,71 @@ def _message_class_str(self) -> str:
         cls_str += '    def __init__(self, message_id: int, target: int, performative: str, contents: Dict, **kwargs):\n'
         cls_str += '        \"\"\"Initialise.\"\"\"\n'
         cls_str += '        super().__init__(message_id=message_id, target=target, performative=performative, contents=contents, **kwargs)\n\n'
-
-        # variables
-        # cls_str += str.format('        self.performatives = {}\n', self._performatives_set())
-        cls_str += str.format('        self.speech_acts = {}\n\n', self._speech_acts_str())
+        cls_str += str.format('        self.speech_acts_definition = {}\n\n', self._speech_acts_str())
```
DavidMinarsch: Can we simply call this `speech_acts`? The definition bit does not add anything for me.

aea/protocols/generator.py
DavidMinarsch: It should also be a private `_var_name` variable and not a property since it is only used by the message object internally

aea/protocols/generator.py
```diff
@@ -161,56 +168,71 @@ def _message_class_str(self) -> str:
...
-        cls_str += '            performative = self.get(\"performative\")\n'
-        cls_str += '            assert type(performative) == str, \"performative is not str\"\n\n'
-
-        cls_str += '            assert self.is_set(\"contents\"), \"contents is not set\"\n'
-        cls_str += '            contents = self.get(\"contents\")\n'
-        cls_str += '            assert type(contents) == dict, \"contents is not a dictionary\"\n'
-        cls_str += '            contents = cast(Dict, contents)\n\n'
+        cls_str += '            assert type(self.message_id) == int, \"message_id must be \'int\' but it is not.\"\n'
+        cls_str += '            assert type(self.target) == int, \"target must be \'int\' but it is not.\"\n'
+        cls_str += '            assert type(self.performative) == str, \"performative must be \'str\' but it is not.\"\n'
```
DavidMinarsch: In our actual implementation, we create an enum for the performatives. We should incorporate this here too!

examples/protocol_specification_ex/output/two_party_negotiation/message.py
DavidMinarsch: I think the `definition` is a bit ambiguous. Naming-wise I prefer `allowed_` or `permitted_` as its clearer. However, we can further improve this by making them class rather than instance variables. The `permitted_variables` is a class variable as it applies to all instances of `SomeProtocolMessage`.

examples/protocol_specification_ex/tac.yaml
DavidMinarsch: You don't need to change existing protocols. We will change them once the generator is finished in its first iteration.

examples/protocol_specification_ex/fipa.yaml
5A11: Done

5A11: Done

examples/protocol_specification_ex/fipa.yaml
5A11: info is dict[str, Any]. The Any part is problematic. What should I change it to?

examples/protocol_specification_ex/ml_trade.yaml
5A11: Done

examples/protocol_specification_ex/oef.yaml
5A11: Done

examples/protocol_specification_ex/tac.yaml
5A11: Done

examples/protocol_specification_ex/tac.yaml
5A11: ok

examples/protocol_specification_ex/tac.yaml
5A11: Done

examples/protocol_specification_ex/output/two_party_negotiation/message.py
5A11: ok I change to 'allowed_X'.

I will make it a class property.

examples/protocol_specification_ex/fipa.yaml
5A11: Done

examples/protocol_specification_ex/output/two_party_negotiation/message.py

aea/protocols/generator.py
5A11: Made it a classmethod but kept it public so could be used by others.

aea/protocols/generator.py

aea/protocols/generator.py
DavidMinarsch: I don't think using a method to expose a class variable is ideal.

PR #630, opened 2020-01-13
AEA-267 fix programmatic guide, add programmatic mode flag to aea/agent

### LINKED ISSUE #619: Hacking an agent guide no longer working
The guide is no longer working.

Make changes so it works again.

DavidMinarsch:
## Proposed changes

This PR fixes the guide to build an agent programmatically. It also adds a flag to an AEA/Agent which lets the developer skip the loading from resource directory stage.

## Fixes

#619 

## Types of changes

## Checklist

## Further comments

-

### DIFF
--- a/aea/aea.py
+++ b/aea/aea.py
@@ -48,7 +48,7 @@ class AEA(Agent):
                  loop: Optional[AbstractEventLoop] = None,
                  timeout: float = 0.0,
                  debug: bool = False,
-                 programmatic: bool = False,
+                 programmatic: bool = True,
                  max_reactions: int = 20,
                  executor: Optional[Executor] = None) -> None:
         """
--- a/aea/aea.py
+++ b/aea/aea.py
@@ -48,6 +48,7 @@ class AEA(Agent):
                  loop: Optional[AbstractEventLoop] = None,
                  timeout: float = 0.0,
                  debug: bool = False,
+                 programmatic: bool = False,
                  max_reactions: int = 20,
                  executor: Optional[Executor] = None) -> None:
         """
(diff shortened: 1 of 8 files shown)


aea/aea.py
```diff
@@ -48,6 +48,7 @@ def __init__(self, name: str,
                  loop: Optional[AbstractEventLoop] = None,
                  timeout: float = 0.0,
                  debug: bool = False,
+                 programmatic: bool = True,
```
panasevychol: I prefer more self explaining naming (`is_programmatic` in this case).

tests/test_aea.py
```diff
@@ -72,13 +73,15 @@ def test_act():
         ledger_apis = LedgerApis({}, 'default')
         address = wallet.addresses['default']
         connections = [OEFLocalConnection(address, node)]
+        resources = Resources(str(Path(CUR_PATH, "data", "dummy_aea")))
```
panasevychol: Could be replaced with `os.path.join(...)`. Returns str so doesn't require an extra action - to lead it to a str.

panasevychol: If that's a common action (I've seen this before) I would request to upgrade `Path` with `to_str()` method.

tests/test_skills/test_base.py
```diff
@@ -61,7 +61,7 @@ def setup_class(cls):
         cls.wallet = Wallet({'default': private_key_pem_path, FETCHAI: private_key_txt_path})
         cls.ledger_apis = LedgerApis({FETCHAI: ["alpha.fetch-ai.com", 80]}, FETCHAI)
         cls.connections = [DummyConnection()]
-        cls.my_aea = AEA("Agent0", cls.connections, cls.wallet, cls.ledger_apis, resources=Resources(str(Path(CUR_PATH, "data", "dummy_aea"))))
+        cls.my_aea = AEA("Agent0", cls.connections, cls.wallet, cls.ledger_apis, resources=Resources(str(Path(CUR_PATH, "data", "dummy_aea"))), programmatic=False)
```
panasevychol: It's quite a common case.

tests/test_aea.py
DavidMinarsch: Thanks, I agree. But this is applied throughout the entire project. So I won't change it in this PR.

aea/aea.py
DavidMinarsch: Good point. I will open an issue as we should change it systematically for all variables of boolean type.

Issue #585, opened 2020-01-02
Carpark demo on ethereum


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
Currently, the carpark demo does not run on ethereum.

**Describe the solution you'd like**
Extend it to run on ethereum as well.

PR #655, opened 2020-01-20
[wip] AEA-159 AEA-247 Verify transaction for ethereum network.

### LINKED ISSUE #574: Check tx_digest against blockchain
Currently, the tx_digest is not used to verify the transaction by the receiving party.

The receiving party should check that the tx_digest is a valid digest of a settled tx and that the transaction references the correct terms. The receiving agent could ask the sending agent to incorporate a nonce-like string into the tx data so the agent can match txs to terms.

Totoual:
## Proposed changes

Adds a random_message generator and a function to validate an ethereum transaction digest.

## Fixes

#574
#585 

## Types of changes

## Checklist


aea/crypto/ethereum.py
```diff
@@ -238,3 +238,15 @@ def is_transaction_settled(self, tx_digest: str) -> bool:
         if tx_status is not None:
             is_successful = True
         return is_successful
+
+    def validate_transaction(self, tx_digest: str, random_message: str) -> bool:
```
DavidMinarsch: This should be part of the ledger api class in my opinion as it is related to the ledger interaction, not the crypto/account management

aea/crypto/helpers.py
```diff
@@ -143,3 +146,27 @@ def _create_ethereum_private_key() -> None:
     account = Account.create()
     with open(ETHEREUM_PRIVATE_KEY_FILE, "w+") as file:
         file.write(account.key.hex())
+
+
+def _generate_ethereum_random_message(
```
DavidMinarsch: This can be used for both ledgers, so I would drop the ethereum reference.

packages/fetchai/skills/carpark_client/behaviours.py
```diff
@@ -41,21 +43,40 @@ def __init__(self, **kwargs):
...
+                        self.context.agent_name, fet_balance
+                    )
                 )
+            else:
+                logger.warning(
+                    "[{}]: you have no starting balance on fetchai ledger!".format(
+                        self.context.agent_name
+                    )
+                )
+                # TODO: deregister skill from filter
```
DavidMinarsch: please remove the TODO here

packages/fetchai/skills/carpark_client/handlers.py
```diff
@@ -211,35 +210,44 @@ def _handle_match_accept(self, msg: FIPAMessage, dialogue: Dialogue) -> None:
...
-            tx_quantities_by_good_id={},
-            ledger_id=proposal.values["ledger_id"],
-            info={"dialogue_label": dialogue.dialogue_label.json},
-        )
-        self.context.decision_maker_message_queue.put_nowait(tx_msg)
-        logger.info(
-            "[{}]: proposing the transaction to the decision maker. Waiting for confirmation ...".format(
-                self.context.agent_name
+            info = msg.info
+            logger.info(info.get("random_message"))
```
DavidMinarsch: see above

packages/fetchai/skills/carpark_client/strategy.py
```diff
@@ -103,3 +111,19 @@ def is_acceptable_proposal(self, proposal: Description) -> bool:
...
         return result
+
+    def is_affordable_proposal(self, proposal: Description) -> bool:
+        """
+        Check whether it is an affordable proposal.
+
+        :return: whether it is affordable
+        """
+        if self.is_ledger_tx:
+            payable = proposal.values["price"] + self._max_price
```
DavidMinarsch: why is payable the sum of these two?

aea/crypto/helpers.py
```diff
@@ -27,7 +27,10 @@
 
 from fetchai.ledger.crypto import Entity  # type: ignore
 
+from web3 import Web3
```
DavidMinarsch: Given that this depends on web3, I am tempted to move this helper function into the ethereum crypto module as a standalone function. Also since we expect it to be used in skills drop the `_` from the name and align the naming to the `tx_nonce`

packages/fetchai/skills/carpark_client/behaviours.py
DavidMinarsch: it's a duplicate of the same todo below

packages/fetchai/skills/carpark_client/strategy.py
DavidMinarsch: it should just be the price imo

packages/fetchai/skills/carpark_detection/handlers.py
```diff
@@ -244,12 +246,23 @@ def _handle_accept(self, msg: FIPAMessage, dialogue: Dialogue) -> None:
                 self.context.agent_name, msg.counterparty[-5:]
             )
         )
+        proposal = cast(Description, dialogue.proposal)
+        identifier = cast(str, proposal.values.get("ledger_id"))
+        self.context.ledger_apis.random_message = _generate_random_message(
+            nonce=new_message_id,
```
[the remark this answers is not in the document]
DavidMinarsch: but this comment is somewhat irrelevant if we drop the `nonce` field

packages/fetchai/skills/carpark_detection/handlers.py
```diff
@@ -244,12 +246,23 @@ def _handle_accept(self, msg: FIPAMessage, dialogue: Dialogue) -> None:
...
+            seller=self.context.agent_address,
+            client=msg.counterparty,
+            time_stamp=int(time.time()),
```
DavidMinarsch: note, you don't need time_stamp as an argument, you can just generate the time stamp in the helper method.

packages/fetchai/skills/weather_station/handlers.py
```diff
@@ -229,12 +231,21 @@ def _handle_accept(self, msg: FIPAMessage, dialogue: Dialogue) -> None:
         )
         proposal = cast(Description, dialogue.proposal)
         identifier = cast(str, proposal.values.get("ledger_id"))
+        self.context.ledger_apis.random_message = _generate_random_message(
```
DavidMinarsch: drop the assignment

aea/crypto/base.py
```diff
@@ -165,3 +166,14 @@ def is_transaction_settled(self, tx_digest: str) -> bool:
         :param tx_digest: the digest associated to the transaction.
         :return: True if the transaction has been settled, False o/w.
         """
+
+    @abstractmethod
+    def validate_transaction(self, tx_digest: str, proposal_msg: Message) -> bool:
```
DavidMinarsch: Why are we passing the entire message? I think we should pass the values we need so we don't create unnecessary coupling between these modules.

aea/crypto/ethereum.py
```diff
@@ -134,6 +137,22 @@ def _generate_private_key(self) -> Account:
         account = Account.create()
         return account
 
+    def generate_tx_nonce(self, seller: Address, client: Address) -> str:
+        """
+        Generate a random str message in order to validate a transaction.
```
DavidMinarsch: `Generate a random str message` > `Generate a random str`

aea/crypto/ethereum.py
```diff
@@ -247,3 +264,17 @@ def is_transaction_settled(self, tx_digest: str) -> bool:
         if tx_status is not None:
             is_successful = True
         return is_successful
+
+    def validate_transaction(self, tx_digest: str, proposal_msg: Message) -> bool:
```
DavidMinarsch: see above, I don't like the dependency on `Message` here

aea/crypto/fetchai.py
```diff
@@ -202,3 +203,14 @@ def is_transaction_settled(self, tx_digest: str) -> bool:
             # TODO: check the amount of the transaction is correct
             is_successful = True
         return is_successful
+
+    def validate_transaction(self, tx_digest: str, proposal_msg: Message) -> bool:
```
DavidMinarsch: see above

aea/crypto/ledger_apis.py
```diff
@@ -194,23 +192,26 @@ def is_tx_settled(self, identifier: str, tx_digest: str) -> bool:
             self._last_tx_statuses[identifier] = ERROR
         return is_successful
 
-    def validate_transaction(self, identifier: str, tx_digest: str) -> bool:
+    def is_tx_valid(
+        self, identifier: str, tx_digest: str, proposal_msg: Message
```
DavidMinarsch: same here, this dependence on `Message` is not ideal

packages/fetchai/protocols/fipa/fipa.proto
```diff
@@ -15,6 +15,7 @@ message FIPAMessage{
     }
     message Propose{
         repeated bytes proposal = 1;
+        bytes tx_nonce=2;
```
DavidMinarsch: no, the tx_nonce should be part of the proposal itself in my opinion. so no change to proto file needed

packages/fetchai/protocols/fipa/message.py
```diff
@@ -111,6 +111,12 @@ def proposal(self) -> List[Description]:
         assert self.is_set("proposal"), "proposal is not set."
         return cast(List[Description], self.get("proposal"))
 
+    @property
```
DavidMinarsch: Please remove the dependency on tx_nonce and vice versa here. `tx_nonce` should be a property of the proposal.

packages/fetchai/protocols/fipa/serialization.py
```diff
@@ -64,6 +64,8 @@ def encode(self, msg: Message) -> bytes:
             proposal = msg.proposal
             p_array_bytes = [pickle.dumps(p) for p in proposal]
             performative.proposal.extend(p_array_bytes)
+            tx_nonce = pickle.dumps(msg.tx_nonce)
```
DavidMinarsch: see above

packages/fetchai/skills/weather_station/handlers.py
```diff
@@ -285,11 +274,8 @@ def _handle_inform(self, msg: FIPAMessage, dialogue: Dialogue) -> None:
             proposal = cast(Description, dialogue.proposal)
             ledger_id = cast(str, proposal.values.get("ledger_id"))
             is_settled = self.context.ledger_apis.is_tx_settled(ledger_id, tx_digest)
-            is_valid = self.context.ledger_apis.validate_transaction(
```
DavidMinarsch: we should discuss why this has been removed here

tests/data/dummy_aea/vendor/fetchai/protocols/fipa/fipa.proto
```diff
@@ -15,6 +15,7 @@ message FIPAMessage{
     }
     message Propose{
         repeated bytes proposal = 1;
+        bytes tx_nonce=2;
```
DavidMinarsch: see above, not needed

packages/fetchai/skills/weather_station/handlers.py
Totoual: I just removed it for now just to be easier for review. Once I have the correct function validate_transaction() I ll update the packages to use it.

aea/crypto/base.py
```diff
@@ -168,11 +167,11 @@ def is_transaction_settled(self, tx_digest: str) -> bool:
         """
 
     @abstractmethod
-    def validate_transaction(self, tx_digest: str, proposal_msg: Message) -> bool:
+    def validate_transaction(self, tx_digest: str, proposal: Dict[str, Any]) -> bool:
```
DavidMinarsch: pass in addresses and tx_nonce

aea/crypto/ethereum.py
```diff
@@ -265,16 +262,15 @@ def is_transaction_settled(self, tx_digest: str) -> bool:
             is_successful = True
         return is_successful
 
-    def validate_transaction(self, tx_digest: str, proposal_msg: Message) -> bool:
+    def validate_transaction(self, tx_digest: str, proposal: Dict[str, Any]) -> bool:
```
DavidMinarsch: see above

aea/crypto/ethereum.py
```diff
@@ -265,16 +262,15 @@ def is_transaction_settled(self, tx_digest: str) -> bool:
...
 
-        :param proposal_msg:
-        :param identifier: the ledger identifier.
+        :param proposal: the proposal we did to the counterparty
         :param tx_digest: the transaction digest.
 
         :return: True if the random_message is equals to tx['input']
         """
-        msg = cast(FIPAMessage, proposal_msg)
+
```
DavidMinarsch: assert sender address is correct & transfer amount is correct

aea/crypto/fetchai.py
```diff
@@ -204,11 +203,11 @@ def is_transaction_settled(self, tx_digest: str) -> bool:
             is_successful = True
         return is_successful
 
-    def validate_transaction(self, tx_digest: str, proposal_msg: Message) -> bool:
+    def validate_transaction(self, tx_digest: str, proposal: Dict[str, Any]) -> bool:
```
DavidMinarsch: see above

aea/crypto/ledger_apis.py
```diff
@@ -193,19 +192,21 @@ def is_tx_settled(self, identifier: str, tx_digest: str) -> bool:
         return is_successful
 
     def is_tx_valid(
-        self, identifier: str, tx_digest: str, proposal_msg: Message
+        self, identifier: str, tx_digest: str, proposal: Dict[str, Any]
```
DavidMinarsch: see above

DavidMinarsch:
I'd suggest to, in the skills, recover the `tx_nonce` from the `proposal` (`proposal = cast(Description, dialogue.proposal)`). Then at the point where you generate the `proposal = strategy.generate_proposal`, i.e. the `generate_proposal` method should assign the `tx_nonce` to the proposal (`Description` object).

packages/fetchai/skills/weather_station/strategy.py
```diff
@@ -92,14 +93,19 @@ def is_matching_supply(self, query: Query) -> bool:
...
         Generate a proposal matching the query.
 
         :param query: the query
         :return: a tuple of proposal and the weather data
         """
+        tx_nonce = self.context.ledger_apis.generate_tx_nonce(
+            identifier=self._ledger_id,
+            seller=self.context.agent_addresses[self._ledger_id],
+            client=counterparty,
+        )
```
Totoual: We generate the tx_nonce inside the strategy before the proposal.

packages/fetchai/skills/weather_station/strategy.py
```diff
@@ -115,9 +121,10 @@ def generate_proposal_and_data(
                 "seller_tx_fee": self._seller_tx_fee,
                 "currency_id": self._currency_id,
                 "ledger_id": self._ledger_id,
+                "tx_nonce": tx_nonce if tx_nonce is not None else "",
```
Totoual: Ledger_apis can return None in case it fails for some reason. There is a try / except block in the implementation.

packages/fetchai/skills/weather_station/handlers.py
```diff
@@ -274,8 +276,16 @@ def _handle_inform(self, msg: FIPAMessage, dialogue: Dialogue) -> None:
...
             ledger_id = cast(str, proposal.values.get("ledger_id"))
             is_settled = self.context.ledger_apis.is_tx_settled(ledger_id, tx_digest)
+            is_valid = self.context.ledger_apis.is_tx_valid(
+                ledger_id,
+                tx_digest,
+                self.context.agent_addresses[ledger_id],
+                msg.counterparty,
+                cast(str, proposal.values.get("tx_nonce")),
+                cast(int, proposal.values.get("price")),
+            )
```
Totoual: Here we are checking if it is_valid and it is_settled.

aea/crypto/base.py
```diff
@@ -165,3 +165,34 @@ def is_transaction_settled(self, tx_digest: str) -> bool:
...
+        """
+        Check whether a transaction is valid or not.
+
+        :param seller: the address of the seller.
+        :param client: the address of the client.
+        :param tx_nonce: the transaction nonce.
+        :param amount: the amount we expect to get from the transaction.
+        :param tx_digest: the transaction digest.
+
+        :return: True if the random_message is equals to tx['input']
```
DavidMinarsch: `:return: True if the transaction referenced by the tx_digest matches the terms`

aea/crypto/ethereum.py
```diff
@@ -192,7 +192,7 @@ def send_transaction(
         destination_address: AddressLike,
         amount: int,
         tx_fee: int,
-        chain_id: int = 1,
+        chain_id: int = 3,
```
DavidMinarsch: I think we should add tx_nonce as a keyword argument here

aea/crypto/ethereum.py
```diff
@@ -216,9 +217,12 @@ def send_transaction(
             "value": amount,
             "gas": tx_fee,
             "gasPrice": self._api.toWei(GAS_PRICE, GAS_ID),
+            "data": kwargs.get("tx_nonce"),
```
DavidMinarsch: see above

aea/crypto/ethereum.py
```diff
@@ -238,3 +242,47 @@ def is_transaction_settled(self, tx_digest: str) -> bool:
         if tx_status is not None:
             is_successful = True
         return is_successful
+
+    def generate_tx_nonce(self, seller: Address, client: Address) -> str:
+        """
+        Generate a random str message.
```
DavidMinarsch: `Generate a unique hash to distinguish txs with the same terms.

aea/crypto/ethereum.py
```diff
@@ -238,3 +242,47 @@ def is_transaction_settled(self, tx_digest: str) -> bool:
...
+        :param client: the address of the client.
+        :param tx_nonce: the transaction nonce.
+        :param amount: the amount we expect to get from the transaction.
+        :param tx_digest: the transaction digest.
+
+        :return: True if the random_message is equals to tx['input']
+        """
+
+        tx = self._api.eth.getTransaction(tx_digest)
+        return (
```
DavidMinarsch: Just in terms of style, I'd assign the result to a variable called `result` or `is_valid` and then have the last line `return is_valid` to make it more readable

aea/crypto/fetchai.py
```diff
@@ -202,3 +202,36 @@ def is_transaction_settled(self, tx_digest: str) -> bool:
...
+
+        :param seller: the address of the seller.
+        :param client: the address of the client.
+        :param tx_nonce: the transaction nonce.
+        :param amount: the amount we expect to get from the transaction.
+        :param tx_digest: the transaction digest.
+
+        :return: True if the random_message is equals to tx['input']
+        """
+
```
DavidMinarsch: the ledger api allows you to recover a list of transactions as well. It would be good if you could have a look at implementing it for fetch.ai too

aea/crypto/fetchai.py
```diff
@@ -202,3 +202,36 @@ def is_transaction_settled(self, tx_digest: str) -> bool:
...
+    def generate_tx_nonce(self, seller: Address, client: Address) -> str:
+        """
+        Generate a random str message.
+
+        :param seller: the address of the seller.
+        :param client: the address of the client.
+        :return: return the hash in hex.
+        """
+
```
DavidMinarsch: the implementation can mirror the one from ethereum, use https://github.com/fetchai/ledger-api-py/blob/b000d20f1375f1b2b2b79ace3616efba18640020/fetchai/ledger/serialisation/sha256.py

aea/crypto/fetchai.py
DavidMinarsch: use: https://github.com/fetchai/ledger-api-py/blob/master/fetchai/ledger/api/tx.py#L113

aea/decision_maker/base.py
```diff
@@ -710,6 +710,8 @@ def _settle_tx(self, tx_message: TransactionMessage) -> Optional[str]:
                 tx_message.tx_counterparty_addr,
                 tx_message.counterparty_amount,
                 tx_message.fees,
+                info=tx_message.info,
+                tx_nonce=tx_message.get("tx_nonce"),
```
DavidMinarsch: see below

aea/decision_maker/messages/transaction.py
```diff
@@ -279,7 +279,7 @@ def check_consistency(self) -> bool:
                 self.Performative.REJECTED_SETTLEMENT,
                 self.Performative.FAILED_SETTLEMENT,
             }:
-                assert len(self.body) == 11
+                assert len(self.body) == 11 or len(self.body) == 12
```
DavidMinarsch: No we should not change that. Please add the `tx_nonce` to the `info` dictionary.

DavidMinarsch: That way the `TransactionMessage` remains unchanged and you can do `tx_message.info.get("tx_nonce")`

packages/fetchai/protocols/fipa/fipa_pb2.py
```diff
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 # Generated by the protocol buffer compiler.  DO NOT EDIT!
```
DavidMinarsch: Please revert this file so it shows no change.

packages/fetchai/skills/carpark_client/handlers.py
```diff
@@ -211,35 +210,40 @@ def _handle_match_accept(self, msg: FIPAMessage, dialogue: Dialogue) -> None:
...
+                skill_callback_ids=["carpark_client"],
+                tx_id="transaction0",
+                tx_sender_addr=self.context.agent_addresses[
+                    proposal.values["ledger_id"]
+                ],
+                tx_counterparty_addr=address,
+                tx_amount_by_currency_id={
+                    proposal.values["currency_id"]: -proposal.values["price"]
+                },
+                tx_sender_fee=0,
```
DavidMinarsch: Check with the weather skills what we did there, but I don't think these should be zero

packages/fetchai/skills/carpark_client/handlers.py
```diff
@@ -211,35 +210,40 @@ def _handle_match_accept(self, msg: FIPAMessage, dialogue: Dialogue) -> None:
...
+                ledger_id=proposal.values["ledger_id"],
+                info={"dialogue_label": dialogue.dialogue_label.json, },
+            )
+            logger.info(tx_msg)
+            self.context.decision_maker_message_queue.put_nowait(tx_msg)
+            logger.info(
+                "[{}]: proposing the transaction to the decision maker. Waiting for confirmation ...".format(
+                    self.context.agent_name
+                )
             )
```
DavidMinarsch: you're missing the `else` case...

packages/fetchai/skills/carpark_client/skill.yaml
```diff
@@ -29,6 +29,9 @@ shared_classes:
       no_find_search_interval: 5
       max_price: 400000000
       max_detection_age: 36000000
+      currency_id: 'FET'
+      ledger_id: 'fetchai'
+      is_ledger_tx: False
```
DavidMinarsch: Hm I think the default should be True for `is_ledger_tx`

packages/fetchai/skills/carpark_client/strategy.py
```diff
@@ -103,3 +111,19 @@ def is_acceptable_proposal(self, proposal: Description) -> bool:
...
+        :return: whether it is affordable
+        """
+        if self.is_ledger_tx:
+            payable = proposal.values["price"]
+            ledger_id = proposal.values["ledger_id"]
+            address = cast(str, self.context.agent_addresses.get(ledger_id))
+            balance = self.context.ledger_apis.token_balance(ledger_id, address)
+            result = balance >= payable
+        else:
+            result = True
```
DavidMinarsch: add a logger.debug("Assuming it is affordable") or something like that

packages/fetchai/skills/carpark_detection/strategy.py
```diff
@@ -37,8 +37,8 @@
 DEFAULT_PRICE = 2000
 DEFAULT_DB_IS_REL_TO_CWD = False
 DEFAULT_DB_REL_DIR = "temp_files_placeholder"
-DEFAULT_CURRENCY_ID = "FET"
-DEFAULT_LEDGER_ID = "fetchai"
+DEFAULT_CURRENCY_ID = "ETH"
```
DavidMinarsch: default should always be FET and fetchai

packages/fetchai/skills/carpark_detection/strategy.py
```diff
@@ -71,10 +71,8 @@ def __init__(self, **kwargs) -> None:
         else:
             db_dir = os.path.join(os.path.dirname(__file__), DEFAULT_DB_REL_DIR)
 
-        self.data_price_fet = (
-            kwargs.pop("data_price_fet")
-            if "data_price_fet" in kwargs.keys()
-            else DEFAULT_PRICE
+        self.data_price = (
+            kwargs.pop("data_price") if "data_price" in kwargs.keys() else DEFAULT_PRICE
```
DavidMinarsch: can abbreviate with `self.data_price = kwargs.pop("data_price_fet", DEFAULT_PRICE)`

packages/fetchai/skills/weather_station/handlers.py
```diff
@@ -274,8 +276,16 @@ def _handle_inform(self, msg: FIPAMessage, dialogue: Dialogue) -> None:
             proposal = cast(Description, dialogue.proposal)
             ledger_id = cast(str, proposal.values.get("ledger_id"))
             is_settled = self.context.ledger_apis.is_tx_settled(ledger_id, tx_digest)
```
DavidMinarsch: I think

DavidMinarsch: But let's do this last. not a priority

packages/fetchai/skills/weather_station/strategy.py
```diff
@@ -92,14 +93,19 @@ def is_matching_supply(self, query: Query) -> bool:
         return True
 
     def generate_proposal_and_data(
-        self, query: Query
+        self, query: Query, counterparty: Address
     ) -> Tuple[Description, Dict[str, List[Dict[str, Any]]]]:
         """
         Generate a proposal matching the query.
 
         :param query: the query
```
DavidMinarsch: missing kwarg in doc string.

packages/fetchai/skills/weather_station/strategy.py
DavidMinarsch: nice!

packages/fetchai/skills/weather_station/strategy.py
DavidMinarsch: Ok but if it fails, we should not send a proposal, so the proposal should be None and we need to adjust the logic to aboard the negotiation.

tests/data/dummy_aea/vendor/fetchai/protocols/fipa/fipa_pb2.py
```diff
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
```
DavidMinarsch: please revert, this should show no change

aea/crypto/ethereum.py
Totoual: The thing here is if I pass it as reference we will have to change a big part of the code to support the keyword. Since we will have to pass it from the handlers to decision maker and then ethereum crypto. The same way we are passing the info.

aea/crypto/fetchai.py
Totoual: I think that we should add fetchai part in a different PR

packages/fetchai/skills/carpark_client/handlers.py
Totoual: It's the same with the weather client.

aea/crypto/ledger_apis.py
```diff
@@ -188,6 +189,63 @@ def is_tx_settled(self, identifier: str, tx_digest: str) -> bool:
...
+        :param identifier: Ledger identifier
+        :param tx_digest:  the transaction digest
+        :param seller: the address of the seller.
+        :param client: the address of the client.
+        :param tx_nonce: the transaction nonce.
+        :param amount: the amount we expect to get from the transaction.
+        :return: True if is valid , False otherwise
+        """
+        assert identifier in self.apis.keys()
+        is_settled = self._is_tx_settled(identifier=identifier, tx_digest=tx_digest)
```
DavidMinarsch: Why do we need that? Can the tx_settled and tx_valid not be merged? I think there should be one function called `is_tx_settled_and_valid`. You really want to know both things and there should be only one call to the blockchain, not two.

DavidMinarsch: or `is_valid_settled_tx` might be more succinct.

packages/fetchai/skills/carpark_client/handlers.py
```diff
@@ -211,35 +210,62 @@ def _handle_match_accept(self, msg: FIPAMessage, dialogue: Dialogue) -> None:
...
-        self.context.decision_maker_message_queue.put_nowait(tx_msg)
-        logger.info(
-            "[{}]: proposing the transaction to the decision maker. Waiting for confirmation ...".format(
-                self.context.agent_name
+            info = msg.info
+            address = cast(str, info.get("address"))
+            proposal = cast(Description, dialogue.proposal)
+            tx_msg = TransactionMessage(
+                performative=TransactionMessage.Performative.PROPOSE_FOR_SETTLEMENT,
+                skill_callback_ids=["carpark_client"],
```
DavidMinarsch: this needs to be changed to `[PublicId("fetchai", "carpark_client", "0.1.0")]`

DavidMinarsch: Alternatively, merge develop as there have been a few other changes.

packages/fetchai/skills/weather_station/strategy.py
DavidMinarsch: See above outstanding comment, the proposal should only be created if `tx_nonce` is not None. But in general, I cannot see why it would ever be None. The return type of `generate_tx_nonce` is not Optional! So just have `"tx_nonce": tx_nonce`

aea/crypto/ethereum.py
DavidMinarsch: We should evaluate this together. For our usecases we will always want the tx_nonce to be set. Otherwise we cannot disambiguate. So having it in kwargs can cause issues with it being missing.

Pipfile
```diff
@@ -50,6 +50,7 @@ web3 = "==5.2.2"
 eth-account = "==0.4.0"
 fetchai-ledger-api = "==1.0.0rc1"
 fetch-p2p-api = {index = "https://test.pypi.org/simple/",version = "==0.0.2"}
+pytest = "*"
```
DavidMinarsch: please remove this here and regenerate lock `pip install --dev`

aea/crypto/ethereum.py
DavidMinarsch: In TransactionMessage I'd add
`
assert isinstance(self.tx_nonce, str), "Tx_nonce must be of type str."
assert len(self.body) == 12
`
then the skill can send the tx_nonce to the DM this way. Then in all the methods which require the tx_digest, we add it to the list of arguments.

PR #660, opened 2020-01-24
add cli commands generate-wealth get-wealth and get-address

### LINKED ISSUE #544: Update versions of sdks to test net
Currently, the sdk is used on alpha net.

Upgrade and point to test net/mainnet 
> swap everything to test net (including wealth creation);

DavidMinarsch:
## Proposed changes

This PR adds:
- cli command `aea generate-wealth [fetchai/ethereum]` which generates wealth on a defined test network
- cli command `aea get-wealth [fetchai/ethereum]` which gets the wealth amount of the agent on the ledger specified in ledger apis
- cli command `aea get-address [fetchai/ethereum/default]` which gets the address associated with the underlying private key
- adds docs re the above
- removes outdated scripts and docs references

## Fixes

#544 

## Types of changes

## Checklist

## Further comments

-


aea/cli/core.py
```diff
@@ -193,6 +197,100 @@ def add_key(ctx: Context, type_, file):
...
+    private_key_paths = dict(
+        [
+            (identifier, config.path)
+            for identifier, config in ctx.agent_config.private_key_paths.read_all()
+        ]
+    )
+    try:
+        wallet = Wallet(private_key_paths)
+        address = wallet.addresses[type_]
+        logger.info("Your address is: {}".format(address))
```
marcofavorito: I propose to remove the prefix `"Your address is:"`. It is useful to have only the address string as output, e.g. for piping of commands etc.

marcofavorito: or, having an optional flag to implement my feature.

docs/car-park-skills.md
```diff
@@ -83,10 +83,23 @@ aea generate-key fetchai
...
-cd ..
-python scripts/fetchai_wealth_generation.py --private-key car_data_buyer/fet_private_key.txt --amount 10000000000 --addr alpha.fetch-ai.com --port 80
+aea get-address fetchai
+```
+
+This will print the address to the console. Copy the address into the clipboard and request test tokens from the faucet [here](https://explore-testnet.fetch.ai/tokentap). It will take a while for the tokens to become available.
+
+Second, after some time, check the wealth associated with the address:
+``` bash
+aea get-wealth fetchai
```
marcofavorito: if I don't follow the step above (i.e. request of test tokens) I get 
```error: Unsupported ledger identifier.```

File: aea/cli/core.py [hunk 3]
```diff
@@ -193,6 +197,100 @@ def add_key(ctx: Context, type_, file):
...
+
+@cli.command()
+@click.argument(
+    "type_",
+    metavar="TYPE",
+    type=click.Choice([FetchAICrypto.identifier, EthereumCrypto.identifier]),
+    required=True,
+)
+@pass_ctx
+def generate_wealth(ctx: Context, type_):
```
marcofavorito: I'd like an option to specify the amount of wealth to generate

docs/car-park-skills.md
marcofavorito: sorry, I forgot the step above (i.e. populating `ledger_apis`).

However, after adding:
```yaml
ledger_apis:
  fetchai:
    address: testnet.fetch-ai.com
    port: 80
```

I get:
```error: Either network or addr and port must be specified.```

File: aea/cli/core.py (hunk 3 above)
Comment (DavidMinarsch) [reply aea/cli/core.py]: That's not possible. Faucets generate an amount determined by the faucet provider, that's industry standard.

File: aea/cli/core.py (hunk 1 above)
Comment (DavidMinarsch) [reply aea/cli/core.py]: Good point, let's just print the address

File: docs/car-park-skills.md (hunk 2 above)
Comment (DavidMinarsch) [reply docs/car-park-skills.md]: Will fix

File: aea/cli/core.py [hunk 4]
```diff
@@ -193,6 +197,100 @@ def add_key(ctx: Context, type_, file):
...
+    private_key_paths = dict(
+        [
+            (identifier, config.path)
+            for identifier, config in ctx.agent_config.private_key_paths.read_all()
+        ]
+    )
+    try:
+        wallet = Wallet(private_key_paths)
+        address = wallet.addresses[type_]
+        logger.info("{}".format(address))
```
marcofavorito: I think we should just use `print`, because this will prepend `info: `

Screenshot from 2020-01-25 12-07-02

aea/cli/core.py
marcofavorito: @DavidMinarsch what about an optional argument to wait until the confirmation?
e.g. `--sync`

aea/cli/core.py
```diff
@@ -193,6 +197,100 @@ def add_key(ctx: Context, type_, file):
...
+            [
+                (identifier, cast(Dict[str, Union[str, int]], config))
+                for identifier, config in ctx.agent_config.ledger_apis.read_all()
+            ]
+        )
+        ledger_apis = LedgerApis(ledger_api_configs, ctx.agent_config.default_ledger)
+
+        address = wallet.addresses[type_]
+        balance = ledger_apis.token_balance(type_, address)
+        logger.info("The wealth for address {} is: {}".format(address, balance))
```
marcofavorito: the same arguments about the prefix hold here.

This is the output I get:
Screenshot from 2020-01-25 12-14-18

This case is different because there's also the address, so the prefix of the output brings some useful information.

However, do I really need that? After all, I just want the amount of wealth in my wallet. The address is identified by the `type_` parameter, so I can do `aea get-address` and get it.

aea/cli/core.py
```diff
@@ -193,6 +197,100 @@ def add_key(ctx: Context, type_, file):
...
+
+@cli.command()
+@click.argument(
+    "type_",
+    metavar="TYPE",
+    type=click.Choice([FetchAICrypto.identifier, EthereumCrypto.identifier]),
+    required=True,
+)
+@pass_ctx
+def get_wealth(ctx: Context, type_):
```
marcofavorito: When will we support different addresses on the same network at the same time? So instead of `type_` as a choice, they should be among user-populated identifiers (e.g. `fetchai_1`, `fetchai_test` etc.), which are of type `fetchai`.

aea/crypto/helpers.py
```diff
@@ -143,3 +147,61 @@ def _create_ethereum_private_key() -> None:
...
+
+    :param identifier: the identifier of the ledger
+    :param address: the address to check for
+    :return: None
+    """
+    try:
+        import requests
+        import json
+
+        if identifier == FETCHAI:
```
marcofavorito: should this be implemented/defined as abstract methods in the `Crypto/LedgerApis` classes? So we avoid if-else pattern and it is more scalable.

docs/car-park-skills.md
marcofavorito: ok `"network": "testnet"` worked

marcofavorito: is there a way to make it work with the pair `addr` and `port` at the moment?

DavidMinarsch: No I couldn't get that to work and I looked at the available hosts and there seems to be some issues with them being resolved correctly

aea/cli/core.py
DavidMinarsch: Good point, will open an issue for that!

aea/cli/core.py
DavidMinarsch: We can start looking into this when it is a user request or we need it for our demos. For now I consider that YAGNI.

DavidMinarsch: Note, I see two cases:
- different node addresses on the same network
- different networks of the same type (e.g. Ropsten and Goerli and Mainnet)

aea/crypto/helpers.py
DavidMinarsch: No I don't think so. The issue is that these are highly specific implementations to the underlying test networks. There is no unified API for accessing faucets across different test networks on Ethereum.

aea/cli/core.py
DavidMinarsch: Feel free to open an issue @marcofavorito if you think this is important. I don't think so for the time being.

aea/cli/core.py
DavidMinarsch: https://github.com/fetchai/agents-aea/issues/663

PR #665, opened 2020-01-25
AEA-89 add docs updates and test fixes

DavidMinarsch:
## Proposed changes

This PR fixes the end-to-end tests of use-cases and updates the docs

## Fixes

-

## Types of changes

## Checklist


packages/fetchai/skills/ml_train/skill.yaml
```diff
@@ -19,7 +19,12 @@ handlers:
   transaction:
     class_name: MyTransactionHandler
     args: {}
-tasks: {}
+tasks:
```
DavidMinarsch: @marcofavorito I had to add this with to make it pass with your changes. Is that intentional?

docs/agent-vs-aea.md
```diff
@@ -0,0 +1,238 @@
...
+    os.remove(INPUT_FILE)
+if os.path.isfile(OUTPUT_FILE):
+    os.remove(OUTPUT_FILE)
+
+# Create a private key
+_create_default_private_key()
+
+# Set up the wallet, stub connection, ledger and (empty) resources
+wallet = Wallet({DEFAULT: DEFAULT_PRIVATE_KEY_FILE})
+stub_connection = StubConnection(input_file_path=INPUT_FILE, output_file_path=OUTPUT_FILE, connection_id=PublicId("fetchai", "stub", "0.1.0"))
```
DavidMinarsch: @marcofavorito I don't understand how this is an improvement over the previous API that we now need to pass the Public id into the connection. In my opinion, the StubConnection defines the public id, so i should not need to assign it...

docs/agent-vs-aea.md
```diff
@@ -0,0 +1,238 @@
...
+
+INPUT_FILE = "input.txt"
+OUTPUT_FILE = "output.txt"
+
+
+class MyAgent(Agent):
+
+    def __init__(self, name: str, connections: List[Connection], wallet: Wallet):
+        super().__init__(name, connections, wallet)
+        self.tick = 0
```
DavidMinarsch: @marcofavorito it feels like `tick` should be a base class variable /property as I can see it being relevant in many usecases. Thoughts?

docs/agent-vs-aea.md
```diff
@@ -0,0 +1,238 @@
+AEAs are more than just agents. In this guide we show some of the differences in terms of code.
```
DavidMinarsch: @marcofavorito I have added this guide as there is an ongoing confusion by some users as to what an AEA and what an agent is. An agent is anything which perceives and acts based on the definition we use. Now this is reflected in our agent base class `Agent`. The benefit of this class is that it gives you access to the protocols and connections concepts without having to know anything about skills and the decision maker. Once the user reaches limits with this approach (as we did in the TAC), they can move to the `AEA`. Thoughts?

DavidMinarsch: We can then, in the intro section, guide users towards the approach which best fits their usecase.

docs/agent-vs-aea.md
marcofavorito: This issue only emerges in the programmatic initialization of the connections.

The problem it tries to solve is the following. Let's say you want to customize the `StubConnection` and want its public id to be `david/stub:0.1.0`. You can literally copy-and-paste the code and just change the field 'author' of the `connection.yaml`. Earlier, instead, you had to change the part of the code that hard-codes the public id.

Both approaches can both live though. Nothing prevents us in `StubConnection.__init__` to have the call to `super().__init__` to include the hard-coded public id:
- with the programmatic initialization, the user does not care about the connection id (as long as the default value makes sense for his purposes)
- nothing changes for the initialization on the framework side (that is when we use `aea run`).

marcofavorito: ## TL;DR

add `super().__ini__(connection_id=PublicId("fetchai", "stub", "0.1.0")` in the constructor of `StubConnection`.

docs/agent-vs-aea.md
marcofavorito: agreed.

docs/agent-vs-aea.md
marcofavorito: yes, it is a good outlook. Worth providing some examples and tests of "raw" agents.

docs/agent-vs-aea.md
DavidMinarsch: Ok, not quite sure I agree with the idea that you'd want to allow for the above change of id with same underlying code but your quick fix works for me so I'll add that for now. To me, the `public_id` is very closely related to the code. In the future i'd like to have a one-to-one mapping between the `public_id` and an IPFS-like hash of the underlying implementation. That way, we can get higher integrity across the stack. I.e. an agent can use the public_id, recover the hash from a securely distributed lookup table, then fetch the package from the registry, initialize it and use it. So it makes sense then that the public id is hard coded in the object. What do you think?

marcofavorito: yes, we should definitely move toward that direction eventually (one-to-one mapping).

However, I think in the short term the current approach is less error-prone. Otherwise, we would have had to guarantee consistency between the configuration file `connection.yaml` and hard-coded public id in the Connection class. Instead, now the information is only in one place: `connection.yaml`.

marcofavorito: > you'd want to allow for the above change of id with same underlying code 

of course, it is not recommended to edit anything inside `vendor/`. But what I'd do if I were a user and wanted to slightly modify the implementation of say `StubConnection` is to copy it in custom `connections/`.

DavidMinarsch: Yeah sure, but then you could change the hard coded id as well :). I think what you say wrt to connection.yaml and connection.py consistency is key. For this hashing will help a lot. We can include an IPFS hash of all files but the `connection.yaml` file in a given folder the `connection.yaml`. Then at load time we check that the files are consistent. Then we have a second hash of the entire directory which will be used for the `public_id` to `hash` mapping. Thoughts?

docs/hacking-an-agent.md
```diff
@@ -145,76 +145,100 @@ If you just want to copy and past the entire script in you can find it here:
...
-t.join()
-t = None
+from aea.skills.base import Skill
 
+ROOT_DIR = "./"
+INPUT_FILE = "input.txt"
+OUTPUT_FILE = "output.txt"
+
+
+def run():
```
DavidMinarsch: @marcofavorito try running this script. I get an issue with the registry.

PR #674, opened 2020-01-27
AEA-307 Adding standard specification Types

5A11:
## Proposed changes

Adding standardised types for protocol specification. 
Converting these types to their equivalents in the target programming language (currently Python) 

## Types of changes

## Checklist

### DIFF
--- a/aea/protocols/generator.py
+++ b/aea/protocols/generator.py
@@ -19,13 +19,14 @@
 """This module contains the protocol generator."""
 
 import os
+import re
 from os import path
 from pathlib import Path
 from typing import Dict, List
 
 from aea.configurations.base import ProtocolSpecification
 
-DEFAULT_TYPES = ["int", "float", "bool", "str", "bytes", "list", "dict", "tuple", "set"]
+CUSTOM_TYPE_PATTERN = "[A-Z_][a-zA-Z0-9]*"
 
 MESSAGE_IMPORT = "from aea.protocols.base import Message"
 SERIALIZER_IMPORT = "from aea.protocols.base import Serializer"
@@ -66,6 +67,11 @@ class ProtocolGenerator:
         self.output_folder_path = os.path.join(output_path, protocol_specification.name)
 
     def _extract_all_contents(self) -> Dict[str, Dict[str, str]]:
+        """
+        Extract all performatives, content names and content types.
+
+        :return: Performatives -> [content names -> content types]
+        """
         all_contents = {}  # type: Dict[str, Dict[str, str]]
         for (
             performative,
@@ -75,6 +81,79 @@ class ProtocolGenerator:
             for content_name, content_type in speech_act_content_config.args.items():
                 all_contents[performative][content_name] = content_type
         return all_contents
+
+    def _specification_type_to_python_type(self, specification_type: str) -> str:
+        """
+        Convert a data type in protocol specification into its Python equivalent.
+
+        :param specification_type: the type described in the specification
+        :return: The Python equivalent
+        """
+        python_type = ""
+        if specification_type.startswith("pt:optional"):
+            element_type = specification_type[
+                specification_type.index("[") + 1 : specification_type.rindex("]")
+            ]
+            element_type_in_python = self._specification_type_to_python_type(
+                element_type
+            )
+            python_type = "Optional[{}]".format(element_type_in_python)
+        else:
+            specification_types = set(specification_type.split(" or "))
+            if len(specification_types) == 1:  # just one type (not a Union[])
+                if specification_type == "pt:bytes":
+                    python_type = "bytes"
+                elif specification_type == "pt:int":
+                    python_type = "int"
+                elif specification_type == "pt:float":
+                    python_type = "float"
+                elif specification_type == "pt:bool":
+                    python_type = "bool"
+                elif specification_type == "pt:str":
+                    python_type = "str"
+                elif specification_type.startswith(
+                    "pt:set"
+                ) or specification_type.startswith("pt:list"):
+                    element_type = specification_type[
+                        specification_type.index("[")
+                        + 1 : specification_type.rindex("]")
+                    ]
+                    element_type_in_python = self._specification_type_to_python_type(
+                        element_type
+                    )
+                    if specification_type.startswith("pt:set"):
+                        python_type = "FrozenSet[{}]".format(element_type_in_python)
+                    else:
+                        python_type = "Tuple[{}]".format(element_type_in_python)
+                elif specification_type.startswith("pt:dict"):
+                    element1_type = specification_type[
+                        specification_type.index("[")
+                        + 1 : specification_type.index(",")
+                    ]
+                    element2_type = specification_type[
+                        specification_type.index(",")
+                        + 1 : specification_type.rindex("]")
+                    ].strip()
+                    element1_type_in_python = self._specification_type_to_python_type(
+                        element1_type
+                    )
+                    element2_type_in_python = self._specification_type_to_python_type(
+                        element2_type
+                    )
+                    python_type = "Dict[{}, {}]".format(
+                        element1_type_in_python, element2_type_in_python
+                    )
+                else:
+                    python_type = specification_type
+            elif len(specification_types) > 1:  # has more than one type 'or' separated
+                python_type = "Union["
+                for t in specification_types:
+                    python_type += "{}, ".format(
+                        self._specification_type_to_python_type(t)
+                    )
+                python_type = python_type[:-2]
+                python_type += "]"
+        return python_type
 
     def _speech_acts_str(self) -> str:
         """
@@ -89,8 +168,14 @@ class ProtocolGenerator:
         ) in self.protocol_specification.speech_acts.read_all():
             speech_act_str += '        "{}": {{'.format(performative)
             if len(speech_act_content_config.args.items()) > 0:
-                for key, value in speech_act_content_config.args.items():
-                    speech_act_str += '"{}": {}, '.format(key, value)
+                for (
+                    content_name,
+                    content_type,
+                ) in speech_act_content_config.args.items():
+                    speech_act_str += '"{}": {}, '.format(
+                        content_name,
+                        self._specification_type_to_python_type(content_type),
+                    )
                 speech_act_str = speech_act_str[:-2]
             speech_act_str += "},\n"
         speech_act_str = speech_act_str[:-1]
@@ -104,7 +189,6 @@ class ProtocolGenerator:
         :return: the string containing class signatures and NotImplemented for every custom type
         """
         cls_str = ""
-        type_set = set()
         custom_types_set = set()
 
         # extract contents' types and separate custom types
@@ -113,9 +197,9 @@ class ProtocolGenerator:
             speech_act_content_config,
         ) in self.protocol_specification.speech_acts.read_all():
             for content_type in speech_act_content_config.args.values():
-                type_set.add(content_type)
-                if content_type not in DEFAULT_TYPES:
-                    custom_types_set.add(content_type)
+                custom_types_set.update(re.findall(CUSTOM_TYPE_PATTERN, content_type))
+                # if content_type not in DEFAULT_TYPES:
+                #     custom_types_set.add(content_type)
         # If no custom class, avoid extra spaces after last custom class
         if len(custom_types_set) == 0:
             return cls_str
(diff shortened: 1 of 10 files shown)


aea/protocols/generator.py
```diff
@@ -19,13 +19,14 @@
...
 import os
+import re
 from os import path
 from pathlib import Path
 from typing import Dict, List
 
 from aea.configurations.base import ProtocolSpecification
 
-DEFAULT_TYPES = ["int", "float", "bool", "str", "bytes", "list", "dict", "tuple", "set"]
+CUSTOM_TYPE_PATTERN = "[A-Z_][a-zA-Z0-9]*"
```
DavidMinarsch: I don't know why we you went for this regex. Care to elaborate?

aea/protocols/generator.py
```diff
@@ -66,6 +67,11 @@ def __init__(
         self.output_folder_path = os.path.join(output_path, protocol_specification.name)
 
     def _extract_all_contents(self) -> Dict[str, Dict[str, str]]:
+        """
+        Extract all performatives, content names and content types.
+
+        :return: Performatives -> [content names -> content types]
```
DavidMinarsch: maybe instead, Dict[performatives, Dict[content names, content types] ?

aea/protocols/generator.py
```diff
@@ -76,6 +82,79 @@ def _extract_all_contents(self) -> Dict[str, Dict[str, str]]:
                 all_contents[performative][content_name] = content_type
         return all_contents
 
+    def _specification_type_to_python_type(self, specification_type: str) -> str:
```
DavidMinarsch: This function would benefit from being broken up into shorter functions so it becomes more readable.

DavidMinarsch: For a start try splitting out each block in an `if, elif, else` and see what that gives you.

aea/protocols/generator.py
```diff
@@ -76,6 +82,79 @@ def _extract_all_contents(self) -> Dict[str, Dict[str, str]]:
...
+                specification_type.index("[") + 1 : specification_type.rindex("]")
+            ]
+            element_type_in_python = self._specification_type_to_python_type(
+                element_type
+            )
+            python_type = "Optional[{}]".format(element_type_in_python)
+        else:
+            specification_types = set(specification_type.split(" or "))
+            if len(specification_types) == 1:  # just one type (not a Union[])
+                if specification_type == "pt:bytes":
```
DavidMinarsch: seems like the next few lines could be significantly shortened by:
if specification_type in ["pt:bytes", ...., "pt:str"]
    python_type = specification_type[3:]

aea/protocols/generator.py
```diff
@@ -184,13 +269,14 @@ def _message_class_str(self) -> str:
 
         # Imports
         cls_str += "from enum import Enum\n"
-        cls_str += "from typing import Set, Tuple, cast\n\n"
+        cls_str += (
+            "from typing import Set, Tuple, cast, Union, Dict, Optional, FrozenSet\n\n"
```
DavidMinarsch: you should only be importing those types which are actually used.

aea/protocols/generator.py
```diff
@@ -274,7 +358,7 @@ def _message_class_str(self) -> str:
                     covered.append(content_name)
                 cls_str += "    @property\n"
                 cls_str += "    def {}(self) -> {}:\n".format(
-                    content_name, content_type
+                    content_name, self._specification_type_to_python_type(content_type)
```
DavidMinarsch: Rather than doing the conversion again and again, why not do it once on the top and then reuse it?

examples/protocol_specification_ex/sample.yaml
```diff
@@ -5,10 +5,10 @@ license: Apache 2.0
 description: 'A protocol for negotiation over a fixed set of resources involving two parties.'
 speech_acts:
   cfp:
-    query: DataModel
+    query: "DataModel"
```
DavidMinarsch: I am wondering whether it would be preferable to have a prefix for user defined types as well. Makes also the logic more easy to read. Something like, `st:DataModel`

examples/protocol_specification_ex/sample.yaml
```diff
@@ -5,10 +5,10 @@ license: Apache 2.0
...
   propose:
-    query: DataModel
-    price: float
+    query: "DataModel"
+    price: "pt:optional[pt:dict[DataModel, pt:bool] or pt:str or pt:dict[pt:int, Unit] or pt:set[pt:int] or pt:dict[_IOTApp, pt:bytes] or pt:list[Ali] or pt:bool]"
```
DavidMinarsch: I don't think we should permit class names starting with an underscore. See here: https://visualgit.readthedocs.io/en/latest/pages/naming_convention.html, the convention is UpperCaseCamelCase (see regex here: https://regexr.com/2rilj)

aea/protocols/generator.py
DavidMinarsch: Also, see comment further below.

5A11: Originally I found this https://stackoverflow.com/questions/10120295/valid-characters-in-a-python-class-name (see second answer for regex) and made some modifications. 

This format is something we can agree together. 

The format of class names from the official Python language reference is here: https://docs.python.org/3/reference/lexical_analysis.html#identifiers

But as you see that is very broad. 

A couple of options:

- [A-Z_][a-zA-Z0-9_]*
- [A-Z][a-zA-Z0-9_]*[a-zA-Z0-9]

After a bit of thinking, I am leaning towards the second regex above.

When choosing this, we should try to be language independent, so not necessarily something very Pythonic.

aea/protocols/generator.py

aea/protocols/generator.py

aea/protocols/generator.py

examples/protocol_specification_ex/sample.yaml
5A11: Yes good idea. 

Maybe "ct:BlaBla" for custom type?

examples/protocol_specification_ex/sample.yaml
5A11: See my first reply.

aea/protocols/generator.py

aea/protocols/generator.py
5A11: The code is recursive so hard to split. We probably end up with more code since we need another function that manages calls to smaller functions etc.

aea/protocols/generator.py
```diff
@@ -65,16 +66,146 @@ def __init__(
...
             speech_act_content_config,
         ) in self.protocol_specification.speech_acts.read_all():
-            all_contents[performative] = {}
+            self._all_performatives.add(performative)
+            self._speech_acts[performative] = {}
             for content_name, content_type in speech_act_content_config.args.items():
-                all_contents[performative][content_name] = content_type
-        return all_contents
+                custom_types = set(re.findall(CUSTOM_TYPE_PATTERN, content_type))
+                for ct in custom_types:
```
DavidMinarsch: please write out variable names, so `ct` -> `custom_type` in this case

examples/protocol_specification_ex/fipa.yaml
```diff
@@ -5,16 +5,16 @@ license: Apache 2.0
...
   cfp:
-    query: Query
+    query: ct:Query
   propose:
-    proposal: Proposal
+    proposal: ct:Proposal
     price: pt:float
   accept_w_inform:
-    info: pt:dict[pt:str, AcceptInfo]
+    info: pt:dict[pt:str, ct:AcceptInfo]
```
DavidMinarsch: For this protocol, I think we could have `ct:Info` for all three as this is what we actually need.

examples/protocol_specification_ex/ml_trade.yaml
```diff
@@ -5,15 +5,15 @@ license: Apache 2.0
 description: 'A protocol for trading data for training and prediction purposes.'
 speech_acts:
   cfp:
-    query: Query
+    query: ct:Query
   terms:
-    terms: Descrition
+    terms: ct:Descrition
```
DavidMinarsch: `Description` rather than `Descrition`

examples/protocol_specification_ex/output/two_party_negotiation/message.py
```diff
@@ -1,18 +1,33 @@
...
+class Unit:
+    """This class represents an instance of Unit."""
+
+    def __init__(self):
+        """Initialise an instance of Unit."""
+        raise NotImplementedError
+
+    def __eq__(self, other):
+        """Compare two Unit instances."""
+        if type(other) is type(self):
```
DavidMinarsch: I'd drop the if, else and just have `raise NotImplementedError`.

examples/protocol_specification_ex/output/two_party_negotiation/message.py
```diff
@@ -1,18 +1,33 @@
...
 class DataModel:
-    """This class represents a DataModel."""
+    """This class represents an instance of DataModel."""
 
     def __init__(self):
-        """Initialise a DataModel."""
+        """Initialise an instance of DataModel."""
         raise NotImplementedError
 
     def __eq__(self, other):
```
DavidMinarsch: see above

examples/protocol_specification_ex/output/two_party_negotiation/message.py
```diff
@@ -23,14 +38,57 @@ def __eq__(self, other):
...
+        else:
+            return False
+
+
 class TwoPartyNegotiationMessage(Message):
     """A protocol for negotiation over a fixed set of resources involving two parties."""
 
     protocol_id = "two_party_negotiation"
 
     _speech_acts = {
```
DavidMinarsch: I don't see why this variable needs to be part of the class. It's only used for its generation so can be dropped.

examples/protocol_specification_ex/output/two_party_negotiation/message.py
```diff
@@ -23,14 +38,57 @@ def __eq__(self, other):
...
+        if type(other) is type(self):
+            raise NotImplementedError
+        else:
+            return False
+
+
 class TwoPartyNegotiationMessage(Message):
     """A protocol for negotiation over a fixed set of resources involving two parties."""
 
     protocol_id = "two_party_negotiation"
```
DavidMinarsch: we changed this, it now needs to be `ProtocolId("author", "two_party_negotiation", "version") where `author` and `version` need to match the config

aea/protocols/generator.py
```diff
@@ -22,11 +22,11 @@
 import re
 from os import path
 from pathlib import Path
-from typing import Dict, List
+from typing import Dict, Set
 
 from aea.configurations.base import ProtocolSpecification
 
-CUSTOM_TYPE_PATTERN = "[A-Z_][a-zA-Z0-9]*"
+CUSTOM_TYPE_PATTERN = "ct:[A-Z_][a-zA-Z0-9]*"
```
DavidMinarsch: In a previous comment I mentioned that this is not the correct regex, please check and replace

examples/protocol_specification_ex/output/two_party_negotiation/message.py
5A11: Keep just __init__

examples/protocol_specification_ex/output/two_party_negotiation/message.py
5A11: Change to set of performatives.

aea/protocols/generator.py
5A11: [A-Z][a-zA-Z0-9]*

examples/protocol_specification_ex/sample.yaml
5A11: Done

aea/protocols/generator.py

examples/protocol_specification_ex/fipa.yaml
5A11: Done

examples/protocol_specification_ex/ml_trade.yaml
5A11: Done

examples/protocol_specification_ex/output/two_party_negotiation/message.py

examples/protocol_specification_ex/output/two_party_negotiation/message.py

examples/protocol_specification_ex/output/two_party_negotiation/message.py
5A11: I remember why we added `_speech_acts`.

It was needed for the shorter version of the code that checks the contents of a message via a loop that goes thought the content definitions in the `_speech_act` and checks the contents of the message match those in the definition in terms of name, number and type.

We could delete the `_speech_act` variable, but if we ever want to simplify the code, we need to put it back in.

examples/protocol_specification_ex/output/two_party_negotiation/message.py

aea/protocols/generator.py

PR #680, opened 2020-01-29
Refactoring attempt of task manager and tasks.

### LINKED ISSUE #635: Task should be submitted by the behaviours and handlers
Currently, tasks are called on each `tick`.

With the new `task` infrastructure we should change this so that tasks need to be enqueued with the task manager by either `handlers` or `behaviours`.

marcofavorito:
## Proposed changes

This PR changes the task manager API and the Task abstraction usage.

## Fixes

Fixes #635 

## Types of changes

## Checklist

## Further comments

This change implies many other changes. E.g. one among others, the task registry loses its usefulness.


aea/aea.py
```diff
@@ -20,7 +20,6 @@
 """This module contains the implementation of an Autonomous Economic Agent."""
 import logging
 from asyncio import AbstractEventLoop
-from concurrent.futures import Executor
```
marcofavorito: we are not using Executor for task manager anymore - `multiprocessing.Pool` gives much more control, especially on cancellation of workers.

aea/aea.py
```diff
@@ -217,9 +213,6 @@ def update(self) -> None:
 
         :return None
         """
-        # TODO: task should be submitted by the behaviours and handlers
-        for task in self.filter.get_active_tasks():
-            task.execute()
```
marcofavorito: this implies removing the task registry.

marcofavorito: and so the dynamic loading and insantiation of `*Task` classes in `skill_folder/tasks.py`.

aea/context/base.py
```diff
@@ -43,7 +43,6 @@ def __init__(
         ownership_state: OwnershipState,
         preferences: Preferences,
         goal_pursuit_readiness: GoalPursuitReadiness,
-        task_queue: Queue,
```
marcofavorito: there's no queue for task manager anymore - tasks can be enqueued by using the method `enqueue_task`

aea/skills/base.py
```diff
@@ -160,14 +160,6 @@ def ledger_apis(self) -> LedgerApis:
         """Get ledger APIs."""
         return self._agent_context.ledger_apis
 
-    @property
```
marcofavorito: see above

aea/context/base.py
marcofavorito: it is a step in the opposite direction of the actor model. However, until that, this is a much simpler solution. We don't need the 'dispatch' thread anymore.

aea/skills/base.py
```diff
@@ -201,11 +193,19 @@ def __init__(self, **kwargs):
         :param skill_context: the skill context
         :param kwargs: keyword arguments
         """
-        self._context = kwargs.pop("skill_context")  # type: SkillContext
-        self._config = kwargs
-        if "name" not in self._config:
+        try:
```
marcofavorito: this adds more control on the `SkillComponent` input. The demos might break after this change.

aea/skills/base.py
```diff
@@ -389,18 +389,70 @@ def parse_module(
 class Task(SkillComponent):
     """This class implements an abstract task."""
 
-    def __init__(self, *args, **kwargs):
+    def __init__(
```
marcofavorito: better having `name` and `skill_context` explicit if possible - to be extended to the other skill component classes eventually.

aea/skills/base.py
```diff
@@ -389,18 +389,70 @@ def parse_module(
...
+        self,
+        name: Optional[str] = None,
+        skill_context: Optional[SkillContext] = None,
```
marcofavorito: both `name` and `skill_context` are obsolete - especially the `skill_context`, since it is not necessary for all the use cases. 

At the end of the day, the user must be able to feed the data he needs to the `execute` function.

Also, `skill_context` is not pickable (due to the queue probably) - which prevents the `multiprocessing.Pool` working properly...

aea/skills/base.py
```diff
@@ -389,18 +389,70 @@ def parse_module(
...
-        :param skill_context: the skill context
+        :param name: the name of the task (mandatory)
+        :param skill_context: the skill context (mandatory)
         :param kwargs: keyword arguments.
         """
-        super().__init__(**kwargs)
-        self.completed = False
+        super().__init__(name=name, skill_context=skill_context, **kwargs)
+
+        self._executed = False
```
marcofavorito: the "functor" pattern is used.

`executed` says whether the function has already been executed.

aea/skills/base.py
```diff
@@ -389,18 +389,70 @@ def parse_module(
...
+        # this is where we store the result.
+        self._result = None
```
marcofavorito: `result` stores the computed result, if any -> accessed through the associated property. Raises error if the task has not been executed yet.

aea/skills/base.py
```diff
@@ -389,18 +389,70 @@ def parse_module(
...
+
+    def __call__(self, *args, **kwargs):
```
marcofavorito: the main entry point for the task. It makes the object `Task` a callable object.

aea/skills/tasks.py
```diff
@@ -18,72 +18,102 @@
...
-from typing import Optional
-
-from aea.skills.base import Task
+from multiprocessing.pool import AsyncResult, Pool
+from typing import Any, Callable, Dict, Optional, Sequence, cast
 
 logger = logging.getLogger(__name__)
 
 
+def init_worker():
```
marcofavorito: this is needed to disable the signal handler for `SIGINT` signals from the main process to children processes.

Without this, there is a bad behaviour of the `KeyboardInterrupt` handler, which is a known issue.

marcofavorito: It prevents graceful teardown of tasks, but at least we guarantee the pool does not hang... it's a trade-off.

aea/skills/tasks.py
```diff
@@ -18,72 +18,102 @@
...
+        self._results_by_task_id = {}  # type: Dict[int, Any]
+
+    @property
+    def nb_workers(self) -> int:
+        """Get the number of workers."""
+        return self._nb_workers
+
+    def enqueue_task(
+        self, func: Callable, args: Sequence = (), kwds: Optional[Dict[str, Any]] = None
+    ) -> AsyncResult:
```
marcofavorito: this is the main function to be used at skill code level to enqueue tasks.

The `TaskManager` sees only callables, so it is not coupled to the `Task` class and so it solves a more general problem.

marcofavorito: it is still useful to have a wrapper, so the user can use it instead of interacting with the "bare" `multiprocessing.Pool`.

aea/skills/tasks.py
```diff
@@ -18,72 +18,102 @@
...
-                return
-
-            future = self._executor.submit(next_task.execute)
-            self.futures.append(future)
+        with self._lock:
+            if self._stopped:
+                raise ValueError("Task manager not running.")
+            self._pool = cast(Pool, self._pool)
+            task_id = self._task_enqueued_counter
+            self._task_enqueued_counter += 1
```
marcofavorito: we assign a unique id to every submitted task. The consistency is guaranteed by the lock.

aea/skills/tasks.py
```diff
@@ -18,72 +18,102 @@
...
+            async_result = self._pool.apply_async(
+                func, args=args, kwds=kwds if kwds is not None else {}
+            )
+            self._results_by_task_id[task_id] = async_result
+            return async_result
```
marcofavorito: the return value is an instance of `AsyncResult` type (described here: https://docs.python.org/3.8/library/multiprocessing.html#multiprocessing.pool.AsyncResult).

It s an handle to access the result of the scheduled task/callable in an asynchronous way. The user can still recover it with the `get_task_result` and `task_id`.

TODO: the task_id should be returned by the `enqueue_task` method somehow. Maybe replace the `async_result` with the `task_id`, since the `async_result` can be recovered with `get_task_result`.

aea/skills/tasks.py
```diff
@@ -18,72 +18,102 @@
...
+
+    def get_task_result(self, task_id: int) -> AsyncResult:
```
marcofavorito: see above

aea/skills/tasks.py
```diff
@@ -18,72 +18,102 @@
...
-            self.thread.join()
-            self.thread = Thread(target=self.dispatch)
+        with self._lock:
+            if self._stopped is True:
+                logger.debug("Task manager already stopped.")
+            else:
+                logger.debug("Stop the task manager.")
+                self._stopped = True
+                self._pool = cast(Pool, self._pool)
+                self._pool.terminate()
```
marcofavorito: this is not a graceful termination because we overwrote the SIGINT handler. See above.

tests/data/exception_skill/behaviours.py
```diff
@@ -18,3 +18,22 @@
 # ------------------------------------------------------------------------------
 
 """This module contains the behaviours for the 'exception' skill."""
+from aea.skills.base import Behaviour
+
+
+class ExceptionBehaviour(Behaviour):
```
marcofavorito: replacement of `ExceptionTask`. It couldn't be used anymore - the `execute()` method is not called automatically by the main loop.

tests/data/exception_skill/tasks.py
```diff
@@ -18,23 +18,3 @@
 # ------------------------------------------------------------------------------
 
 """This module contains the tasks for the 'exception' skill."""
-
-from aea.skills.base import Task
-
-
-class ExceptionTask(Task):
```
marcofavorito: see above `ExceptionBehaviour`

tests/test_aea.py
```diff
@@ -306,12 +304,12 @@ def test_initialize_aea_programmatically(self):
...
         assert dummy_behaviour.nb_act_called > 0
 
-        dummy_task_name = "dummy"
-        dummy_task = self.aea.resources.task_registry.fetch(
-            (dummy_skill_id, dummy_task_name)
-        )
-        assert dummy_task is not None
-        assert dummy_task.nb_execute_called > 0
+        # TODO the previous code caused an error:
+        #      _pickle.PicklingError: Can't pickle <class 'tasks.DummyTask'>: import of module 'tasks' failed
```
marcofavorito: I mentioned this before. I can see it as a problem in all the cases where the `Task` object contains data that are non-pickable (e.g. `skill_context`), or that they are quite cumbersome in terms of memory (Gigabytes? probably not a problem in the short term).

aea/skills/base.py
marcofavorito: it returns the instance itself. `self`

marcofavorito: alternative: return the return value of `execute`

tests/test_task.py
```diff
@@ -0,0 +1,134 @@
...
+        expected_return_value = 42
+        my_task = MyTask(
+            return_value=expected_return_value, name="my_task", skill_context=object()
+        )
+        task_result = self.task_manager.enqueue_task(
+            my_task, args=expected_args, kwds=expected_kwargs
+        )
+        assert isinstance(task_result, AsyncResult)
+
+        expected_task = task_result.get(2.0)
```
marcofavorito: pay attention to this. `expected_task` is a *copy* of `my_task`. It might not be desirable in all cases.

BTW, worth mentioning that the result is indeed the `Task` copy because in the `Task.__call__` we return `self`.
Please see https://github.com/fetchai/agents-aea/pull/680/files#r372136135

tests/test_task.py
```diff
@@ -0,0 +1,134 @@
...
+        assert not my_task.executed
+        assert not my_task.setup_called
+        assert not my_task.execute_called
+        assert not my_task.teardown_called
+        assert my_task.execute_args is None
+        assert my_task.execute_kwargs is None
+
+    def test_task_manager_task_object_fails_when_not_pickable(self):
+        """Test task manager with task object fails when the task is not pickable."""
+        expected_args = [lambda x: x]
```
marcofavorito: the task argument in this test is a lambda, which is not pickable.

aea/registries/base.py
```diff
@@ -518,7 +518,9 @@ def __init__(self, directory: Optional[Union[str, os.PathLike]] = None):
         self.protocol_registry = ProtocolRegistry()
         self.handler_registry = HandlerRegistry()
         self.behaviour_registry = ComponentRegistry[Behaviour]()
-        self.task_registry = ComponentRegistry[Task]()
+        self.task_registry = ComponentRegistry[
```
DavidMinarsch: Unless we use it to keep track of the initialized tasks. But that doesn't really make sense as it limits the programmatic usage. So I think it can be removed.

aea/skills/base.py
DavidMinarsch: Ok so then why do you add it? Remove it :)

aea/skills/base.py
DavidMinarsch: you should mention the `functor` patterns in the docs

aea/skills/base.py
DavidMinarsch: This should be documented too

aea/skills/base.py
DavidMinarsch: ok that's fine

aea/skills/tasks.py
DavidMinarsch: Could also be mentioned in docs

aea/skills/tasks.py
DavidMinarsch: Worth adding a comment in code to have this info maintained

tests/test_aea.py
DavidMinarsch: Ok, add to docs imo

packages/fetchai/skills/gym/skill.yaml
```diff
@@ -8,13 +8,9 @@ behaviours: {}
 handlers:
   gym:
     class_name: GymHandler
-    args:
-      foo: bar
-tasks:
-  gym:
-    class_name: GymTask
```
marcofavorito: since the `Task` class is not loaded, there's no need to have this part in the configuration file.

packages/fetchai/skills/gym/handlers.py
```diff
@@ -41,6 +42,12 @@ def __init__(self, **kwargs):
         logger.info("GymHandler.__init__: arguments: {}".format(kwargs))
         super().__init__(**kwargs)
 
+        nb_steps = kwargs.get("nb_steps", DEFAULT_NB_STEPS)
+        self.task = GymTask(self.context, nb_steps)
+        # launch the task
+        self.task.setup()
+        self.task.execute()
```
marcofavorito: instead of submitting the task to the task manager, we instantiate the `GymTask`, which in turn create a thread (as before) to run the fit of the agent.

We can't submit this to the multiprocess version of the TaskManager, since the ProxyEnv.step() (called by the RLAgent.fit) cannot put messages to the outbox.

Issue #578, opened 2019-12-31
Integrate with remote registry


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
Currently, the system is not integrated with the remote registry.

**Describe the solution you'd like**
Integrate with the remote registry.

DavidMinarsch:
As part of this issue:
- [ ] ensure each command of the cli (https://fetchai.github.io/agents-aea/cli-commands/) works with the registry, make changes on registry and aea cli where necessary
- [ ] add login command to above list in `docs/cli-commands.md` and test it too
- [ ] once registry and aea cli are fully compatible, prepare a release on registry repo.

PR #683, opened 2020-01-29
AEA-311 AEA-264 AEA-243 Bugfix/registry compatibility

panasevychol:
## Fixes

- item unpacking path on add fixed (added `vendor` subpath)
- logic command call changed to `aea login USERNAME --password=[password]`
- tests fixed
- docs updated

#578 

## Types of changes

## Checklist

### DIFF
--- a/aea/cli/common.py
+++ b/aea/cli/common.py
@@ -379,7 +379,7 @@ def try_get_item_target_path(
 
     :return: the item target path
     """
-    target_path = os.path.join(path, author_name, item_type_plural, item_name)
+    target_path = os.path.join(path, "vendor", author_name, item_type_plural, item_name)
     if os.path.exists(target_path):
         raise click.ClickException(
             'Item "{}" already exists in target folder.'.format(item_name)
--- a/aea/cli/common.py
+++ b/aea/cli/common.py
@@ -366,7 +366,7 @@ def try_get_item_source_path(
     return source_path
 
 
-def try_get_item_target_path(
+def try_get_vendorized_item_target_path(
     path: str, author_name: str, item_type_plural: str, item_name: str
 ) -> str:
     """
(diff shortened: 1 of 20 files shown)


aea/cli/common.py
```diff
@@ -379,7 +379,7 @@ def try_get_item_target_path(
 
     :return: the item target path
     """
-    target_path = os.path.join(path, author_name, item_type_plural, item_name)
+    target_path = os.path.join(path, "vendor", author_name, item_type_plural, item_name)
```
DavidMinarsch: The function is called `try_get_item_target_path`. I am worried that this is ambiguous. It should probably be renamed to `try_get_vendorized_item_target_path` to make clear that this is only for items in the vendor folder.

aea/cli/registry/fetch.py
```diff
@@ -44,6 +44,7 @@ def fetch_agent(ctx: Context, public_id: PublicId) -> None:
     target_folder = os.path.join(ctx.cwd, name)
     os.makedirs(target_folder, exist_ok=True)
 
+    click.echo('Fetching dependencies...')
```
DavidMinarsch: I notice, that in some places we use click built in print functionality and in other places the logger.info. Can we unify this as part of this PR and use `click.echo` for all `logger.info`? @marcofavorito thoughts? I think `click.echo` might be preferable as it means the `info` bit is missing from the output.

panasevychol: I would prefer `click.echo` for messages that should reach user. Logging probably sould be used for dev related output.

aea/cli/common.py
panasevychol: But we came to a conclusion that target (the thing that we add/fetch) is always a published thing so it should go to `vendor` folder. Maybe that assumption is enough and we don't need that long naming in this case?)

aea/cli/registry/fetch.py
DavidMinarsch: Ok, so please apply `click.echo` throughout where `logger.info` is used atm. `logger.debug` of course remains.

aea/cli/common.py
DavidMinarsch: Well but another developer might easily confuse it and use it wrongly, I think in this case it would help to replace the names

PR #691, opened 2020-01-31
Adding test to cover 100%

Totoual:
## Proposed changes

Adding tests to get back to 100%

## Fixes

 - 

## Types of changes

## Checklist


tests/test_packages/test_skills/test_thermometer.py
```diff
@@ -142,16 +201,31 @@ def test_themrometer(self, pytestconfig):
                 "fetchai/oef:0.1.0",
             ],
             stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
             env=os.environ.copy(),
         )
 
-        # check the gym run ends
+        tty_read_thread = threading.Thread(target=_read_tty, args=(process_one,))
+        tty_read_thread.start()
```
DavidMinarsch: these threads are never stopped?

tests/test_packages/test_skills/test_thermometer.py
```diff
@@ -52,7 +64,7 @@ def setup_class(cls):
         cls.t = tempfile.mkdtemp()
         os.chdir(cls.t)
 
-    def test_themrometer(self, pytestconfig):
+    def test_weather(self, pytestconfig):
```
DavidMinarsch: this is the thermometer not the weather, check for occurrence throughout and replace

tests/test_packages/test_skills/test_thermometer.py
Totoual: These threads should stop automatically. At least this is what @marcofavorito told me :)

DavidMinarsch: Is that so @marcofavorito . If so, we should have a comment. It's not obvious that they must stop automatically and the next person reviewing this might be confused.

aea/cli/core.py
```diff
@@ -278,7 +278,7 @@ def _wait_funds_release(agent_config, wallet, type_):
     end_time = time.time() + FUNDS_RELEASE_TIMEOUT
     while time.time() < end_time:
         if start_balance != _try_get_balance(agent_config, wallet, type_):
-            break
+            break  # pragma: no cover
```
DavidMinarsch: please test the `sync` case as well

tests/test_crypto/test_helpers.py
```diff
@@ -55,3 +71,57 @@ def tests_private_keys(self):
...
+        result = Response()
+        result.status_code = 500
+        with patch.object(requests, "get", return_value=result):
+            _try_generate_testnet_wealth(identifier=ETHEREUM, address=address)
+            assert mock_logging.error.called
+
+        result.status_code = 200
+        with pytest.raises(SystemExit):
+            with patch.object(requests, "get", return_value=result):
+                _try_generate_testnet_wealth(identifier=ETHEREUM, address=address)
```
Totoual: Since you are using the ResponseMock() function to test the other parts of the _try_generate_testnet_wealth, should be better to use the same thing here? ( I mean don't mind to delete existing code if you have a better solution).

PR #703, opened 2020-02-04
AEA-353 Checking elements of compositional types

5A11:
Added type checking for compositional types.

## Types of changes

## Checklist


examples/protocol_specification_ex/output/two_party_negotiation/message.py
```diff
@@ -1,93 +1,54 @@
...
-            return False
-
 
 class TwoPartyNegotiationMessage(Message):
     """A protocol for negotiation over a fixed set of resources involving two parties."""
 
-    protocol_id = "two_party_negotiation"
+    protocol_id = ProtocolId("fetchai", "two_party_negotiation", "0.1.0")
 
     _speech_acts = {
```
DavidMinarsch: We had said previously we'd not store this here as only the keys are required...So if anything, store the keys.

examples/protocol_specification_ex/output/two_party_negotiation/message.py
```diff
@@ -163,35 +124,46 @@ def query(self) -> DataModel:
...
-            Dict[int, Unit],
-            FrozenSet[int],
-            Tuple[Ali],
-            bool,
-            str,
-            Dict[_IOTApp, bytes],
-            Dict[DataModel, bool],
-        ]
-    ]:
+    def number(self) -> int:
```
DavidMinarsch: Can we sort the properties by alphabet? I.e. first the block of meta-data ones by alphabet, then the ones specific to the protocol by alphabet.

examples/protocol_specification_ex/output/two_party_negotiation/message.py
```diff
@@ -97,11 +58,11 @@ class TwoPartyNegotiationMessage(Message):
     class Performative(Enum):
         """Performatives for the two_party_negotiation protocol."""
 
-        PROPOSE = "propose"
         DECLINE = "decline"
```
DavidMinarsch: should be sorted by alphabet

examples/protocol_specification_ex/output/two_party_negotiation/message.py

examples/protocol_specification_ex/output/two_party_negotiation/message.py

examples/protocol_specification_ex/output/two_party_negotiation/message.py
5A11: I replied to your previous comment as well.

The reason why we added _speech_acts was to have a shorter version of the code that checks the contents of a message (under ## Check correct contents). 

Currently we spell out all performatives (CFP, PROPOSE, etc) and under each we manually state how many contents there should be (expected_nb_of_contents), and then check the message has the correct number, name and type of contents.

I used to have an implementation where you loop through the contents in the _speech_act variable (for the performative entry used in the message) and check the contents of the message match those in _speech_acts in terms of name, number and type.

We could delete the _speech_act variable, but if we ever want to simplify the code, we need to put it back in.

DavidMinarsch: Yes I am aware of that discussion. Given that the current code is what it is, we should take it out/only keep what we need. At a later stage we can think about implementing a loop. Taking it out is quicker at this stage and fully functional.

PR #705, opened 2020-02-04
AEA-354 add identity object to agent and aea

### LINKED ISSUE #699: Create identity object for agent
Currently, the agent does not have this object.

Replace the name at the constructor level and pass in:
`agent_identity = AgentIdentity(agent_name=“agent_name”) (with optional addresses and public_key fields)`
`agent = Agent(agent_identity, [connection1, connection2])`
`aea = AEA(agent_identity, …)`

DavidMinarsch:
## Proposed changes

This PR introduces an `Identity` object. This is used by in the agent `Agent(identity, [connections])` and the AEA `AEA(identity, ...)`.

For now, it just contains the name and public key and address (if set). Later, it will also include other identity claims. It is introduced now to be forward compatible to these changes and not break the AEA and Agent constructor when they are extended.

## Fixes

#699 

## Types of changes

## Checklist

## Further comments

-

### DIFF
--- a/aea/aea.py
+++ b/aea/aea.py
@@ -22,7 +22,7 @@ import logging
 from asyncio import AbstractEventLoop
 from typing import List, Optional, cast
 
-from aea.agent import Agent
+from aea.agent import Agent, Identity
 from aea.connections.base import Connection
 from aea.context.base import AgentContext
 from aea.crypto.ledger_apis import LedgerApis
@@ -43,7 +43,7 @@ class AEA(Agent):
 
     def __init__(
         self,
-        name: str,
+        identity: Identity,
         connections: List[Connection],
         wallet: Wallet,
         ledger_apis: LedgerApis,
@@ -57,7 +57,7 @@ class AEA(Agent):
         """
         Instantiate the agent.
 
-        :param name: the name of the agent
+        :param identity: the identity of the agent
         :param connections: the list of connections of the agent.
         :param loop: the event loop to run the connections.
         :param wallet: the wallet of the agent.
@@ -71,7 +71,7 @@ class AEA(Agent):
         :return: None
         """
         super().__init__(
-            name=name,
+            identity=identity,
             connections=connections,
             loop=loop,
             timeout=timeout,
@@ -79,6 +79,8 @@ class AEA(Agent):
             programmatic=programmatic,
         )
 
+        self.identity.public_keys = wallet.public_keys
+        self.identity.addresses = wallet.addresses
         self._wallet = wallet
         self.max_reactions = max_reactions
         self._task_manager = TaskManager()
@@ -87,8 +89,8 @@ class AEA(Agent):
         )
         self._context = AgentContext(
             self.name,
-            self.wallet.public_keys,
-            self.wallet.addresses,
+            self.identity.public_keys,
+            self.identity.addresses,
             ledger_apis,
             self.multiplexer.connection_status,
             self.outbox,
--- a/aea/aea.py
+++ b/aea/aea.py
@@ -22,12 +22,13 @@ import logging
 from asyncio import AbstractEventLoop
 from typing import List, Optional, cast
 
-from aea.agent import Agent, Identity
+from aea.agent import Agent
 from aea.connections.base import Connection
 from aea.context.base import AgentContext
 from aea.crypto.ledger_apis import LedgerApis
 from aea.crypto.wallet import Wallet
 from aea.decision_maker.base import DecisionMaker
+from aea.identity.base import Identity
 from aea.mail.base import Envelope
 from aea.protocols.default.message import DefaultMessage
 from aea.registries.base import Filter, Resources
@@ -79,18 +80,13 @@ class AEA(Agent):
             programmatic=programmatic,
         )
 
-        self.identity.public_keys = wallet.public_keys
-        self.identity.addresses = wallet.addresses
-        self._wallet = wallet
         self.max_reactions = max_reactions
         self._task_manager = TaskManager()
         self._decision_maker = DecisionMaker(
-            self.name, self.max_reactions, self.outbox, self.wallet, ledger_apis
+            self.name, self.max_reactions, self.outbox, wallet, ledger_apis
         )
         self._context = AgentContext(
-            self.name,
-            self.identity.public_keys,
-            self.identity.addresses,
+            self.identity,
             ledger_apis,
             self.multiplexer.connection_status,
             self.outbox,
@@ -102,11 +98,6 @@ class AEA(Agent):
         )
         self._resources = resources
         self._filter = Filter(self.resources, self.decision_maker.message_out_queue)
-
-    @property
-    def wallet(self) -> Wallet:
-        """Get the wallet."""
-        return self._wallet
 
     @property
     def decision_maker(self) -> DecisionMaker:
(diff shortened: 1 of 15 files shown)


aea/aea.py
```diff
@@ -71,14 +71,16 @@ def __init__(
...
-            name=name,
+            identity=identity,
             connections=connections,
             loop=loop,
             timeout=timeout,
             debug=debug,
             programmatic=programmatic,
         )
 
+        self.identity.public_keys = wallet.public_keys
```
DavidMinarsch: we set them here as they are derived from the wallet.

aea/aea.py
```diff
@@ -71,14 +71,16 @@ def __init__(
...
+        self.identity.addresses = wallet.addresses
         self._wallet = wallet
```
DavidMinarsch: the wallet is now only used by the decision maker, as intended. we could therefore pass it to the decision maker and not keep it in the AEA.

aea/agent.py
```diff
@@ -54,12 +54,70 @@ def is_stopped(self) -> bool:
...
+
+    It can include:
+    - the agent name
+    - the public half of private / public key pairs
+    - the addresses
+    """
+
+    def __init__(self, name: str):
+        """
+        Instantiate the agent.
```
DavidMinarsch: Fix, needs to say `identity`

aea/agent.py
```diff
@@ -110,7 +173,7 @@ def outbox(self) -> OutBox:
     @property
     def name(self) -> str:
```
DavidMinarsch: I keep this for backwards compatibility of the usage `self.name` and because there is no cost to keep it.

docs/agent-vs-aea.md
```diff
@@ -199,7 +200,8 @@ def run():
     )
 
     # Create our Agent
-    my_agent = MyAgent("my_agent", [stub_connection])
+    identity = Identity("my_agent")
```
DavidMinarsch: the identity needs to be constructed outside the agent.

aea/aea.py
marcofavorito: I feel it is a bit counterintuitive. It means the `Identiy` object gets modified when passed to the constructor. This may lead to weird behaviour of the identity instance, that might be used outside the constructor.

What if the constructor accepts name, and the `Identity` object is instantiated and populated inside the constructor?

aea/agent.py
```diff
@@ -54,12 +54,70 @@ def is_stopped(self) -> bool:
...
+        self._addresses = addresses
+
+
 class Agent(ABC):
     """This class implements a template agent."""
 
     def __init__(
         self,
-        name: str,
+        identity: Identity,
```
marcofavorito: see above

aea/agent.py
```diff
@@ -54,12 +54,70 @@ def is_stopped(self) -> bool:
...
+    """
+    An identity are the public elements identifying an agent.
+
+    It can include:
+    - the agent name
+    - the public half of private / public key pairs
+    - the addresses
+    """
+
+    def __init__(self, name: str):
```
marcofavorito: I think the constructor should include public keys and addresses. It seems more natural, as the `Identity` is a data holder, and its content should be immutable, e.g. read-only. 

You can optionally have empty dictionaries as defaults, both for public keys and addresses. They can then be populated manually (by accessing the properties `public_keys` and `addresses`), or by invoking ad-hoc methods (e.g. `add_address`), that make some consistency check before adding/removing items, potentially raising `ValueError`.

docs/agent-vs-aea.md
marcofavorito: must it be the case, always?

marcofavorito: if the public keys and addresses must match with the `Wallet`'s ones, then it would be easier to initialize them in the `Agent`/`AEA` constructor.

aea/agent.py
marcofavorito: also, on another note: should the `Agent` have a crypto-identity?
If not (that is, `Agent` won't use the fields `address` and `public_keys`), then I propose two alternatives:

1. only `AEA` instances hold an `Identity` object. `Agent` still has a `name` in the constructor, but in case the `Agent` is actually the `AEA`, `Agent.name == AEA.identity.name`
2. Having a simpler identity for Agent, e.g. `AgentIdentity`, or simply `Identity`, which only holds a `name` (but later it can be extended). The Identity class for the `AEA` would be called `AEAIdentity`, which is a subclass of the previous identity, plus the addresses and public keys..

aea/aea.py
DavidMinarsch: I don't know how else you suggest we should modify the identity object. The identity object must be modifiable with the details from the wallet. I am not a fan of "name" in the constructor, the reason of introducing `identity` is that it makes it trivial to add credentials to the identity in the future without breaking the agent api. The agent api is bound to be broken otherwise. "name" is too specific, imo only object should be passed to the Agent/AEA

aea/aea.py
DavidMinarsch: @marcofavorito keep it in AEA or move to DM? I tend to think we should do the latter

docs/agent-vs-aea.md
DavidMinarsch: yes in my opinion, otherwise, as we enrich the identity concept we will break the Agent/AEA API.

DavidMinarsch: The public keys and addresses must not match the wallet ones in general

aea/aea.py
DavidMinarsch: We could not modify the identity here and instead pass the details from the wallet to the identity outside the agent, at identity construction time.

aea/agent.py
DavidMinarsch: The `Agent` should not have a crypto-identity as we suggest that it would be better to use an AEA for that. Of course, you could always extend our `Agent` base class to add a wallet.

DavidMinarsch: I don't like 1. The identity concept will be pretty quickly much broader than the `name`. The `name` is really just a logging helper, it is not used anywhere else, so we should not keep it as a first class argument. The `Agent` will need some `address` to connect to the OEF for instance. So (also see above) we might want to make `name` and `addresses` the required arguments in the identity constructor and drop `public_keys` as per suggestion by @Totoual . The subclass approach from 2. could be trivially added as we enrich the identity concept without breaking the AEA/Agent api

aea/agent.py
DavidMinarsch: How about we just add `addresses` and drop `public_keys`? Public keys aren't needed by the agents.

docs/agent-vs-aea.md
Totoual: How comes that the public keys and addresses must not match the wallet ones? I mean, I don't see the point having different pbk and addresses from the wallet.

aea/aea.py
marcofavorito: ok do the latter

aea/agent.py
marcofavorito: agreed

aea/agent.py
marcofavorito: yes

docs/agent-vs-aea.md
DavidMinarsch: The question is why they should have to match. I could see a scenario where I want to use one address for interacting with AEAs in public and another to do payments with the subset of agents who I actually transact with. So forcing the two to align avoids this.

Totoual: Though both of the keys should be inside the wallet and must be a pair private/public key. Otherwise we need to find an other way to identify the agents. I mean, if the address does not divert from the public key there is no reason to be in the wallet. And also it must not be of type address :)

PR #721, opened 2020-02-06
Changes based on bandit report B608

Totoual:
## Proposed changes

Changed the way we are adding the data in the database.

## Fixes

 -

## Types of changes

## Checklist

### DIFF
--- a/packages/fetchai/skills/carpark_detection/detection_database.py
+++ b/packages/fetchai/skills/carpark_detection/detection_database.py
@@ -60,7 +60,6 @@ class DetectionDatabase:
         ret = self.get_system_status("db", False) == "Exists"
         return ret
 
-    # This is not used in the AEA example.
     def reset_database(self):
         """Reset the database and remove all data."""
         # If we need to reset the database, then remove the table and any stored images
@@ -79,7 +78,6 @@ class DetectionDatabase:
         self.initialise_backend()
         logger.info("Finished initialising backend!")
 
-    # This is not used in the AEA example
     def reset_mask(self):
         """Just reset the detection mask."""
         # If we need to reset the database, then remove the table and any stored images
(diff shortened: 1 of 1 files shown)


packages/fetchai/skills/carpark_detection/detection_database.py
```diff
@@ -129,11 +129,11 @@ def initialise_backend(self):
...
     def set_fet(self, amount, t):
         """Record how much FET we have and when we last read it from the ledger."""
-        self.execute_single_sql(
-            "INSERT OR REPLACE INTO fet_table(id, amount, last_updated) values(0, '{0}', '{1}')".format(
-                amount, t
-            )
+        command = (
+            "INSERT OR REPLACE INTO fet_table(id, amount, last_updated) values(0, ?, ?)"
         )
+        variables = (str(amount), str(t))
```
panasevychol: Why we making strings from input data here? Why don't accept only strings on input?

packages/fetchai/skills/carpark_detection/detection_database.py
```diff
@@ -172,20 +172,17 @@ def get_lat_lon(self):
...
     def set_system_status(self, system_name, status):
         """Record the status of one of the systems."""
-        self.execute_single_sql(
-            "INSERT OR REPLACE INTO status_table(system_name, status) values('{}', '{}')".format(
-                system_name, status
-            )
+        command = (
+            "INSERT OR REPLACE INTO status_table(system_name, status) values(?, ?)"
         )
+        variables = (str(system_name), str(status))
```
panasevychol: Same as above. And again couple of times below.

packages/fetchai/skills/carpark_detection/detection_database.py
Totoual: Because I just wanted to make as less changes as I could to the specific skill. And the way it was implemented it was converting them inside the database query.

Issue #575, opened 2019-12-31
IPFS hash the packages


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
Currently, we do not ensure consistency of a package and it is not content addressable.

**Describe the solution you'd like**
Hashing a package folder with IPFS allows us to ensure that it has not been tampered with and makes it content addressable.

DavidMinarsch:
Relevant article: https://rsci.app.link/NbdkbPolJ2?_p=c20a36c09d1c65f0e4038cff

Issue #542, opened 2019-12-20
Command to update the fingerprint.


Totoual:
**Is your feature request related to a problem? Please describe.**
No.

**Describe the solution you'd like**
We can introduce an aea command for updating / generating the fingerprint of the package we are developing. For example : `aea generate fingerprint`. Will generate the hash of the package author:package_name:version:nonce and store it in the .yaml file.

**Describe alternatives you've considered**
None.

PR #785, opened 2020-02-17
AEA-248 AEA-345 add ipfs hashing and replace fingerprint of packages

DavidMinarsch:
## Proposed changes

This PR adds IPFS hashing to the stack:
- it hashes each `.py` file of a package and adds the hash to the `package.yaml`
- it hashes the entire package and adds the hash to `packages/hashes.csv`
- it allows for "IPFS only hashing", that is hashing of files only (!) w/o usage of the IPFS daemon via the `aea.helpers.ipfs` module.
- it adds a new `aea fingerprint [skills/connections/protocols] [PUBLIC_ID]` to generate IPFS hashes for files in non-vendorized packages and adding them to `fingerprint` field in config.

Limitations:
- Hashing of the entire package needs the daemon. But that’s ok as one can do that registry side where we will integrate an IPFS daemon (or if one wants no reliance on the registry get the daemon running)

Other Benefits:
- The IPFS hashes of the files can be used to do consistency checks.
- The hashes of a package can be used to retrieve the package from IPFS once we make the packages available there.

## Fixes

Addresses #575, #542

## Types of changes

## Checklist

## Further comments

Helpful resources:
- https://discuss.ipfs.io/t/how-to-calculate-file-directory-hash/777/46
- https://github.com/ipfs/py-ipfs
- https://www.npmjs.com/package/ipfs-only-hash
- https://ethereum.stackexchange.com/questions/44506/ipfs-hash-algorithm
- https://github.com/bmcorser/py-multihash
- https://github.com/libp2p/py-libp2p/
- https://stackoverflow.com/questions/40998621/how-to-create-an-ipfs-compatible-multihash
- https://pypi.org/project/pymultihash/
- https://pymultihash.readthedocs.io/en/latest/
- https://github.com/ipfs/py-datastore


Pipfile
```diff
@@ -51,6 +51,7 @@ python-dotenv = "*"
 pyyaml = ">=4.2b1"
 watchdog = "*"
 web3 = "==5.2.2"
+ipfshttpclient = "*"
```
DavidMinarsch: note, this is also a dependency of web3!

aea/connections/stub/connection.yaml
```diff
@@ -2,7 +2,7 @@ name: stub
 author: fetchai
 version: 0.1.0
 license: Apache-2.0
-fingerprint: ""
+fingerprint: {'__init__.py': 'QmWwepN9Fy9gHAp39vUGFSLdnB9JZjdyE3STnbowSUhJkC', 'connection.py': 'QmYBKp9zhAmv21u7JJ7XsTgQFtjfY2FMtfhBDJTCJxcSVW'}
```
DavidMinarsch: All finished packages now have the fingerprints of their files as part of the config. This will be helpful for ensuring their integrity.

aea/helpers/base.py
```diff
@@ -141,26 +140,3 @@ def add_agent_component_module_to_sys_modules(
     item_type_plural = item_type + "s"
     dotted_path = "packages.{}.{}.{}".format(author_name, item_type_plural, item_name)
     import_module(dotted_path, module_obj)
-
-
-def generate_fingerprint(
```
DavidMinarsch: This is no longer needed

packages/hashes.csv
```diff
@@ -0,0 +1,28 @@
+fetchai/connections/gym,Qmd8Z9wKPTBxXXGaB6cZRPn1UhNvDQ4T8T72LiiMNR2omW
```
DavidMinarsch: Note, IPFS hashes make the packages content addressable. This hash can be used to upload the package to IPFS. Then the package is uniquely addressable with the hash. So at some point in the future one can retrieve the package via IPFS directly rather than via the registry. The registry could run an IPFS node.

aea/connections/stub/connection.yaml
Totoual: This is generated automatically right ? (Just a sanity check :D )

DavidMinarsch:
> LGTM; How is this ipfs is generated? For example if I create a new skill is this going to be generated automatically, or I should run the generator?

At the moment it is only generated via the script. There is an open issue to add generation functionality as a CLI command!

tests/data/dummy_connection/connection.yaml
```diff
@@ -2,7 +2,7 @@ name: dummy
 author: fetchai
 version: 0.1.0
 license: Apache-2.0
-fingerprint: ""
+fingerprint: {}
```
marcofavorito: I suggest adding fingerprints also here. Tests should be also documentation.

tests/data/dummy_skill/skill.yaml
```diff
@@ -2,7 +2,7 @@ name: dummy
 author: dummy_author
 version: 0.1.0
 license: Apache-2.0
-fingerprint: ""
+fingerprint: {}
```
marcofavorito: see above

tests/data/exception_skill/skill.yaml
```diff
@@ -2,7 +2,7 @@ name: exception
 author: fetchai
 version: 0.1.0
 license: Apache-2.0
-fingerprint: ""
+fingerprint: {}
```
marcofavorito: see above

tests/data/gym-connection.yaml
```diff
@@ -2,7 +2,7 @@ name: gym
 author: fetchai
 version: 0.1.0
 license: Apache-2.0
-fingerprint: ""
+fingerprint: {}
```
marcofavorito: see above

scripts/generate_ipfs_hashes.py
```diff
@@ -0,0 +1,196 @@
...
+
+        # confirm ipfs only generates same hash:
+        for file_name, ipfs_hash in fingerprints_dict.items():
+            path = os.path.join(target_dir, file_name)
+            ipfsho_hash = ipfs_hash_only.get(path)
+            if ipfsho_hash != ipfs_hash:
+                print("WARNING, hashes don't match for: {}".format(path))
+
+        # update fingerprints
+        file_name = package_type[:-1] + ".yaml"
```
marcofavorito: I think this part could be simplified. I'd like to make some suggestions if you don't mind by opening another PR.

scripts/generate_ipfs_hashes.py
```diff
@@ -0,0 +1,196 @@
...
+    package_type: str,
+    package_name: str,
+    ipfs_hash_only: IPFSHashOnly,
+):
+    """Hashes a package and its components."""
+    print("Processing package {} of type {}".format(package_name, package_type))
+
+    if package_type != "agents":
+        # hash inner components (all `.py` files)
+        result_list = client.add(target_dir, pattern="*.py")
```
marcofavorito: Here we are hashing only Python modules. However, there might be other files of different format that are somehow part of the package.

scripts/generate_ipfs_hashes.py
marcofavorito: For instance, we could update the dictionary, rather than working with the content of the file as string.

scripts/generate_ipfs_hashes.py
```diff
@@ -0,0 +1,196 @@
...
+                    )
+
+        # output the package hashes
+        to_csv(package_hashes)
+
+    except Exception as e:
+        print(e)
+
+    finally:
+        # terminate the ipfs daemon
```
marcofavorito: I would try to call `process.terminate()` (i.e. `SIGINT` on Posix OSs), and wait for some time before `process.kill()` (i.e. `SIGTERM` on Posix OSs).

https://docs.python.org/3/library/subprocess.html#subprocess.Popen.terminate

scripts/generate_ipfs_hashes.py
DavidMinarsch: I think we don't need to hash files other than the `.py` files for now. The package folder itself will be hashed so it will include other files.

scripts/generate_ipfs_hashes.py
DavidMinarsch: I am sure there is much room for improvement, since it's just a script I did not pay much attention to making it neat. Feel free to improve it.

scripts/generate_ipfs_hashes.py
DavidMinarsch: Resolved it

PR #789, opened 2020-02-18
Fix tx_nonce generator

### LINKED ISSUE #788: Tx_nonce_generator isn't working without a ledger_api.
**Describe the bug**
Currently the `tx_nonce` cannot work without a specified identifier. Results to assertion error due to `assert identifier in self.apis.keys()`

**Expected behavior**
Return null if we don't need a ledger.

Totoual:
## Proposed changes

The way we generating the transaction nonce it asserts if the identifier exists in the `ledger_apis.keys()`. But when we don't want to transact with a ledger we cannot generate a `tx_nonce` because we implemented this functionality inside the `LedgerApi` of each token. I added the assertion inside the try/catch block so it will return an empty tx_nonce.

## Fixes

#788

## Types of changes

## Checklist

### DIFF
--- a/aea/crypto/ledger_apis.py
+++ b/aea/crypto/ledger_apis.py
@@ -248,7 +248,7 @@ class LedgerApis(object):
                 )
                 tx_nonce = ""
         else:
-            logger.warning(" You didn't specify a ledger so the tx_nonce will be Empty.")
+            logger.warning("You didn't specify a ledger so the tx_nonce will be Empty.")
             tx_nonce = ""
         return tx_nonce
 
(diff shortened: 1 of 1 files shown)


aea/crypto/ledger_apis.py
```diff
@@ -238,13 +238,13 @@ def generate_tx_nonce(
         :param client: the address of the client.
         :return: return the hash in hex.
         """
-        assert identifier in self.apis.keys()
```
DavidMinarsch: I don't like the solution. Can we instead have an if else block?
`if identifier in self.apis.keys():
     try:
             api = self.apis[identifier]
             tx_nonce = api.generate_tx_nonce(seller=seller, client=client)		             tx_nonce = api.generate_tx_nonce(seller=seller, client=client)
         except Exception:
             logger.warning(
                 "An error occurred while attempting to generate the tx_nonce"
             )
 else:
    logger.warning(" You didn't specify a ledger so the tx_nonce will be Empty.")
    tx_nonce= ""
return tx_nonce
`

PR #792, opened 2020-02-18
AEA-364 Matching the generator's result with existing protocols

5A11:
## Proposed changes

Small changes to ensure the generated protocols match existing protocols.

## Types of changes

## Checklist


aea/protocols/generator.py
```diff
@@ -51,6 +51,26 @@
     "bool": "bool",
     "str": "string",
 }
+FETCHAI_COPYRIGHT = (
```
DavidMinarsch: I don't think we should add fetchai as the copy holder. Instead, it should be the `author`

aea/protocols/generator.py
```diff
@@ -53,12 +55,55 @@
...
+
+    :param author: the author of the protocol.
+    :return: The copyright header text.
+    """
+    copy_right_str = (
+        "# -*- coding: utf-8 -*-\n"
+        "# ------------------------------------------------------------------------------\n"
+        "#\n"
+    )
+    copy_right_str += "#   Copyright {} {} Limited\n".format(date.today().year, author)
```
DavidMinarsch: The `Limited` should be dropped

DavidMinarsch:
Also please merge the latest develop so that tests pass :)

PR #811, opened 2020-02-20
Fix weather test

Totoual:
## Proposed changes

Fix test weather skill (No ledger) to not trying to interact with ledger.

## Fixes

-

## Types of changes

## Checklist

### DIFF
--- a/tests/test_packages/test_skills/test_weather.py
+++ b/tests/test_packages/test_skills/test_weather.py
@@ -58,7 +58,7 @@ class TestWeatherSkills:
             pytest.skip("Skipping the test since it doesn't work in CI.")
         # add packages folder
         packages_src = os.path.join(self.cwd, "packages")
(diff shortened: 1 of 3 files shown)


tests/test_packages/test_skills/test_weather.py
```diff
@@ -89,6 +89,23 @@ def test_weather(self, pytestconfig):
...
+        whole_file = file.read()
+
+        whole_file = whole_file.replace("is_ledger_tx: True", "is_ledger_tx: False")
+
+        # close the file
+        file.close()
+
+        with open(yaml_path, "w") as f:
+            f.write(whole_file)
+
```
Totoual: Since we changed the weather_station skill to is_ledger_tx to be True by default in the test we need to turned to False when we don't want to interact with the ledger. Otherwise it will fail since we don't have a ledger_api.

PR #829, opened 2020-02-23
try to fix pipenv

DavidMinarsch:
## Proposed changes

This PR tries to fix the Pipenv install steps.

## Fixes

- #831, #832

## Types of changes

## Checklist


Pipfile
```diff
@@ -12,6 +12,8 @@ name = "test-pypi"
 bandit = "==1.6.2"
 black = "==19.10b0"
 bs4 = "==0.0.1"
+colorlog = "==4.1.0"
```
DavidMinarsch: these are not core dependencies, only needed for tests

setup.py
```diff
@@ -68,15 +68,16 @@ def get_all_extras() -> Dict:
 
     cli_deps = [
         "click",
-        "PyYAML",
-        "jsonschema",
+        "pyyaml>=4.2b1",
```
DavidMinarsch: we should mirror Pipfile[packages] here imo

tox.ini
```diff
@@ -64,10 +64,10 @@ deps = black
 commands=black aea examples packages scripts tests --check --verbose
 
 [testenv:flake8]
-deps = flake8
-       flake8-docstrings
-       flake8-import-order
-       flake8-bugbear
+deps = flake8==3.7.9
```
DavidMinarsch: to mirror Pipfile

setup.py
marcofavorito: not applicable anymore

tox.ini
```diff
@@ -26,98 +26,106 @@ commands =
...
+    pytest-asyncio==0.10.0
+    docker
+    colorlog==4.1.0
+    oef==0.8.1
+    gym==0.15.6
+    numpy==1.18.1
+    tensorflow==1.14.0
 commands =
-    pipenv install
-    pip install -e .[all]
```
DavidMinarsch: isn't `pip install -e .[all]` required?

tox.ini
```diff
@@ -26,98 +26,106 @@ commands =
...
+    docker
+    colorlog==4.1.0
+    oef==0.8.1
+    gym==0.15.6
+    numpy==1.18.1
+    tensorflow==1.14.0
 
 commands =
-    pipenv install
-    pip install -e .[all]
```
DavidMinarsch: see above

tox.ini
marcofavorito: it is not required. `tox` provides to install the package by default.
However, a problem I'm encountering is that the coverage is not computed only on the main code base but also on the copy of the codebase `tox` does.

PR #840, opened 2020-02-24
Fix several issues in ml skills.

marcofavorito:
## Proposed changes

The ML Skill didn't work properly. The model was not actually trained, due to problems with Keras models when trained in multiple threads.

This PR improves the training process in the `ml_train` skill. In particular, the model component `MLModel` starts a `training_loop`, running in a separate thread, and waits for data from a data queue, which in turn is populated by the main thread.

So the training is done in a separate thread, but at the end of every `fit` it updates the weights stored as an attribute in `MLModel`. This operation is made thread-safe by using a lock. The weights are then loaded whenever there's the need for `predict` or `evaluate` the model, which is compiled every time but updated with the latest version of the weights.

I leave this link as a future reference:
https://github.com/keras-team/keras/issues/5640#issuecomment-345613052

Noticeably, I discovered there was no `model_registry`, so the PR also introduces that. 

## Types of changes

## Checklist

## Further comments

The fact we build and compile the model at every predict/evaluate call is not ideal, although it is not a big issue in the ML demo. It could be optimized by having a dedicated thread to accomplish such predict/evaluate requests, and limiting the number of loadings of weights by checking at the time elapsed since the last weight update. However, there might be better patterns in order to share the model across multiple threads, and at the end, this is just a demo. Maybe PyTorch is more flexible on that.

### DIFF
--- a/aea/registries/base.py
+++ b/aea/registries/base.py
@@ -42,7 +42,7 @@ from aea.configurations.loader import ConfigLoader
 from aea.decision_maker.messages.base import InternalMessage
 from aea.decision_maker.messages.transaction import TransactionMessage
 from aea.protocols.base import Message, Protocol
-from aea.skills.base import AgentContext, Behaviour, Handler, Skill, Model
+from aea.skills.base import AgentContext, Behaviour, Handler, Model, Skill
 from aea.skills.tasks import Task
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


aea/registries/base.py
```diff
@@ -42,7 +42,7 @@
 from aea.decision_maker.messages.base import InternalMessage
 from aea.decision_maker.messages.transaction import TransactionMessage
 from aea.protocols.base import Message, Protocol
-from aea.skills.base import AgentContext, Behaviour, Handler, Skill
+from aea.skills.base import AgentContext, Behaviour, Handler, Model, Skill
```
DavidMinarsch: Why are we making Model a core skill component now?

PR #857, opened 2020-02-26
AEA-386 First draft of integrated erc1155

### LINKED ISSUE #801: Create ERC1155 contract package
-

Port from research repo to AEA packages and implement APIs

Totoual:
## Proposed changes

Adds the erc1155 as a package to the AEA-framework ready to be used from an agent's behaviour. 

## Fixes

#801 

## Types of changes

## Checklist


aea/contracts/base.py
```diff
@@ -21,30 +21,18 @@
...
+from aea.configurations.base import PublicId
 
 
 class Contract(ABC):
     """Abstract definition of a contract."""
 
-    def __init__(
-        self, contract_id: ContractId, config: ContractConfig,
-    ):
+    contract_id = None  # type: PublicId
```
DavidMinarsch: It should be `ContractId` which is a `PublicId`

aea/contracts/base.py
```diff
@@ -21,30 +21,18 @@
...
         """
-        self._contract_id = contract_id
-        self._config = config
-
-    @property
-    def id(self) -> ContractId:
-        """Get the name."""
-        return self._contract_id
-
-    @property
```
DavidMinarsch: why has this been removed?

aea/registries/base.py
```diff
@@ -143,10 +143,10 @@ def register(self, contract_id: ContractId, contract: Contract) -> None:
             raise ValueError(
                 "Contract already registered with contract id '{}'".format(contract_id)
             )
-        if contract.id != contract_id:
```
DavidMinarsch: id was more succinct

aea/contracts/base.py
DavidMinarsch: id change it to `id` as it is more succinct

aea/registries/base.py
```diff
@@ -243,8 +243,18 @@ def _add_contract(self, contract_directory: Path):
             open(contract_directory / DEFAULT_CONTRACT_CONFIG_FILE)
         )
 
-        # instantiate the protocol manager.
-        contract = Contract(contract_config.public_id, contract_config)
+        contract_spec = importlib.util.spec_from_file_location(
+            "contracts", contract_directory / "contract.py"
+        )
+
+        contract_module = importlib.util.module_from_spec(contract_spec)
```
DavidMinarsch: this should be in a try: except: block as the `SomeContract` class might not be available.

packages/fetchai/contracts/erc1155/contract.py
```diff
@@ -35,3 +60,424 @@ def __init__(self, **kwargs):
         :param performative: the type of message.
         """
         super().__init__(**kwargs)
+        self.w3 = Web3(Web3.HTTPProvider(PROVIDER))
```
DavidMinarsch: this should not be instantiated here, instead, I would pass in the relevant ledger_api as an argument of the function which needs it.

packages/fetchai/contracts/erc1155/contract.py
```diff
@@ -35,3 +60,424 @@ def __init__(self, **kwargs):
...
+        self.address = None
+        self.deployed = False
+        self.abi = None
+        self.bytecode = None
+        self.instance = None
+        batch_token_ids = []
```
DavidMinarsch: this is very verbose, can't the helper method return a list of item_ids?

packages/fetchai/contracts/erc1155/contract.py
```diff
@@ -35,3 +60,424 @@ def __init__(self, **kwargs):
...
+        for j in range(10):
+            batch_token_ids.append(Helpers().generate_id(FT, j))
+        self.item_ids = batch_token_ids
+        self._account = eth.Account.from_key(PRIVATE_KEY_1)
```
DavidMinarsch: same as I said about web3

packages/fetchai/contracts/erc1155/contract.py
```diff
@@ -35,3 +60,424 @@ def __init__(self, **kwargs):
...
+
+    def load_from_json(self):
```
DavidMinarsch: add `w3` as an arg

packages/fetchai/contracts/erc1155/contract.py
```diff
@@ -35,3 +60,424 @@ def __init__(self, **kwargs):
...
+        )
+        with open(path, "r") as f_contract:
+            contract_interface = json.load(f_contract)
+
+        self.abi = contract_interface["abi"]
+        self.bytecode = contract_interface["bytecode"]
+        self.instance = self.w3.eth.contract(abi=self.abi, bytecode=self.bytecode)
+
+    #  This should go to the Decision Maker.
+    def _sign_transaction(self, tx):
```
DavidMinarsch: same as above, add w3 arg

DavidMinarsch: sign transaction must happen in the decision maker!

packages/fetchai/contracts/erc1155/contract.py
```diff
@@ -35,3 +60,424 @@ def __init__(self, **kwargs):
...
+        tx_signed = self.w3.eth.account.signTransaction(
+            transaction_dict=tx, private_key=self._account.key
+        )
+        return tx_signed
+
+    def deploy_contract(self):
```
DavidMinarsch: add w3 as an arg

aea/crypto/ethereum.py
```diff
@@ -114,6 +114,12 @@ def sign_message(self, message: bytes) -> bytes:
         signature = self.entity.sign_message(signable_message=signable_message)
         return signature["signature"]
 
+    def sign_transaction(self, transaction):
+        signature = Account.sign_transaction(transaction_dict=transaction,
```
DavidMinarsch: I suggest
```suggestion
        signature = self.entity.sign_transaction(transaction_dict=transaction,
```

aea/crypto/ethereum.py
```diff
@@ -176,12 +182,18 @@ def __init__(self, address: str, gas_price: str = DEFAULT_GAS_PRICE):
         """
         self._api = Web3(HTTPProvider(endpoint_uri=address))
         self._gas_price = gas_price
+        self._contract_address = None
```
DavidMinarsch: The contract address has no reason to be here. The ledger api might interact with many contracts.

aea/crypto/ethereum.py
```diff
@@ -176,12 +182,18 @@ def __init__(self, address: str, gas_price: str = DEFAULT_GAS_PRICE):
...
 
     @property
     def api(self) -> Web3:
         """Get the underlying API object."""
         return self._api
 
+    @property
```
DavidMinarsch: please remove, see above

aea/crypto/ethereum.py
```diff
@@ -243,6 +255,27 @@ def send_transaction(
                 time.sleep(3.0)
         return tx_digest
 
+    def send_raw_transaction(self, tx_signed, tx_type) -> Optional[str]:
```
DavidMinarsch: This is almost exactly the same as above. Please refactor to avoid code duplication. For instance, the above method can call this method at line 244.

aea/decision_maker/base.py
```diff
@@ -841,3 +847,35 @@ def _handle_state_update_message(
                 quantities_by_good_id=state_update_message.quantities_by_good_id,
             )
             self._ownership_state = new_ownership_state
+
+    def _handle_tx_message_for_deployment(self, tx_message) -> None:
```
DavidMinarsch: We should discuss this in person. I think it could be merged with the signing handler

aea/decision_maker/messages/transaction.py
```diff
@@ -291,6 +292,7 @@ def _is_consistent(self) -> bool:
             elif self.performative in {
                 self.Performative.PROPOSE_FOR_SIGNING,
                 self.Performative.REJECTED_SIGNING,
+                self.performative.PROPOSE_FOR_CONTRACT
```
DavidMinarsch: I think this is a propose for signing as well.

packages/fetchai/contracts/erc1155/contract.py
```diff
@@ -31,17 +31,17 @@
...
+from web3 import Web3
 
-from aea.configurations.base import PublicId
+from aea.configurations.base import ContractId
 from aea.contracts.base import Contract
+from aea.crypto.base import LedgerApi
+from aea.decision_maker.messages.transaction import TransactionMessage
 from aea.mail.base import Address
 
 CHAIN_ID = 3
```
DavidMinarsch: These should be set via the ledger apis, not here

aea/crypto/ethereum.py
Totoual: I know. It is more of an experimental variable. Cause I was facing some problems with the contract and wanted to discuss it with you :)

aea/decision_maker/messages/transaction.py
Totoual: Looks like that the PROPOSE_FOR_SIGNING is working too. The only thing that I have to figure out is how I am going to send the transaction after signing

packages/fetchai/contracts/erc1155/contract.py
Totoual: As I say in the comment I can actually remove the token_id from there and pass it only when the transaction happens. But this is something that can remove once I have all the functionality.

packages/fetchai/contracts/erc1155/contract.py
```diff
@@ -18,20 +18,464 @@
...
+                "nonce": nonce,
+            }
+        )
+        logger.info(tx)
+        logger.info(ledger_api.api.provider)
+        # gas_estimate = self.w3.eth.estimateGas(transaction=tx)
+        # logger.info("gas estimate create_batch: {}".format(gas_estimate))
+        return tx
+
+    def mint_batch(self, deployer_address: Address, recipient_address: Address, mint_quantities: List[int], ledger_api: LedgerApi):
```
Totoual: I split the functionalities to different functions so we have a function for create and an other function for minting.

packages/fetchai/contracts/erc1155/contract.py
```diff
@@ -18,20 +18,464 @@
...
+            terms.item_ids,
+            terms.from_supplies,
+            terms.to_supplies,
+            terms.value_eth_wei,
+            terms.trade_nonce,
+            signature,
+            data,
+        ).buildTransaction(
+            {
+                "chainId": 3,
```
Totoual: I will find a way to pass it from the `ledger_api.api.provider`. But not sure yet.

packages/fetchai/contracts/erc1155/contract.py
```diff
@@ -18,20 +18,464 @@
...
+                "gas": 2818111,
+                "from": terms.from_address,
+                "value": terms.value_eth_wei,
+                "gasPrice": ledger_api.api.toWei("50", "gwei"),
+                "nonce": nonce,
+            }
+        )
+        return tx
+
+    def _get_balance(self, contract, from_address: Address, item_id: int):
```
Totoual: Introduce private function _get_balance() in case we want to do checks later on. ( The original version has assertions based on the balances but this must not be a case here.)

aea/crypto/ethereum.py
```diff
@@ -268,7 +262,8 @@ def send_raw_transaction(self, tx_signed, tx_type) -> Optional[str]:
                 result = self._api.eth.getTransactionReceipt(hex_value)
                 logger.info("transaction validated - exiting")
                 if result.contractAddress is not None:
-                    self._contract_address = result.contractAddress
+                    # Checking the address so I can pass it manually for now.
+                    logger.info(result.contractAddress)
```
DavidMinarsch: why don't you just return the transaction digest `tx_digest` and the `result`? It seems the generic use case is that I want to get these two values back and then if it happens to be a transaction associated with contract deployment the contract address can be recovered in the business logic where this is called.

aea/decision_maker/base.py
DavidMinarsch: I think the skill should send a `TransactionMessage` with performative `PROPOSE_FOR_SIGNING` to the DM which contains the tx. The actual deployment should then be done from the skill once it receives the callback.

aea/decision_maker/messages/transaction.py
DavidMinarsch: You can access the ledger apis on the skill context, hence this will work: `ledger_apis.apis.get(tx_message.ledger_id).send_raw_transaction(tx_signed=signature,
                                                                                  tx_type=tx_message.tx_id)`

DavidMinarsch: `tx_type=tx_message.tx_id` not sure about this part.. what does it do?

packages/fetchai/contracts/erc1155/contract.py
DavidMinarsch: well I think its helpful to actually have access to the token ids here. where else would you store them?

packages/fetchai/contracts/erc1155/contract.py
```diff
@@ -18,20 +18,464 @@
...
+            "erc1155.json",
+        )
+        with open(path, "r") as f_contract:
+            contract_interface = json.load(f_contract)
+
+        self.abi = contract_interface["abi"]
+        self.bytecode = contract_interface["bytecode"]
+        self.instance = ledger_api.api.eth.contract(abi=self.abi, bytecode=self.bytecode)
+
+    def deploy_contract(self, deployer_address: Address, ledger_api: LedgerApi) -> TransactionMessage:
```
DavidMinarsch: this method should be called `get_deploy_transaction`

packages/fetchai/contracts/erc1155/contract.py
```diff
@@ -18,20 +18,464 @@
...
+
+        # estimate the gas and update the transaction dict
+        gas_estimate = ledger_api.api.eth.estimateGas(transaction=tx)
+        logger.info("gas estimate deploy: {}".format(gas_estimate))
+        tx["gas"] = gas_estimate
+        return tx
+
+    def update_contract_instance(self, contract_address: Address, ledger_api: LedgerApi):
+        """Update the local instance of the smart contract with the deployed one."""
+        logger.info("Updating the local instance of the contract...")
```
DavidMinarsch: This should have some assertions that the contract is not already deployed.

packages/fetchai/contracts/erc1155/contract.py
DavidMinarsch: This should have some assertions that the contract is not already deployed.

packages/fetchai/contracts/erc1155/contract.py
```diff
@@ -18,20 +18,464 @@
...
+            tx_counterparty_fee=0,
+            tx_quantities_by_good_id={},
+            info={},
+            ledger_id="ethereum",
+            signing_payload=tx
+        )
+
+        return tx_message
+
+    def sign_single_transaction(self, terms) -> TransactionMessage:
```
DavidMinarsch: it should be called `get_single_transaction_proposal`, or something like that

packages/fetchai/contracts/erc1155/contract.py
```diff
@@ -18,20 +18,464 @@
...
+            tx_counterparty_fee=0,
+            tx_quantities_by_good_id={},
+            info={},
+            ledger_id="ethereum",
+            signing_payload=tx_hash
+        )
+
+        return tx_message
+
+    def sign_batch_transaction(self, terms):
```
DavidMinarsch: `get_batch_transaction_proposal`

packages/fetchai/contracts/erc1155/contract.py
```diff
@@ -18,20 +18,464 @@
...
+
+        return tx_message
+
+    def _get_balance_of_batch(self, contract, address, item_ids):
+        """Get the balance for a batch of items"""
+        return contract.instance.functions.balanceOfBatch(
+            [address] * 10, item_ids
+        ).call()
+
+    def atomic_swap_batch(self, deployer_address, contract, terms, signature) -> TransactionMessage:
```
DavidMinarsch: `get_atomic_swap_batch_transaction_proposal` ?

aea/registries/base.py
Totoual: Should we consider exiting in case we don't find the file ? 

>         try:
            contract_spec = importlib.util.spec_from_file_location(
                "contracts", contract_directory / "contract.py"
            )
            contract_module = importlib.util.module_from_spec(contract_spec)
        except FileNotFoundError:
            logger.warning("File not found.")
            sys.exit(1)

packages/fetchai/contracts/erc1155/contract.py
Totoual: I would just generate them in the skill and pass them to the contract. Then we should create an other method inside the contract that returns the ids. It makes sense to have access to the ids if you are the one who deployed the contract but if you are just a user you have no idea what the contracts contains. With that being said the user must be able to fetch the ids from the contract it self.

aea/crypto/ethereum.py
Totoual: So can we consider to always returns the result and if we need the `tx_hash` just get it from the AttributeDictionary? The Dictionary is in the form of 
```
info: AttributeDict({
   'blockHash': HexBytes('...'),
   'blockNumber': 7418505, 
   'contractAddress': '0x03D2B44B573418B8ACfe53810379FfdF6E21E0F1',                 'cumulativeGasUsed': 5185623, 
   'from': '0x21795d753752ccc1ac728002d23ba33cbf13b8b0', 
   'gasUsed': 3603991, 
   'logs': [],
   'logsBloom': HexBytes('a lot of 0'), 
   'status': 1, 'to': None, 
   'transactionHash':         HexBytes('0xa1920cb04cbd645f7f81bd3b745cd847e442a33335a7b3dc18ebfdfb36602af6'),
   'transactionIndex': 28})
```

packages/fetchai/contracts/erc1155/contract.py
Totoual: It is more of the signing payload here. And not about the proposal. This is the second part that we sign the items we want to trade. So I wouldn't say that this is a transaction_proposal. More like `get_single_signing_proposal`. What do you think ?

aea/registries/base.py
DavidMinarsch: Yes you can do that. In this case also add the try and similar behaviour for the other registries.

packages/fetchai/contracts/erc1155/contract.py
DavidMinarsch: Well I think the methods to create them should be attached to the contract but callable from the skill, potentially with some optional ids to be fed in from the skill

aea/crypto/ethereum.py
DavidMinarsch: yes ok

packages/fetchai/contracts/erc1155/contract.py
DavidMinarsch: ok

aea/decision_maker/messages/transaction.py
```diff
@@ -300,9 +302,9 @@ def _is_consistent(self) -> bool:
                 assert isinstance(self.signing_payload, dict) and all(
                     isinstance(key, str) for key in self.signing_payload.keys()
                 ), "Signing_payload must be of type Dict[str, Any]"
-                assert isinstance(
-                    self.tx_signature, bytes
-                ), "Tx_signature must be of type bytes"
+                # assert isinstance(
+                #     self.tx_signature, bytes
+                # ), "Tx_signature must be of type bytes"
```
Totoual: Must provide alternatives here. The signature from the web3 is a dictionary.

aea/crypto/base.py
```diff
@@ -22,6 +22,8 @@
 from abc import ABC, abstractmethod
 from typing import Any, BinaryIO, Optional, Union
 
+from eth_account.datastructures import AttributeDict
```
DavidMinarsch: The base class cannot contain any ledger specifics

aea/decision_maker/base.py
```diff
@@ -752,50 +752,43 @@ def _handle_tx_message_for_signing(self, tx_message: TransactionMessage) -> None
...
-                tx_message_response = TransactionMessage.respond_signing(
-                    tx_message,
-                    performative=TransactionMessage.Performative.REJECTED_SIGNING,
-                )
-            self.message_out_queue.put(tx_message_response)
-        else:
-            if self._is_acceptable_for_signing(tx_message):
-                tx_signature = self._sign_tx(tx_message)
+            if self._is_valid_tx(tx_message):
+                tx_signed = self._sign_ledger_tx(tx_message)
```
Totoual: Why we are checking this thing twice ?

aea/decision_maker/base.py
```diff
@@ -752,50 +752,43 @@ def _handle_tx_message_for_signing(self, tx_message: TransactionMessage) -> None
...
-            else:
-                tx_message_response = TransactionMessage.respond_signing(
-                    tx_message,
-                    performative=TransactionMessage.Performative.REJECTED_SIGNING,
-                )
-            self.message_out_queue.put(tx_message_response)
-        else:
-            if self._is_acceptable_for_signing(tx_message):
-                tx_signature = self._sign_tx(tx_message)
+            if self._is_valid_tx(tx_message):
```
Totoual: We are checking the same thing in the line 755. I can see that we change the `tx_message` in the previous block in case it is true.

DavidMinarsch: Yes but for now I prefer it this way until I am refactoring the dm this week

Totoual: Ok. Makes sense :)

PR #859, opened 2020-02-26
Fix a flaky oef connection test

marfvr:
## Proposed changes

the issue seemed to be due to some timing problem between the send and the disconnect.

## Types of changes

## Checklist

### DIFF
--- a/tests/test_packages/test_connections/test_oef/test_communication.py
+++ b/tests/test_packages/test_connections/test_oef/test_communication.py
@@ -58,7 +58,6 @@ from packages.fetchai.protocols.fipa.serialization import FIPASerializer
 from packages.fetchai.protocols.oef.message import OEFMessage
 from packages.fetchai.protocols.oef.serialization import DEFAULT_OEF, OEFSerializer
 
-from ....conftest import CUR_PATH
 
 logger = logging.getLogger(__name__)
 
--- a/tests/test_packages/test_connections/test_oef/test_communication.py
+++ b/tests/test_packages/test_connections/test_oef/test_communication.py
@@ -21,7 +21,6 @@
 
 import asyncio
 import logging
-import os
 import sys
 import time
 import unittest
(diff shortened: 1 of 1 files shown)


tests/test_packages/test_connections/test_oef/test_communication.py
```diff
@@ -973,9 +971,7 @@ def _node(self):
 @pytest.mark.asyncio
 async def test_send_oef_message(network_node):
     """Test the send oef message."""
-    private_key_pem_path = os.path.join(CUR_PATH, "data", "priv.pem")
-    wallet = Wallet({FETCHAI: private_key_pem_path})
-    address = wallet.addresses[FETCHAI]
+    address = FetchAICrypto().address
```
marcofavorito: we now use a new address for every test, rather than the same address associated with the private key `test/data/priv.pem`. The same has been applied to other tests.

The reason is that, on the OEF side, using the same address for testing might cause some problems due to delays. This should not happen if all the tests clean up the resources correctly. However, we have better test independence with this approach.

tests/test_packages/test_connections/test_oef/test_communication.py
```diff
@@ -1011,16 +1007,18 @@ def _node(self):
         message=msg_bytes,
     )
     await oef_connection.send(envelope)
+    search_result = await oef_connection.receive()
+    assert isinstance(search_result, Envelope)
+    await asyncio.sleep(2.0)
```
marcofavorito: the main contribution is to add a delay between the `send`/`receive` and the disconnection.
Indeed, I noticed the disconnection didn't complete due to `AsyncioCore.stop` that in some case never terminated.

marcofavorito:
Although I'm not completely satisfied with the workaround, worth noticing the test set up is not the one that will be used in practice. That is, the developer never interacts with the Connection object directly.

For the record, the guilty test was `test_send_oef_message`. In `tests/test_packages/test_connections/test_oef/test_communication.py`

The guilty function was the `disconnect()`:
https://github.com/fetchai/agents-aea/blob/master/packages/fetchai/connections/oef/connection.py#L674-L689

In particular, the final call to `disconnect` seemed to stuck in `self._core.stop()`.

DavidMinarsch (approved):
LGTM

OjusWiZard (approved):
Approve: I reviewed the PR diff and full changed file context. No confirmed bugs found in the changes.

The behavior change in test_send_oef_message (waiting for one receive before disconnect) appears intentional and consistent with the stated flaky timing issue.

Optional follow-up: if flakiness reappears, consider replacing the fixed sleep(2.0) with a bounded condition/wait pattern.

PR #860, opened 2020-02-27
Feature/conn

DavidMinarsch:
## Proposed changes

Added the following:

- HTTP connection and config: /packages/packages/fetchai/*
- Pet Store YAML file example: /examples/http_ex/*

## Types of changes

## Checklist

## Further comments

The HTTP connection.py is still largely a WIP with only the baseline functions.


packages/fetchai/connections/http/connection.py
```diff
@@ -0,0 +1,384 @@
...
+
+class HTTPConnection(Connection):
+    """Proxy to the functionality of the web RESTful API."""
+
+    def __init__(
+        self,
+        address: Address,
+        host: str = "",
+        port: int = 10000,
+        api_path: str = None,  # Directory path of the API YAML file.
```
DavidMinarsch: I think you'd want it to be `api_specification_path: Optional[str]`

packages/fetchai/connections/http/connection.py
```diff
@@ -0,0 +1,384 @@
...
+        *args,
+        **kwargs,
+    ):
+        """
+        Initialize a connection to an RESTful API.
+
+        :param api_spec_path: the directory path of the API specification YAML source file.
```
DavidMinarsch: docstrings not matching the signature

packages/fetchai/connections/http/connection.py
```diff
@@ -0,0 +1,384 @@
...
+            try:
+                with open(api_path, "r") as f:
+                    api_spec = yaml.safe_load(f)
+            except FileNotFoundError:
+                logger.error(
+                    "API specification YAML source file not found. Please double-check filename and path."
+                )
+                return
+        else:
+            api_spec = api_path
```
DavidMinarsch: All of this should be moved to the channel, just forward the `api_specification_path`

packages/fetchai/connections/http/connection.py
```diff
@@ -0,0 +1,384 @@
...
+        Send an envelope.
+
+        :param envelope: the envelop
+        :return: None
+        """
+        if not self.connection_status.is_connected:
+            raise ConnectionError(
+                "Connection not established yet. Please use 'connect()'."
+            )  # pragma: no cover
+        # self.channel.send(envelope)
```
DavidMinarsch: Note, this send function is called by the multiplexer to `send` outgoing envelopes via the connection.

packages/fetchai/connections/http/connection.py
```diff
@@ -0,0 +1,384 @@
...
+        :param address: the address of the agent.
+        :param connection_configuration: the connection configuration object.
+            :host - RESTful API hostname / IP address
+            :port - RESTful API port number
+            :api - Directory API path and filename of the API spec YAML source file.
+        :return: the connection object
+        """
+        host = cast(str, connection_configuration.config.get("host"))
+        port = cast(int, connection_configuration.config.get("port"))
+        api = cast(str, connection_configuration.config.get("api")) if not "" else None
```
DavidMinarsch: should be
```suggestion
        api_specification_path = cast(str, connection_configuration.config.get("api_specification_path"), "")
```

DavidMinarsch: the `get` lets you specify a default value

packages/fetchai/connections/http/connection.py
```diff
@@ -0,0 +1,384 @@
...
+        :return: None
+        """
+        if not self.connection_status.is_connected:
+            raise ConnectionError(
+                "Connection not established yet. Please use 'connect()'."
+            )  # pragma: no cover
+        # self.channel.send(envelope)
+
+    async def receive(self, *args, **kwargs) -> Optional["Envelope"]:
+        """
```
DavidMinarsch: Similarly, the receive method is called by the multiplexer to `receive` incoming envelopes via the connection. (So in the context of your `http` connection these are the envelopes containing the `requests`).

packages/fetchai/connections/http/connection.py
```diff
@@ -0,0 +1,384 @@
...
+    """A wrapper for an RESTful API with an internal HTTPServer."""
+
+    def __init__(
+        self,
+        address: Address,
+        host: str,
+        port: int,
+        api_spec: str,
+        connection_id: PublicId,
+        excluded_protocols: Optional[Set[PublicId]] = None,
```
DavidMinarsch: you don't need the excluded protocols here. It is just needed on the `connection` object

packages/fetchai/connections/http/connection.py
```diff
@@ -0,0 +1,384 @@
...
+
+        :param address: the address
+        """
+        self.address = address
+        self.host = host
+        self.port = port
+        self.connection_id = connection_id
+        self.in_queue = None  # type: Optional[asyncio.Queue]
+        self.loop = None  # type: Optional[asyncio.AbstractEventLoop]
+        self.excluded_protocols = excluded_protocols
```
DavidMinarsch: not needed

packages/fetchai/connections/http/connection.py
```diff
@@ -0,0 +1,384 @@
...
+            self.send_header("Content-type", "text/html")
+            self.end_headers()
+
+            parsed_path = urlparse(self.path)
+            url = parsed_path.path
+            # url = parsed_path.geturl()
+            param = parse_qs(parsed_path.query)
+            param = json.dumps(param)
+
+            self._channel._process_request("GET", url, param)
```
DavidMinarsch: Note, since this is a get request, you'd need to block here and wait for a response from this call.

So something like `response = self._channel._process_request("GET", url, param)` and then serve the response: https://pymotw.com/2/BaseHTTPServer/

packages/fetchai/connections/http/connection.py
```diff
@@ -0,0 +1,384 @@
...
+            parsed_path = urlparse(self.path)
+            url = parsed_path.path
+            # url = parsed_path.geturl()
+            param = parse_qs(parsed_path.query)
+            param = json.dumps(param)
+
+            content_length = int(self.headers["Content-Length"])
+            body = self.rfile.read(content_length).decode()
+
+            self._channel._process_request("POST", url, param, body)
```
DavidMinarsch: same as for get request, I'd wait for a response to determine the status code and then serve the status code https://pymotw.com/2/BaseHTTPServer/

So `response = self._channel._process_request("POST", url, param, body)` and then serve it.

packages/fetchai/connections/http/connection.py
```diff
@@ -0,0 +1,384 @@
...
+        """
+        with self.lock:
+            if not self.stopped:
+                self.httpd.shutdown()
+                logger.info(f"HTTP Server has shutdown on port {self.port}.")
+                self.stopped = True
+                # self.thread.join()
+
+
+def HTTPHandlerFactory(channel: HTTPChannel):
```
DavidMinarsch: Is the factory pattern needed here? Also, PascalCase is never used for functions afaik.

packages/fetchai/connections/http/connection.py
```diff
@@ -0,0 +1,384 @@
...
+                    # self.httpd = HTTPServer((self.host, self.port), HTTPHandler(self))
+                    # handler = SimpleHTTPServer.SimpleHTTPRequestHandler
+                    # handler = HTTPHandler.set_channel(self)
+                    # handler.set_channel(self)
+                    self.httpd = HTTPServer(
+                        (self.host, self.port), HTTPHandlerFactory(self)
+                    )
+                    # self.httpd = HTTPServer((self.host, self.port), handler)
+                    logger.info(f"HTTP Server has connected to port {self.port}.")
+                    self.httpd.serve_forever()
```
DavidMinarsch: Isn't this blocking here?

packages/fetchai/connections/http/connection.py
```diff
@@ -0,0 +1,384 @@
...
+        uri = URI(f"http://{self.host}:{self.port}{url}")
+        context = EnvelopeContext(connection_id=self.connection_id, uri=uri)
+        # Prepare the Envelope's message body and encode it into bytes.
+        msg = {
+            "performative": http_method,
+            "path": url,
+            "params": param,
+            "payload": body,
+        }
+        msg_bytes = json.dumps(msg).encode()
```
DavidMinarsch: This is ok for now, but we will need to create an actual protocol to represent these messages once the other issues are resolved.

packages/fetchai/connections/http/connection.py
```diff
@@ -0,0 +1,384 @@
...
+            to=self.address,
+            sender=self._client_id,
+            protocol_id=protocol_id,
+            context=context,
+            message=msg_bytes,
+        )
+        # Send the Envelope to the Agent's InBox.
+        assert self.in_queue is not None, "Input queue not initialized."
+        assert self.loop is not None, "Loop not initialized."
+        asyncio.run_coroutine_threadsafe(self.in_queue.put(envelope), self.loop)
```
DavidMinarsch: You need to get the correct response envelope here as well, unpack it and return its payload etc.

packages/fetchai/connections/http/connection.py
```diff
@@ -0,0 +1,384 @@
...
+                assert self.in_queue is not None, "Input queue not initialized."
+                assert self.loop is not None, "Loop not initialized."
+                # Ok to put these Envelopes in_queue during HTTPChannel initialization?
+                asyncio.run_coroutine_threadsafe(self.in_queue.put(envelope), self.loop)
+            except ValueError:
+                logger.error(f"Bad formatted path: {path_name}")
+            except Exception as e:
+                logger.error(f"Error when processing a path. Message: {str(e)}")
+
+    def _decode_path(self, path_name: str, path_value: dict, separator: str = "/"):
```
DavidMinarsch: There is no point to this. Unnecessary overhead as you can just feed the specification file to the skill directly. I'd remove this method.

packages/fetchai/connections/http/connection.py
```diff
@@ -0,0 +1,384 @@
...
+        self.excluded_protocols = excluded_protocols
+        # self.thread = Thread(target=self.receiving_loop)
+        self.lock = threading.Lock()
+        self.stopped = True
+
+        if api_spec is not None:
+            self._process_api(api_spec)
+        logger.info(f"Initialized the HTTP Server on HOST: {host} and PORT: {port}")
+
+    def _process_api(self, api_spec):
```
DavidMinarsch: Here I'd consume the api_specification_path as an argument and construct a helper object which lets me verify any given path

packages/fetchai/connections/http/connection.py
```diff
@@ -0,0 +1,384 @@
...
+            message=msg_bytes,
+            context=context,
+        )
+
+    def _process_request(
+        self, http_method: str, url: str, param: str, body: str = "{}"
+    ):
+        """Process incoming API request by packaging into Envelope and sending it in-queue.
+
+        """
```
DavidMinarsch: the first thing I'd do is take the path, and feed it to the helper object for verification. All that does is it checks if the path satisfy the api specification. For now you can stub it by returning true always.

examples/http_ex/README.md
```diff
@@ -0,0 +1,9 @@
+# HTTP agent example
```
DavidMinarsch: Please move to `docs/http_connection_and_skill.md`

packages/fetchai/connections/http/__init__.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# ------------------------------------------------------------------------------
+#
+#   Copyright 2018-2019 Fetch.AI Limited
```
DavidMinarsch: should be
```suggestion
#   Copyright 2018-2020 Fetch.AI Limited
```

DavidMinarsch: same for following files

tests/test_cli_gui/test_search.py
```diff
@@ -139,12 +139,18 @@ def test_real_search():
...
 
     assert data[i]["id"] == "fetchai/gym:0.1.0"
     assert data[i]["description"] == "The gym connection wraps an OpenAI gym."
     i += 1
+    assert data[i]["id"] == "fetchai/http:0.1.0"
+    assert (
+        data[i]["description"]
+        == "The HTTP connection that wraps a web-based, RESTful API specification."
+    )
+    i += 1
```
DavidMinarsch: You have duplicated the same block making your tests fail. Removing this one and keeping the one below should make the tests pass.

DavidMinarsch:
Description:

> HTTP connection: It accepts HTTP requests for the registered swagger api. It turns requests into Envelopes (to: agent_address, from: uuid, protocol_id: fetchai/http:0.1.0, uri: path, message: body). The message consists of the following fields (performative: HTTP verb, payload: HTTP payload, params: HTTP parameters, path: path). The message is implemented as one of our protocols. Envelopes are enqueued with the agent. A response is served once either a) a response envelope arrives from the agent or b) a timeout is hit. If the incoming response does not match the swagger api spec then it is rejected and the appropriate HTTP status code is returned.

packages/fetchai/connections/http/__init__.py
Kevin-Chen0: fixed locally, will be in next commit

packages/fetchai/connections/http/connection.py
Kevin-Chen0: I changed it to `api_spec_path: Optional[str]`, will be in next commit.

packages/fetchai/connections/http/connection.py
Kevin-Chen0: left it as api_spec_path

packages/fetchai/connections/http/connection.py
Kevin-Chen0: I have moved the following to HTTPChannel.

packages/fetchai/connections/http/connection.py
Kevin-Chen0: fixed as `api_spec_path = cast(str, connection_configuration.config.get("api_spec_path"), "")`, will be there next commit.

packages/fetchai/connections/http/connection.py
Kevin-Chen0: I have removed from HTTPChannel, in next commit.

packages/fetchai/connections/http/connection.py
Kevin-Chen0: it has been removed

packages/fetchai/connections/http/connection.py
Kevin-Chen0: Yes it is needed, otherwise, I would not be able to pass in the channel arg into the handler when initializing HTTPServer: `self.httpd = HTTPServer((self.host, self.port), HTTPHandlerFactory(self))`. I tried without factory but failed because HTTPHandler also need to inherit from BaseHTTPRequestHandler.

I am following the  **MakeHandlerClassFromArgv** example here, including the naming convention:

https://stackoverflow.com/questions/21631799/how-can-i-pass-parameters-to-a-requesthandler

packages/fetchai/connections/http/connection.py
Kevin-Chen0: Will try that, I'll probably have questions depending on the implementation structure.

packages/fetchai/connections/http/connection.py
Kevin-Chen0: Deleted, the skill will just receive the path of the specification file.

tests/test_cli_gui/test_search.py
Kevin-Chen0: that is fixed, thanks

packages/fetchai/connections/http/connection.py
Kevin-Chen0: I have included send function in HTTPChannel.

examples/http_ex/README.md
Kevin-Chen0: moved to `docs/http-connection-and-skill.md`

packages/fetchai/connections/http/connection.py
Kevin-Chen0: To start, I'll create a **PetStoreHelper** class within connection.py that emulates an "echo" handler. It takes an init Envelope from _process_api() that specifies what are the valid path and stores in an internal dict. It then uses that to verify incoming requests and returns the appropriate response.

packages/fetchai/connections/http/connection.py
Kevin-Chen0: See my comment regarding the **APIHelper** class above.

packages/fetchai/connections/http/connection.py
Kevin-Chen0: For HTTPConnection, I changed it back so that it gives the user flexibility not to pass in anything into `api_spec_path` (although that may not be a good idea in terms of verification: https://stackoverflow.com/questions/51848288/in-python-should-we-set-optionalstr-none-in-the-argument-list 

But still good for demo purposes to showcase the differences between passing in the YAML file versus not.

packages/fetchai/connections/http/connection.py
Kevin-Chen0: Update, it is now in _process_initial_api, but there is a check HTTPChannel that `api_spec_path` is not null.

packages/fetchai/connections/http/connection.py
Kevin-Chen0: Did not change this `receive` function yet as the requests are already being picked up by do_GET, do_POST methods of HTTPHandler. Not sure whether this `receive` function will need any additional modification.

packages/fetchai/connections/http/connection.py
Kevin-Chen0: I renamed it to `_process_incoming_request`. I first changed this function so that it returns an Envelope. Next I input this Envelope into verification helper object **APIVerificationHelper** (see below) that will then output the response.

packages/fetchai/connections/http/connection.py
Kevin-Chen0: Same thing as GET. Although now both calls the function `_send` in **HTTPHandlerFactory** that takes care of the Envelope and response so that `do_GET` and `do_POST` don't have to individually.

packages/fetchai/connections/http/connection.py
Kevin-Chen0: Not sure what you mean. If you mean by socket, no as long as you pick an available port? I'm currently running on localhost.

packages/fetchai/connections/http/connection.py
Kevin-Chen0: indeed, next iteration or so

packages/fetchai/connections/http/connection.py
Kevin-Chen0: Update, I currently have the **APIVerificationHelper** (see below) that directly returns the response as it is. Maybe next iteration it will return the response in an Envelope format.

packages/fetchai/connections/http/connection.py
Kevin-Chen0: I hope it is fine **APIVerificationHelper** class it nested under HTTPConnection. Next step, it may probably be its own Skill handler that receives and responds both in Envelope format through the Multiplexer.

packages/fetchai/connections/http/connection.py
Kevin-Chen0: Update, revert back to `api_spec_path = cast(str, connection_configuration.config.get("api_spec_path"))` as mypy has been given me the following errors:

```
connection.py:400: error: 'cast' expects 2 arguments
connection.py:400: error: No overload variant of "cast" matches argument types "Any", "Any", "str"
connection.py:400: note: Possible overload variants:
connection.py:400: note:     def [_T] cast(tp: Type[_T], obj: Any) -> _T
connection.py:400: note:     def cast(tp: str, obj: Any) -> Any
```

Kevin-Chen0:
**Updates:**

1) Added **APIVerificationHelper** helper object to HTTPConnection.
2) Addressed the comments on this PR.

**Next steps:**

1) Migrate the **APIVerificationHelper** out of HTTPConnection to become a HTTP Skill handler?
2) Start working on the HTTP protocol.
3) Fix any bugs arising from running AEA using the HTTPConnection, including:
`
  File ".../agents-aea/aea/decision_maker/base.py", line 529, in stop
    self._thread.join()
AttributeError: 'NoneType' object has no attribute 'join'
`

packages/fetchai/connections/http/connection.py
```diff
@@ -0,0 +1,416 @@
...
+    return HTTPHandler
+
+
+class APIVerificationHelper:
+    def __init__(self, e: Optional[Envelope]):
+        self.valid_paths = {}
+        if e is not None:
+            self.valid_paths = json.loads(e.message.decode())
+
+    def _verify(self, e: Envelope) -> str:
```
DavidMinarsch: this should not consume an envelope, instead it should consume the url, path, params

packages/fetchai/connections/http/connection.py
```diff
@@ -0,0 +1,416 @@
...
+        path_key = msg["path"]
+        params = json.loads(msg["params"])
+        if bool(params):  # Check param dict is not empty
+            param_key, param_value = list(params.items())[0]
+            param_value = param_value[0]
+            path_key += r"/{" + param_key + r"}"
+        method_key = msg["performative"].lower()
+
+        try:
+            status_code = "200"
```
DavidMinarsch: constants should be declared at the top of the file.

packages/fetchai/connections/http/connection.py
```diff
@@ -0,0 +1,416 @@
...
+            # Send the Envelope to the Agent's InBox.
+            # assert self._channel.in_queue is not None, "Input queue not initialized."
+            # assert self._channel.loop is not None, "Loop not initialized."
+            # self._channel.send(api_req_envelope)
+
+    return HTTPHandler
+
+
+class APIVerificationHelper:
+    def __init__(self, e: Optional[Envelope]):
```
DavidMinarsch: this should consume the api_specification_path, not the envelopes

packages/fetchai/connections/http/connection.py
DavidMinarsch: Still, your entire docstring does not match the signature.

packages/fetchai/connections/http/connection.py
DavidMinarsch: You are right, it does not need modification at this point afaikt. Just explaining to you what the supposed behaviour is.

packages/fetchai/connections/http/connection.py
```diff
@@ -0,0 +1,416 @@
...
+            self.send_response(200)
+            self.send_header("Content-type", "text/html")
+            self.end_headers()
+
+            parsed_path = urlparse(self.path)
+            url = parsed_path.path
+            # url = parsed_path.geturl()
+            param = parse_qs(parsed_path.query)
+            param = json.dumps(param)
+
```
DavidMinarsch: at this point I'd have something like:
```
is_valid_request, response_code = verifier.verify("GET", url, param)
if is_valid_request:
    # create envelope
    # send envelope into agent
    # wait for response envelope
   if # response envelope arrives within timeout window:
      # serve its payload
   else:
    response_code = 408 # request timeout
    self.send_response(response_code)
    self.send_header("Content-type", "text/html")
    self.end_headers()
else:
    self.send_response(response_code)
    self.send_header("Content-type", "text/html")
    self.end_headers()
```

packages/fetchai/connections/http/connection.py
DavidMinarsch: ok

packages/fetchai/connections/http/connection.py
DavidMinarsch: as in, the call to `connect()` will never return in this setup I believe. So your multiplexer would block here. I think you need to run this in a thread.

packages/fetchai/connections/http/connection.py
DavidMinarsch: As discussed offline, the APIVerificationHelper does not deal with the envelopes.

DavidMinarsch:
> **Updates:**
> 
> 1. Added **APIVerificationHelper** helper object to HTTPConnection.
> 2. Addressed the comments on this PR.
> 
> **Next steps:**
> 
> 1. Migrate the **APIVerificationHelper** out of HTTPConnection to become an HTTP Skill handler? Edit*: It will be left in HTTPConnection.
> 2. Start working on the HTTP protocol.
> 3. Fix any bugs arising from running AEA using the HTTPConnection, including:
>    `  File ".../agents-aea/aea/decision_maker/base.py", line 529, in stop self._thread.join() AttributeError: 'NoneType' object has no attribute 'join'`

1. is not a relevant step at this point. Please see my comments and the discussion we had regarding this.

packages/fetchai/connections/http/connection.py
Kevin-Chen0: that is fixed, will be in next commit.

packages/fetchai/connections/http/connection.py
Kevin-Chen0: done, will be in next commit

packages/fetchai/connections/http/connection.py
Kevin-Chen0: fixed, also removed function _process_initial_api() as it is not needed anymore.

packages/fetchai/connections/http/connection.py
Kevin-Chen0: Fixed, I'm still using the function `_send()`, which I renamed to `_process()` in HTTPHandler.

Kevin-Chen0:
**Updates:**

1) Address the PR comments regarding APIVerificationHelper.
2) Written the send and receive Envelope functions for HTTP Channel.

**Next steps:**

1) Write basic pet store skill/handler that returns the response Envelope.
2) Finalize the receive function in HTTP channel.

DavidMinarsch:
Closing this and reopening to simplify code reviews.

PR #861, opened 2020-02-27
AEA-253 Custom type protobuf description in protocol specification

5A11:
## Proposed changes

Protocol buffer serialisation code of every custom type in a protocol specification is included in a second yaml document of the specification itself.

## Types of changes

## Checklist


DavidMinarsch (approved):
LGTM

OjusWiZard (commented):
There is a potential infinite loop and memory exhaustion bug in the new `_varint_encode` function when handling negative integers.

Because Python integers have arbitrary precision, right-shifting a negative number preserves the sign bit (e.g., `-1 >> 7 == -1`). If a negative integer is passed to `_varint_encode`, `number` will never reach `0`, causing the `while True` loop to run indefinitely while continuously appending to `buf`.

While the framework currently only passes known positive integers (lengths and predefined hash codes) to this function, adding a guard is strongly recommended to prevent future hangs if this utility is reused or exposed to user input.

**Suggested Fix (`aea/helpers/multiformat.py`):**
```python
def _varint_encode(number: int) -> bytes:
    """Encode an integer as an unsigned varint (LEB128)."""
    if number < 0:
        raise ValueError("Cannot encode negative integers as unsigned varints")
    buf = bytearray()
    # ...
```

**Regression Test (`tests/test_helpers/test_multiformat.py`):**
```python
class TestVarint:
    def test_varint_encode_negative(self) -> None:
        """Test that negative numbers cannot be varint encoded."""
        from aea.helpers.multiformat import _varint_encode

        with pytest.raises(ValueError, match="Cannot encode negative integers"):
            _varint_encode(-1)
```

PR #863, opened 2020-02-27
Contracts added to CLI, tests updated. AEA-391

panasevychol:
## Proposed changes

Contracts added to CLI functionality (add, search, push, etc.)
Tests updated.

## Types of changes

## Checklist


cyenyxe:
Is this still WIP even though a review has been requested? The code coverage decrease in `search.py` is pretty high (more than 10%), and the test file associated with it doesn't seem to have been updated.

aea/cli/search.py
```diff
@@ -139,6 +140,38 @@ def connections(ctx: Context, query):
     print(format_items(sorted(result, key=lambda k: k["name"])))
 
 
+@search.command()
+@click.option("--query", default="", help="Query string to search Contracts by name.")
+@pass_ctx
+def contracts(ctx: Context, query):
+    """Search for Contracts."""
+    if ctx.config.get("is_registry"):
```
DavidMinarsch: Please refactor this and the other search commands do avoid code duplication and allow for code reuse.

DavidMinarsch: Rows 148-172 are generic apart from the `contracts` string afaikt

DavidMinarsch: Could have something like `_search_item(ctx, query, pluralized_item_type)`

aea/cli/common.py
```diff
@@ -248,29 +252,6 @@ def format_items(items):
...
-def format_skills(items):
-    """Format list of skills to a string for CLI output."""
-    list_str = ""
-    for item in items:
-        list_str += (
-            "{line}\n"
-            "Public ID: {public_id}\n"
-            "Name: {name}\n"
-            "Description: {description}\n"
-            "Protocols: {protocols}\n"
```
DavidMinarsch: Note, @panasevychol you removed the dependency info from the search command output for skills now. That's ok but we should open an issue that dependencies should be shown for skills (protocols, contracts) and agents (skills, contracts, protocols, connections).

aea/cli/search.py
```diff
@@ -103,121 +103,136 @@ def _get_details_from_dir(
...
+            configs[item_type_plural]["loader"],
+            registry,
+            "*/{}".format(item_type_plural),
+            configs[item_type_plural]["config_file"],
+            result,
+        )
+
+    _get_details_from_dir(
+        configs[item_type_plural]["loader"],
+        lookup_dir,
```
DavidMinarsch: I would change here `lookup_dir` to registry.

aea/cli/search.py
```diff
@@ -103,121 +103,136 @@ def _get_details_from_dir(
...
+        lookup_dir = registry
+    else:
+        lookup_dir = AEA_DIR
+        _get_details_from_dir(
+            configs[item_type_plural]["loader"],
+            registry,
+            "*/{}".format(item_type_plural),
+            configs[item_type_plural]["config_file"],
+            result,
+        )
```
DavidMinarsch: I'd change here to:
```suggestion
    if item_type_plural != "agents":
        _get_details_from_dir(
            configs[item_type_plural]["loader"],
            AEA_DIR,
            "*/{}".format(item_type_plural),
            configs[item_type_plural]["config_file"],
            result,
        )
```

DavidMinarsch: plus change in following comment make this bit easier to read

aea/cli/search.py
panasevychol: That's a bit different cause we look for items in registry and AEA_DIR, and for agents only in registry. Looks at those 2 calls.

aea/cli/search.py
panasevychol: Not a big deal, can change. I just prefer in cases where we have one condition and everything else goes under `else` to check that very condition in `if` statement. But I won't insist.

aea/cli/common.py
panasevychol: @DavidMinarsch we discussed that before and you told that we won't need protocols to be shown in skills search results. That just wasn't done that time.
IMO we should have an opened issue about output format. Now it looks a bit poor, maybe we should update it, add more info, better formatting. Will create one.

PR #872, opened 2020-03-01
Replace default protocol with auto generated protocol

DavidMinarsch:
## Proposed changes

This PR replaces the `default` protocol with the one which the generator generates when using `aea generate protocol ../examples/protocol_specification_ex/default.yaml`

## Fixes

- #873 

## Types of changes

## Checklist

## Further comments

-


aea/protocols/default/serialization.py
```diff
@@ -17,53 +17,91 @@
...
+        performative_id = msg.performative
+        if performative_id == DefaultMessage.Performative.BYTES:
+            performative = default_pb2.DefaultMessage.Bytes()  # type: ignore
+            content = msg.content
+            performative.content = content
+            default_msg.bytes.CopyFrom(performative)
+        elif performative_id == DefaultMessage.Performative.ERROR:
+            performative = default_pb2.DefaultMessage.Error()  # type: ignore
+            error_code = msg.error_code
+            performative.error_code = error_code
```
DavidMinarsch: @5A11 this assignment does not actually work.

aea/skills/error/handlers.py
```diff
@@ -67,10 +68,13 @@ def send_unsupported_protocol(self, envelope: Envelope) -> None:
         """
         logger.warning("Unsupported protocol: {}".format(envelope.protocol_id))
         reply = DefaultMessage(
-            type=DefaultMessage.Type.ERROR,
-            error_code=DefaultMessage.ErrorCode.UNSUPPORTED_PROTOCOL,
+            dialogue_reference=("", ""),
```
DavidMinarsch: Note, @5A11 and @marcofavorito we now have these default arguments which are `dialogue_reference`, `message_id` and `target`. In this scenario this creates an unnecessary overhead. We might want to allow for protocol specifications to specify that these arguments are not included. Thoughts?

aea/skills/error/handlers.py
```diff
@@ -67,10 +68,13 @@ def send_unsupported_protocol(self, envelope: Envelope) -> None:
...
+            message_id=1,
+            target=0,
+            performative=DefaultMessage.Performative.ERROR,
+            error_code=ErrorCode.UNSUPPORTED_PROTOCOL,
```
DavidMinarsch: Note, @marcofavorito since ErrorCode is not defined fully by the specification we had to move it into its own file (to be able to completely generate `message.py`). What do you think of the current implementation?

examples/protocol_specification_ex/default.yaml
```diff
@@ -9,16 +10,16 @@ speech_acts:
...
+...
 ---
 ct:ErrorCode: |
   enum ErrorCodeEnum {
-      UNSUPPORTED_PROTOCOL = -10001;
-      DECODING_ERROR = -10002;
-      INVALID_MESSAGE = -10003;
-      UNSUPPORTED_SKILL = -10004;
-      INVALID_DIALOGUE = -10005;
+      UNSUPPORTED_PROTOCOL = 0;
```
DavidMinarsch: @5A11 negative values are not permitted and enums must start from 0 for protobuf 3

aea/protocols/default/models.py
```diff
@@ -0,0 +1,32 @@
+# -*- coding: utf-8 -*-
```
marcofavorito: why `models.py`? In fact, there's no model. So better having another module for that?

examples/protocol_specification_ex/default.yaml
marcofavorito: 1. This is a backwards-incompatible change
2.  we don't have to use the Protobuf `enum`. The value of the error code can be checked at runtime, so we have more flexibility on the values and we are implementation independent.

aea/skills/error/handlers.py
marcofavorito: That is fine.

aea/skills/error/handlers.py
marcofavorito: Ideally, shouldn't the dialogue be the same as the incoming message? 
Some thoughts:
- we could move the dialogue system to `Envelope`
- `dialogue_message=0` and `target=-1` as default values for undefined dialogues.

DavidMinarsch: Not sure what you mean with `the dialogue be the same as the incoming message`. We don't keep track of the dialogue on the `Envelope` level as the `dialogue` is a concept only maintained by the agents, not by the peers which route the envelope. This also makes sense when thinking about encryption as you can easily encrypt all the message body but not the envelope fields.

DavidMinarsch: We start indexing messages at 1 to avoid negative indeces on target.

examples/protocol_specification_ex/default.yaml
DavidMinarsch: Fair point. @5A11 thoughts?

aea/protocols/default/models.py
DavidMinarsch: @marcofavorito agreed, how about `custom_types.py`?

marcofavorito: Ok, but it is not a standard module name

aea/skills/error/handlers.py
marcofavorito: The commented lines correspond to the reply to an envelope whose protocol id is not supported.
Let's say the sender sends two envelopes with a protocol not supported by the recipient. With this code, the recipient answers with two messages. How the sender can distinguish between the two replies? This is an argument to include some identification system for envelopes.

DavidMinarsch: Actually, on discussion with @5A11 we realise that the error handler starts a new high level interaction in the default protocol. So we should set the LHS dialogue reference. The dialogue reference will allow the receiver to tell the two messages apart. Envelopes are part of another 'meta' protocol, we should be careful about having dependencies between the two.

aea/protocols/default/serialization.py
5A11: Upon inspection, this line cannot be fixed quickly with the current setup (I didn't realise at first it is a custom type enum, so mistakenly told @DavidMinarsch I have a quick fix for it). 

In general, custom types are Not taken care of in the current generator. They appear in:
- `serialisation.py`
- `message.py`
- `X.proto`

They are an outstanding problem in `serialisation.py`. And even in the other two, we don't really properly tackle them.
- In `message.py`, we decided to take the class definitions out and put them in another module (`models.py` or `custom_types.py`) and have them `raise NotImplemented`.
- In `X.proto` we decided that their Protobuf definition is provided in the specification.

I think if a specification contains custom types we should either:
- generate the protocol package but warn the user and tell them to update the three files to cover the custom types, or 
- come up with a proper way to handle them (describing custom types in yaml/ontology) and generating the code in all the three files (`serialisation.py`, `message.py`, `X.proto`).

aea/protocols/default/models.py
5A11: Changed to `custom_types.py`.

aea/connections/stub/connection.py
```diff
@@ -143,13 +143,11 @@ def read_envelopes(self) -> None:
         """Receive new envelopes, if any."""
         line = self.input_file.readline()
         logger.debug("read line: {!r}".format(line))
+        lines = b""
```
DavidMinarsch: @marcofavorito I had to change this implementation. It turns out the old one was problematic with protobuf as protobuf encoded messages could contain a newline character causing this logic not to work. Now we are constrained to single envelopes, but I think this is ok as the connection will pick them off the file very quickly

aea/protocols/default/custom_types.py
```diff
@@ -0,0 +1,46 @@
...
+#   limitations under the License.
+#
+# ------------------------------------------------------------------------------
+
+"""This module contains class representations corresponding to every custom type in the protocol specification."""
+
+from enum import Enum
+
+
+class ErrorCode(Enum):
```
DavidMinarsch: @5A11 this is much simplified relative to before. and avoids nested custom classes.

aea/protocols/default/message.py
```diff
@@ -17,22 +17,27 @@
...
 
 
 class DefaultMessage(Message):
-    """The Default message class."""
+    """A protocol for exchanging any bytes message."""
+
+    protocol_id = ProtocolId("fetchai", "default", "0.1.0")
 
-    protocol_id = PublicId("fetchai", "default", "0.1.0")
+    ErrorCode = CustomErrorCode
```
DavidMinarsch: @5A11 all custom classes should by default be exposed here. Then they are accessible via `DefaultMessage.ErrorCode` which means only the `DefaultMessage` needs to be imported to access all its attributes and member classes.

PR #878, opened 2020-03-02
AEA-386 ERC1155 smart contract

### LINKED ISSUE #801: Create ERC1155 contract package

Totoual:
## Proposed changes

Adds the erc1155 smart contract. 
## Fixes

#801 

## Types of changes

## Checklist


aea/contracts/ethereum.py
```diff
@@ -73,17 +73,18 @@ def set_instance(self, ledger_api: EthereumApi) -> None:
         :param ledger_api: the ethereum ledger api
         """
         assert self._instance is None, "Instance already set!"
-        self._instance = ledger_api.api.web3.Contract(
+        self._instance = ledger_api.api.eth.contract(
             abi=self.abi, bytecode=self.bytecode
         )
 
-    def set_address(self, ledger_api: EthereumApi, contract_address: str) -> None:
+    def set_address(self, ledger_api: EthereumApi, tx_digest: str) -> None:
```
DavidMinarsch: For separation of concerns I'd keep this as previously implemented.

packages/fetchai/contracts/erc1155/contract.py
```diff
@@ -472,10 +511,9 @@ def generate_id(self, token_id, item_id):
...
         return final_id_int
 
-    #
-    # def generate_trade_nonce(self, contract, address):
-    #     """Generate a valid trade nonce."""
-    #     trade_nonce = random.randrange(0, 10000000)
-    #     while contract.instance.functions.is_nonce_used(address, trade_nonce).call():
-    #         trade_nonce = random.randrange(0, 10000000)
-    #     return trade_nonce
+    def generate_trade_nonce(self, contract, address):
```
DavidMinarsch: already defined above

packages/fetchai/skills/erc1155_skill/handlers.py
```diff
@@ -48,17 +48,19 @@ def handle(self, message: Message) -> None:
         if tx_msg_response.tx_id == "contract_deploy":
             tx_signed = tx_msg_response.signed_payload.get("tx_signed")
             ledger_api = self.context.ledger_apis.apis.get("ethereum")
-            result = ledger_api.send_raw_transaction(  # type: ignore
-                tx_signed
+            result = ledger_api.send_signed_transaction(  # type: ignore
+                True, tx_signed
```
DavidMinarsch: please don't use magic variables, makes it hard to read

packages/fetchai/skills/erc1155_skill/handlers.py
```diff
@@ -48,17 +48,19 @@ def handle(self, message: Message) -> None:
...
             )
             self.context.logger.info(result)
-            while contract.instance.address is None:
-                contract.update_contract_instance(
-                    contract_address=result.contractAddress,
-                    ledger_api=self.context.ledger_apis.apis.get("ethereum"),
-                )
-                time.sleep(3)
-            contract.is_deployed = True
+            contract.set_address(ledger_api, result)
```
DavidMinarsch: this should not be fed the `result` var, it should be fed the `address`

aea/contracts/ethereum.py
Totoual: But I will not be able to get the contract address...

packages/fetchai/contracts/erc1155/contract.py
Totoual: It is a copy, I am trying to remove the helpers class.

packages/fetchai/skills/erc1155_skill/handlers.py
Totoual: At this point you don't have the address. You are returning the tx_digest from send_signed_transaction and not the attributeDict or the contract address.

packages/fetchai/skills/erc1155/__init__.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# ------------------------------------------------------------------------------
+#
+#   Copyright 2018-2019 Fetch.AI Limited
```
DavidMinarsch: Careful, your dates are not correct :)

packages/fetchai/skills/erc1155/contract.py
```diff
@@ -0,0 +1,536 @@
+# -*- coding: utf-8 -*-
+# ------------------------------------------------------------------------------
+#
+#   Copyright 2018-2019 Fetch.AI Limited
```
DavidMinarsch: Why is the contract duplicated here?

packages/fetchai/skills/erc1155/contract.yaml
```diff
@@ -0,0 +1,11 @@
+name: erc1155
+author: fetchai
+version: 0.1.0
```
DavidMinarsch: see above

packages/fetchai/skills/erc1155/contracts/Migrations.sol
```diff
@@ -0,0 +1,23 @@
+pragma solidity >=0.4.21 <0.6.0;
+
+contract Migrations {
+  address public owner;
+  uint public last_completed_migration;
+
+  constructor() public {
+    owner = msg.sender;
```
DavidMinarsch: see above

packages/fetchai/skills/erc1155/contracts/erc1155.vy
```diff
@@ -0,0 +1,710 @@
+# Author: Sören Steiger, github.com/ssteiger
+# Author: Fetch.ai, github.com/fetchai
+# License: MIT
+
```
DavidMinarsch: see above

packages/fetchai/skills/erc1155/migrations/1_initial_migration.js
```diff
@@ -0,0 +1,5 @@
+var Migrations = artifacts.require("./Migrations.sol");
+
```
DavidMinarsch: see above

packages/fetchai/skills/erc1155/migrations/2_deploy_contracts.js
```diff
@@ -0,0 +1,5 @@
+var ERC1155 = artifacts.require("erc1155");
+
```
DavidMinarsch: see above

packages/fetchai/skills/erc1155_client/__init__.py
```diff
@@ -17,4 +17,4 @@
 #
 # ------------------------------------------------------------------------------
 
-"""This module contains the implementation of the default skill."""
+"""This module contains the implementation of the generic buyer skill."""
```
DavidMinarsch: wrong description

packages/fetchai/skills/erc1155_client/behaviours.py
```diff
@@ -0,0 +1,127 @@
...
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+# ------------------------------------------------------------------------------
+
+"""This package contains the behaviour for the generic buyer skill."""
```
DavidMinarsch: description not aligned

packages/fetchai/skills/erc1155_client/dialogues.py
```diff
@@ -0,0 +1,62 @@
...
+from typing import Optional
+
+from aea.helpers.dialogue.base import DialogueLabel
+from aea.helpers.search.models import Description
+from aea.skills.base import Model
+
+from packages.fetchai.protocols.fipa.dialogues import FIPADialogue, FIPADialogues
+
+
+class Dialogue(FIPADialogue):
```
DavidMinarsch: Not needed as you're not keeping track of a dialogue

packages/fetchai/skills/erc1155_client/handlers.py
```diff
@@ -0,0 +1,237 @@
...
+        self.context.logger.info(
+            "[{}]: received INFORM from sender={}".format(
+                self.context.agent_name, msg.counterparty[-5:]
+            )
+        )
+        if len(msg.info.keys()) >= 1:
+            data = msg.info
+            if "contract" in data.keys():
+                contract = self.context.contracts.erc1155
+                contract.set_instance(
```
DavidMinarsch: This usage suggests it would be good to also expose an abstract method on the base contract which performs `set_instance` and `set_address` in one; i.e. `set_instance_w_address`.

packages/fetchai/skills/erc1155_client/handlers.py
```diff
@@ -0,0 +1,237 @@
...
+                    ledger_api=self.context.ledger_apis.apis.get("ethereum")
+                )
+                contract.set_address(
+                    ledger_api=self.context.ledger_apis.apis.get("ethereum"),
+                    contract_address=data["contract"],
+                )
+                contract.item_ids = data["item_ids"]
```
DavidMinarsch: would be good to provide some logging statements throughout to explain what's happening

packages/fetchai/skills/erc1155_client/skill.yaml
```diff
@@ -0,0 +1,37 @@
...
+  strategy:
+    class_name: Strategy
+    args:
+      ledger_id: 'ethereum'
+      is_ledger_tx: True
+      search_query:
+        search_term: country
+        search_value: UK
+        constraint_type: '=='
+  dialogues:
```
DavidMinarsch: can be dropped

packages/fetchai/skills/erc1155_client/handlers.py
Totoual: Yes. That's true. All the set_address does is actually reinstantiate the contract. So I don't think that we should introduce a new function. Just remove the assert from the set_address? Thoughts?

packages/fetchai/skills/erc1155_client/handlers.py
```diff
@@ -0,0 +1,237 @@
...
+        :param msg: the message
+        :return: None
+        """
+        self.context.logger.info(
+            "[{}]: received INFORM from sender={}".format(
+                self.context.agent_name, msg.counterparty[-5:]
+            )
+        )
+        if len(msg.info.keys()) >= 1:
+            data = msg.info
```
Totoual: We are expecting to receive the following from the inform message:
- contract_address
- item_ids
- trade_nonce.
Currently the trade_nonce is hardcoded but now I am going to store it in the shared_state and pass it around.

packages/fetchai/skills/erc1155_client/handlers.py
Totoual: Since we don't have a way to get the item_ids from the contract (And actually we don't store them inside the contract but we are mapping the item_ids with the addresses) we need to inform the client for the item_ids. The other way around would be to mint the items to the client and then ask the balance. That means we will not pass the item_ids to the message.

packages/fetchai/skills/erc1155_client/handlers.py
```diff
@@ -0,0 +1,237 @@
...
+                    item_id=contract.item_ids[0],
+                    from_supply=10,
+                    to_supply=0,
+                    value=0,
+                    trade_nonce=500,
+                    ledger_api=self.context.ledger_apis.apis.get("ethereum"),
+                    skill_callback_id=self.context.skill_id,
+                )
+
+                self.context.decision_maker_message_queue.put_nowait(tx)
```
Totoual: We request from the contract to hash a transaction and then we pass this to the decision maker for signing.

packages/fetchai/skills/erc1155_client/handlers.py
```diff
@@ -0,0 +1,237 @@
...
+            new_message_id = 2
+            new_target = 1
+
+            inform_msg = FIPAMessage(
+                message_id=new_message_id,
+                dialogue_reference=("", ""),
+                target=new_target,
+                performative=FIPAMessage.Performative.INFORM,
+                info={"signature": tx_signed},
+            )
```
Totoual: Create the inform message. for the deployer once we signed the transaction. The deployer will send the transaction to the smart contract. And the smart contract will check the signatures. If the signatures match the trade will be completed.

packages/fetchai/skills/erc1155_client/strategy.py
```diff
@@ -0,0 +1,110 @@
...
+"""This module contains the strategy class."""
+
+from typing import cast
+
+from aea.helpers.search.models import Constraint, ConstraintType, Description, Query
+from aea.skills.base import Model
+
+DEFAULT_MAX_PRICE = 5
+DEFAULT_MAX_BUYER_TX_FEE = 2
+DEFAULT_CURRENCY_PBK = "FET"
```
Totoual: I will drop these.

packages/fetchai/skills/erc1155_client/strategy.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def __init__(self, **kwargs) -> None:
+        """
+        Initialize the strategy of the agent.
+
+        :return: None
+        """
+        self._max_price = kwargs.pop("max_price", DEFAULT_MAX_PRICE)
+        self.max_buyer_tx_fee = kwargs.pop("max_buyer_tx_fee", DEFAULT_MAX_BUYER_TX_FEE)
+        self._currency_id = kwargs.pop("currency_id", DEFAULT_CURRENCY_PBK)
```
Totoual: And these.

packages/fetchai/skills/erc1155_deploy/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+# ------------------------------------------------------------------------------
+
+"""This module contains the implementation of the generic seller skill."""
```
Totoual: I'll change the doctstring to match.

packages/fetchai/skills/erc1155_deploy/behaviours.py
```diff
@@ -0,0 +1,234 @@
...
+    """This class implements a behaviour."""
+
+    def __init__(self, **kwargs):
+        """Initialise the behaviour."""
+        services_interval = kwargs.pop(
+            "services_interval", DEFAULT_SERVICES_INTERVAL
+        )  # type: int
+        super().__init__(tick_interval=services_interval, **kwargs)
+        self._registered_service_description = None  # type: Optional[Description]
+        self.contract_address = "0x7939eb9BF181bAE263BCb74bfb87464652A6dfE3"
```
Totoual: I am setting an address of an already deployed contract in order to avoid deployment all the time (I am short on eth :) )

packages/fetchai/skills/erc1155_deploy/behaviours.py
```diff
@@ -0,0 +1,234 @@
...
+            )
+            self.context.decision_maker_message_queue.put_nowait(dm_message_for_deploy)
+        else:
+            self.context.logger.info("Setting the address of the deployed contract")
+            contract.set_address(
+                ledger_api=self.context.ledger_apis.apis.get("ethereum"),
+                contract_address=self.contract_address,
+            )
+            create_items = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
+            contract.create_item_ids(token_type=strategy.ft, token_ids=create_items)
```
Totoual: I ll remove this once I have a fully working example.

packages/fetchai/skills/erc1155_deploy/behaviours.py
```diff
@@ -0,0 +1,234 @@
...
+
+    def act(self) -> None:
+        """
+        Implement the act.
+
+        :return: None
+        """
+        contract = self.context.contracts.erc1155
+        strategy = cast(Strategy, self.context.strategy)
+        if not contract.is_deployed or self.contract_address is None:
```
Totoual: I ll remove this once I have a fully working example.

packages/fetchai/skills/erc1155_deploy/generic_data_model.py
```diff
@@ -0,0 +1,52 @@
...
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+# ------------------------------------------------------------------------------
+
+"""This package contains the dataModel for the generic seller aea."""
```
Totoual: Change the docstring.

packages/fetchai/skills/erc1155_deploy/handlers.py
```diff
@@ -0,0 +1,203 @@
...
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+# ------------------------------------------------------------------------------
+
+"""This package contains the handlers of a generic seller AEA."""
```
Totoual: Docstring.

packages/fetchai/skills/erc1155_deploy/generic_data_model.py
```diff
@@ -0,0 +1,52 @@
...
+            ), "Wrong type!! is_required must be bool"
+            self.attributes.append(
+                Attribute(
+                    name=values["name"],  # type: ignore
+                    type=SUPPORTED_TYPES[values["type"]],
+                    is_required=values["is_required"],
+                )
+            )
+
+        super().__init__("weather_station_datamodel", self.attributes)
```
Totoual: Change the name.

packages/fetchai/skills/erc1155_deploy/generic_data_model.py
```diff
@@ -0,0 +1,52 @@
...
+"""This package contains the dataModel for the generic seller aea."""
+from typing import Any, Dict
+
+from aea.helpers.search.models import Attribute, DataModel
+
+SUPPORTED_TYPES = {"str": str, "int": int, "float": float, "bool": bool}
+
+
+class Generic_Data_Model(DataModel):
+    """Data model for the generic seller aea."""
```
Totoual: Docstring.

packages/fetchai/skills/erc1155_deploy/handlers.py
```diff
@@ -0,0 +1,203 @@
...
+            target=new_target,
+            performative=FIPAMessage.Performative.INFORM,
+            info={
+                "contract": contract.instance.address,
+                "item_ids": contract.item_ids,
+                "trade_nonce": contract.generate_trade_nonce(
+                    self.context.agent_address
+                ),
+            },
+        )
```
Totoual: Generating the first message for the counterparty once we receive an cfp.

packages/fetchai/skills/erc1155_deploy/handlers.py
```diff
@@ -0,0 +1,203 @@
...
+            from_supply=10,
+            to_supply=0,
+            value=0,
+            trade_nonce=500,
+            ledger_api=self.context.ledger_apis.apis.get("ethereum"),
+            skill_callback_id=self.context.skill_id,
+            signature=msg.info.get("signature"),
+        )
+
+        self.context.decision_maker_message_queue.put(tx)
```
Totoual: Creating the atomic swap tx and queuing it to the Decision Maker.

packages/fetchai/skills/erc1155_client/handlers.py
Totoual: I ll add the third function ass we discussed.

aea/contracts/ethereum.py
```diff
@@ -73,17 +73,31 @@ def set_instance(self, ledger_api: EthereumApi) -> None:
...
+    ) -> None:
+        """
+                Set the contract address.
+
+                :param ledger_api: the ledger_api we are using.
+                :param contract_address: the contract address
+                """
+        self._instance = ledger_api.api.eth.contract(
             address=contract_address, abi=self.abi
         )
```
DavidMinarsch: this is identical to set address... I was thinking this should also set the bytecode and have two asserts:
```suggestion
        assert self._instance is None, "Instance already set!"
        assert self.instance.address is None, "Address already set!"
        self._instance = ledger_api.api.eth.contract(
            address=contract_address, abi=self.abi, bytecode=self.bytecode
        )
```

packages/fetchai/contracts/erc1155/contract.py
```diff
@@ -357,59 +392,85 @@ def get_atomic_swap_batch_transaction_proposal(
...
+        trade_nonce,
+        ledger_api,
+    ):
         """Sign the transaction before send them to agent1."""
         # assert self.address == terms.to_address
-        from_address_hash = self.instance.functions.getAddress(
-            terms.from_address
-        ).call()
-        to_address_hash = self.instance.functions.getAddress(terms.to_address).call()
+        from_address_hash = self.instance.functions.getAddress(from_address).call()
```
DavidMinarsch: Aren't we able to generate these hashes client side now without a contract call?

packages/fetchai/contracts/erc1155/contract.py
```diff
@@ -56,18 +57,20 @@ def __init__(
         super().__init__(contract_id, contract_config, contract_interface)
         self.is_items_created = False
         self.is_items_minted = False
-        self.token_ids = []  # type: List[int]
+        self.is_trade = False
```
DavidMinarsch: What is this for?

packages/fetchai/skills/erc1155_client/handlers.py
```diff
@@ -0,0 +1,234 @@
...
+        pass
+
+    def handle(self, message: Message) -> None:
+        """
+        Implement the reaction to a message.
+
+        :param message: the message
+        :return: None
+        """
+        # convenience representations
```
DavidMinarsch: the comments here are unnecessary

packages/fetchai/skills/erc1155_client/handlers.py
DavidMinarsch: so if you are expecting it, add some assertions ..then the code is more correct and needs less explanation :)

packages/fetchai/skills/erc1155_client/handlers.py
DavidMinarsch: could add a logger statement here with a similar message so its in code

packages/fetchai/skills/erc1155_deploy/behaviours.py
DavidMinarsch: this suggests adding a field in config of the contract, so a contract can already be preloaded with the address

packages/fetchai/skills/erc1155_deploy/behaviours.py
```diff
@@ -0,0 +1,235 @@
+# -*- coding: utf-8 -*-
+# ------------------------------------------------------------------------------
+#
+#   Copyright 2018-2020 Fetch.AI Limited
```
DavidMinarsch: I'd call the skills

packages/fetchai/skills/erc1155_deploy/behaviours.py
```diff
@@ -0,0 +1,235 @@
...
+        """
+        Implement the act.
+
+        :return: None
+        """
+        contract = self.context.contracts.erc1155
+        strategy = cast(Strategy, self.context.strategy)
+        if not contract.is_deployed or self.contract_address is None:
+            if contract.is_deployed and not contract.is_items_created:
+                create_items = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
```
DavidMinarsch: Don't do these magic vars..everything should be configurable in config files and then there should be `DEFAULT` constants

packages/fetchai/skills/erc1155_deploy/behaviours.py
```diff
@@ -0,0 +1,235 @@
...
+                    skill_callback_id=self.context.skill_id,
+                )
+                self.context.decision_maker_message_queue.put_nowait(creation_message)
+                contract.is_items_created = True
+            if (
+                contract.is_deployed
+                and contract.is_items_created
+                and not contract.is_items_minted
+            ):
+                mint_items = [20, 20, 20, 20, 26, 24, 22, 21, 23, 22]
```
DavidMinarsch: see above

packages/fetchai/skills/erc1155_deploy/behaviours.py
```diff
@@ -0,0 +1,235 @@
...
+                skill_callback_id=self.context.skill_id,
+            )
+            self.context.decision_maker_message_queue.put_nowait(dm_message_for_deploy)
+        else:
+            self.context.logger.info("Setting the address of the deployed contract")
+            contract.set_instance_w_address(
+                ledger_api=self.context.ledger_apis.apis.get("ethereum"),
+                contract_address=self.contract_address,
+            )
+            create_items = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
```
DavidMinarsch: see below

packages/fetchai/skills/erc1155_deploy/generic_data_model.py
```diff
@@ -0,0 +1,52 @@
...
+
+"""This package contains the dataModel for the erc1155 deploy skill aea."""
+from typing import Any, Dict
+
+from aea.helpers.search.models import Attribute, DataModel
+
+SUPPORTED_TYPES = {"str": str, "int": int, "float": float, "bool": bool}
+
+
+class Generic_Data_Model(DataModel):
```
DavidMinarsch: CamelCase

packages/fetchai/skills/erc1155_deploy/handlers.py
DavidMinarsch: All these magic values..really not ideal. Please comb through your code and make it configurable :)

aea/contracts/ethereum.py
Totoual: What's the point though ? Just not to be identical ?

packages/fetchai/contracts/erc1155/contract.py
Totoual: For handling the sequence in the Transaction handler of the deployer.

packages/fetchai/contracts/erc1155/contract.py
Totoual: No, cause the web3 hashes differently than the vyper. I will check again though.

packages/fetchai/contracts/erc1155/contract.py
DavidMinarsch: So then it should not be here. It should be in the skill imo.

packages/fetchai/contracts/erc1155/contract.py
DavidMinarsch: Check what we have in the research repo please, just to be sure.

docs/erc1155-skills.md
```diff
@@ -0,0 +1,214 @@
...
+``` bash
+aea create my_erc1155_client
+cd my_erc1155_client
+aea add connection fetchai/oef:0.1.0
+aea add skill fetchai/erc1155_client:0.1.0
+aea add contract fetchai/erc1155:0.1.0
+aea install
+```
+
+Additionally, create the private key for the client AEA.
```
DavidMinarsch: I don't think this step is needed? The keys will be generated automatically when running and this AEA does not make any transactions?

docs/erc1155-skills.md
```diff
@@ -0,0 +1,214 @@
...
+      ledger_id: 'ethereum'
+      is_ledger_tx: True
+      nft: 1
+      ft: 2
+      item_ids: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
+      mint_stock: [100, 100, 100, 100, 100, 100, 100, 100, 100, 100]
+      from_supply: 10
+      to_supply: 0
+      value: 0
+      search_schema:
```
DavidMinarsch: This search schema seems incorrect?

docs/erc1155-skills.md
```diff
@@ -0,0 +1,214 @@
...
+```
+The `search_schema` and the `search_data` are used to register the service in the OEF and make your agent discoverable. The name of each attribute must be a key in the `search_data` dictionary.
+
+
+### Fund the deployer AEA
+
+To create some wealth for your deployer AEA for the Ethereum `ropsten` network:
+
+``` bash
+aea generate-wealth ethereum
```
DavidMinarsch: Add that this needs to be executed from deployer AEA project

docs/erc1155-skills.md
```diff
@@ -0,0 +1,214 @@
...
+```
+
+## Run the AEAs
+
+You can change the endpoint's address and port by modifying the connection's yaml file (my_seller_aea/connection/oef/connection.yaml)
```
DavidMinarsch: Let's just assume this is run from the same host so we can skip this.

docs/erc1155-skills.md
```diff
@@ -0,0 +1,214 @@
...
+        participant Blockchain
+    
+        activate Deployer_AEA
+        activate Search
+        activate Client_AEA
+        activate Erc1155_contract
+        activate Blockchain
+        
+        Deployer_AEA->>Blockchain: deployes smart contract
+        Deployer_AEA->>Erc1155_contract: creates tokens
```
DavidMinarsch: ERC1155_Contract

mkdocs.yml
```diff
@@ -47,6 +47,7 @@ nav:
         - ORM integration: 'orm-integration-to-generic.md'
         - Frontend intergration: 'connect-a-frontend.md'
         - CLI vs programmatic AEAs: 'cli-vs-programmatic-aeas.md'
+        - Erc1155 deploy and interact: 'erc1155-skills.md'
```
DavidMinarsch: ERC1155 deploy and interact

tests/test_packages/test_contracts/__init__.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# ------------------------------------------------------------------------------
+#
+#   Copyright 2018-2019 Fetch.AI Limited
```
DavidMinarsch: date

tests/test_packages/test_contracts/conftest.py
```diff
@@ -0,0 +1,197 @@
+# -*- coding: utf-8 -*-
+# ------------------------------------------------------------------------------
+#
+#   Copyright 2018-2019 Fetch.AI Limited
```
DavidMinarsch: does that work having a second `conftest.py` file?

Pipfile
```diff
@@ -39,6 +39,8 @@ pytest-randomly = "==3.2.1"
 safety = "==1.8.5"
 tox = "==3.14.5"
 vyper = "==0.1.0b12"
+eth-tester = "==0.3.0b1"
```
DavidMinarsch: I'd remove this and the tests for the contract. Since the contract is already compiled here it does not make much sense to include this.

aea/contracts/ethereum.py
```diff
@@ -73,17 +73,31 @@ def set_instance(self, ledger_api: EthereumApi) -> None:
...
+        :param ledger_api: the ledger_api we are using.
         :param contract_address: the contract address
         """
         assert self.instance.address is None, "Address already set!"
-        self._instance = ledger_api.api.web3.Contract(
+        self._instance = ledger_api.api.eth.contract(
+            address=contract_address, abi=self.abi
+        )
+
+    def set_instance_w_address(
```
DavidMinarsch: maybe we call this `set_deployed_instance` instead? This might make its usage clearer.

aea/contracts/ethereum.py
DavidMinarsch: Actually, we don't need line 102, so just 
```suggestion
assert self._instance is None, "Instance already set!"
self._instance = ledger_api.api.eth.contract(
  address=contract_address, abi=self.abi, bytecode=self.bytecode
)
```
that way, we ensure it isn't accidentally called a second time

docs/erc1155-skills.md
```diff
@@ -0,0 +1,214 @@
...
+## Preparation instructions
+ 
+### Dependencies
+
+Follow the <a href="../quickstart/#preliminaries">Preliminaries</a> and <a href="../quickstart/#installation">Installation</a> sections from the AEA quick start.
+
+##Discussion
+
+The scope of the specific demo is to demonstrate how to deploy a smart contract and interact with it. For the specific use-case, we create two AEAs one that deploys and creates tokens inside the smart contract and the other that signs a transaction so we can complete an atomic swap. The smart contract we are using is an ERC1155 smart contract
+with a one-step atomic swap functionality. That means the trade between the two AEAs can be trustless.
```
DavidMinarsch: We should add a `note` below saying:
`This demo serves demonstrative purposes only.  Since the AEA deploying the contract also has the ability to mint tokens, in reality the transfer of tokens from the AEA signing the transaction is pointless.`

packages/fetchai/contracts/erc1155/contract.py
```diff
@@ -357,59 +418,85 @@ def get_atomic_swap_batch_transaction_proposal(
...
+        to_address,
+        item_ids,
+        from_supplies,
+        to_supplies,
+        value,
+        trade_nonce,
+        ledger_api,
+    ):
         """Sign the transaction before send them to agent1."""
         # assert self.address == terms.to_address
```
DavidMinarsch: please remove comment

packages/fetchai/contracts/erc1155/contract.py
```diff
@@ -357,59 +418,85 @@ def get_atomic_swap_batch_transaction_proposal(
...
+            _nonce=trade_nonce,
         )
 
         tx_message = TransactionMessage(
             performative=TransactionMessage.Performative.PROPOSE_FOR_SIGNING,
-            skill_callback_ids=[ContractId("fetchai", "erc1155_skill", "0.1.0")],
-            tx_id="contract_deployment",
-            tx_sender_addr=terms.from_address,
+            skill_callback_ids=[skill_callback_id],
+            tx_id="contract-sign-hash",
```
DavidMinarsch: No magic values like this please; create a constant and/or make it configurable.

packages/fetchai/contracts/erc1155/contract.py
```diff
@@ -334,17 +375,37 @@ def get_balance_of_batch(self, address):
...
+        to_supplies,
+        value,
+        trade_nonce,
+        signature,
+        skill_callback_id,
     ) -> TransactionMessage:
         """Make a trust-less trade for a batch of items between 2 agents."""
-        assert deployer_address == terms.from_address, "Wrong 'from' address"
-        tx = contract.get_trade_batch_tx(terms=terms, signature=signature)
+        assert deployer_address == from_address, "Wrong 'from' address"
```
DavidMinarsch: I don't think we want that assertion here.

packages/fetchai/contracts/erc1155/contract.py
```diff
@@ -357,59 +418,85 @@ def get_atomic_swap_batch_transaction_proposal(
...
+        item_id,
+        from_supply,
+        to_supply,
+        value,
+        trade_nonce,
+        ledger_api,
+        skill_callback_id,
+    ) -> TransactionMessage:
         """Sign the transaction before send them to agent1."""
         # assert self.address == terms.to_address
```
DavidMinarsch: please remove comment

packages/fetchai/contracts/erc1155/contract.py
```diff
@@ -303,18 +324,38 @@ def get_balance(self, from_address: Address, item_id: int):
...
+        from_supply,
+        to_supply,
+        value,
+        trade_nonce,
+        signature,
+        ledger_api: LedgerApi,
+        skill_callback_id: ContractId,
     ) -> TransactionMessage:
         """Make a trustless trade between to agents for a single token."""
         # assert self.address == terms.from_address, "Wrong from address"
```
DavidMinarsch: please remove comment

packages/fetchai/contracts/erc1155/contract.py
```diff
@@ -232,66 +233,86 @@ def _create_mint_batch_tx(
 
         return tx
 
-    def _create_trade_tx(self, terms, signature, ledger_api: LedgerApi) -> str:
+    def _create_trade_tx(
+        self,
+        from_address,
```
DavidMinarsch: Almost all function signatures are missing type declarations!

packages/fetchai/contracts/erc1155/contract.py
```diff
@@ -54,20 +52,19 @@ def __init__(
         :param contract_interface: the contract interface.
         """
         super().__init__(contract_id, contract_config, contract_interface)
-        self.is_items_created = False
-        self.is_items_minted = False
-        self.token_ids = []  # type: List[int]
         self.item_ids = []  # type: List[int]
 
     def create_item_ids(self, token_type: int, token_ids: List[int]) -> None:
```
DavidMinarsch: I still don't understand why we have the `token_ids` here ? Why can we not simple have a signature like this:
```
def create_token_ids(self, token_type: int, nb_tokens: int) -> None:
```

DavidMinarsch: And please remove the reference to `items` throughout this file. we should only use the term `token`

packages/fetchai/skills/erc1155_client/behaviours.py
```diff
@@ -0,0 +1,127 @@
...
+                        self.context.agent_name, eth_balance
+                    )
+                )
+            else:
+                self.context.logger.warning(
+                    "[{}]: you have no starting balance on ethereum ledger!".format(
+                        self.context.agent_name
+                    )
+                )
+                # TODO: deregister skill from filter
```
DavidMinarsch: This deregistering is now possible: https://github.com/fetchai/agents-aea/blob/develop/aea/skills/base.py#L99...just set it to `False`

packages/fetchai/skills/erc1155_client/behaviours.py
```diff
@@ -0,0 +1,127 @@
...
+                        self.context.agent_name, fet_balance
+                    )
+                )
+            else:
+                self.context.logger.warning(
+                    "[{}]: you have no starting balance on fetchai ledger!".format(
+                        self.context.agent_name
+                    )
+                )
+                # TODO: deregister skill from filter
```
DavidMinarsch: see below

packages/fetchai/skills/erc1155_deploy/generic_data_model.py
```diff
@@ -0,0 +1,52 @@
...
+
+"""This package contains the dataModel for the erc1155 deploy skill aea."""
+from typing import Any, Dict
+
+from aea.helpers.search.models import Attribute, DataModel
+
+SUPPORTED_TYPES = {"str": str, "int": int, "float": float, "bool": bool}
+
+
+class GenericDataModel(DataModel):
```
DavidMinarsch: We should extract this into `aea.helpers.search.models` and then use it from there rather than have it here. It is generic after all.

packages/fetchai/skills/erc1155_deploy/handlers.py
```diff
@@ -0,0 +1,216 @@
...
+        new_message_id = msg.message_id + 1
+        new_target = msg.message_id
+        self.context.logger.info(
+            "[{}]: received CFP from sender={}".format(
+                self.context.agent_name, msg.counterparty[-5:]
+            )
+        )
+        strategy = cast(Strategy, self.context.strategy)
+        contract = self.context.contracts.erc1155
+        self.context.logger.info(contract.item_ids)
```
DavidMinarsch: I really don't like these logger statements..no-one other than you will understand what this is once it prints to console.

packages/fetchai/skills/erc1155_deploy/handlers.py
```diff
@@ -0,0 +1,216 @@
...
+        contract_nonce = contract.generate_trade_nonce(self.context.agent_address)
+        self.context.shared_state["contract_nonce"] = contract_nonce
+        info_msg = FIPAMessage(
+            message_id=new_message_id,
+            dialogue_reference=("", ""),
+            target=new_target,
+            performative=FIPAMessage.Performative.INFORM,
+            info={
+                "contract": contract.instance.address,
+                "item_ids": contract.item_ids,
```
DavidMinarsch: see above, use `tokens` rather than `items`

packages/fetchai/skills/erc1155_deploy/handlers.py
```diff
@@ -0,0 +1,216 @@
...
+                self.context.agent_name, msg.counterparty[-5:]
+            )
+        )
+        self.context.logger.info(self.context.ledger_apis.apis.get("ethereum"))
+        strategy = cast(Strategy, self.context.strategy)
+        contract = self.context.contracts.erc1155
+        contract_nonce = cast(int, self.context.shared_state.get("contract_nonce"))
+        tx = contract.get_atomic_swap_single_proposal(
+            from_address=self.context.agent_address,
+            to_address=msg.counterparty,
```
DavidMinarsch: I think which  one of the single proposals is used should be configurable.

packages/fetchai/skills/erc1155_deploy/handlers.py
```diff
@@ -0,0 +1,216 @@
...
+        Implement the reaction to a message.
+
+        :param message: the message
+        :return: None
+        """
+        tx_msg_response = cast(TransactionMessage, message)
+        contract = self.context.contracts.erc1155
+        ledger_api = cast(LedgerApi, self.context.ledger_apis.apis.get("ethereum"))
+        tx_digest = ""
+        if tx_msg_response.tx_id == "contract_deploy":
```
DavidMinarsch: See above, these tx_ids could be at least defined in constants, but maybe better in an enum as part of the `contract.py`.

packages/fetchai/skills/erc1155_deploy/skill.yaml
```diff
@@ -0,0 +1,48 @@
...
+        attribute_one:
+          name: country
+          type: str
+          is_required: True
+        attribute_two:
+          name: city
+          type: str
+          is_required: True
+      search_data:
+        country: UK
```
DavidMinarsch: this seems odd as the search data. I'd prefer something relative to the erc1155 ..

tests/test_packages/test_contracts/conftest.py
Totoual: it kind of works but ci generates an error. But as we discussed I ll remove this since it is contract related test.

docs/erc1155-skills.md
Totoual: This is actually something that we add in every guide. That's why it is there :)

docs/erc1155-skills.md
Totoual: This is the default schema and we are using only to find each other.

packages/fetchai/contracts/erc1155/contract.py
```diff
@@ -52,13 +52,15 @@ def __init__(
...
-    def create_item_ids(self, token_type: int, token_ids: List[int]) -> None:
+    def create_item_ids(self, token_type: int, nb_goods: int) -> Dict[int, int]:
         """Populate the item_ids list."""
-        assert self.item_ids == [], "Item ids already created."
-        for token_id in token_ids:
-            self.item_ids.append(Helpers().generate_id(token_type, token_id))
+        assert self.token_ids == [], "Item ids already created."
+        for i in range(nb_goods):
+            print(self.instance.functions.is_token_id_exists(i).call())
+            self.token_ids[Helpers().generate_id(token_type, i)] = 0
```
DavidMinarsch:
```suggestion
        lowest_valid_integer = 0
        token_id = Helpers().generate_id(token_type, lowest_valid_integer)
        for _i in range(nb_goods):
            while self.instance.functions.is_token_id_exists(token_id).call():
                # id already taken
                lowest_valid_integer += 1
                token_id = Helpers().generate_id(token_type, lowest_valid_integer)
             self.token_ids[token_id] = token_type
```

DavidMinarsch: also change `nb_goods` to `nb_tokens` please

packages/fetchai/contracts/erc1155/contracts/erc1155.vy
```diff
@@ -319,6 +320,17 @@ def is_nonce_used(addr: address, nonce: uint256) -> bool:
...
+@public
+@constant
+def is_token_id_exists(token_id: uint256) -> bool:
+    """
+    @notice Checks if the given token_id is already created.
+    @param token_id: uint256 the id of the token.
+    """
+    if self.token_ids[token_id]:
+        return True
+    return False
```
DavidMinarsch:
```suggestion
    return self.token_ids[token_id]
```

DavidMinarsch: Also, in the create call you need to set it: self.token_ids[token_id] = True

packages/fetchai/contracts/erc1155/contract.py
```diff
@@ -553,7 +574,7 @@ def get_hash(
         m_list.append(_nonce.to_bytes(32, "big"))
         return keccak256(b"".join(m_list))
 
-    def generate_id(self, token_id, item_id):
+    def generate_id(self, token_id: int, item_id: int):
```
DavidMinarsch: let's replace item_id with index

DavidMinarsch: `token_id` should be `token_type`

packages/fetchai/skills/erc1155_deploy/strategy.py
```diff
@@ -29,7 +29,7 @@
 DEFAULT_IS_LEDGER_TX = True
 DEFAULT_NFT = 1
 DEFAULT_FT = 2
-DEFAULT_ITEMS_IDS = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
```
DavidMinarsch: Note, token_ids cannot be set anymore, we autogenerate them. So I'd remove the constant and variable.

DavidMinarsch: What can be set is the number of tokens, which needs to be added instead.

docs/erc1155-skills.md
DavidMinarsch: There is no such thing as a default schema. You can define a new one here. just provide different values?

packages/fetchai/skills/erc1155_deploy/handlers.py
Totoual: Do you mean that I have to make one function `get_atomic_swap_proporal(proposal_type: int, ....., signature)`
And on the background call the single or batch ?

aea/helpers/search/models.py
```diff
@@ -83,6 +84,30 @@ def __eq__(self, other) -> bool:
         )
 
 
+class GenericDataModel(DataModel):
+    """Data model for the the erc1155 deploy skill aea."""
+
+    def __init__(self, data_model_attributes: Dict[str, Any], service_name: str):
```
DavidMinarsch: def __init__(self, data_model_name: str, data_model_attributes: Dict[str, Any])

aea/helpers/search/models.py
```diff
@@ -83,6 +84,30 @@ def __eq__(self, other) -> bool:
...
+            ), "Wrong type!! is_required must be bool"
+            self.attributes.append(
+                Attribute(
+                    name=values["name"],  # type: ignore
+                    type=SUPPORTED_TYPES[values["type"]],
+                    is_required=values["is_required"],
+                )
+            )
+
+        super().__init__(service_name, self.attributes)
```
DavidMinarsch: super().__init__(data_model_name, self.attributes)

aea/helpers/search/models.py
DavidMinarsch: just suggest changing this to align it with work I am doing on another branch

PR #879, opened 2020-03-02
[wip] IPFS Consistency check 

### LINKED ISSUE #637: Consistency checks for all configs
Currently, when executing any `aea` cli command we assume the agent project has a consistent structure (i.e. one generated from previous `aea` cli commands of the same `aea` version.

We should ensure that the configs and project are internally consistent on each `aea` cli command usage. However, to avoid having to run extensive checks each time we can utilize a caching strategy involving hashing the packages. Since we are interested in distributing packages via IPFS eventually we could utilize IPFS for this.

marcofavorito:
## Proposed changes

Add consistency check for agent components.

## Fixes

Fixes #637 

## Types of changes

## Checklist


aea/skills/base.py
```diff
@@ -317,6 +317,12 @@ def parse_module(
         )
 
         name_to_class = dict(behaviours_classes)
+        _print_warning_message_for_non_declared_skill_components(
```
DavidMinarsch: note, you should merge develop as this is not in sync with it anymore

aea/cli/common.py
```diff
@@ -83,19 +80,33 @@ class Context(object):
...
-        self.contract_loader = ConfigLoader(
-            "contract-config_schema.json", ContractConfig
-        )
-        self.protocol_loader = ConfigLoader(
-            "protocol-config_schema.json", ProtocolConfig
-        )
         self.cwd = cwd
 
+    @property
+    def agent_loader(self) -> ConfigLoader:
```
marcofavorito: we avoid multiple instantiations of the configuration loaders by having them in a single class `ConfigLoaders` (see below).

aea/cli/common.py
```diff
@@ -83,19 +80,33 @@ class Context(object):
     def __init__(self, cwd: str = "."):
         """Init the context."""
         self.config = dict()  # type: Dict
-        self.agent_loader = ConfigLoader("aea-config_schema.json", AgentConfig)
```
marcofavorito: rewritten as properties, see below

aea/cli/common.py
```diff
@@ -496,3 +507,77 @@ def _find_item_locally(ctx, item_type, item_public_id) -> Path:
...
+    # find all valid files of the package
+    all_files = [
+        x
+        for x in package_directory.glob("**/*")
+        if x.is_file()
+        and not x.match("*__pycache__/*")
+        and not x.match(
+            "*.yaml"
+        )  # TODO this should only match "protocol.yaml", "connection.yaml" etc.
+    ]
```
marcofavorito: this finds all files in a directory, excluding:
- the ones in `__pycache__`
- all the `*.yml` files (including the package config files)

aea/cli/common.py
```diff
@@ -496,3 +507,77 @@ def _find_item_locally(ctx, item_type, item_public_id) -> Path:
...
+
+    for file in all_files:
+        file_hash = hasher.get(str(file))
+        key = str(file.relative_to(package_directory))
+        assert key not in actual_fingerprints  # nosec
+        actual_fingerprints[key] = file_hash
```
marcofavorito: populate the fingerprint dictionary with the found files and their hashes.

aea/cli/common.py
```diff
@@ -496,3 +507,77 @@ def _find_item_locally(ctx, item_type, item_public_id) -> Path:
...
+                "vendor", public_id.author, item_type.to_plural(), public_id.name
+            )
+
+        loader = ConfigLoaders.from_configuration_type(item_type)
+        config_file_name = _get_default_configuration_file_name_from_type(item_type)
+
+        package_configuration = loader.load(
+            (package_directory / config_file_name).open("r")
+        )
+        _check_package_fingerprints(package_configuration, package_directory)
```
marcofavorito: this loop iterates over all the agent packages (including vendor) and runs the check of fingerprint (see above)

aea/cli/create.py
```diff
@@ -64,12 +64,12 @@ def _check_is_parent_folders_are_aea_projects_recursively() -> None:
     return
 
 
-def _setup_package_folder(ctx, item_type_plural):
+def _setup_package_folder(path: Path):
```
marcofavorito: simplify the function

aea/cli/create.py
```diff
@@ -95,6 +95,15 @@ def create(click_context, agent_name):
...
         path.mkdir(exist_ok=False)
 
+        # set up packages directories.
+        _setup_package_folder(Path(agent_name, "protocols"))
+        _setup_package_folder(Path(agent_name, "connections"))
+        _setup_package_folder(Path(agent_name, "skills"))
+
+        # set up a vendor directory
+        Path(agent_name, "vendor").mkdir(exist_ok=False)
+        Path(agent_name, "vendor", "__init__.py").touch(exist_ok=False)
```
marcofavorito: these lines have been moved from below, although not needed now. It will be useful when we will check the fingerprint also in the `aea-config.yaml` file.

aea/cli/run.py
```diff
@@ -258,13 +259,27 @@ def _setup_connection(
...
     help="Install all the dependencies before running the agent.",
 )
+@click.option(
+    "--skip-consistency-check",
+    "skip_check",
+    is_flag=True,
+    required=False,
+    default=False,
+    help="Skip consistency check..",
+)
```
marcofavorito: optionally. for backward-compatibility, `aea run` can skip the checks introduced in this PR by using the option `--skip-consistency-check`

aea/configurations/base.py
```diff
@@ -273,11 +291,20 @@ def __lt__(self, other):
...
     """This class represent a package configuration."""
 
-    def __init__(self, name: str, author: str, version: str):
+    def __init__(
+        self,
+        name: str,
+        author: str,
+        version: str,
+        license: str,
+        fingerprint: Optional[Dict[str, str]] = None,
```
marcofavorito: add fingerprint to the abstract class `PackageConfiguration`

marcofavorito: ... also license, which was missing.

aea/configurations/base.py
```diff
@@ -352,6 +378,7 @@ def from_json(cls, obj: Dict):
             author=cast(str, obj.get("author")),
             version=cast(str, obj.get("version")),
             license=cast(str, obj.get("license")),
+            fingerprint=cast(Dict[str, str], obj.get("fingerprint")),
```
marcofavorito: in this case and in other cases, the `from_json` ignore the `fingerprint` field

aea/configurations/loader.py
```diff
@@ -24,7 +24,7 @@
 import os
```
marcofavorito: the changes in this file were not strictly needed. Happy to revert them if we don't like the result.

tests/test_cli/test_run.py
```diff
@@ -754,7 +754,7 @@ def setup_class(cls):
 
         shutil.copytree(
             Path(CUR_PATH, "data", "exception_skill"),
-            Path(cls.t, cls.agent_name, "skills", "exception"),
+            Path(cls.t, cls.agent_name, "vendor", "fetchai", "skills", "exception"),
```
marcofavorito: In this test, the exception skill was not placed correctly. Since the author is `fetchai` it should have been put in the vendor directory.

tests/test_cli/test_run.py
```diff
@@ -805,7 +805,7 @@ def setup_class(cls):
         os.chdir(Path(cls.t, cls.agent_name))
 
         try:
-            cli.main([*CLI_LOG_OPTION, "run"])
+            cli.main([*CLI_LOG_OPTION, "run", "--skip-consistency-check"])
```
marcofavorito: the flag helped to integrate the feature with the old tests.

aea/configurations/loader.py
```diff
@@ -66,11 +66,11 @@ def __init__(self, schema_filename: str, configuration_type: Type[T]):
...
         """
-        self.schema = json.load(open(os.path.join(_SCHEMAS_DIR, schema_filename)))
+        self._schema = json.load(open(os.path.join(_SCHEMAS_DIR, schema_filename)))
         root_path = "file://{}{}".format(Path(_SCHEMAS_DIR).absolute(), os.path.sep)
-        self.resolver = jsonschema.RefResolver(root_path, self.schema)
-        self.validator = Draft4Validator(self.schema, resolver=self.resolver)
-        self.configuration_type = configuration_type  # type: Type[T]
+        self._resolver = jsonschema.RefResolver(root_path, self._schema)
+        self._validator = Draft4Validator(self._schema, resolver=self._resolver)
+        self._configuration_type = configuration_type  # type: Type[T]
```
DavidMinarsch: Some of them should be exposed as properties I believe

aea/cli/config.py
```diff
@@ -184,7 +184,7 @@ def get(ctx: Context, json_path: List[str]):
     configuration_file_path = cast(str, ctx.config.get("configuration_file_path"))
 
     configuration_object = yaml.safe_load(open(configuration_file_path))
-    config_loader.validator.validate(instance=configuration_object)
+    config_loader._validator.validate(instance=configuration_object)
```
DavidMinarsch: Please expose `validator` as a property of `config_loader`.

aea/cli/common.py
```diff
@@ -496,3 +507,77 @@ def _find_item_locally(ctx, item_type, item_public_id) -> Path:
...
+                lambda x: (x, ConfigurationType.CONNECTION),
+                ctx.agent_config.connections,
+            ),
+            *map(lambda x: (x, ConfigurationType.SKILL), ctx.agent_config.skills),
+            *map(lambda x: (x, ConfigurationType.CONTRACT), ctx.agent_config.contracts),
+        ]
+    )  # type: Dict[PublicId, ConfigurationType]
+
+    for public_id, item_type in packages_public_ids_to_types.items():
+        if public_id.author == ctx.agent_config.public_id.author:
```
DavidMinarsch: It is worth checking, but I think if one adds a package from the registry it always ends up in vendors, even if it is of the same author. This makes sense as vendors contains all finished packages even self authored ones.

aea/cli/run.py
```diff
@@ -258,13 +259,27 @@ def _setup_connection(
...
-    click_context, connection_ids: List[PublicId], env_file: str, is_install_deps: bool
+    click_context,
+    connection_ids: List[PublicId],
+    env_file: str,
+    is_install_deps: bool,
+    skip_check: bool,
 ):
     """Run the agent."""
     ctx = cast(Context, click_context.obj)
     try_to_load_agent_config(ctx)
```
DavidMinarsch: I'd prefer a solution where `try_to_load_agent_config` and `validate_config_consistency` are by default run on each `aea` subcommand. Then, also apply the `skip_consistency_check` at the `aea` level so it can be skipped for each run.

aea/cli/scaffold.py
```diff
@@ -129,6 +130,7 @@ def _scaffold_item(ctx: Context, item_type, item_name):
         )
         config = loader.load(config_filepath.open())
         config.name = item_name
+        config.fingerprint = _compute_fingerprint(config_filepath.parent)
```
DavidMinarsch: Let's precompute it for scaffold packages.

aea/cli/fingerprint.py
```diff
@@ -48,10 +41,10 @@
 
 
 @click.group()
-@pass_ctx
-def fingerprint(ctx: Context):
+@click.pass_context
+@check_aea_project
```
DavidMinarsch: Ok we cannot actually run it here, otherwise you won't be able to fingerprint a resource.

Basically, the flow would be:
1. scaffold or develop your package
2. try to run aea > it points out that the fingerprint of these packages is no longer correct
3. you update the fingerprint for the packages you developed with these commands.
4. you successfully run.

aea/cli/common.py
```diff
@@ -496,3 +508,117 @@ def _find_item_locally(ctx, item_type, item_public_id) -> Path:
...
+
+def _check_package_fingerprints(
+    package_configuration: PackageConfiguration, package_directory: Path
+):
+    """Given a package configuration and a package directory, check whether
+    the hash of the files in the package directory match the one in the package configurations."""
+    expected_fingerprints = package_configuration.fingerprint
+    actual_fingerprints = _compute_fingerprint(package_directory)
+
+    if expected_fingerprints != actual_fingerprints:
```
DavidMinarsch: Can we have two types of warnings? I'd like it to say:
- in the case of non-matching fingerprints for vendorized projects: "Fingerprints for package {} do not match:\nExpected: {}\nActual:{}\nVendorized projects should not be tampered with, please revert any changes."
- in the case of non-matching fingerprints for root projects: "Fingerprints for package {} do not match:\nExpected: {}\nActual:{}\nPlease fingerprint the package in {} before continuing."

aea/cli/fingerprint.py
marcofavorito: true... I'll fix that.

aea/cli/common.py
marcofavorito: yes, good idea!

aea/cli/scaffold.py
marcofavorito: yes!

aea/cli/fingerprint.py
marcofavorito: that should be documented... I'll open an issue.

aea/cli/common.py
marcofavorito: see if the provided solution works. We give precedence to custom packages.

Issue #739, opened 2020-02-09
Build a HTTP connection


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
Currently, we do not have a generic http connection package.

**Describe the solution you'd like**
Build one, which does the following:
- consumes a swagger 
- https://github.com/swagger-api/swagger-codegen
- creates messages from requests
- blocks until the response is returned (o/w times out)

DavidMinarsch:
- can be configured for web hooks

PR #895, opened 2020-03-04
Add a http connection package

### LINKED ISSUE #592: Provide an example of the URI usage in an envelope
Currently, non of the agent examples use the URI field.

Provide an example of the URI usage.

DavidMinarsch:
## Proposed changes

Reopened #860 , fixes #592, #739

## Types of changes

## Checklist

## Further comments

-


Pipfile
```diff
@@ -18,6 +18,7 @@ flake8 = "==3.7.9"
 flake8-bugbear = "==20.1.4"
 flake8-docstrings = "==1.5.0"
 flake8-import-order = "==0.18.1"
+flex = "==6.14.1"
```
DavidMinarsch: @marcofavorito not a fan of having to introduce this, but we need it for the swagger/openapi spec verification

tests/test_packages/test_connections/test_http/test_http.py
```diff
@@ -0,0 +1,347 @@
...
+
+        envelope = await self.http_connection.receive()
+        assert (
+            envelope is not None
+            and envelope.to == self.address
+            and len(self.http_connection.channel.timed_out_request_ids) == 1
+        )
+
+    # @pytest.mark.asyncio
+    # async def test_get_202(self):
```
DavidMinarsch: @marcofavorito struggling with this one. The challenge here is to get the agent_processing to happen after the envelope has been enqueued and before the request times out...

Pipfile
marcofavorito: What about these projects?
-  https://github.com/p1c2u/openapi-core
- https://github.com/p1c2u/openapi-spec-validator

Also, we could check how `connexion` validates specs files and requests/responses.
- https://github.com/zalando/connexion/blob/master/setup.py#L23-L47

tests/data/petstore_sim.yaml
```diff
@@ -0,0 +1,32 @@
+swagger: '2.0'
```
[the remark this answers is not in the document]
marcofavorito: The ones I linked seem to work with OpenAPI v3.

packages/fetchai/connections/http/connection.py
```diff
@@ -0,0 +1,500 @@
...
+
+        if envelope.to in self.timed_out_request_ids:
+            self.timed_out_request_ids.remove(envelope.to)
+            logger.warning(
+                "Dropping envelope for request id {} which has timed out.".format(
+                    envelope.to
+                )
+            )
+        else:
+            self.dispatch_ready_envelopes.update({envelope.to: envelope})
```
marcofavorito: This should be executed only when the agent handler produces a response to a previous request. However, it seems we can dispatch a message even if it is not associated to a request id (although it will never be consumed by the `get_response` method).

So I propose to also check that there is a pending request to be answered.

tests/test_packages/test_connections/test_http/test_http.py
```diff
@@ -0,0 +1,347 @@
...
+        client_id = "to_key"
+        envelope = Envelope(
+            to=client_id,
+            sender="from_key",
+            protocol_id=self.protocol_id,
+            message=b"some message",
+        )
+        await self.http_connection.send(envelope)
+        assert (
+            self.http_connection.channel.dispatch_ready_envelopes.get(client_id)
```
marcofavorito: If the above comment is relevant, this is not ideal. We should put a "dummy request-id" such that we "fool" the `send` method making him believe that there's a pending request.
https://github.com/fetchai/agents-aea/pull/895/files#r388648114

tests/test_packages/test_connections/test_http/test_http.py
```diff
@@ -0,0 +1,347 @@
...
+        )
+
+    @classmethod
+    def teardown_class(cls):
+        """Teardown the class."""
+        value = cls.loop.run_until_complete(cls.http_connection.disconnect())
+        assert value is None
+
+
+def test_gym_from_config():
```
marcofavorito: fix test name

packages/fetchai/connections/http/connection.py
```diff
@@ -0,0 +1,500 @@
...
+            """Deal with header only requests."""
+            self.send_response(SUCCESS)
+            self.send_header("Content-Type", "text/plain; charset=utf-8")
+            self.end_headers()
+
+        def do_GET(self):
+            """Respond to a GET request."""
+            request = self.build_request("get")
+
+            response = self.channel.loop.run_until_complete(
```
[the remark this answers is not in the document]
marcofavorito: because in the tests you are not using the connection via the `Multiplexer`, so `self.channel.loop` is not running.
If it is not running, you can use `run_until_complete`, but if it is already running (which is the case when you run the connection with the Multiplexer) you'll get an error!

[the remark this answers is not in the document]
marcofavorito: https://github.com/fetchai/agents-aea/pull/901

tests/test_packages/test_connections/test_http/test_http.py
```diff
@@ -0,0 +1,347 @@
...
+    #         response = conn.getresponse()
+    #         return response.status, response.read()
+
+    #     async def client_thread() -> Tuple[int, bytes]:
+    #         executor = concurrent.futures.ThreadPoolExecutor(max_workers=3)
+    #         loop = asyncio.get_event_loop()
+    #         result = await loop.run_in_executor(executor, request_response_cycle)
+    #         return result
+
+    #     response_code, response_body = await client_thread()
```
marcofavorito: this line is not correct. That won't execute `agent_processing()` because it will `await` the result from the corouting.

What you can do is to "schedule" the task and run it concurrently with the `agent_processing`. thanks to the delay you wrote, the `client_thread` will send the HTTP connection before `agent_processing`reads it.

marcofavorito: see here one way to do it:
https://github.com/fetchai/agents-aea/pull/901/files#diff-9424a9106678c3c13e4c9501f9f458b7R243-R250

tests/test_packages/test_connections/test_http/test_http.py
```diff
@@ -0,0 +1,347 @@
...
+    #                 sender=envelope.to,
+    #                 protocol_id=envelope.protocol_id,
+    #                 context=envelope.context,
+    #                 message=json.dumps({"status_code": 200, "message": "Response"}).encode(),
+    #             )
+    #         await self.http_connection.send(response_envelope)
+
+    #     agent_processing()
+
+    #     assert response_code == 200 and b"Response"
```
marcofavorito: this should have `response_body == b"Response"`

however, in the other branch, I didn't manage to make it return the bytes, I have to investigate more. Feel free to take the token.

tests/test_packages/test_connections/test_http/test_http.py
DavidMinarsch: Thanks :)

Pipfile
DavidMinarsch: Used that now, thanks for the suggestion.

examples/protocol_specification_ex/http.yaml
```diff
@@ -0,0 +1,20 @@
...
+version: 0.1.0
+license: Apache-2.0
+description: 'A protocol for HTTP requests and responses.'
+speech_acts:
+  request:
+    method: pt:str
+    url: pt:str
+    version: pt:str
+    headers: pt:str
+    bodyy: pt:bytes
```
DavidMinarsch: Due to bugs and design errors in the generator (https://github.com/fetchai/agents-aea/issues/905) we cannot chose `body` as a content and we cannot make it optional. Once this is fixed, we can change to `body: pt:optional[pt:bytes]`.

examples/protocol_specification_ex/http.yaml
```diff
@@ -0,0 +1,20 @@
+---
+name: http
+author: fetchai
+version: 0.1.0
+license: Apache-2.0
+description: 'A protocol for HTTP requests and responses.'
+speech_acts:
```
DavidMinarsch: Following https://developer.mozilla.org/en-US/docs/Web/HTTP/Messages for design of the content fields

packages/fetchai/connections/http/connection.py
```diff
@@ -0,0 +1,565 @@
...
+            path=path,
+        )
+        request = Request(
+            full_url_pattern=url,
+            method=method,
+            parameters=parameters,
+            body=body,
+            mimetype=mimetype,
+        )
+        request.id = uuid4().hex
```
DavidMinarsch: Doing this via a property was preferred after trying several attempts of satisfying mypy.

packages/fetchai/connections/http/connection.py
DavidMinarsch: Done now via `self.pending_request_ids`.

tests/test_packages/test_connections/test_http/test_http.py
DavidMinarsch: done

examples/protocol_specification_ex/http.yaml
cyenyxe: Note also the typo `bodyy` instead of `body`.

PR #902, opened 2020-03-06
Wrong assigning registry folder fixed.

panasevychol:
## Proposed changes

Wrong assigning registry folder fixed.

## Types of changes

## Checklist

### DIFF
--- a/aea/cli/search.py
+++ b/aea/cli/search.py
@@ -71,7 +71,7 @@ def search(ctx: Context, registry):
             # agent_config = ctx.agent_loader.load(fp)
             registry_directory = ctx.agent_config.registry_path
         except Exception:
-            registry_directory = os.path.join(ctx.cwd, DEFAULT_REGISTRY_PATH)
+            registry_directory = os.path.join("..", DEFAULT_REGISTRY_PATH)
 
         ctx.set_config("registry_directory", registry_directory)
         logger.debug("Using registry {}".format(registry_directory))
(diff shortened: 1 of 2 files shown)


aea/cli/search.py
```diff
@@ -71,7 +71,7 @@ def search(ctx: Context, registry):
             # agent_config = ctx.agent_loader.load(fp)
             registry_directory = ctx.agent_config.registry_path
         except Exception:
-            registry_directory = os.path.join(ctx.cwd, DEFAULT_REGISTRY_PATH)
+            registry_directory = os.path.join("..", DEFAULT_REGISTRY_PATH)
```
DavidMinarsch: The original functionality is correct.

tests/test_cli/test_search.py
```diff
@@ -333,13 +333,6 @@ def test_correct_output(self,):
             self.result.output == 'Searching for ""...\n'
             "Skills found:\n\n"
             "------------------------------\n"
-            "Public ID: fetchai/echo:0.1.0\n"
```
DavidMinarsch: I am not sure why this should be removed.

DavidMinarsch:
Rejecting as this is not introducing fixes.

aea/cli/search.py
panasevychol: Is that? Case: loading agent config fails from agent's folder, it leads us to `./packages` path inside the agent's directory which is not existing. Exactly the case we had on Friday's presentation.

PR #908, opened 2020-03-08
AEA-266 Prototype of Agent Builder

marcofavorito:
## Proposed changes

Add some refactoring ideas.

## Types of changes

## Checklist


aea/aea_builder.py
```diff
@@ -0,0 +1,164 @@
...
+
+PackageId = Tuple[AgentComponentType, PublicId]
+
+
+class _Package:
+
+    def __init__(self, package_type, package_configuration):
+        """"""
+
+        self._package_type = package_type  # one of protocol, skill, connection etc.
```
DavidMinarsch: We should have an enum somewhere for the package types so we can use it throughout rather than the magic strings we use now.

aea/aea_builder.py
```diff
@@ -0,0 +1,164 @@
...
+        self._package_config = package_configuration
+        self._modules = {}
+
+    @classmethod
+    def load(self, package_type) -> "_Package":
+        """Load package from dir"""
+        # read config file
+        # load modules
```
DavidMinarsch: where does this load to?..i.e. what does `load` entail here.

aea/aea_builder.py
```diff
@@ -0,0 +1,164 @@
...
+from aea.crypto.wallet import Wallet
+from aea.identity.base import Identity
+from aea.mail.base import Address
+from aea.protocols.base import Protocol
+from aea.registries.base import Resources
+
+PackageId = Tuple[AgentComponentType, PublicId]
+
+
+class _Package:
```
DavidMinarsch: This class should live in the base folder imo. Something like `aea/common.py`?

aea/aea_builder.py
```diff
@@ -0,0 +1,164 @@
...
+        """Load package from dir"""
+        # read config file
+        # load modules
+
+
+class AEABuilder:
+    """This class helps to build an AEA."""
+
+    def __init__(self):
+        self._wallet = Wallet({})  # TODO make Wallet mutable (add/remove_private_key)
```
DavidMinarsch: Would we want to allow for removal? I think from a safety perspective, you'd not want to allow for removal.

DavidMinarsch: As part of this PR I'd not touch the primitives (AEA, Wallet, LedgerApis, Resources, Identity) unless where necessary to make the PR work.

aea/aea_builder.py
```diff
@@ -0,0 +1,164 @@
...
+        self._ledger_apis = LedgerApis({}, "")  # TODO make it mutable (add/remove api)
+
+        self._resources = Resources()
+        self._connections = []
+
+        # identity
+        self._name = None
```
DavidMinarsch: I think this should be subsumed into identity as well

aea/aea_builder.py
```diff
@@ -0,0 +1,164 @@
...
+        self._addresses = {}  # type: Dict[str, Address]
+        self._default_key = None  # set by the user, or instantiate a default one.
+
+        # from package id to packages it depends on.
+        # we should detect loops at "add" time, by inspecting the component config file.
+        self._dependency_graph = {}  # type: Dict[PackageId, Set[_Package]]
```
DavidMinarsch: Are we keeping track of versions here?

aea/aea_builder.py
```diff
@@ -0,0 +1,164 @@
...
+
+    def remove_address(self, identifier: str):
+        self._addresses.pop(identifier, None)
+
+    def add_package(self, package_type, directory):
+        """Load a package, given its type and the directory."""
+        # load config, check fingerprint, handle errors
+        # from config, check package dependencies
+        # visit dependency graph, load modules in sys.modules when going backward
+        # try to load the package (e.g. the same we do for Skill.from_dir, but for any type
```
DavidMinarsch: we already also have `Package.from_dir` just fyi..so both will be made generic

aea/aea_builder.py
DavidMinarsch: ok, saw below now

aea/aea_builder.py
```diff
@@ -0,0 +1,164 @@
...
+
+    # def add_connection(self, path):
+    # def remove_connection(self):
+    # def add_skill(self, path):
+    # def remove_skill(self):
+
+    def build(self) -> AEA:
+        """Get the AEA."""
+        aea = AEA(
+            Identity(self._name, addresses=self._addresses),
```
DavidMinarsch: would instantiate this as part of builder. Note. addresses come straight out of the wallet

aea/aea_builder.py
```diff
@@ -0,0 +1,164 @@
...
+            is_programmatic=True,
+            max_reactions=20,
+        )
+        return aea
+
+    def dump_config(self) -> AgentConfig:
+        """Dump configurations"""
+
+    def dump(self, directory):
+        """Dump agent project."""
```
DavidMinarsch: Not sure what this would entail

aea/aea_builder.py
```diff
@@ -0,0 +1,164 @@
...
+        # identity
+        self._name = None
+        self._addresses = {}  # type: Dict[str, Address]
+        self._default_key = None  # set by the user, or instantiate a default one.
+
+        # from package id to packages it depends on.
+        # we should detect loops at "add" time, by inspecting the component config file.
+        self._dependency_graph = {}  # type: Dict[PackageId, Set[_Package]]
+
+        # add default protocol
```
DavidMinarsch: I think the builder should be adding the default packages during initialisation

aea/aea_builder.py
```diff
@@ -0,0 +1,164 @@
...
+        return aea
+
+    def dump_config(self) -> AgentConfig:
+        """Dump configurations"""
+
+    def dump(self, directory):
+        """Dump agent project."""
+
+
+class AEAProject:
```
DavidMinarsch: Not sure about the name.. otherwise a useful class.

aea/aea_builder.py
DavidMinarsch: Note, the primitives can be built from the aea config directly. I would keep them immutable.

aea/aea_builder.py
marcofavorito: we do have an enum. But it is called `ConfigurationType`, whereas it should be called `ItemType` in my opinion/

aea/aea_builder.py
marcofavorito: yes, I think we can initialize those object only when `build` is called, without requiring the mutability of those objects.

aea/aea_builder.py
marcofavorito: yes, it will be used to build `Identity` object, but we don't instantiate it until `build` is called

aea/aea_builder.py
marcofavorito: yes. `PackageId` is defined as `Tuple[ItemType, PublicId]` (at the top of the module)

aea/aea_builder.py
marcofavorito: don't know, let's see

aea/aea_builder.py
DavidMinarsch: Oh I see and `PublicId` now references versions :)

aea/configurations/base.py
```diff
@@ -304,12 +336,102 @@ def __lt__(self, other):
         return str(self) < str(other)
 
 
+class PackageId:
+    """A component identifier."""
```
DavidMinarsch: "A package identifier" is clearer?

aea/configurations/base.py
```diff
@@ -52,13 +53,15 @@
 Dependencies = Dict[str, Dependency]
 
 
+# TODO rename this to "PackageType"
```
DavidMinarsch: Yes!

aea/configurations/base.py
```diff
@@ -333,10 +455,89 @@ def public_id(self) -> PublicId:
...
+        Check that the fingerprint are correct against a directory path.
+
+        :raises ValueError if:
+            - the argument is not a valid package directory
+            - the fingerprint do not match.
+        """
+        if not directory.exists() or not directory.is_dir():
+            raise ValueError("Directory {} is not valid.".format(directory))
+
+        # TODO take from release/v0.3 branch.
```
DavidMinarsch: Once this is in develop, merge develop into release v0.3 branch

aea/configurations/base.py
```diff
@@ -643,6 +881,23 @@ def __init__(
...
+            for public_id in self.protocols
+        )
+        connections = set(
+            ComponentId(ComponentType.CONNECTION, public_id)
+            for public_id in self.connections
+        )
+        skills = set(
+            ComponentId(ComponentType.CONNECTION, public_id)
+            for public_id in self.skills
+        )
```
DavidMinarsch: Contracts?

DavidMinarsch: Even though this is against develop, you mention contracts above, so maybe just add a todo

aea/configurations/components.py
```diff
@@ -0,0 +1,178 @@
...
+        return connection
+
+    @classmethod
+    def load_from_directory(
+        cls, component_type: ComponentType, directory: Path
+    ) -> "Component":
+        """Load a component from the directory."""
+        configuration = ComponentConfiguration.load(component_type, directory)
+        if component_type == ComponentType.CONNECTION:
+            # the connection case is an exception,
```
DavidMinarsch: good point. I think we did that because of the gym connection. Should be solvable.

aea/connections/base.py
```diff
@@ -59,6 +66,9 @@ def __init__(
         :param restricted_to_protocols: the set of protocols ids of the only supported protocols for this connection.
         :param excluded_protocols: the set of protocols ids that we want to exclude for this connection.
         """
+        # TODO for backward compatibility, the configuration attribute will be set after initialization
```
DavidMinarsch: We should break backward compatibility for v0.3?

aea/connections/base.py
```diff
@@ -116,6 +126,11 @@ def loop(self, loop: AbstractEventLoop) -> None:
         ), "Cannot set the loop while it is running."
         self._loop = loop
 
+    @property
```
DavidMinarsch: Isn't that a property of the Component base class already?

aea/configurations/components.py
```diff
@@ -0,0 +1,178 @@
...
+    def configuration(self) -> ComponentConfiguration:
+        """Get the component configuration."""
+        return self._configuration
+
+    @property
+    def directory(self) -> Optional[Path]:
+        """Get the directory, or None if the component is not associated with any directory."""
+        return self._directory
+
+    def setup(self, *args, **kwargs):
```
DavidMinarsch: What about teardown?

aea/protocols/base.py
```diff
@@ -225,15 +226,18 @@ def decode(self, obj: bytes) -> Message:
...
     """
     This class implements a specifications for a protocol.
 
     It includes a serializer to encode/decode a message.
     """
 
     def __init__(
-        self, protocol_id: ProtocolId, serializer: Serializer, config: ProtocolConfig
+        self,
+        protocol_id: Optional[ProtocolId] = None,
```
DavidMinarsch: why all optional?

aea/protocols/base.py
```diff
@@ -292,3 +299,23 @@ def from_dir(cls, directory: str) -> "Protocol":
         )
         protocol = Protocol(protocol_id, serializer_class(), protocol_config)
         return protocol
+
+    def setup(self, *args, **kwargs):
```
DavidMinarsch: maybe, load would be a better name? Also avoids confusion with the setup/teardown methods in the skill modules

aea/skills/base.py
```diff
@@ -60,13 +61,13 @@
 class SkillContext:
     """This class implements the context of a skill."""
 
-    def __init__(self, agent_context: AgentContext):
+    def __init__(self, agent_context: Optional[AgentContext] = None):
```
DavidMinarsch: why?

aea/skills/base.py
```diff
@@ -60,13 +61,13 @@
...
         """
         Initialize a skill context.
 
         :param agent_context: the agent's context
         """
-        self._agent_context = agent_context
+        self._agent_context = cast(AgentContext, agent_context)
```
DavidMinarsch: makes no sense to me that first you declare it optional and then you cast it to non-optional...

aea/skills/base.py
```diff
@@ -530,30 +531,52 @@ def parse_module(
...
 
     def __init__(
         self,
-        config: SkillConfig,
-        skill_context: SkillContext,
-        handlers: Optional[Dict[str, Handler]],
-        behaviours: Optional[Dict[str, Behaviour]],
-        models: Optional[Dict[str, Model]],
+        configuration: SkillConfig,
+        skill_context: Optional[SkillContext] = None,
```
DavidMinarsch: why optional?

examples/programmatic.py
```diff
@@ -0,0 +1,44 @@
...
+
+"""Example of programmatic initialization of an AEA using the builder."""
+
+import logging
+
+from aea.aea_builder import AEABuilder
+from aea.crypto.fetchai import FetchAICrypto
+
+
+if __name__ == "__main__":
```
DavidMinarsch: awesome!

aea/aea_builder.py
```diff
@@ -0,0 +1,418 @@
...
+        for skill in self._resources.get_all_skills():
+            logger_name = "aea.{}.skills.{}.{}".format(
+                context.agent_name, skill.configuration.author, skill.configuration.name
+            )
+            skill.skill_context._agent_context = context
+            skill.skill_context._logger = logging.getLogger(logger_name)
+
+
+#
+# class AEAProject:
```
DavidMinarsch: ?

aea/aea_builder.py
```diff
@@ -0,0 +1,418 @@
...
+        # add stub connection
+        self.add_connection(Path(AEA_DIR, "connections", "stub"))
+        # add error skill
+        self.add_skill(Path(AEA_DIR, "skills", "error"))
+
+    def set_name(self, name: str) -> "AEABuilder":
+        self._name = name
+        return self
+
+    def add_address(self, identifier: str, address: Address) -> "AEABuilder":
```
DavidMinarsch: I think the `add_address` should be replaced with an `add_private_key`. As all addresses are derived from a private key (via the wallet)

aea/aea_builder.py
```diff
@@ -0,0 +1,418 @@
...
+        self._addresses[identifier] = address
+        return self
+
+    def remove_address(self, identifier: str) -> "AEABuilder":
```
DavidMinarsch: See above, should be `remove_private_key`

aea/aea_builder.py
DavidMinarsch: And it should consume a file_path to the private key file

DavidMinarsch: Then, when I add encryption of the private key file, we can add a password without further changes to the builder

aea/aea_builder.py
```diff
@@ -0,0 +1,418 @@
...
+
+class AEABuilder:
+    """This class helps to build an AEA."""
+
+    def __init__(self):
+        self._resources = Resources()
+
+        # identity
+        self._name = None
+        self._addresses = {}  # type: Dict[str, Address]
```
DavidMinarsch: see below

aea/aea_builder.py
```diff
@@ -0,0 +1,418 @@
...
+
+    def _get_import_order(self) -> List[Tuple[str, types.ModuleType]]:
+        """
+        Get the import order.
+
+        At the moment:
+        - a protocol has no dependencies.
+        - a connection can depend on protocols.
+        - a skill can depend on protocols.
+        - a contract ...
```
DavidMinarsch: a contract can depend on skills

aea/configurations/base.py
```diff
@@ -333,10 +455,89 @@ def public_id(self) -> PublicId:
...
+        self.check_fingerprint(directory)
+        self.check_aea_version()
+
+    def check_fingerprint(self, directory: Path) -> None:
+        """
+        Check that the fingerprint are correct against a directory path.
+
+        :raises ValueError if:
+            - the argument is not a valid package directory
+            - the fingerprint do not match.
```
marcofavorito: fix typo

aea/configurations/base.py
```diff
@@ -367,6 +568,19 @@ def __init__(
...
+    def component_type(self) -> ComponentType:
+        """Get the component type."""
+        return ComponentType.CONNECTION
+
+    @property
+    def package_dependencies(self) -> Set[ComponentId]:
+        """Get the connection dependencies."""
+        return set(
+            ComponentId(ComponentType.PROTOCOL, protocol_id)
+            for protocol_id in self.protocols
```
marcofavorito: in our model, connections can only have protocols as dependencies. In the future, they might depend on some "skills"?
Think about packages that only contain utilities. Instead of deploying them to PyPI, they can be deployed to the registry and used by other agents as local packages.

aea/configurations/base.py
```diff
@@ -643,6 +881,23 @@ def __init__(
...
+        protocols = set(
+            ComponentId(ComponentType.PROTOCOL, public_id)
+            for public_id in self.protocols
+        )
+        connections = set(
+            ComponentId(ComponentType.CONNECTION, public_id)
+            for public_id in self.connections
+        )
+        skills = set(
+            ComponentId(ComponentType.CONNECTION, public_id)
```
marcofavorito: fix this

aea/configurations/components.py
```diff
@@ -0,0 +1,178 @@
...
+    def is_vendor(self) -> bool:
+        """Get whether the component is vendorized or not."""
+        return self._is_vendor
+
+    @property
+    def prefix_import_path(self):
+        """Get the prefix import path for this component."""
+        if self.is_vendor:
+            return "packages.{}.{}.{}".format(
+                self.public_id.author,
```
marcofavorito: should we forbid some author strings that could make some conflicts with the import paths below (non-vendor case)? E.g. `skills`, `protocols`...

aea/configurations/components.py
marcofavorito: we should discuss. `setup` is used below in `load_from_directory`

aea/configurations/components.py
marcofavorito: I was thinking of using the same paradigm we used for `Protocol` and `Serialization`.
For connection, we could have something similar but with `Channel`.

aea/configurations/base.py
DavidMinarsch: Open issue for this?

aea/configurations/components.py
DavidMinarsch: Yes definitely! :)

DavidMinarsch: Open issue

aea/configurations/components.py
DavidMinarsch: Let's use `setup` and ask developer to implement it. Then call it here.

aea/skills/base.py
```diff
@@ -609,6 +632,38 @@ def from_dir(cls, directory: str, agent_context: AgentContext) -> "Skill":
...
         return skill
 
+    def setup(self, *args, **kwargs):
+        """
+        Set the component up.
+
+        In the case of a skill, we load the 'serialization.py' module
+        to instantiate an instance of the Serializer.
+        """
+        skill_context = SkillContext(None)
```
marcofavorito: this is a bit ugly...

aea/connections/base.py
DavidMinarsch: cleaner and backwards compatibility is broken in multiple instances anyway

aea/protocols/base.py
DavidMinarsch: ok let's break backwards compatibility here too and just initialize as `None` below (with type declaration)

aea/skills/base.py
DavidMinarsch: brake backwards compatibility, initialize as none ...

aea/skills/base.py
DavidMinarsch: ok

aea/skills/base.py
DavidMinarsch: remove optional fields from constructor

aea/aea_builder.py
```diff
@@ -0,0 +1,418 @@
...
+
+        return: True if all the dependencies are covered, False otherwise.
+        """
+        not_supported_packages = component_configuration.package_dependencies.difference(
+            self.all_dependencies
+        )
+        return len(not_supported_packages) == 0
+
+    @contextmanager
+    def load_dependencies(self):
```
DavidMinarsch: only add the highest version for two otherwise identical component ids

DavidMinarsch: provide a useful error message

aea/aea_builder.py
```diff
@@ -0,0 +1,418 @@
...
+        return self
+
+    def build(self) -> AEA:
+        """Get the AEA."""
+        connections = list(self._package_dependency_graph.connections.values())
+        aea = AEA(
+            Identity(self._name, addresses=self._addresses),
+            connections,
+            Wallet({}),
+            LedgerApis({}, ""),
```
DavidMinarsch: need way to add ledger api config.

aea/aea_builder.py
```diff
@@ -0,0 +1,418 @@
...
+        """Dump agent project."""
+
+    def _set_context_to_all_skills(self, context: AgentContext):
+        """Set a skill context to all skills"""
+        for skill in self._resources.get_all_skills():
+            logger_name = "aea.{}.skills.{}.{}".format(
+                context.agent_name, skill.configuration.author, skill.configuration.name
+            )
+            skill.skill_context._agent_context = context
+            skill.skill_context._logger = logging.getLogger(logger_name)
```
DavidMinarsch: from_aea_config

aea/aea_builder.py
DavidMinarsch: for building a project

aea/configurations/base.py
marcofavorito: https://github.com/fetchai/agents-aea/issues/960

aea/configurations/components.py
marcofavorito: https://github.com/fetchai/agents-aea/issues/961

aea/skills/base.py
marcofavorito: solved by removing `AgentContext` from the `SkillContext.__init__` arguments.

aea/configurations/base.py
marcofavorito: #934

aea/configurations/components.py
marcofavorito: replace with `load`

aea/protocols/base.py
marcofavorito: yes

aea/aea.py
```diff
@@ -137,8 +136,6 @@ def setup(self) -> None:
 
         :return: None
         """
-        if not self.is_programmatic:
-            self.resources.load(self.context)
```
marcofavorito: we don't have anymore "programmatic" usage, it is always programmatic, and require that resources are already populated.

aea/aea_builder.py
```diff
@@ -0,0 +1,736 @@
...
+            component_path = cls._find_component_directory_from_component_id(
+                aea_project_path, component_id
+            )
+            builder.add_component(component_id.component_type, component_path)
+
+        return builder
+
+
+def _verify_or_create_private_keys(aea_project_path: Path) -> None:
+    """Verify or create private keys."""
```
marcofavorito: This is copied from `aea.cli.common._verify_or_create_private_keys`, with some variation. It is used in `AEABuilder.from_aea_config` in order to emulate the behaviour of `aea run`.  In particular, the CLI command instantiates the private key if it is not in the project path. Of course it can be further improved and refactored.

aea/aea_builder.py
```diff
@@ -0,0 +1,736 @@
...
+            _try_validate_ethereum_private_key_path(
+                str(aea_project_path / ethereum_private_key_path), exit_on_error=False
+            )
+        except FileNotFoundError:  # pragma: no cover
+            logger.error(
+                "File {} for private key {} not found.".format(
+                    repr(ethereum_private_key_path), ETHEREUM,
+                )
+            )
+            raise
```
marcofavorito: This function is copied from aea.cli.common._verify_or_create_private_keys, with some variation. It is used in AEABuilder.from_aea_config in order to emulate the behaviour of aea run. In particular, the CLI command instantiates the private key if it is not in the project path. Of course it can be further improved and refactored.

aea/aea_builder.py
marcofavorito: OK, I missed this comment for some reason.
The current approach lets the user decide if the `AEABuilder` should be initialized with the default packages or not. Look at `__init__` method.

aea/aea_builder.py
```diff
@@ -0,0 +1,736 @@
...
+
+class AEABuilder:
+    """
+    This class helps to build an AEA.
+
+    It follows the fluent interface. Every method of the builder
+    returns the instance of the builder itself.
+    """
+
+    def __init__(self, with_default_packages: bool = True):
```
marcofavorito: if `with_default_packages` is true, the default packages are added in the constructor. Since this is the most common use case the default value is `True`.

tests/test_packages/test_connections/test_local/test_search_services.py
```diff
@@ -399,29 +366,6 @@ def teardown_class(cls):
         cls.node.stop()
 
 
-class TestOEFConnectionFromJson:
```
DavidMinarsch: not supported anymore?

tests/test_packages/test_connections/test_oef/test_communication.py
```diff
@@ -847,13 +821,6 @@ def test_connection(self):
     #     with pytest.raises(ConnectionError):
     #         await con.connect()
 
-    def test_oef_from_config(self):
```
DavidMinarsch: not supported anymore?

tests/test_packages/test_connections/test_tcp/test_communication.py
```diff
@@ -254,14 +214,6 @@ class TestTCPClientConnection:
         await tcp_client.disconnect()
         await tcp_server.disconnect()
 
-    @pytest.mark.asyncio
-    async def test_from_config(self):
```
DavidMinarsch: not supported anymore?

tests/test_protocols/test_base.py
```diff
@@ -103,15 +102,6 @@ def test_body_setter(self):
         self.message2.body = m_dict
         assert "Hello" in self.message2.body.keys()
 
-    def test_protocols_config(self):
```
DavidMinarsch: not supported anymore?

tests/test_packages/test_connections/test_local/test_search_services.py
marcofavorito: No, because `Component.load_from_directory(ComponentType.CONNECTION, directory)` can be used.

aea/aea_builder.py
```diff
@@ -0,0 +1,746 @@
...
+from aea.protocols.base import Protocol
+from aea.registries.resources import Resources
+from aea.skills.base import Skill
+
+PathLike = Union[os.PathLike, Path, str]
+
+logger = logging.getLogger(__name__)
+
+
+class _DependenciesManager:
```
marcofavorito: this is for internal usage only, hence the underscore

aea/aea_builder.py
marcofavorito: > provide a useful error message
We load only the highest version (see the property `dependencies_highest_version`), but no warning message yet (TODO)

aea/aea_builder.py
```diff
@@ -0,0 +1,746 @@
...
+        ]
+        return cast(List[Tuple[str, types.ModuleType]], module_by_import_path)
+
+    @staticmethod
+    def _build_dotted_part(component, relative_import_path) -> str:
+        """Given a component, build a dotted path for import."""
+        if relative_import_path == "":
+            return component.prefix_import_path
+        else:
+            return component.prefix_import_path + "." + relative_import_path
```
marcofavorito: in fact, this will only be used when there are Python sub-packages in a package directory.

aea/aea_builder.py
```diff
@@ -0,0 +1,746 @@
...
+        return self
+
+    def remove_private_key(self, identifier: str) -> "AEABuilder":
+        """
+        Remove a private key path by identifier, if present.
+
+        :param identifier: the identifier of the private key.
+
+        """
+        self._private_key_paths.pop(identifier, None)
```
marcofavorito: it does not give error if the identifier is not used

tests/test_registries.py
```diff
@@ -133,27 +135,23 @@ def setup_class(cls):
         shutil.copytree(os.path.join(CUR_PATH, "data", "dummy_aea"), cls.agent_folder)
         os.chdir(cls.agent_folder)
 
-        # make fake protocol
```
DavidMinarsch: do we have a test elsewhere for adding poorly formatted protocols/components?

tests/test_registries.py
```diff
@@ -213,32 +211,27 @@ def setup_class(cls):
         shutil.copytree(os.path.join(CUR_PATH, "data", "dummy_aea"), cls.agent_folder)
         os.chdir(cls.agent_folder)
 
-        cls.error_skill_public_id = PublicId("fetchai", "error", "0.1.0")
-        cls.dummy_skill_public_id = PublicId.from_str("dummy_author/dummy:0.1.0")
-
-        # # make fake skill
```
DavidMinarsch: see above

tests/test_packages/test_connections/test_http_client/test_http_client.py
```diff
@@ -158,11 +155,3 @@ def setup_class(cls):
 
     await http_client_connection.disconnect()
     assert http_client_connection.connection_status.is_connected is False
-
-
-def test_http_client_from_config():
```
DavidMinarsch: not supported anymore?

tests/test_packages/test_connections/test_http/test_http.py
```diff
@@ -708,17 +691,3 @@ def teardown_class(cls):
         cls.t.join()
         value = cls.loop.run_until_complete(cls.http_connection.disconnect())
         assert value is None
-
-
-def test_conn_from_config():
```
DavidMinarsch: same

tests/test_packages/test_connections/test_gym.py
```diff
@@ -91,11 +88,3 @@ def test_decode_envelope_error(self):
         """Test receive connection error and Cancel Error."""
         with pytest.raises(ConnectionError):
             await self.gym_con.receive()
-
-
-def test_gym_from_config():
```
DavidMinarsch: same

aea/aea_builder.py
```diff
@@ -0,0 +1,746 @@
...
+    def add_component(
+        self, component_type: ComponentType, directory: PathLike
+    ) -> "AEABuilder":
+        """
+        Add a component, given its type and the directory.
+
+        :param component_type: the component type.
+        :param directory: the directory path.
+        :raises ValueError: if a component is already registered with the same component id.
+        """
```
marcofavorito: In summary:
- we first load the configuration in the directory, and do preliminary checks (e.g. right aea version, consistent fingerprints etc.)
- Then, we load the package by including all the dependencies added so far to the agent builder (`manager.dependencies()`). They are handled with a context manager that removes the additional packages imported by the loaded module. For instance, if `connection.py` imports some protocol it will be removed after loading. Other library packages like `requests` will be left in `sys.modules` (they could be removed too anyway).
-

tests/test_connections/test_stub.py
```diff
@@ -211,23 +222,6 @@ def teardown_class(cls):
         cls.multiplexer.disconnect()
 
 
-def test_connection_from_config():
```
DavidMinarsch: removed?

aea/aea_builder.py
```diff
@@ -0,0 +1,746 @@
...
+        self._add_component_to_resources(component)
+
+        return self
+
+    def _add_component_to_resources(self, component: Component):
+        """Add component to the resources."""
+        if component.component_type == ComponentType.CONNECTION:
+            # Do nothing - we don't add connections to resources.
+            return
+        self._resources.add_component(component)
```
marcofavorito: this method will dispatch the component to the right registry

tests/test_connections/test_scaffold_connection.py
```diff
@@ -1,33 +0,0 @@
-# -*- coding: utf-8 -*-
-# ------------------------------------------------------------------------------
-#
-#   Copyright 2018-2019 Fetch.AI Limited
-#
-#   Licensed under the Apache License, Version 2.0 (the "License");
```
DavidMinarsch: why removed?

aea/aea_builder.py
```diff
@@ -0,0 +1,746 @@
...
+
+    def add_contract(self, directory: PathLike) -> "AEABuilder":
+        """Add a contract to the agent."""
+        self.add_component(ComponentType.CONTRACT, directory)
+        return self
+
+    def remove_contract(self, public_id: PublicId) -> "AEABuilder":
+        """Remove protocol"""
+        self.remove_component(ComponentId(ComponentType.CONTRACT, public_id))
+        return self
```
marcofavorito: the above are just wrappers to `add/remove_component`

tests/test_cli/test_run.py
```diff
@@ -1348,15 +1292,60 @@ def teardown_class(cls):
...
-
-    @mock.patch("aea.cli.run.Path.exists", return_value=False)
-    @mock.patch("aea.cli.run.load_agent_component_package", raise_stoptest)
-    def test__setup_connection_no_dir(self, *mocks):
-        """Test for _setup_connection no connection dir."""
-        public_id = PublicIdMock.from_str("author/name:version")
-        ctx = ContextMock(connections=[public_id])
-        with self.assertRaises(StopTest):
-            _setup_connection(public_id, "address", ctx)
+# class TestRunFailsWhenProtocolNotComplete:
```
DavidMinarsch: what's the plan here?

aea/aea_builder.py
```diff
@@ -0,0 +1,746 @@
...
+
+    def _build_identity_from_wallet(self, wallet: Wallet) -> Identity:
```
marcofavorito: this is taken from the old `aea.cli.run`

aea/aea_builder.py
```diff
@@ -0,0 +1,746 @@
...
+                addresses=wallet.addresses,
+                default_address_key=self._default_ledger,
+            )
+        else:  # pragma: no cover
+            identity = Identity(
+                self._name, address=wallet.addresses[self._default_ledger],
+            )
+        return identity
+
+    def _process_connection_ids(
```
marcofavorito: this emulates the option `--connections` in `aea run`

tests/test_cli/test_run.py
```diff
@@ -1012,15 +1009,19 @@ def setup_class(cls):
...
             cls.t,
             cls.agent_name,
             "vendor",
             cls.connection_author,
             "connections",
             cls.connection_name,
             DEFAULT_CONNECTION_CONFIG_FILE,
-        ).unlink()
+        )
+        cls.connection_configuration_path.unlink()
```
DavidMinarsch: what's that doing here, not sure i undertstand..

aea/aea_builder.py
marcofavorito: it is called in build. 
It makes me think that maybe the `connection_ids` argument should be handled by `Agent.start()` instead.

aea/aea_builder.py
```diff
@@ -0,0 +1,746 @@
...
+                )
+            )
+
+    def _check_package_dependencies(self, configuration):
+        self._package_dependency_manager.check_package_dependencies(configuration)
+
+    @staticmethod
+    def _find_component_directory_from_component_id(
+        aea_project_directory: Path, component_id: ComponentId
+    ):
```
marcofavorito: this is useful to find the component directory from the component id, either vendor or not

aea/aea_builder.py
```diff
@@ -0,0 +1,746 @@
...
+            / component_id.public_id.name
+        )
+        if custom_package_path.exists() and custom_package_path.is_dir():
+            return custom_package_path
+
+        raise ValueError("Package {} not found.".format(component_id))
+
+    @staticmethod
+    def _try_to_load_agent_configuration_file(aea_project_path: Path):
+        """Try to load the agent configuration file.."""
```
marcofavorito: this is the same code in the `aea.cli.helpers` module

marcofavorito: it could be refactored eventually.

marcofavorito: I'm thinking that `@check_aea_project` in `aea.cli` might be implemented with the builder, and if the builder manages to add all the components then it means the project is consistent. 

However, that would raise the following issue: should the component be loaded at "add" time (that is, when `add_component` is called) or at `build` time, before instantiating the `AEA` object?
Currently, we do that at "add" time, so the client of this class gets notified soon if he tried to add a bad component. However, that's unpractical if we want to use it in place of `check_aea_project`, because every command would need to load all the packages.

A solution would be to try to load the component at "add" time, but without instantiating a component object.

aea/aea_builder.py
```diff
@@ -0,0 +1,746 @@
...
+            )
+        except jsonschema.exceptions.ValidationError:
+            raise Exception(
+                "Agent configuration file '{}' is invalid. Please check the documentation.".format(
+                    DEFAULT_AEA_CONFIG_FILE
+                )
+            )
+
+    @classmethod
+    def from_aea_project(cls, aea_project_path: PathLike):
```
marcofavorito: this is used in `aea.cli.run` as a replacement of the old implementation.
Basically, what was done in `aea.cli.run`, `aea.registries` and `aea.skill.base.Skill.from_dir` now it is done here and `aea.configurations.components`. Every component class is then in charge of loading what's relevant for it.

marcofavorito: see `Component.load_from_directory`

aea/agent.py
```diff
@@ -79,7 +79,7 @@ def __init__(
         loop: Optional[AbstractEventLoop] = None,
         timeout: float = 1.0,
         is_debug: bool = False,
-        is_programmatic: bool = True,
+        is_programmatic: bool = True,  # TODO to remove
```
marcofavorito: see https://github.com/fetchai/agents-aea/pull/908/files#r400856959

aea/cli/common.py
```diff
@@ -21,36 +21,31 @@
...
 from aea.cli.loggers import default_logging_config
 from aea.configurations.base import (
     AgentConfig,
     ConfigurationType,
     DEFAULT_AEA_CONFIG_FILE,
     Dependencies,
-    PackageConfiguration,
     PublicId,
+    _check_aea_version,
+    _compare_fingerprints,
```
marcofavorito: they have been moved to `aea.configurations.base` since now this functionality is implemented in the configuration objects.

aea/cli/common.py
```diff
@@ -81,17 +75,6 @@
...
-    ".DS_Store",
-    "*__pycache__/*",
-    "*.pyc",
-    "aea-config.yaml",
-    "protocol.yaml",
-    "connection.yaml",
-    "skill.yaml",
-    "contract.yaml",
-]
-
```
marcofavorito: moved in `aea.configurations.base` for the same reason of above

aea/cli/common.py
```diff
@@ -222,16 +205,6 @@ def try_to_load_agent_config(ctx: Context, is_exit_on_except: bool = True) -> No
...
 
 
-def _load_env_file(env_file: str):
-    """
-    Load the content of the environment file into the process environment.
-
-    :param env_file: path to the env file.
-    :return: None.
-    """
-    load_dotenv(dotenv_path=Path(env_file), override=False)
```
marcofavorito: see above: now this is potentially needed to the builder. I say "potentially" because the call to this function is commented in `from_aea_proejct`

aea/aea_builder.py
```diff
@@ -0,0 +1,746 @@
...
+        :param aea_project_path: path to the AEA project.
+        :return: an AEA agent.
+        """
+        aea_project_path = Path(aea_project_path)
+        cls._try_to_load_agent_configuration_file(aea_project_path)
+        _verify_or_create_private_keys(aea_project_path)
+        builder = AEABuilder(with_default_packages=False)
+
+        # TODO isolate environment
+        # load_env_file(str(aea_config_path / ".env"))
```
marcofavorito: this might conflict with other builders.

tests/test_aea.py
```diff
@@ -94,47 +82,41 @@ def test_initialise_aea():
...
-        connection = OEFLocalConnection(
-            identity.address, node, connection_id=LOCAL_CONNECTION_PUBLIC_ID
+        builder = AEABuilder()
+        builder.set_name(agent_name)
+        builder.add_private_key(FETCHAI, private_key_path)
+        builder.add_connection(
+            Path(ROOT_DIR, "packages", "fetchai", "connections", "local")
         )
-        connections = [connection]
-        resources = Resources(str(Path(CUR_PATH, "data", "dummy_aea")))
```
DavidMinarsch: So we cannot build an entire AEA from a path?

tests/test_connections/test_stub.py
```diff
@@ -25,18 +25,38 @@
...
+    ConnectionConfig,
+    PublicId,
+)
 from aea.connections.stub.connection import StubConnection
 from aea.mail.base import Envelope, Multiplexer
 from aea.protocols.default.message import DefaultMessage
 from aea.protocols.default.serialization import DefaultSerializer
 
 
+def _make_stub_connection(input_file_path: str, output_file_path: str):
```
DavidMinarsch: maybe also in conftest.py?

scripts/generate_ipfs_hashes.py
```diff
@@ -40,12 +40,17 @@
...
 
-from aea.cli.common import _compute_fingerprint
+from aea.configurations.base import _compute_fingerprint
 from aea.helpers.ipfs.base import IPFSHashOnly
 
 AUTHOR = "fetchai"
 CORE_PATH = "aea"
-CORE_PACKAGES = {"connections": "stub", "protocols": "default", "skills": "error"}
+CORE_PACKAGES = {
+    "contracts": ["scaffold"],
```
DavidMinarsch: scaffold should not be fingerprinted imo. Or what's the rational?

packages/fetchai/skills/echo/handlers.py
```diff
@@ -30,11 +30,6 @@ class EchoHandler(Handler):
 
     SUPPORTED_PROTOCOL = DefaultMessage.protocol_id
 
-    def __init__(self, **kwargs):
```
DavidMinarsch: Why has this been removed? Then we need to update the docs as well

packages/fetchai/skills/echo/behaviours.py
```diff
@@ -25,11 +25,6 @@
 class EchoBehaviour(TickerBehaviour):
     """Echo behaviour."""
 
-    def __init__(self, **kwargs):
```
DavidMinarsch: see below

packages/fetchai/contracts/erc1155/contract.yaml
```diff
@@ -16,3 +16,4 @@ description: "The erc1155 contract implements an ERC1155 contract package."
 path_to_contract_interface: "build/erc1155.json"
 dependencies:
   vyper: { version: "==0.1.0b12"}
+class_name: ERC1155Contract
```
DavidMinarsch: Not sure why we are adding this, we should use the `*Contract` convention like for the other components. If we do introduce a change to this it should be consistent.

packages/fetchai/contracts/erc1155/contract.py
```diff
@@ -48,22 +48,9 @@ class Performative(Enum):
         CONTRACT_ATOMIC_SWAP_BATCH = "contract_atomic_swap_batch"
         CONTRACT_SIGN_HASH = "contract_sign_hash"
 
-    def __init__(
```
DavidMinarsch: not sure why this has to change

packages/fetchai/connections/local/connection.py
```diff
@@ -315,35 +315,16 @@ class OEFLocalConnection(Connection):
...
-        :param restricted_to_protocols: the only supported protocols for this connection.
-        :param excluded_protocols: the excluded protocols for this connection.
-        """
-        if kwargs.get("connection_id") is None:
-            kwargs["connection_id"] = PublicId("fetchai", "local", "0.1.0")
-        super().__init__(*args, **kwargs)
-        self._address = address
-        self._local_node = local_node
-
+    def load(self) -> None:
```
DavidMinarsch: missing docstring

packages/fetchai/connections/http_client/connection.py
```diff
@@ -167,34 +167,15 @@ def disconnect(self) -> None:
...
-        :param provider_address: server hostname / IP address
-        :param provider_port: server port number
-        """
-
-        if kwargs.get("connection_id") is None:
-            kwargs["connection_id"] = PublicId("fetchai", "http_client", "0.1.0")
-
-        super().__init__(*args, **kwargs)
-        self.agent_address = agent_address
+    def load(self) -> None:
```
DavidMinarsch: missing docstring

packages/fetchai/connections/oef/connection.py
```diff
@@ -606,28 +606,17 @@ def send_oef_message(self, envelope: Envelope) -> None:
 class OEFConnection(Connection):
     """The OEFConnection connects the to the mailbox."""
 
-    def __init__(
-        self, address: Address, oef_addr: str, oef_port: int = 10000, *args, **kwargs
-    ):
+    def load(self) -> None:
         """
-        Initialize.
+        Load the connection.
```
DavidMinarsch: Should probably be.."Load the connection configuration"?

packages/fetchai/connections/p2p_client/connection.py
```diff
@@ -155,24 +155,10 @@ def disconnect(self) -> None:
...
-        Initialize a connection to an SDK or API.
-
-        :param address: the address used in the protocols.
-        """
-        if kwargs.get("connection_id") is None:
-            kwargs["connection_id"] = PublicId("fetchai", "p2p", "0.1.0")
-        super().__init__(*args, **kwargs)
-        self.channel = PeerToPeerChannel(address, provider_addr, provider_port, excluded_protocols=self.excluded_protocols)  # type: ignore
-        self.address = address
+    def load(self) -> None:
```
DavidMinarsch: missing doc string

packages/fetchai/connections/tcp/base.py
```diff
@@ -22,33 +22,22 @@
...
-        :param port: the port to connect to.
-        """
-        if kwargs.get("connection_id") is None:
-            kwargs["connection_id"] = PublicId("fetchai", "tcp", "0.1.0")
-        super().__init__(*args, **kwargs)
-        self.address = address
-
-        self.host = host
-        self.port = port
+    def load(self) -> None:
```
DavidMinarsch: missing docstring

packages/fetchai/connections/tcp/tcp_client.py
```diff
@@ -39,17 +37,8 @@
...
-        Initialize a TCP channel.
-
-        :param address: address.
-        :param host: the socket bind address.
-        :param port: the socket bind port.
-        :param connection_id: the identifier for the connection object.
-        """
-        super().__init__(address, host, port, *args, **kwargs)
-
+    def load(self) -> None:
```
DavidMinarsch: missing docstring

packages/fetchai/connections/tcp/tcp_server.py
```diff
@@ -38,15 +36,8 @@
...
-    def __init__(self, address: Address, host: str, port: int, *args, **kwargs):
-        """
-        Initialize a TCP channel.
-
-        :param address: address.
-        :param host: the socket bind address.
-        """
-        super().__init__(address, host, port, *args, **kwargs)
-
+    def load(self):
```
DavidMinarsch: missing docstring

packages/fetchai/connections/gym/connection.py
```diff
@@ -145,20 +145,10 @@ def disconnect(self) -> None:
...
-        :param gym_env: the gym environment.
-        :param connection_id: the connection id.
-        """
-        if kwargs.get("connection_id") is None:
-            kwargs["connection_id"] = PublicId("fetchai", "gym", "0.1.0")
-        super().__init__(*args, **kwargs)
-        self.address = address
-        self.channel = GymChannel(address, gym_env)
-
+    def load(self) -> None:
```
DavidMinarsch: missing docstring

aea/skills/base.py
```diff
@@ -196,12 +211,6 @@ def contracts(self) -> SimpleNamespace:
         assert self._skill is not None, "Skill not initialized."
         return SimpleNamespace(**self._skill.contracts)
 
-    @property
```
DavidMinarsch: what happened with the logger?

aea/skills/base.py
```diff
@@ -61,34 +53,57 @@
...
+        """Get the logger."""
+        assert self._logger is not None, "Logger not set."
+        return self._logger
+
+    @logger.setter
+    def logger(self, logger_: Logger) -> None:
+        assert self._logger is None, "Logger already set."
+        self._logger = logger_
+
+    def _get_agent_context(self) -> AgentContext:
```
DavidMinarsch: why not a property?

DavidMinarsch: i see because it needs to be private..don't quite follow the change.

aea/protocols/base.py
```diff
@@ -231,69 +226,66 @@ def decode(self, obj: bytes) -> Message:
...
-        return self._protocol_id
+        return self._configuration.public_id
 
     @property
     def serializer(self) -> Serializer:
         """Get the serializer."""
+        assert self._serializer is not None, "Serializer not initialized yet."
         return self._serializer
 
+    @serializer.setter
```
DavidMinarsch: do we actually use the serializer property anywhere?

examples/programmatic.py
```diff
@@ -0,0 +1,57 @@
+# -*- coding: utf-8 -*-
```
marcofavorito: this file was only to show an example to the reviewer. Now it can be removed.

tests/test_connections/test_stub.py
marcofavorito: yes

tests/test_aea.py
marcofavorito: yes, I just wanted to build it partially. But probably it's not a bad idea to build from the directory in this case.

aea/cli/common.py
```diff
@@ -568,94 +541,6 @@ def _find_item_locally(ctx, item_type, item_public_id) -> Path:
     return package_path
 
 
-def _compute_fingerprint(
```
marcofavorito: moved in `aea.configurations.base`

aea/cli/run.py
```diff
@@ -19,169 +19,28 @@
...
-from aea.identity.base import Identity
-from aea.registries.resources import Resources
-
+from aea.configurations.base import PublicId
+from aea.helpers.base import load_env_file
 
 AEA_DIR = str(Path("."))
 
 
-def _setup_connection(
```
marcofavorito: integrated in `AEABuilder.from_aea_project`

aea/configurations/base.py
```diff
@@ -396,13 +439,137 @@ def __lt__(self, other):
...
+
+    A package can be one of:
+    - agents
+    - protocols
+    - connections
+    - skills
+    - contracts
+    """
+
+    default_configuration_filename: str
```
marcofavorito: the default configuration file name is now a class attribute of a package configuration class.

aea/configurations/base.py
```diff
@@ -491,6 +778,7 @@ def __init__(
             aea_version,
             fingerprint,
             fingerprint_ignore_patterns,
+            dependencies,
```
marcofavorito: now `dependencies` are defined at `ComponentConfiguration` level.

marcofavorito: should it be included in an agent configuration too? Not sure it makes sense.

aea/configurations/base.py
```diff
@@ -1142,4 +1473,96 @@ def from_json(cls, obj: Dict):
             path_to_contract_interface=cast(
                 str, obj.get("path_to_contract_interface", "")
             ),
+            class_name=obj.get("class_name", ""),
+        )
+
+
+"""The following functions are called from aea.cli.common."""
+
+
```
marcofavorito: the following functions have been moved from `aea.cli.common`

aea/contracts/ethereum.py
```diff
@@ -23,30 +23,17 @@
...
-        Initialize the contract.
-
-        :param contract_id: the contract id.
-        :param config: the contract configurations.
-        :param contract_interface: the contract interface.
-        """
-        super().__init__(contract_id, config, contract_interface)
-        self._abi = contract_interface["abi"]
-        self._bytecode = contract_interface["bytecode"]
+    def load(self) -> None:
```
DavidMinarsch: docstring missing

aea/contracts/base.py
```diff
@@ -18,44 +18,53 @@
...
+        super().__init__(config)
+        self._contract_interface = None  # type: Optional[Dict[str, Any]]
 
     @property
     def id(self) -> ContractId:
         """Get the name."""
-        return self._contract_id
+        return self.public_id
 
+    # TODO to remove
```
DavidMinarsch: ?

aea/configurations/schemas/contract-config_schema.json
```diff
@@ -8,7 +8,8 @@
     "author",
     "version",
     "license",
-    "aea_version"
+    "aea_version",
+    "class_name"
```
DavidMinarsch: see below, not a fan

packages/fetchai/contracts/erc1155/contract.yaml
marcofavorito: for connections, we have the class name.

aea/aea_builder.py
```diff
@@ -0,0 +1,746 @@
...
+    @property
+    def contracts(self) -> Dict[ComponentId, Any]:
+        """Get the contracts."""
+        return cast(
+            Dict[ComponentId, Any],
+            self._all_dependencies_by_type.get(ComponentType.CONTRACT, {}),
+        )
+
+    def add_component(self, component: Component) -> None:
+        """Add a component."""
```
DavidMinarsch: quite a few incomplete docstrings

scripts/generate_ipfs_hashes.py
marcofavorito: because `AEABuilder.add_component`, in the current implementation, always checks the consistency of the configuration when a package is added. If the scaffold packages are a package, then they should be consistent since the beginning.

packages/fetchai/skills/echo/handlers.py
marcofavorito: It was a mistake, I'll reintroduce them

aea/contracts/base.py
marcofavorito: Contract inherits `configuration` from parent class `Component`, hence `config` is obsolete.

aea/protocols/base.py
marcofavorito: in `Protocol.load`

aea/skills/base.py
marcofavorito: to handle the case when `self._agent_context` is None. It avoids to repeat `assert self._get_agent_context is not None` in other properties.

packages/fetchai/contracts/erc1155/contract.py
marcofavorito: subclasses don't define custom constructors anymore, but can add custom attributes in `load`

PR #916, opened 2020-03-11
Functionality for single tokens creation

Totoual:
## Proposed changes

Adds functionality to the smart contract for creating and minting single tokens.

## Fixes

- 

## Types of changes

## Checklist


packages/fetchai/contracts/erc1155/contract.py
```diff
@@ -79,6 +90,13 @@ def create_token_ids(self, token_type: int, nb_tokens: int) -> Dict[int, int]:
 
         return self.token_ids
 
+    def generate_single_item_id(self, token_type: int, game_currency: int) -> int:
```
DavidMinarsch: There is no need for this function. You already have `create_token_id` with which you can generate a single token.

packages/fetchai/contracts/erc1155/contract.py
```diff
@@ -63,21 +63,34 @@ def __init__(
...
+        self._token_ids = {}  # type: Dict[int, int]
+        self._game_currency_id = 0  # type: int
 
-    def create_token_ids(self, token_type: int, nb_tokens: int) -> Dict[int, int]:
+    @property
+    def token_ids(self) -> Dict[int, int]:
+        """The generated token ids."""
+        return self._token_ids
+
+    @property
```
DavidMinarsch: @Totoual I think this property and associated attribute should be removed. They should be kept on the business logic side of the skill, not here.

DavidMinarsch: From the perspective of the erc1155 they are all tokens, so the `token_ids` dict is enough

Issue #640, opened 2020-01-15
Introduce a filter class which intelligently selects the appropriate handler for a message.


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
Currently, a message of a given protocol gets routed to all handlers registered for that protocol.

**Describe the solution you'd like**
We could extend this in two directions:
- handlers can define a set of rules regarding when they want to receive a message for the protocol they implement.
- messages can contain routing information in their optional URI field which determines the handler which receives the message.

PR #919, opened 2020-03-11
AEA-342 attempt to add routing to filter

DavidMinarsch:
## Proposed changes

Attempt to satisfy #https://github.com/fetchai/agents-aea/issues/640 in a minimal way.

## Fixes

https://github.com/fetchai/agents-aea/issues/640

## Types of changes

## Checklist

## Further comments

-


aea/registries/base.py
```diff
@@ -728,16 +729,32 @@ def decision_maker_out_queue(self) -> Queue:
...
         :param protocol_id: the protocol id
+        :param envelope context: the envelope context
         :return: the list of handlers currently active
         """
-        # TODO naive implementation
-        handlers = self.resources.handler_registry.fetch_by_protocol(protocol_id)
-        active_handlers = list(filter(lambda h: h.context.is_active, handlers))
+        if envelope_context is not None and envelope_context.uri is not None:
+            uri_path = envelope_context.uri.path
+            skill_id = PublicId.from_str(uri_path)
```
DavidMinarsch: @marcofavorito we have this uri available in the envelope context. In the HTTP connection this will be filled with the request URL. It would be nice if we could use the URI to map to a skill. But our skill ids are not compatible with the path (`:`) unless we do something like `fetchai/echo/0.1.0` mapping to `fetchai/echo:0.1.0`. Thoughts?

marcofavorito: I think it is not too bad to have `fetchai/echo/0.1.0` as URI.
We could add methods like `public_id.to_uri()` and `PublicId.from_uri()` to parse them.

aea/registries/base.py
```diff
@@ -728,16 +729,36 @@ def decision_maker_out_queue(self) -> Queue:
...
         :return: the list of handlers currently active
         """
-        # TODO naive implementation
-        handlers = self.resources.handler_registry.fetch_by_protocol(protocol_id)
-        active_handlers = list(filter(lambda h: h.context.is_active, handlers))
+        skill_id = None  # Optional[PublicId]
+        if envelope_context is not None and envelope_context.uri is not None:
+            uri_path = envelope_context.uri.path
+            try:
+                skill_id = PublicId.from_uri_path(uri_path)
```
DavidMinarsch: @marcofavorito , thoughts?

PR #924, opened 2020-03-15
AEA-397 Add 'aea_version' field to jsonschema for packages config

marcofavorito:
## Proposed changes

Make configuration files include a field named 'aea_version'. It has the following format:
- either a string that matches the semver format: https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string
- or a string that matches the version specifier (PEP 440): https://www.python.org/dev/peps/pep-0440/#version-scheme

The former is both for backward compatibility and easy of use. The latter lets the user to specify version ranges. It is an optional field (mostly for backward-compatibility).

## Fixes

Fixes #764 

## Types of changes

What types of changes does your code introduce to agents-aea?
_Put an `x` in the boxes that apply_

- [ ] Bugfix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

_Put an `x` in the boxes that apply._

- [ ] I have read the [CONTRIBUTING](../CONTRIBUTING.md) doc
- [ ] I am making a pull request against the `develop` branch (left side). Also you should start your branch off our `develop`.
- [ ] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have checked that code coverage does not decrease.
- [ ] I have checked that the documentation about the `aea cli` tool works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

## Further comments

I think it is recommended to add the field by default, maybe in later changes, in the form `aea_version: >=[current-aea-version]`

PR #929, opened 2020-03-16
Add checks for `aea_version` of packages.

marcofavorito:
## Proposed changes

This PR adds checks regarding the AEA version compatibility that packages and agent project have w.r.t. the version of the CLI `aea` that is used.

For instance, if we are using `aea` version `0.2.0`, and we are trying to add a package with the attribute `aea_version` of value `>=0.3.0`, the command is going to fail.

## Fixes

Linked to #924, but it adds consistency checks on the `aea` version.

## Types of changes

## Checklist


aea/configurations/base.py
```diff
@@ -301,14 +307,42 @@ def __init__(
...
         self.author = author
         self.version = version
         self.license = license
         self.fingerprint = fingerprint if fingerprint is not None else {}
+        self.aea_version = aea_version if aea_version != "" else aea.__version__
+        self._aea_version_specifiers = self._parse_aea_version_specifier(aea_version)
+
+    def _parse_aea_version_specifier(self, aea_version_specifiers: str) -> SpecifierSet:
+        try:
+            Version(aea_version_specifiers)
```
marcofavorito: we first check the case `aea_version: <version_number>`.
It might fail since we allow for version specifiers.

Pipfile
```diff
@@ -41,6 +41,7 @@ pytest-randomly = "==3.2.1"
 safety = "==1.8.5"
 tox = "==3.14.5"
 vyper = "==0.1.0b12"
+packaging = "==20.3"
```
DavidMinarsch: This dependency must be defined in setup.py, not here.

aea/connections/scaffold/connection.yaml
```diff
@@ -2,6 +2,7 @@ name: scaffold
 author: fetchai
 version: 0.1.0
 license: Apache-2.0
+aea_version: ">=0.2.2"
```
DavidMinarsch: In my opinion it should be `==0.2.2` since we are still having breaking changes between versions. We should aim to move to `>=0.2.2, <0.3.0`. But I think `>=0.2.2` is wrong.

DavidMinarsch: Applies to all files!

Pipfile
DavidMinarsch: I notice it is in setup.py, so remove here please. We have the convention now that all framework dependencies are only listed in setup.py

PR #945, opened 2020-03-18
AEA-346 Changes to message_consistency

Totoual:
## Proposed changes

Addresses the message we send to the participants.

## Fixes

- 

## Types of changes

## Checklist


packages/fetchai/skills/tac_control_contract/behaviours.py
```diff
@@ -56,6 +56,7 @@ def __init__(self, **kwargs):
         self._oef_msg_id = 0
         self._registered_desc = None  # type: Optional[Description]
         self.context.shared_state["is_items_created"] = False
```
DavidMinarsch: I am not quite sure why you use shared_state for intra skill state. It is intended for inter skill state. For intra skill you can just access the module, like `self.context.behaviours.behaviour_name`

packages/fetchai/protocols/tac/serialization.py
```diff
@@ -27,16 +26,38 @@
...
+        if key_type == int and value_type == int:
+            pair = tac_pb2.IntIntPair()
+        elif key_type == int and value_type == float:
+            pair = tac_pb2.IntFloatPair()
+        elif key_type == int and value_type == str:
+            pair = tac_pb2.IntStrPair()
+        elif key_type == str and value_type == int:
             pair = tac_pb2.StrIntPair()
-        elif type(value) == str:
+        elif key_type == str and type(value) == str:
```
DavidMinarsch: can use `value_type` var here and in line below

packages/fetchai/protocols/tac/serialization.py
```diff
@@ -27,16 +26,38 @@
...
+    return {key_map(k): value_map(v) for k, v in ob.items()}
+
+
+def keys_to_strings(ob: dict):
+    """Convert keys to strings."""
+    return map_dict(ob, key_map=str)
+
+
+def keys_to_integers(ob: dict):
+    """Convert keys to strings."""
```
DavidMinarsch: docstring wrong

packages/fetchai/skills/tac_control_contract/game.py
```diff
@@ -889,7 +889,7 @@ def holdings_summary(self) -> str:
                 + "\n"
             )
             for good_id, quantity in agent_state.quantities_by_good_id.items():
-                result += "    " + good_id + ": " + str(quantity) + "\n"
+                result += "    " + str(good_id) + ": " + str(quantity) + "\n"
```
DavidMinarsch: since it is string now, the conversion is no longer needed

packages/fetchai/skills/tac_control_contract/game.py
```diff
@@ -901,7 +901,7 @@ def equilibrium_summary(self) -> str:
         """Get equilibrium summary."""
         result = "\n" + "Equilibrium prices: \n"
         for good_id, eq_price in self.initialization.good_id_to_eq_prices.items():
-            result = result + good_id + " " + str(eq_price) + "\n"
+            result = result + str(good_id) + " " + str(eq_price) + "\n"
```
DavidMinarsch: see above

packages/fetchai/skills/tac_control_contract/game.py
```diff
@@ -915,7 +915,7 @@ def equilibrium_summary(self) -> str:
                 + ":\n"
             )
             for good_id, quantity in eq_allocations.items():
-                result = result + "    " + good_id + ": " + str(quantity) + "\n"
+                result = result + "    " + str(good_id) + ": " + str(quantity) + "\n"
```
DavidMinarsch: see above

packages/fetchai/skills/tac_control_contract/behaviours.py
Totoual: I find it easier just to use the shared_state. But if this is a bad practice I can change it.

DavidMinarsch: Yes I consider it bad practice as it gives other skills the ability to affect that skill.

packages/fetchai/skills/tac_control_contract/behaviours.py
```diff
@@ -149,16 +146,14 @@ def act(self) -> None:
                 self.context.logger.info("Mint objects after registration.")
                 for agent in self.context.configuration.agent_addr_to_name.keys():
                     self._mint_objects(
-                        is_batch=True,
-                        address=agent,
-                        nonce_index=self.context.shared_state["agent_counter"],
+                        is_batch=True, address=agent, nonce_index=self.agent_counter,
```
DavidMinarsch: why do we not handle the nonce on the contract.py. Seems like this is unnecessary complexity to expose here?

packages/fetchai/skills/tac_control_contract/behaviours.py
```diff
@@ -343,6 +338,6 @@ def _mint_objects(
                 mint_quantity=parameters.money_endowment,
                 ledger_api=self.context.ledger_apis.apis.get("ethereum"),
                 skill_callback_id=self.context.skill_id,
-                token_id=self.context.shared_state["token_ids"][0],
+                token_id=token_id,
```
DavidMinarsch: where is this defined?

packages/fetchai/skills/tac_control_contract/behaviours.py
Totoual: No we handle it in the contract. Though, because we are creating the tx before we send it the transaction, due to the loop need to add the counter so I manually update the nonce here.

packages/fetchai/skills/tac_control_contract/behaviours.py
Totoual: The token_id is a parameter in the function. None by default. If we want to mint a single token we need to provide the. token id in the function.

PR #949, opened 2020-03-19
[wip] AEA-346 Tac participation contract skill

Totoual:
## Proposed changes

Introduces the tac participation contract skill for the contract version of the TAC game.

## Fixes

-

## Types of changes

## Checklist


packages/fetchai/skills/tac_participation_contract/handlers.py
```diff
@@ -0,0 +1,517 @@
...
+            str(key): value for key, value in tac_message.quantities_by_good_id.items()
+        }
+        exchange_params_by_currency_id = {
+            str(key): value
+            for key, value in tac_message.exchange_params_by_currency_id.items()
+        }
+        utility_params_by_good_id = {
+            str(key): value
+            for key, value in tac_message.utility_params_by_good_id.items()
+        }
```
Totoual: The main change is here that I am changing the type of the keys in the dictionaries. For some reason I receive them as ints. Thought I am investigating in order to remove this conversion.

packages/fetchai/skills/tac_participation/handlers.py
```diff
@@ -329,12 +330,43 @@ def _on_start(self, tac_message: TACMessage) -> None:
...
+                for key, value in tac_message.quantities_by_good_id.items()
+            }
+            exchange_params_by_currency_id = {
+                str(key): value
+                for key, value in tac_message.exchange_params_by_currency_id.items()
+            }
+            utility_params_by_good_id = {
+                str(key): value
+                for key, value in tac_message.utility_params_by_good_id.items()
+            }
```
Totoual: This is not ideal but for now I have to make the conversion.

packages/fetchai/contracts/erc1155/contract.py
```diff
@@ -196,9 +196,6 @@ def _get_create_batch_tx(
         """Create a batch of items."""
         # create the items
         self.nonce += 1
-        nonce = ledger_api.api.eth.getTransactionCount(deployer_address)
-        if self.nonce != nonce:
-            self.nonce = nonce
```
Totoual: Removed the check here. Since we are keeping track of the nonce. Actually creates a problem because the nonce is always less than the supposed nonce.

packages/fetchai/contracts/erc1155/contract.py
```diff
@@ -254,9 +251,6 @@ def _get_create_single_tx(
     ) -> str:
         """Create an item."""
         self.nonce += 1
-        nonce = ledger_api.api.eth.getTransactionCount(deployer_address)
-        if self.nonce != nonce:
-            self.nonce = nonce
```
Totoual: See above.

packages/fetchai/contracts/erc1155/contract.py
```diff
@@ -378,11 +371,6 @@ def _create_mint_single_tx(
         """Mint a batch of items."""
         # mint batch
         self.nonce += 1
-        nonce = ledger_api.api.eth.getTransactionCount(
-            ledger_api.api.toChecksumAddress(deployer_address)
-        )
-        if self.nonce != nonce:
-            self.nonce = nonce
```
Totoual: See above.

packages/fetchai/skills/tac_control_contract/behaviours.py
```diff
@@ -59,6 +59,7 @@ def __init__(self, **kwargs):
         self.can_start = False
         self.agent_counter = 0
         self.token_ids = []
+        self.currency_id = 0
```
Totoual: We keep track of the game currency on the behaviour level. We generate the currency along with the token_ids. That. means we have to create eleven items at the beginning of the controller and split the list before we create the items on the contract side.

packages/fetchai/skills/tac_control_contract/behaviours.py
```diff
@@ -113,8 +114,13 @@ def act(self) -> None:
             )
             self.token_ids = [int(token_id) for token_id in token_ids_dictionary.keys()]
             self.context.logger.info("Creating the items.")
+            self.currency_id = self.token_ids[0]
+            self.token_ids = self.token_ids[1:11]
```
Totoual: Splitting the token_ids, it is not ideal but it is the fastest way. Else we have to change the assertion when we are creating the items.

packages/fetchai/skills/tac_control_contract/behaviours.py
```diff
@@ -113,8 +114,13 @@ def act(self) -> None:
...
             transaction_message = self._create_items(self.token_ids)
             self.context.decision_maker_message_queue.put_nowait(transaction_message)
+
+            transaction_message = self._create_items(self.currency_id)
+            self.context.decision_maker_message_queue.put_nowait(transaction_message)
```
Totoual: Transaction message for game currency.

packages/fetchai/skills/tac_control_contract/behaviours.py
```diff
@@ -147,6 +153,10 @@ def act(self) -> None:
                     self._mint_objects(
                         is_batch=True, address=agent,
                     )
+                    # Mint the game currency.
+                    self._mint_objects(
+                        is_batch=False, address=agent,
+                    )
```
Totoual: Minting the game currency.

packages/fetchai/skills/tac_control_contract/behaviours.py
```diff
@@ -309,12 +320,12 @@ def _create_items(self, token_ids: Union[List[int], int]) -> TransactionMessage:
                 token_id=token_ids,
             )
 
-    def _mint_objects(self, is_batch: bool, address: Address, token_id: int = None):
+    def _mint_objects(self, is_batch: bool, address: Address):
```
Totoual: Since we are keeping track of the game currency in the behaviour no need to pass it to the function anymore.

packages/fetchai/skills/tac_control_contract/behaviours.py
```diff
@@ -309,12 +320,12 @@ def _create_items(self, token_ids: Union[List[int], int]) -> TransactionMessage:
...
         self.context.logger.info("Minting the items")
         contract = self.context.contracts.erc1155
         parameters = cast(Parameters, self.context.parameters)
         if is_batch:
-            minting = [parameters.base_good_endowment] * parameters.nb_goods
+            minting = [parameters.base_good_endowment] * (parameters.nb_goods - 1)
```
Totoual: The nb_goods that we are passing from the parameter has value `11`. It contains the game currency. The minting list must be of size `10`. We are minting the game currency later on a separate transaction.

packages/fetchai/skills/tac_control_contract/handlers.py
```diff
@@ -321,11 +321,10 @@ def handle(self, message: Message) -> None:
                     )
                 )
                 contract.set_address(ledger_api, transaction.contractAddress)
-        elif (
-            tx_msg_response.tx_id == "contract_create_batch"
-            or tx_msg_response.tx_id == "contract_create_single"
-        ):
-            self.context.logger.info("Sending creation transaction to the ledger!")
+        elif tx_msg_response.tx_id == "contract_create_single":
```
Totoual: Had to split the check here. So I can control the game flow better.

packages/fetchai/skills/tac_participation/skill.yaml
```diff
@@ -29,6 +31,10 @@ models:
     class_name: Search
     args:
       search_interval: 5
+  parameters:
+    class_name: Parameters
+    args:
+      is_using_contract: False
```
Totoual: We can specify if we are using the contract or not in the skill config. We will need to add this also in the negotiation skill or save it in the `shared state`.

packages/fetchai/contracts/erc1155/contract.py
DavidMinarsch: Well it makes sense to check the nonce. Just this particular check does not make sense. Leave it for me to finish.

packages/fetchai/skills/tac_participation/parameters.py
```diff
@@ -0,0 +1,36 @@
...
+#   limitations under the License.
+#
+# ------------------------------------------------------------------------------
+
+"""This package contains a class representing the game parameters."""
+
+from aea.skills.base import Model
+
+
+class Parameters(Model):
```
DavidMinarsch: I'd add this to `game` class to reduce lines of code

packages/fetchai/skills/tac_participation/handlers.py
```diff
@@ -329,12 +330,48 @@ def _on_start(self, tac_message: TACMessage) -> None:
         game = cast(Game, self.context.game)
         game.init(tac_message, tac_message.counterparty)
         game.update_game_phase(Phase.GAME)
+        parameters = cast(Parameters, self.context.parameters)
+
+        if parameters.is_using_contract:
+            amount_by_currency_id = {
+                str(key): value
```
DavidMinarsch: why the conversion? below it is not needed

packages/fetchai/skills/tac_participation/handlers.py
```diff
@@ -329,12 +330,48 @@ def _on_start(self, tac_message: TACMessage) -> None:
...
+            }
+            utility_params_by_good_id = {
+                str(key): value
+                for key, value in tac_message.utility_params_by_good_id.items()
+            }
+
+            contract = self.context.contracts.erc1155
+            contract.set_deployed_instance(
+                self.context.ledger_apis.apis.get("ethereum"),
+                tac_message.get("contract_address"),
```
DavidMinarsch: contract address should be in the info dict imo

packages/fetchai/skills/tac_control_contract/skill.yaml
```diff
@@ -32,7 +32,7 @@ models:
     args:
       min_nb_agents: 2
       money_endowment: 2000000
-      nb_goods: 10
+      nb_goods: 11
```
DavidMinarsch: need to think about this

packages/fetchai/contracts/erc1155/contract.py
DavidMinarsch: self.nonce is your off-chain reference to the nonce, which is generally ahead of the ledger api nonce. Hence, I would just check that `nonce <= self.nonce`, the assignment makes no sense in my opinion.

packages/fetchai/contracts/erc1155/contract.py
DavidMinarsch: see above

packages/fetchai/contracts/erc1155/contract.py
DavidMinarsch: see above

packages/fetchai/skills/tac_control_contract/behaviours.py
DavidMinarsch: So I'd initially set `self._currency_id = None  # type: Optional[int]` and `self._token_ids = None. # type: Optional[List[int]]` and add two properties, one for each, which assert that the value is set and return the non-optional type.

packages/fetchai/skills/tac_control_contract/behaviours.py
DavidMinarsch: See above, then here you assign to `self._*`

packages/fetchai/skills/tac_control_contract/behaviours.py
DavidMinarsch: I think parameter value should still be 10 for goods. It is clear that there must be a single game currency.

packages/fetchai/skills/tac_control_contract/skill.yaml
DavidMinarsch: see above, this should be reverted to 10, as the use of a single game currency is implicit.

Totoual: The reason this is 11 and not 10 is that when we are using the generate method in the contract we are checking if we created the items. If we create them separate and not once then we have problem with the assertion.

` assert self.token_ids == {}, "Item ids already created."`

DavidMinarsch: I'd suggest using a different way to check that then. There aren't 11 goods so this should be reverted.

packages/fetchai/skills/tac_participation/handlers.py
Totoual: The problem here is that you don't support info in the `GAME_DATA` message.

packages/fetchai/contracts/erc1155/contract.py
```diff
@@ -74,7 +74,7 @@ def token_ids(self) -> Dict[int, int]:
 
     def create_token_ids(self, token_type: int, nb_tokens: int) -> List[int]:
         """Populate the token_ids dictionary."""
-        assert self.token_ids == {}, "Item ids already created."
+        # assert self.token_ids == {}, "Item ids already created."
```
Totoual: I commented out the assertion because we are not generating anything if the `_i` is not in the range of token_id_list  and we don't recreate it if it is already generated. I was thinking to check if the id exists in the dictionary but I am not sure here.

Issue #873, opened 2020-03-01
Replace all packages with generated versions


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
Currently, packages are the hand written versions.

**Describe the solution you'd like**
Replace all packages with the generated equivalents.

PR #964, opened 2020-03-22
WIP: AEA-425 generated oef protocol

DavidMinarsch:
## Proposed changes

This PR changes the OEF protocol to the generated version.

## Fixes

- #873 

## Types of changes

## Checklist

## Further comments

-


aea/context/base.py
```diff
@@ -29,6 +29,8 @@
 from aea.mail.base import Address, OutBox
 from aea.skills.tasks import TaskManager
 
+DEFAULT_OEF = "default_oef"
```
DavidMinarsch: Note, @marcofavorito we need to expose the choice of OEF address on the configuration. For now it is stubbed.

packages/fetchai/connections/oef/connection.py
```diff
@@ -503,17 +510,20 @@ def on_dialogue_error(
         """
         assert self.in_queue is not None
         assert self.loop is not None
-        msg = OEFMessage(
-            type=OEFMessage.Type.DIALOGUE_ERROR,
-            id=answer_id,
-            dialogue_id=dialogue_id,
-            origin=origin,
+        msg = DefaultMessage(
```
DavidMinarsch: Note, @marcofavorito it makes little sense to use the OEF protocol for notifying of dialogue errors as these would be caused by protocols other than the OEF protocol. It makes sense to use  the default protocol instead so that the error handler can kick in. Otherwise, each agent needs the oef protocol just for error handling, which will confuse people.

packages/fetchai/protocols/oef/custom_types.py
```diff
@@ -0,0 +1,142 @@
...
+
+from aea.helpers.search.models import Description as BaseDescription
+from aea.helpers.search.models import Query as BaseQuery
+
+
+class Description(BaseDescription):
+    """This class represents an instance of Description."""
+
+    @classmethod
+    def encode(cls, performative, description_from_message: "Description"):
```
DavidMinarsch: Eventually, BaseDescription and BaseQuery should come with their own serizliation logic using protobuf

packages/fetchai/connections/oef/connection.py
marcofavorito: But can't a protocol have its own error messages and error handling, like in the OEF protocol?

DavidMinarsch: The dialogue error would result from a different protocol (i.e. not oef protocol) and should therefore error out on base protocol (default protocol). Ex: I send a message in fipa via oef which causes a dialogue error on oef node..it's now not ok for the oef node to send me an error in oef protocol as I might not have that protocol, instead it should use default protocol

aea/helpers/search/models.py
```diff
@@ -108,6 +110,34 @@ def __iter__(self):
...
+        """
+        Encode an instance of this class into the protocol buffer object.
+
+        The content in the 'performative' argument must be matched with the message content in the 'description_from_message' argument.
+
+        :param performative: the performative protocol buffer object containing a content whose type is this class.
+        :param description_from_message: the message content to be encoded in the protocol buffer object.
+        :return: the 'performative' protocol buffer object encoded with the message content in the 'description_from_message' argument.
+        """
+        description_from_message_bytes = pickle.dumps(description_from_message)  # nosec
```
DavidMinarsch: Imperfect, but a temporary working solution. Eventually, we make these serializable with protobuf and remove the dependency on performative.

packages/fetchai/connections/oef/connection.py
```diff
@@ -296,6 +303,8 @@ def __init__(
         self.in_queue = None  # type: Optional[asyncio.Queue]
         self.loop = None  # type: Optional[AbstractEventLoop]
         self.excluded_protocols = excluded_protocols
+        self.oef_msg_id = 0
+        self.oef_msg_it_to_dialogue_reference = {}  # type: Dict[int, Tuple[str, str]]
```
DavidMinarsch: we now keep track of the dialogue reference to be able to respond appropriately

packages/fetchai/connections/oef/connection.py
```diff
@@ -553,28 +574,31 @@ def send_oef_message(self, envelope: Envelope) -> None:
...
         """
-        oef_message = OEFSerializer().decode(envelope.message)
-        oef_message = cast(OEFMessage, oef_message)
-        oef_type = oef_message.type
-        oef_msg_id = oef_message.id
-        if oef_type == OEFMessage.Type.REGISTER_SERVICE:
+        oef_message = OefSearchSerializer().decode(envelope.message)
+        oef_message = cast(OefSearchMessage, oef_message)
+        self.oef_msg_id += 1
+        self.oef_msg_it_to_dialogue_reference[self.oef_msg_id] = (
```
DavidMinarsch: we create the dialogue reference part for the oef (RHS)

Issue #954, opened 2020-03-20
Consistency checks of protocols must bubble up errors


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
Currently, some message consistency checks do not show the relevant error message.

**Describe the solution you'd like**
Show it.

Issue #913, opened 2020-03-09
Provide dialogue reference helper methods and default values for message meta data


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
Currently, all messages need to provided with a dialogue reference and values for the meta data (target, message_id, dialogue_reference)

**Describe the solution you'd like**
Set these by default (i.e. new dialogue by default) and provide a helper method to generate the dialogue references.

Issue #905, opened 2020-03-07
Generator Bugs


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
A small bugs are present in the generator:
- [x] errors in the specification cause a stack trace rather than just a clean warning message
- [x] the generated files are not linted with black
- [x] if the message defines speech act contents already used by the default properties (e.g. target, body), then the generator does not show this as an error
- [x] optional fields not decoding: https://github.com/googleapis/google-cloud-python/issues/1402
<img width="709" alt="Screenshot 2020-03-08 at 10 50 18" src="https://user-images.githubusercontent.com/19314578/76161594-c896c200-612c-11ea-96f2-be1a076e14a9.png">

- [x] repeat usage of variables ("key, value"). Mypy complains and we should use unique and descriptive names for these variables:
![Screenshot 2020-03-11 at 14 34 56](https://user-images.githubusercontent.com/19314578/76430317-179c5b80-63a8-11ea-9dd6-9b72b9e1eb35.png)

- [x] custom types not correctly declared when tac protocol is generated:
![Screenshot 2020-03-11 at 14 34 14](https://user-images.githubusercontent.com/19314578/76430399-37338400-63a8-11ea-90a4-5c37d6fff4cd.png)

- [x] in serialisation.py: 'packages.fetchai.protocols.tac.custom_types.ErrorInfo' imported but unused

- [x] `pt:list[pt:str]` generates a tuple in oef protocol. explain this in the docs!

- [x] add type declaration in decoding logic or serialisation `performative_content = dict()  # type: Dict[str, Any]`

**Describe the solution you'd like**
Fix the above.

PR #966, opened 2020-03-22
WIP: AEA-425 AEA-440 add black and default values to generator

DavidMinarsch:
## Proposed changes

This PR adds black code formatting and default values for `target`, `message_id` and `dialogue_reference` to the protocol generator.

## Fixes

It partially addresses #905 and #913, it addresses #954

## Types of changes

## Checklist

## Further comments

-


aea/cli/generate.py
```diff
@@ -50,13 +51,21 @@ def generate(ctx: Context):
...
+    # check protocol buffer compiler is installed
+    res = shutil.which("protoc")
+    if res is None:
+        print(
+            "Please install protocol buffer first! See the following link: https://developers.google.com/protocol-buffers/"
+        )
+        sys.exit(1)
+
+    # check black code formatter is installed
+    res = shutil.which("black")
```
marcofavorito: an idea: since below it is used with `-m black`, we should 
```python
try:
    import black
except ImportError:
    print(...)
```

aea/cli/generate.py
```diff
@@ -136,6 +145,24 @@ def _generate_item(ctx: Context, item_type, specification_path):
...
+                os.path.join(item_type_plural, protocol_spec.name),
+                "--quiet",
+            ]
+        )
+        subp.wait(5.0)
+    finally:
+        poll = subp.poll()
+        if poll is None:  # pragma: no cover
+            subp.terminate()
+            subp.wait(2)
```
marcofavorito: in theory, this could fail again if SIGTERM is not enough.

marcofavorito: but for `black` it should be OK

DavidMinarsch: hm ok..we have this pattern all over the place.

aea/cli/generate.py
DavidMinarsch: removed `-m` instead

aea/protocols/generator.py
```diff
@@ -416,7 +427,12 @@ def _setup(self) -> None:
             all_performatives_set.add(performative)
             self._speech_acts[performative] = {}
             for content_name, content_type in speech_act_content_config.args.items():
-                custom_types = set(re.findall(CUSTOM_TYPE_PATTERN, content_type))
+                if not _is_valid_content_name(content_name):
+                    raise ProtocolSpecificationParseError(
+                        "Incorrect content name: '{}'".format(content_name)
```
DavidMinarsch: Maybe say, `"Reserved content name:"`?

aea/protocols/generator.py
```diff
@@ -826,6 +852,7 @@ def _message_class_str(self) -> str:
             cls_str += self._import_from_custom_types_module()
         else:
             cls_str += "\n{}\n".format(self._import_from_custom_types_module())
+        cls_str += "\nlogger = logging.getLogger(__name__)"
```
DavidMinarsch: @marcofavorito this is an issue I believe as the logger will have to be given the correct name. Can you advise?

marcofavorito: Yes, maybe set the name as:
```
packages.protocols.<protocol_name>.serializatino
```

5A11: So the generator should produce this line?
logger = logging.getLogger(packages.protocols.<protocol_name>.serialization)

Note, this logger is for the message class though.

DavidMinarsch: @marcofavorito it should be the following?:
```
logger = logging.getLogger(packages.<author>.protocols.<protocol_name>.message)
```

marcofavorito: so that means the protocols generated by the tool are vendorized by convention?

marcofavorito: No, my bad, we now have `author` in every import path, regardless of vendorized or not.

aea/protocols/generator.py
5A11: Changed the message. Done

aea/protocols/generator.py

aea/protocols/generator.py
```diff
@@ -128,6 +129,36 @@ def _get_indent_str(no_of_indents: int) -> str:
     return indents_str
 
 
+def _is_composition_type_with_custom_type(content_type: str) -> bool:
+    """
+    Evaluate whether the content_type is a composition type (FrozenSet, Tuple, Dict) and contains a custom type as a sub-type.
```
DavidMinarsch: docstring does not seem to match code?

DavidMinarsch: Optional, Union not mentioned

PR #971, opened 2020-03-23
Correctly split between semantic versioning and package versioning.

### LINKED ISSUE #931: Discuss SemVer vs PyPI compatible version in jsonschemas.

We are in a situation where there are different types of version numbers:
1. one used for dependency checks, in the `dependencies.version` field (as version specifier, see [PEP440](https://www.python.org/dev/peps/pep-0440/): https://github.com/fetchai/agents-aea/blob/master/aea/configurations/schemas/definitions.json#L29
2. one used for declaring the version of a package. This follows  [semver.org/](semver.org/) format. https://github.com/fetchai/agents-aea/blob/master/aea/configurations/schemas/definitions.json#L51
3. the other one is the `aea_version`, introduced by #924. This can be both a version specifiers (PEP440) or a semantic version number. The idea is that we reuse the comparison of PEP440 version numbers (using the `packaging` package). But, as stated in the Python proposal, [SemVer and PEP440 specifiers are not compatible in some cases.](https://www.python.org/dev/peps/pep-0440/#semantic-versioning)

I think this contrast might lead to confusion.

- `dependencies.version` must, of course, be compliant with PEP 440.
- `aea_version` must definitely compliant with PEP 440, since `aea` is a PyPI package (so far?). However, we support the simple notation without `==`, so we have to extend the regular expression to support that case (which is actually the most common).
- a package/aea-project `version` should follow SemVer, since they are not necessarily coupled with the Python stack.

**Additional context**

marcofavorito:
## Proposed changes

Both at JSON schema level and code level, now we have a clear separation of the version numbers.
In particular:
- dependencies.version is compliant with PEP 440 version specifiers
- aea_version is compliant either with PEP 440 version specifiers or one version number (as recommended by PEP 440)
- version is compliant with the SemVer format, since an AEA package is not coupled with Python.

## Fixes

Fixes  #931

## Types of changes

## Checklist

## Further comments
N/A

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -100,7 +100,7 @@ base_deps = [
     "base58>=1.0.3",
     "jsonschema>=3.0.0",
     "packaging>=20.3",
-    "semver",
+    "semver>=2.9.1",
     "protobuf",
     "pyyaml>=4.2b1",
     "watchdog",
--- a/setup.py
+++ b/setup.py
@@ -100,6 +100,7 @@ base_deps = [
     "base58>=1.0.3",
     "jsonschema>=3.0.0",
     "packaging>=20.3",
+    "semver",
     "protobuf",
     "pyyaml>=4.2b1",
     "watchdog",
--- a/aea/configurations/base.py
+++ b/aea/configurations/base.py
@@ -21,11 +21,25 @@
 import re
 from abc import ABC, abstractmethod
 from enum import Enum
-from typing import Dict, Generic, List, Optional, Set, Tuple, TypeVar, Union, cast
+from typing import (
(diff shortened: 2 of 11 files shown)


setup.py
```diff
@@ -100,6 +100,7 @@ def get_all_extras() -> Dict:
     "base58>=1.0.3",
     "jsonschema>=3.0.0",
     "packaging>=20.3",
+    "semver",
```
DavidMinarsch: want to set a lower boung?

marcofavorito: ok

PR #989, opened 2020-03-27
Compare test protocol package with latest generator

5A11:
WIP Added a test that compares the test protocol package with the output of the latest generator.

Various bug fixes.

This is a WIP PR. There is a broken test in the test_geenrate.py file, which is worked on.

## Types of changes

## Checklist

### DIFF
--- a/tests/test_protocols/test_generator.py
+++ b/tests/test_protocols/test_generator.py
@@ -83,9 +83,9 @@ PORT = 10000
 class TestEndToEndGenerator:
     """Test that the generating a protocol works correctly in correct preconditions."""
 
-    # @pytest.fixture(autouse=True)
-    # def _start_oef_node(self, network_node):
-    #     """Start an oef node."""
+    @pytest.fixture(autouse=True)
+    def _start_oef_node(self, network_node):
+        """Start an oef node."""
 
     @classmethod
     def setup_class(cls):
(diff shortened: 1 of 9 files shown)


tests/test_protocols/test_generator.py
```diff
@@ -86,11 +91,104 @@ def _start_oef_node(self, network_node):
...
         cls.runner = CliRunner()
-        cls.agent_name = "agent_1"
         cls.cwd = os.getcwd()
         cls.t = tempfile.mkdtemp()
         os.chdir(cls.t)
 
+    def test_compare_latest_generator_output_with_test_protocol(self):
+        """Test that the "t_protocol" test protocol matches with what the latest generator generates based on the specification."""
+        # check protoc is installed
+        res = shutil.which("protoc")
```
DavidMinarsch: @5A11 why are you not just calling the cli command to generate the protocol? It would save you code duplication. The below will easily diverge from the generator command logic and is therefore not recommended.

5A11: If I do it using generate cli command, then I cannot supply the `path_to_protocol_package` argument so the generated protocol package would not quite match the actual test protocol due to import path differences.

Unless I open up that argument to the generate cli command as well, which means whenever you want to generate, you have to also give it the `path_to_protocol_package` argument which does not really makes sense and is not at all a clean solution.

DavidMinarsch: Making the `path_to_protocol_package` an optional argument in the cli `aea generate protocol` command solves the problem.

PR #1011, opened 2020-03-30
AEA-472 Add URI to envelop serialisation

5A11:
## Proposed changes

This PR adds URI field to envelop serialisation.

## Types of changes

## Checklist


aea/mail/base.proto
```diff
@@ -7,4 +7,6 @@ message Envelope{
     string sender = 2;
     string protocol_id = 3;
     bytes message = 4;
+    string uri = 5;
+    bool uri_is_set = 6;
```
DavidMinarsch: Do we need this 6th field? A raw uri cannot be an empty string by definition, so the default "" would mean the uri is not set.

5A11: In general yes; same issue as we discussed exists in proto3. 
But you are right. Since in this case the default value, i.e. empty string, in fact means the field is not set, we can drop it.

aea/mail/base.py
```diff
@@ -215,6 +215,11 @@ def encode(self, envelope: "Envelope") -> bytes:
         envelope_pb.sender = envelope.sender
         envelope_pb.protocol_id = str(envelope.protocol_id)
         envelope_pb.message = envelope.message
+        if envelope.context is not None:
+            # envelope_pb.uri_is_set = True
+            envelope_pb.uri = envelope.context.uri_raw
+        # else:
```
DavidMinarsch: Please remove the comments

5A11: Oh, sorry it was overlooked. Done

PR #1023, opened 2020-04-01
auto generate api docs and fix accordingly

DavidMinarsch:
## Proposed changes

This PR updates the api docs and makes small adjustments to the code base where necessary.

## Fixes

-

## Types of changes

## Checklist

## Further comments

-


aea/aea.py
```diff
@@ -34,12 +35,15 @@
 from aea.protocols.default.message import DefaultMessage
 from aea.registries.filter import Filter
 from aea.registries.resources import Resources
-from aea.skills.error import ERROR_SKILL_ID
 from aea.skills.error.handlers import ErrorHandler
 from aea.skills.tasks import TaskManager
 
 logger = logging.getLogger(__name__)
 
+ERROR_SKILL_ID = PublicId(
```
DavidMinarsch, line 43: There was a circular import issue with this. Now looking at it more carefully, I think we need to make this configurable anyway. The user should be forced to specify the error handler (by default we use ours).

aea/connections/base.py
```diff
@@ -34,9 +34,6 @@
     from aea.mail.base import Envelope, Address  # pragma: no cover
 
 
-logger = logging.getLogger(__name__)
```
DavidMinarsch, line 37: unused

aea/connections/scaffold/connection.py
```diff
@@ -19,14 +19,11 @@
 
 """Scaffold connection and channel."""
 
-import logging
 from typing import Optional
 
 from aea.connections.base import Connection
 from aea.mail.base import Envelope
 
-logger = logging.getLogger(__name__)
```
DavidMinarsch, line 28: unused

aea/skills/error/__init__.py
```diff
@@ -18,6 +18,3 @@
 # ------------------------------------------------------------------------------
 
 """This module contains the implementation of the error skill."""
-from aea.configurations.base import PublicId
```
DavidMinarsch, line 21: this caused issues

DavidMinarsch: It is also not particularly neat, as the constant easily can get out of line with the config file.

DavidMinarsch: Ideally, the public id would be constructed from the config file.

aea/skills/error/handlers.py
```diff
@@ -29,8 +29,6 @@
 from aea.protocols.default.serialization import DefaultSerializer
 from aea.skills.base import Handler
 
-logger = logging.getLogger(__name__)
```
DavidMinarsch, line 32: moved to context based logger

aea/aea.py
marcofavorito: Ok, but maybe not. The error skill should be the same for every agent (although we can't enforce that anyway).

aea/skills/error/__init__.py
marcofavorito: yes, it is how it works now. The public id is retrieved from the configuration object.

Issue #961, opened 2020-03-22
Forbid some author names


marcofavorito:
**Is your feature request related to a problem? Please describe.**

Some author names should be forbidden because they might cause conflicts in the import paths of the vendorized modules.

In particular, a generic import path for a vendorized package is:
```python
import packages.author_name.component_type_plural.component_name
```
Whereas, for a custom package:
```python
import packages.component_type_plural.component_name
```

Now, consider the following (corner) case:
- a vendorized skill whose name is `skills` and whose author name is `skills`
- a custom skill with the same name, `skills` (the author name doesn't matter)
```python
import packages.skills.skills # this is the import for the custom skill
import packages.skills.skills.skills # this is the import for the vendorized skill
```
There is an evident conflict. The same applies to all the other component types.

**Describe the solution you'd like**

Some author names must be forbidden, like `skills`, `protocols` etc.

**Describe alternatives you've considered**

**Additional context**

PR #1025, opened 2020-04-01
init update to verify author

DavidMinarsch:
## Proposed changes

Fixes init to avoid forbidden author names

## Fixes

- https://github.com/fetchai/agents-aea/issues/961

## Types of changes

## Checklist

## Further comments

-

### DIFF
--- a/aea/cli/common.py
+++ b/aea/cli/common.py
@@ -74,6 +74,7 @@ DEFAULT_SKILL = PublicId.from_str("fetchai/error:" + DEFAULT_VERSION)
 DEFAULT_LEDGER = FETCHAI
 DEFAULT_REGISTRY_PATH = str(Path("./", "packages"))
 DEFAULT_LICENSE = "Apache-2.0"
+NOT_PERMITTED_AUTHORS = ["skills", "connections", "protocols", "contracts", "vendor"]
 
 
 from_string_to_type = dict(str=str, int=int, bool=bool, float=float)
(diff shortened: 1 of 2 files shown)


aea/cli/init.py
```diff
@@ -38,11 +42,56 @@
...
+                    PublicId.AUTHOR_REGEX
+                )
+            )
+        elif not _is_permitted_author_handle(author_prompt):
+            is_acceptable_author = False
+            click.echo(
+                "Not a permitted author handle. The following author handles are not allowed: {}".format(
+                    NOT_PERMITTED_AUTHORS
+                )
+            )
```
panasevychol, line 82: I would recommend to have an `else:` statement here.

aea/cli/init.py
```diff
@@ -38,11 +42,56 @@
 from aea.configurations.base import PublicId
 
 
-def _registry_init(author):
-    username = author
+def _local_init(author: Optional[str] = None) -> str:
```
panasevychol, line 45: Maybe it's better to call it `_validate_author`?

aea/cli/init.py
DavidMinarsch: that would be unreachable

aea/cli/init.py
DavidMinarsch: for symmetry with _registry_init, i think it makes sense to keep that name, although i agree it would be a candidate too

aea/cli/common.py
```diff
@@ -74,6 +74,7 @@
 DEFAULT_LEDGER = FETCHAI
 DEFAULT_REGISTRY_PATH = str(Path("./", "packages"))
 DEFAULT_LICENSE = "Apache-2.0"
+NOT_PERMITTED_AUTHORS = ["skills", "connections", "protocols", "contracts", "vendor"]
```
DavidMinarsch, line 77: @marcofavorito anything else? (https://github.com/fetchai/agents-aea/issues/961)

PR #1060, opened 2020-04-06
[AEA-482] Fix stub connection

lrahmani:
### Changes
- Fix input file parsing: file was read as a whole and parsed as one line
  + otherwise produces and errors everytime there is more than one message in the file
- Fix truncating: after reading from input file, add seek call to move fd pointer to the start of the file
  + otherwise misses messages
- Fix output format: to allow another process to use the output file as its input file
  + needs cleaning
- Add file locking/unlocking:  avoids race conditions

### How to test
Examples at https://github.com/fetchai/agents-p2p#stub-connection


aea/connections/stub/connection.py
```diff
@@ -79,6 +84,25 @@ def _decode(e: bytes, separator: bytes = SEPARATOR):
 
     return Envelope(to=to, sender=sender, protocol_id=protocol_id, message=message)
 
+def _lock_fd(fd: IO[AnyStr]) -> bool:
```
DavidMinarsch, line 87: @lrahmani please don't use abbreviations like `fd` as they make it harder to read for other devs. Use full names where possible.

PR #1068, opened 2020-04-08
Feature/tac

DavidMinarsch:
## Proposed changes

This PR fixes the existing tac skills.

## Fixes

-

## Types of changes

## Checklist

## Further comments

-


aea/aea_builder.py
```diff
@@ -559,10 +559,22 @@ def _process_connection_ids(
             ]
         else:
             selected_connections_ids = [
-                k.public_id for k in self._package_dependency_manager.connections.keys()
+                component_id.public_id
+                for component_id in self._package_dependency_manager.connections.keys()
             ]
 
-        return selected_connections_ids
+        # sort default id to be first
```
DavidMinarsch, line 566: @marcofavorito note, you forgot to ensure that the default connection is at index 0. I added this now.

aea/decision_maker/base.py
```diff
@@ -467,19 +486,14 @@ def _split_tx_fees(tx_fee: int) -> Dict[str, int]:
...
+        :param access_code: the access code to read from the queue
         """
         super().__init__()
-        self._permitted_caller = permitted_caller
-
-    @property
-    def permitted_caller(self) -> "DecisionMaker":
-        """Get the permitted caller."""
-        return self._permitted_caller
+        self._access_code_hash = _hash(access_code)
```
DavidMinarsch, line 496: Note, we now use a hash here, the previous implementation meant the caller could be picked from the object and passed back in. Now one needs the access code which only the DM has. We could do something similar for the preferences and ownership states. Thoughts?

packages/fetchai/connections/oef/connection.py
```diff
@@ -534,7 +534,7 @@ def on_dialogue_error(
         envelope = Envelope(
             to=self.address,
             sender=DEFAULT_OEF,
-            protocol_id=OefSearchMessage.protocol_id,
+            protocol_id=DefaultMessage.protocol_id,
```
DavidMinarsch, line 537: So technically, we now need to bump the version number of the oef connection. @marcofavorito should we bump it with each release which implements fixes? It is a lot of bumping potentially but also the cleanest way.

marcofavorito: yes I think that's a good approach.

aea/aea_builder.py
marcofavorito: so that means the default connection is always run, even if I specify the `--connections` option?

aea/decision_maker/base.py
marcofavorito: OK although It is not true isolation, as it can be accessed anyway from the outside of the object. Also the `protected_get` can be circumvented.

Also `Inbox`/`Outbox`/`Multiplexer` have the same issue.

aea/aea_builder.py
DavidMinarsch: Yes, it seems problematic to not require the `default_connection` to be present even when the flag is used and I don't see what's the benefit in supporting such a "misconfiguration"

aea/decision_maker/base.py
DavidMinarsch: I agree, not ideal.

docs/tac-skills.md
```diff
@@ -64,8 +65,8 @@ You must set the start time to a point in the future `start_time: 12 11 2019  15
 Alternatively, use the command line to get and set the start time:
 
 ``` bash
-aea config get skills.tac_control.models.parameters.args.start_time
-aea config set skills.tac_control.models.parameters.args.start_time '21 12 2019  07:14'
+aea config get vendor.fetchai.skills.tac_control.models.parameters.args.start_time
+aea config set vendor.fetchai.skills.tac_control.models.parameters.args.start_time '01 01 2020  00:01'
 ```
 
 ### Run the TAC controller AEA
```
marcofavorito, line 72: the line below makes the tutorial to fail with the following error:
```error: Default connection not a dependency. Please add it and retry.```

marcofavorito: with the new semantics, we probably need to add:
```aea config set agent.default_connection fetchai/oef:0.1.0```

... not only here, but in all the tutorials that use the oef connection. We should change the `packages/fetchai/agents/*/aea-config.yaml` as well

DavidMinarsch: Done, thanks

PR #1074, opened 2020-04-09
[AEA-529] Feature - Add p2p stub

lrahmani:
### Changes

Implement a p2p stub connection based on the regular stub connection.
This will be useful for testing p2p protocols locally.

### Missing/TOFIXs

- generate fingerprint for yaml configuration and __init__ files

## How to test

Example https://github.com/fetchai/agents-p2p/tree/ec7d5c6c16b078410b22039060e3fa944ccaa863#p2p-stub-connection


aea/connections/p2p/stub/__init__.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# ------------------------------------------------------------------------------
+#
+#   Copyright 2018-2020 Fetch.AI Limited
+#
```
DavidMinarsch, line 5: @lrahmani the folder structure in the AEA framework needs to be strictly adhered to for it to work. Hence, you need to create  a folder `stub_p2p` in the `connections dir which contains the `connection.py` and connection.yaml` files.

aea/connections/p2p/stub/connection.yaml
```diff
@@ -0,0 +1,20 @@
+name: stub
+author: fetchai
+version: 0.1.0
+description: The stub p2p connection implements a local p2p connection allowing agents
+  to communicate with each other through files created in the namespace directory.
+license: Apache-2.0
+aea_version: '>=0.3.0, <0.4.0'
+# TOFIX(LR) which algo is used to compute fingerprints?
```
DavidMinarsch, line 8: you can use `aea fingerprint` You should also add the file to `scripts/generate_ipfs_hashes.py` and run that

aea/connections/p2p/stub/connection.yaml
```diff
@@ -0,0 +1,20 @@
+name: stub
```
DavidMinarsch, line 1: the name needs to match the folder, i.e. `stub_p2p`

aea/connections/p2p/stub/connection.py
```diff
@@ -0,0 +1,123 @@
...
+import logging
+import os
+import tempfile
+from pathlib import Path
+from typing import Union
+
+from aea.configurations.base import ConnectionConfig, PublicId
+from aea.connections.base import Connection
+from aea.connections.stub.connection import (
+    StubConnection,
```
DavidMinarsch, line 31: @lrahmani at the moment the framework does not support having one connection depend on another. In this particular case it will work though as the stub connection is distributed in the framework

aea/connections/stub/connection.py
```diff
@@ -20,10 +20,11 @@
 """This module contains the stub connection."""
 
 import asyncio
+import fcntl
 import logging
 import os
```
DavidMinarsch, line 25: Note, since you are changing the stub connection code, you will have to bump the stub connection version in its yaml file to 0.2.0. Also all references in the tests and the docs to the stub connection need to be bumped. I can do that once the PR is in a clean state.

aea/connections/p2p/stub/__init__.py
DavidMinarsch: Also, we only allow for the default stub connection here. This package needs to move to `packages` dir so the user imports it. Will explain.

PR #1075, opened 2020-04-09
[AEA-481 AEA-484] Feature - add p2p noise connection

lrahmani:
## Changes

Add a P2P Noise Connection that implement an AEA Connection that deploys an noise node (go code/binary) and exchanges envelopes through named pipes.

## Missing/TOFIXs

- detect that noise node has disconnected
- run a fully featured noise node
- verify thread safety
- check if connection integrates well with AEA multiplexer
- testing

## How to test

Note: for pr history check https://github.com/fetchai/agents-aea/pull/1075


aea/connections/p2p/noise/__init__.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# ------------------------------------------------------------------------------
+#
```
DavidMinarsch, line 3: This cannot be added here, It needs to be under `packages` directory. I will explain you how it works in the call

PR #1076, opened 2020-04-09
[AEA-482] Fix stub connection

lrahmani:
### Changes
- Fix input file parsing: file was read as a whole and parsed as one line
  + otherwise produces and errors everytime there is more than one message in the file
- Fix truncating: after reading from input file, add seek call to move fd pointer to the start of the file
  + otherwise misses messages
- Fix output format: to allow another process to use the output file as its input file
  + needs cleaning
- Add file locking/unlocking:  avoids race conditions

### How to test
Examples at https://github.com/fetchai/agents-p2p#stub-connection

Note: for PR history check https://github.com/fetchai/agents-aea/pull/1060

### flaky tests 
- https://github.com/fetchai/agents-aea/pull/1076/checks?check_run_id=604898661#step:6:11337


aea/connections/stub/connection.py
```diff
@@ -79,11 +80,33 @@ def _decode(e: bytes, separator: bytes = SEPARATOR):
     to = split[0].decode("utf-8").strip()
     sender = split[1].decode("utf-8").strip()
     protocol_id = PublicId.from_str(split[2].decode("utf-8").strip())
-    message = split[3].decode("unicode_escape").encode("utf-8")
+    message = split[3]
 
     return Envelope(to=to, sender=sender, protocol_id=protocol_id, message=message)
 
 
+def _lock_file(fd: IO[AnyStr]) -> bool:
```
marcofavorito, line 88: I strongly suggest using context managers for resource acquisition. It is a pretty standard idiom in Python (e.g. like `with some_lock`)

https://docs.python.org/3/library/contextlib.html
https://docs.python.org/3/library/threading.html#using-locks-conditions-and-semaphores-in-the-with-statement

aea/connections/stub/connection.py
```diff
@@ -145,15 +168,47 @@ def __init__(
...
         """Receive new envelopes, if any."""
-        line = self.input_file.readline()
-        logger.debug("read line: {!r}".format(line))
-        lines = b""
-        while len(line) > 0:
-            lines += line
-            line = self.input_file.readline()
-        if lines != b"":
-            self._process_line(lines)
+        ok = _lock_file(self.input_file)
```
marcofavorito, line 171: see above. I'd like to see something like:

`with lock_file(self.input_file):`

aea/connections/stub/connection.py
```diff
@@ -145,15 +168,47 @@ def __init__(
...
+        if not ok:
+            # TOFIX(LR) how to handle locking errors
```
marcofavorito, line 173: error handling will be done in the context manager

aea/connections/stub/connection.py
```diff
@@ -145,15 +168,47 @@ def __init__(
...
+            return
+
+        lines = self.input_file.readlines()
+        if len(lines) > 0:
             self.input_file.truncate(0)
+            self.input_file.seek(0)
+
+        ok = _unlock_file(self.input_file)
```
marcofavorito, line 181: see above

aea/connections/stub/connection.py
```diff
@@ -223,10 +278,17 @@ async def send(self, envelope: Envelope):
...
         encoded_envelope = _encode(envelope, separator=SEPARATOR)
         logger.debug("write {}".format(encoded_envelope))
+        ok = _lock_file(self.output_file)
         self.output_file.write(encoded_envelope)
         self.output_file.flush()
+        ok = _unlock_file(self.output_file)
+        # TOFIX(LR) handle (un)locking errors.
+        #  should exception be propagated?
+        if not ok:
+            logger.error("while locking/unlocking file")
```
marcofavorito, line 291: see above on resource acquisition

aea/connections/stub/connection.py
```diff
@@ -145,15 +168,47 @@ def __init__(
...
+                if split == b"\n" or split == b"":
+                    parts = 0
+                    continue
+            message += split + separator
+            if parts == 1:
+                message = message.decode("utf-8").strip().encode("utf-8")
+
+        if 5 > parts > 0:
+            logger.debug("processing: {!r}".format(message))
+            self._process_line(message)
```
marcofavorito, line 211: the fact we use `\n` as message separator is a bit unfortunate since it can occur within the message fields.

Is there a way to improve the parsing? Maybe instead of `readlines()` we could use `read()` and then split with `separator`. 

Please take the following code as a suggestion or a source of inspiration. No action required, I don't want to add non-necessary work:

```python
# assuming 'content' is the content of the file (notice: it's not readlines()):
#    content = input_file.read()  
# The following is a regex over bytes: ".*,.*,.*,.*,"
regex = re.compile((b".*" + SEPARATOR) * 4, re.DOTALL)
iterator = re.finditer(regex, content)
for match in iterator:
    # remove trailing comma?
    message = match.group(0)[:-1]
    logger.debug("processing: {!r}".format(message))
    self._process_line(message)

```

lrahmani: Absolutely we can.
Indeed, if we use `,` as a separator the parsing will be straight forward.
Though, I would rather say that the issue here is that we have to allow adding a trailing `\n` as we want to support people `echo`ing to the input file.
Right, regex is the implementation to go with.

PR #1078, opened 2020-04-09
[AEA-481 AEA-484 AEA-485 AEA-550] Feature - add p2p noise connection

lrahmani:
## Changes

Add a P2P Noise Connection that implement an AEA Connection that deploys an noise node (go code/binary) and exchanges envelopes through named pipes.

## Missing/TOFIXs
- To do

## How to test
1. `git fetch && git checkout feature/add-p2p-noise`
2. `python3 -m pip install -e .[all]`
3. [`aea init &&`]`aea create my_agent && cd my_agent`
4. `aea add --local connection fetchai/p2p_noise:0.1.0`
5. `aea remove connection fetchai/stub:0.2.0`
6. `aea config set agent.default_connection fetchai/p2p_noise:0.1.0`
7. `aea install`
8. `aea run`

- Noise node logs are saved to "noise_node.log" file by default

Note: 
- for code history: https://github.com/fetchai/agents-p2p/tree/251ffaacafe4eba264df093d19d3357c1c852dc0/pocs/p2p_noise_pipe
- for pr history check https://github.com/fetchai/agents-aea/pull/1075


packages/fetchai/connections/p2p_noise/connection.yaml
```diff
@@ -0,0 +1,23 @@
...
+license: Apache-2.0
+aea_version: '>=0.3.0, <0.4.0'
+fingerprint:
+  __init__.py: QmbPzrjd27coFfS2vN9xDL4uARKZWCCmtWvmEuzGKSjxf7
+  connection.py: QmeTeiceuNVLBP1nKDPXLiUsWPTt18xEfh1wCsyFi3iQ8j
+fingerprint_ignore_patterns: []
+protocols: []
+class_name: P2PNoiseConnection
+config:
+  entry_addr: 192.168.1.200
```
DavidMinarsch, line 16: For all of these, can we have syntax like this `${OEF_ADDR:127.0.0.1}` so environment vars can be used as well?

lrahmani: so, does this means use OEF_ADDR env var if set, else use 127.0.0.1?

packages/fetchai/connections/p2p_noise/connection.yaml
```diff
@@ -0,0 +1,27 @@
...
+author: fetchai
+version: 0.1.0
+description: The p2p noise connection implements an interface to standalone golang
+  noise node that can exchange aea envelopes with other agents participating in the
+  same p2p network.
+license: Apache-2.0
+aea_version: '>=0.3.0, <0.4.0'
+fingerprint:
+  __init__.py: QmbPzrjd27coFfS2vN9xDL4uARKZWCCmtWvmEuzGKSjxf7
+  aea/api.go: QmYYRrMemRA1x9UsSg2VrofprHyH5oZv9S3FrF6G9jzCC1
```
DavidMinarsch, line 11: minor detail, but why not `aea_api.go`? More importantly, I assume we expect `aea/api.go` and `noise_node.go` to be reusable by other libraries and applications?

packages/fetchai/connections/p2p_noise/connection.yaml
```diff
@@ -0,0 +1,27 @@
...
+  connection.py: QmaayZzvz5WY6D5SD8MDwqRoscxrxB8fQfUDy3S8JG84Eu
+  noise_node.go: QmUf63vTTndzgqDeogsQ3fEERtYJ5hQotgiWgb7rtqMSzZ
+  publickey.py: QmPyfGWQJAhHV1SQoNiCkS5uDhFXSJfPrPe7igQ1uxcPsa
+fingerprint_ignore_patterns: []
+protocols: []
+class_name: P2PNoiseConnection
+config:
+  entry_peers:
```
DavidMinarsch, line 19: entry_peers is maybe not very explanatory? How about seed_nodes or something like that?

packages/fetchai/connections/p2p_noise/publickey.py
```diff
@@ -0,0 +1,71 @@
+import os
+import sys
+from typing import Optional
+
+import nacl.signing
+import nacl.encoding
+
+
```
DavidMinarsch, line 8: Let's discuss on a call.

packages/fetchai/connections/p2p_noise/noise_node.go
```diff
@@ -0,0 +1,188 @@
...
+	// Wait until Ctrl+C or a termination call is done.
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, os.Interrupt)
+	<-c
+
+	// Empty println.
+	println()
+}
+
+// Deliver an envelope
```
DavidMinarsch, line 102: Maybe to clarify directionality: `// Deliver an envelope from agent to receiver peer`

packages/fetchai/connections/p2p_noise/noise_node.go
```diff
@@ -0,0 +1,188 @@
...
+	<-c
+
+	// Empty println.
+	println()
+}
+
+// Deliver an envelope 
+func  send(envel aea.Envelope, node *noise.Node, overlay *kademlia.Protocol) error {
+    //fmt.Printf("[noise-p2p][debug] Looking for %s...\n", envel.To)
+    discover(overlay)
```
DavidMinarsch, line 105: Is this required? Already on 81.

packages/fetchai/connections/p2p_noise/noise_node.go
```diff
@@ -0,0 +1,188 @@
...
+		check(bootstrap(node, agent.EntryUris()...))
+	}
+
+	// Attempt to discover peers if we are bootstrapped to any nodes.
+	discover(overlay)
+
+	// Once overlay setup, connect to agent
+	check(agent.Connect())
+
+	// Receive envelopes from agent
```
DavidMinarsch, line 86: maybe slightly more explicit: `// Receive envelopes from agent and forward to peer`

packages/fetchai/connections/p2p_noise/noise_node.go
```diff
@@ -0,0 +1,188 @@
...
+    	}
+  	}()
+
+	// Wait until Ctrl+C or a termination call is done.
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, os.Interrupt)
+	<-c
+
+	// Empty println.
+	println()
```
DavidMinarsch, line 99: ?

packages/fetchai/connections/p2p_noise/noise_node.go
```diff
@@ -0,0 +1,188 @@
...
+
+	// Release resources associated to node at the end of the program.
+	defer node.Close()
+
+	// Register Envelope message
+	node.RegisterMessage(aea.Envelope{}, aea.UnmarshalEnvelope)
+
+	// Register a message handler to the node.
+	node.Handle(func (ctx noise.HandlerContext) error {
+    	return handle(ctx, agent)
```
DavidMinarsch, line 51: (minor) > indentation

packages/fetchai/connections/p2p_noise/noise_node.go
```diff
@@ -0,0 +1,188 @@
...
+			envel.To,
+			err,
+		)
+		return errors.New("Failed to send message")
+	}
+
+	return nil
+}
+
+// Handle envelope from other peers
```
DavidMinarsch, line 136: maybe: `// Handle envelope from other peers for agent`

packages/fetchai/connections/p2p_noise/noise_node.go
DavidMinarsch: minor: would be nice to apply linters > https://github.com/golangci/awesome-go-linters

packages/fetchai/connections/p2p_noise/connection.py
```diff
@@ -0,0 +1,416 @@
...
+    cmd = ["go", "get", "-v", "-d", "."]
+
+    try:
+        logger.debug(cmd, loop)
+        proc = await asyncio.create_subprocess_exec(
+            *cmd, cwd=os.path.dirname(src), loop=loop
+        )
+    except Exception as e:
+        logger.error("While executing go get : {}".format(str(e)))
+        raise e
```
DavidMinarsch, line 61: on exception, are we sure proc is not becoming a zombie proc? I.e. do we need a finally?

packages/fetchai/connections/p2p_noise/connection.py
```diff
@@ -0,0 +1,416 @@
...
+
+def _golang_get_deps(src: str):
+    cmd = ["go", "get", "-v", "-d", "."]
+
+    try:
+        logger.debug(cmd)
+        proc = subprocess.Popen(cmd, cwd=os.path.dirname(src))
+    except Exception as e:
+        logger.error("While executing go get : {}".format(str(e)))
+        raise e
```
DavidMinarsch, line 74: see above

packages/fetchai/connections/p2p_noise/connection.py
```diff
@@ -0,0 +1,416 @@
...
+            *cmd, cwd=os.path.dirname(src), loop=loop
+        )
+    except Exception as e:
+        logger.error("While executing go get : {}".format(str(e)))
+        raise e
+
+    return proc
+
+
+def _golang_get_deps(src: str):
```
DavidMinarsch, line 66: missing return signature

DavidMinarsch: missing docstring

packages/fetchai/connections/p2p_noise/connection.py
```diff
@@ -0,0 +1,416 @@
...
+
+
+NOISE_NODE_SOURCE = os.path.join(
+    os.path.abspath(os.path.dirname(__file__)), "noise_node.go"
+)
+NOISE_NODE_CLARGS = list()
+
+
+# TOFIX(LR) error: Cannot add child handler, the child watcher does not have a loop attached
+async def _async_golang_get_deps(src: str, loop: AbstractEventLoop):
```
DavidMinarsch, line 51: missing return signature, missing docstring

packages/fetchai/connections/p2p_noise/connection.py
```diff
@@ -0,0 +1,416 @@
...
+        proc = subprocess.Popen(cmd, cwd=os.path.dirname(src))
+    except Exception as e:
+        logger.error("While executing go get : {}".format(str(e)))
+        raise e
+
+    return proc
+
+
+# TOFIX(LR) add typing
+def _golang_run(src: str, args=[], env={}):
```
DavidMinarsch: see above, signature and docstring

packages/fetchai/connections/p2p_noise/connection.py
```diff
@@ -0,0 +1,416 @@
...
+    cmd = ["go", "run", src]
+
+    cmd.extend(args)
+
+    try:
+        logger.debug(cmd)
+        proc = subprocess.Popen(cmd, env=env)
+    except Exception as e:
+        logger.error("While executing go run {} {} : {}".format(src, args, str(e)))
+        raise e
```
DavidMinarsch, line 91: see above

packages/fetchai/connections/p2p_noise/connection.py
```diff
@@ -0,0 +1,416 @@
...
+
+    return proc
+
+
+class Uri:
```
DavidMinarsch, line 96: We have an actual `URI` class which follows `https://tools.ietf.org/html/rfc3986.html`. Could think about using that instead

DavidMinarsch: See `aea/mail/base.py`

packages/fetchai/connections/p2p_noise/connection.py
```diff
@@ -0,0 +1,416 @@
...
+        return self._addr
+
+    @property
+    def port(self) -> int:
+        return self._port
+
+
+# TOFIX(LR) NOT thread safe
+class NoiseNode:
+    r"""Noise p2p node as a subprocess with named pipes interface
```
DavidMinarsch, line 130: No `r` needed here.

packages/fetchai/connections/p2p_noise/connection.py
```diff
@@ -0,0 +1,416 @@
...
+        entry_peers: Sequence[Uri],
+        source: Union[Path, str],
+        clargs: Optional[List[str]] = [],
+    ):
+        """
+        Initialize a p2p noise node.
+
+        :param key: ec25519 curve private key.
+        :param uri: noise node ip address and port number in format ipaddress:port.
+        :param entry_peers: noise entry peers ip address and port numbers.
```
DavidMinarsch, line 147: incomplete params in docstring

packages/fetchai/connections/p2p_noise/connection.py
```diff
@@ -0,0 +1,416 @@
...
+        self.clargs = clargs
+
+        # named pipes (fifos)
+        self.noise_to_aea_path = "/tmp/{}-noise_to_aea".format(self.pub[:5])
+        self.aea_to_noise_path = "/tmp/{}-aea_to_noise".format(self.pub[:5])
+        self._noise_to_aea = None
+        self._aea_to_noise = None
+        self._connection_attempts = 10
+
+        #
```
DavidMinarsch, line 174: remove comment or add one

packages/fetchai/connections/p2p_noise/connection.py
```diff
@@ -0,0 +1,416 @@
...
+        if os.path.exists(in_path):
+            os.remove(in_path)
+        if os.path.exists(out_path):
+            os.remove(out_path)
+        os.mkfifo(in_path)
+        os.mkfifo(out_path)
+
+        # setup config
+        env = os.environ
+        env["ID"] = self.key + self.pub
```
DavidMinarsch, line 202: I am worried that these very generic Env variables might easily clash with existing ones. Can we have much more explicit ones, e.g. `AEA_P2P_...`

packages/fetchai/connections/p2p_noise/connection.py
```diff
@@ -0,0 +1,416 @@
...
+        env["AEA_TO_NOISE"] = out_path
+
+        # run node
+        self.proc = _golang_run(self.source, self.clargs, env)
+
+        await self._connect()
+
+    async def _connect(self) -> None:
+        if self._connection_attempts == 1:
+            logger.error("couldn't connect to noise p2p process")
```
DavidMinarsch, line 217: not sure we need the duplication in logger and exception?

packages/fetchai/connections/p2p_noise/connection.py
```diff
@@ -0,0 +1,416 @@
...
+            return None
+
+    def stop(self) -> None:
+        # TOFIX(LR) wait is blocking and proc can ignore terminate
+        self.proc.terminate()
+        self.proc.wait()
+
+
+class P2PNoiseConnection(Connection):
+    r"""A noise p2p node connection.
```
DavidMinarsch, line 303: see above

packages/fetchai/connections/p2p_noise/aea/api.go
```diff
@@ -0,0 +1,395 @@
...
+  return msgin_path, msgout_path, id, host, port, nil
+}
+
+
+func run_aea_sandbox(msgin_path string, msgout_path string) error {
+  // open pipe
+  msgout, erro := os.OpenFile(msgout_path, os.O_RDONLY, os.ModeNamedPipe)
+  msgin,  erri := os.OpenFile(msgin_path, os.O_WRONLY, os.ModeNamedPipe)
+  if erri != nil || erro != nil {
+    fmt.Println("[aea-api][error][sandbox] error while opening pipes:", erri, erro)
```
DavidMinarsch: `[sandbox]` bit can be dropped?

packages/fetchai/connections/p2p_noise/aea/api.go
```diff
@@ -78,14 +78,14 @@ func (aea *AeaApi) Init() error {
   uri := os.Getenv("URI")
 
   if aea.msgin_path == "" || aea.msgout_path == "" || aea.id == "" || uri == "" {
-    fmt.Println("[aea-api][error] couldn't get configuration")
+    fmt.Println("[aea-api  ][error] couldn't get configuration")
```
DavidMinarsch, line 81: don't understand why all the whitespace is added

packages/fetchai/connections/p2p_noise/connection.yaml
```diff
@@ -21,6 +21,7 @@ config:
   key_file: secret.txt
   noise_host: 127.0.0.1
   noise_port: 9999
+  noise_log_file: noise_node.log
```
DavidMinarsch, line 23: Great!

packages/fetchai/connections/p2p_noise/aea/api.go
```diff
@@ -0,0 +1,395 @@
...
+
+
+/*
+  
+  Protobuf generated Envelope - Edited
+
+*/
+
+// Code generated by protoc-gen-go. DO NOT EDIT.
+// source: pocs/p2p_noise_pipe/envelope.proto
```
marcofavorito, line 308: not sure my comment makes sense. I believe `api.go` has been written by you. However, the code below this comment seems to be autogenerated. I wonder if it would be possible to put this in a different file and then add it into the compilation, so to make updates of the package easier (instead of copying & pasting the generated code into this).

packages/fetchai/connections/p2p_noise/connection.yaml
```diff
@@ -0,0 +1,27 @@
...
+  same p2p network.
+license: Apache-2.0
+aea_version: '>=0.3.0, <0.4.0'
+fingerprint:
+  __init__.py: QmbPzrjd27coFfS2vN9xDL4uARKZWCCmtWvmEuzGKSjxf7
+  aea/api.go: QmdatRSHu3Nuz18NDMk9PGSPi34o1ATggYRX962nP6jSXm
+  connection.py: QmPi2M1uaexKUGjZxpSuoDTZ6J6wVYYCpEREp9TZYY3jwu
+  noise_node.go: QmP92aiWBd2LNgEbH1gSjr3H575c4XUPkZ2uZVTq3QgUWJ
+  publickey.py: QmPyfGWQJAhHV1SQoNiCkS5uDhFXSJfPrPe7igQ1uxcPsa
+fingerprint_ignore_patterns: []
```
marcofavorito, line 15: shouldn't we ignore the log file? Or other files generated from the `.go` source code?

packages/fetchai/connections/p2p_noise/noise_node.go
```diff
@@ -0,0 +1,193 @@
...
+// discover uses Kademlia to discover new peers from nodes we already are aware of.
+func discover(overlay *kademlia.Protocol) {
+	ids := overlay.Discover()
+
+	var str []string
+	for _, id := range ids {
+		str = append(str, fmt.Sprintf("%s(%s)", id.Address, id.ID.String()))
+	}
+
+	// TOFIX(LR) keeps printing already known peers
```
marcofavorito, line 187: I think we should decide a common way to define TODO tags and stick with it team-wide.

I use `TODO` because by default IDEs can detect them in comments, but that's not the same for other variants.

Notice, nothing to do now, just a thought I had in mind.

packages/fetchai/connections/p2p_noise/noise_node.go
```diff
@@ -0,0 +1,193 @@
...
+	if len(ids) > 0 {
+		//fmt.Printf("[noise-p2p][debug] Discovered %d peer(s): [%v]\n", len(ids), strings.Join(str, ", "))
+	} else {
+		//fmt.Printf("[noise-p2p][debug] Did not discover any peers.\n")
```
marcofavorito, line 191: is it correct that they are commented?

packages/fetchai/connections/p2p_noise/aea/api.go
```diff
@@ -0,0 +1,395 @@
...
+
+  // get config
+  aea.msgin_path  = os.Getenv("AEA_TO_NOISE")
+  aea.msgout_path = os.Getenv("NOISE_TO_AEA")
+  aea.id  = os.Getenv("ID")
+  entry_uris := os.Getenv("ENTRY_URI")
+  uri := os.Getenv("URI")
+
+  if aea.msgin_path == "" || aea.msgout_path == "" || aea.id == "" || uri == "" {
+    fmt.Println("[aea-api  ][error] couldn't get configuration")
```
marcofavorito, line 81: instead of just printing to stdout, would be useful to employ some Go logging library like https://golang.org/pkg/log/. That would let the developer have finer control over the messages printed by the package, at _configuration level_.

packages/fetchai/connections/p2p_noise/publickey.py
```diff
@@ -0,0 +1,71 @@
+import os
```
marcofavorito, line 1: There are a few things missing:
- encoding, shebang, copyright notice (copy & paste from here: https://github.com/fetchai/agents-aea/blob/master/setup.py#L1-L19)
- docstring that states the purpose of this module

packages/fetchai/connections/p2p_noise/publickey.py
```diff
@@ -0,0 +1,71 @@
...
+        return self._ed25519.encode(encoder=nacl.encoding.HexEncoder).decode("ascii")
+
+    def pub(self) -> PubKey:
+        return PubKey(naclkey=self._ed25519.verify_key)
+
+
+if __name__ == "__main__":
+    if len(sys.argv) < 3:
+        print("Usage: {} <swarm> <number>".format(sys.argv[0]), file=sys.stderr)
+        sys.exit(-1)
```
marcofavorito, line 53: this is OK. However, I'm not a fan of parsing argument by hand.

Check this for an example using `argparse`: https://github.com/fetchai/agents-aea/blob/master/scripts/parse_main_dependencies_from_lock.py#L28-L41

Docs of the standard library: https://docs.python.org/3/library/argparse.html

packages/fetchai/connections/p2p_noise/connection.py
```diff
@@ -0,0 +1,431 @@
...
+    cmd = ["go", "get", "-v", "-d", "."]
+
+    try:
+        logger.debug(cmd, loop)
+        proc = await asyncio.create_subprocess_exec(
+            *cmd, cwd=os.path.dirname(src), loop=loop
+        )
+    except Exception as e:
+        logger.error("While executing go get : {}".format(str(e)))
+        raise e
```
marcofavorito, line 61: a nice `TODO` is to put this into `aea.helpers`, because it might be useful for other code.

packages/fetchai/connections/p2p_noise/connection.py
marcofavorito: it's a private function. In theory it's legal for linters :-)

marcofavorito: but good to have for better readability.

packages/fetchai/connections/p2p_noise/connection.py
marcofavorito: also with the code below

packages/fetchai/connections/p2p_noise/connection.yaml
lrahmani: it is a workaround to keep only one package in the connection directory to be able to run `go get -d -v .`

packages/fetchai/connections/p2p_noise/connection.yaml
lrahmani: I don't know if seed_nodes is any better. For now I change it to `noise_entry_peers`

packages/fetchai/connections/p2p_noise/noise_node.go
lrahmani: 81 was run only once, hence the need to discover before a send.
Removed it, as now discover is run regularly in background

packages/fetchai/connections/p2p_noise/noise_node.go
lrahmani: run gofmt

packages/fetchai/connections/p2p_noise/connection.py
lrahmani: yes, no zombie proc, because no proc was created as there was an error while *starting* the process

packages/fetchai/connections/p2p_noise/connection.py
lrahmani: see above

packages/fetchai/connections/p2p_noise/connection.py
lrahmani: see above

packages/fetchai/connections/p2p_noise/connection.py
lrahmani: substituting `Uri` by `URI` is not straightforward.
Will do that in subsequent pr

packages/fetchai/connections/p2p_noise/aea/api.go
lrahmani: for `aea-api` logs to align with `noise-p2p`, a hack in the absence of a proper logger.
will be fixed in a subsequent pr

packages/fetchai/connections/p2p_noise/aea/api.go
lrahmani: This part is generated by protoc and *edited* by me to be able to use the struct as a message type for noise.

packages/fetchai/connections/p2p_noise/connection.yaml
lrahmani: The log file won't live in the package but rather in the agent root directory.
I am not aware of go source code generating other files.

packages/fetchai/connections/p2p_noise/noise_node.go
lrahmani: Personally, I use `TOFIX` for TODOs as well.
But I agree, a convention is better.

packages/fetchai/connections/p2p_noise/noise_node.go
lrahmani: yes, otherwise `discover` will pollute the node's logs.
In subsequent pr, I'll make `discover` ignore already discovered pr.
This is already marked as a `TOFIX`

packages/fetchai/connections/p2p_noise/aea/api.go
lrahmani: I agree. 
I was planning to use the default one (the one you point it out), but *due to the lack of warn/debug/trace levels* I kept printing to stdout until I find a better logger.

packages/fetchai/connections/p2p_noise/publickey.py
lrahmani: file deleted

packages/fetchai/connections/p2p_noise/publickey.py
lrahmani: file deleted

packages/fetchai/connections/p2p_noise/connection.py
lrahmani: sure.
However, this function raise an exception when run with python3.6.
I already  have a TOFIX to address it.

packages/fetchai/connections/p2p_noise/connection.py
DavidMinarsch: Ok no worries, just leave it then for now.

packages/fetchai/connections/p2p_noise/connection.py
```diff
@@ -0,0 +1,524 @@
...
+    async def read(self) -> Optional[bytes]:
+        assert (
+            self._stream_reader is not None
+        ), "StreamReader not set, call connect first!"
+        try:
+            logger.debug("Waiting for messages...")
+            size = await self._stream_reader.readexactly(4)
+            if not size:
+                return None
+            size = struct.unpack("!I", size)[0]
```
DavidMinarsch, line 353: size is bytes but readexactly expects int!

packages/fetchai/connections/p2p_noise/connection.py
```diff
@@ -0,0 +1,524 @@
...
+            # TOFIX(LR) use proper exception
+        self._connection_attempts -= 1
+
+        logger.debug(
+            "Attempt opening pipes {}, {}...".format(
+                self.noise_to_aea_path, self.aea_to_noise_path
+            )
+        )
+
+        self._noise_to_aea = posix.open(
```
DavidMinarsch, line 310: Please use `os` rather than `posix`: https://docs.python.org/2/library/posix.html

aea/aea_builder.py
```diff
@@ -604,6 +606,42 @@ def build(self, connection_ids: Optional[Collection[PublicId]] = None) -> AEA:
         self._load_and_add_skills(aea.context)
         return aea
 
+    # TODO: remove and replace with a clean approach (~noise based crypto module or similar)
+    def _update_identity(
```
DavidMinarsch, line 610: Note, @marcofavorito and @lrahmani this is a hack which only affects the case when the p2p_noise connection is used.

The problem with this connection is that its public key needs to be set as the agents default address. Currently, we do not support noise crypto and we are not sure we do want to. So this is a hack until a better solution is found.

aea/aea_builder.py
```diff
@@ -604,6 +606,42 @@ def build(self, connection_ids: Optional[Collection[PublicId]] = None) -> AEA:
...
+                    self._name,
+                    addresses={**wallet.addresses, **noise_addresses},
+                    default_address_key=p2p_noise_connection.noise_address_id,  # type: ignore
+                )
+            else:  # pragma: no cover
+                identity = Identity(self._name, address=p2p_noise_connection.noise_address)  # type: ignore
+            return identity
+        else:
+            logger.error(
+                "The p2p-noise connection can only be used as a single connection. "
```
DavidMinarsch, line 639: We cannot allow running p2p_noise with other connection atm as this would cause unexpected side effects due to the hack.

aea/connections/stub/connection.py
```diff
@@ -275,14 +275,6 @@ def from_config(
...
-            p.name for p in configuration.restricted_to_protocols
-        }
-        excluded_protocols_names = {p.name for p in configuration.excluded_protocols}
         return StubConnection(
-            input_file,
-            output_file,
-            connection_id=configuration.public_id,
-            restricted_to_protocols=restricted_to_protocols_names,
-            excluded_protocols=excluded_protocols_names,
+            input_file, output_file, address=address, configuration=configuration,
```
DavidMinarsch, line 279: Note, this was wrong, the configuration is needed to recover for instance the public. id

packages/fetchai/connections/p2p_noise/connection.py
```diff
@@ -266,8 +270,10 @@ async def start(self) -> None:
         # TOFIX(LR) async version
         # proc = await _async_golang_get_deps(self.source, loop=self._loop)
         # await proc.wait()
+        logger.info("Downloading goland dependencies. This may take a while...")
```
DavidMinarsch, line 273: @lrahmani I cannot reach this.

packages/fetchai/connections/p2p_noise/aea/api.go
```diff
@@ -69,13 +71,24 @@ func (aea *AeaApi) Init() error {
 	if aea.sandbox {
 		return nil
 	}
+	env_file := os.Args[1]
+	fmt.Println("[aea-api  ][debug] env_file:", env_file)
 
 	// get config
+	err := godotenv.Load(env_file)
```
DavidMinarsch, line 78: Using an environment file avoids the issues of dynamically set environment variables not being available in subshells etc.

packages/fetchai/connections/p2p_noise/noise_node.go
```diff
@@ -62,31 +64,35 @@ func main() {
...
+			}
+		}()
+	}
 
-	// Once overlay setup, connect to agent
-	check(agent.Connect())
+	// Have the node start listening for new peers.
+	// if len(agent.EntryUris()) > 0 {
+	check(node.Listen())
+	// }
```
DavidMinarsch: @marcofavorito  and @lrahmani this code is still not sound!
things to note:

- `check(node.Listen())` blocks

packages/fetchai/connections/p2p_noise/connection.py
DavidMinarsch: Resolved

PR #1080, opened 2020-04-09
AEA-511 Webhook connection

### LINKED ISSUE #862: Web hooks support for a connection
The HTTP connection does for now not support web hooks ("user-defined HTTP callbacks"). Web hooks require both to be `servers`. Add websocket for sending info to client? https://websockets.readthedocs.io/en/stable/intro.html

5A11:
## Proposed changes

Adds a web hook connection and associated web hook channel

WIP: The connection works (tested with running Aries agent that sends web hook notifications). What is missing is a generic test method that is independent of Aries.

Fixes #862

## Types of changes

## Checklist


packages/fetchai/connections/webhook/connection.py
```diff
@@ -0,0 +1,267 @@
...
+        webhook_port: int,
+        webhook_url_path: str,
+        connection_id: PublicId,
+    ):
+        """
+        Initialize a webhook channel.
+
+        :param agent_address: the address of the agent
+        :param webhook_address: webhook hostname / IP address
+        :param webhook_port: webhook port number
```
DavidMinarsch, line 63: missing webhook_url_path here

packages/fetchai/connections/webhook/connection.py
```diff
@@ -0,0 +1,267 @@
...
+            self.app = web.Application()
+            self.app.add_routes(
+                [web.post(self.webhook_url_path, self._receive_webhook)]
+            )
+            self.runner = web.AppRunner(self.app)
+            await self.runner.setup()
+            self.webhook_site = web.TCPSite(
+                self.runner, self.webhook_address, self.webhook_port
+            )
+            await self.webhook_site.start()
```
DavidMinarsch, line 101: Is this blocking? I hope not but hard to tell without looking at `aiohttp`

packages/fetchai/connections/webhook/connection.py
```diff
@@ -0,0 +1,267 @@
...
+        Disconnect.
+
+        Shut-off and cleanup the webhook site, the runner and the web app, then stop the channel.
+        """
+        assert (
+            self.webhook_site is not None
+        ), "Application not connected, call connect first!"
+
+        if not self.is_stopped:
+            logger.info("Webhook app is shutdown.")
```
DavidMinarsch, line 115: Should be moved to below line 120

packages/fetchai/connections/webhook/connection.py
```diff
@@ -0,0 +1,267 @@
...
+        Convert a webhook request object into an Envelope containing an HttpMessage (from the 'http' Protocol).
+
+        :param connection_id: the connection id
+        :param request: the webhook request
+        """
+
+        payload_bytes = await request.read()
+        version = str(request.version[0]) + "." + str(request.version[1])
+
+        context = EnvelopeContext(connection_id=connection_id)
```
DavidMinarsch, line 149: you don't need the `connection_id` here. the `connection_id` in the envelope context is just used for agent > multiplexer direction; here you could add the URI instead (`aea/mail/base.py`)

packages/fetchai/connections/webhook/connection.py
```diff
@@ -0,0 +1,267 @@
...
+            performative=HttpMessage.Performative.REQUEST,
+            method=request.method,
+            url=str(request.url),
+            version=version,
+            headers=json.dumps(dict(request.headers)),
+            bodyy=payload_bytes if payload_bytes is not None else b"",
+        )
+        envelope = Envelope(
+            to=self.agent_address,
+            sender=request.remote,
```
DavidMinarsch, line 160: What does `request.remote` return?

packages/fetchai/connections/webhook/connection.py
```diff
@@ -0,0 +1,267 @@
...
+
+        self.webhook_site = None  # type: Optional[web.TCPSite]
+        self.runner = None  # type: Optional[web.AppRunner]
+        self.app = None  # type: Optional[web.Application]
+
+        self.is_stopped = True
+
+        self.connection_id = connection_id
+        self.in_queue = None  # type: Optional[asyncio.Queue]  # pragma: no cover
+        self.loop = (
```
DavidMinarsch, line 80: note, you are not using this...you would want to run the `aiohttp` against that imo

packages/fetchai/connections/webhook/connection.yaml
```diff
@@ -0,0 +1,23 @@
...
+aea_version: '>=0.3.0, <0.4.0'
+fingerprint:
+  __init__.py: QmWUKSmXaBgGMvKgdmzKmMjCx43BnrfW6og2n3afNoAALq
+  connection.py: QmZKFwYYW3dLb8uY3V7Y8pc7B5t3Dcbxvh1pbqanGdnqB5
+fingerprint_ignore_patterns: []
+protocols:
+- fetchai/http:0.1.0
+class_name: WebhookConnection
+config:
+  host: ${addr:127.0.0.1}
```
DavidMinarsch, line 15: this doesn't match with the kwargs the connection expects

tests/test_packages/test_connections/test_webhook/test_webhook.py
```diff
@@ -0,0 +1,162 @@
...
+        """Test the disconnect functionality of the webhook connection."""
+        await self.webhook_connection.connect()
+        assert self.webhook_connection.connection_status.is_connected is True
+
+        await self.webhook_connection.disconnect()
+        assert self.webhook_connection.connection_status.is_connected is False
+
+
+@pytest.mark.asyncio
+async def test_webhook_receive():
```
DavidMinarsch: this can be commented as nothing useful is being tested

packages/fetchai/connections/webhook/connection.py
5A11: Done

packages/fetchai/connections/webhook/connection.py
5A11: Does not seem to be blocking. Looking at aiohttp, the `start` method is async and has its own loop. Also from examining the Aries code it, the call does not block.

packages/fetchai/connections/webhook/connection.py
5A11: Done

packages/fetchai/connections/webhook/connection.py
5A11: Added URI. 

What is the URI field? I couldn't find any documentation on it.

packages/fetchai/connections/webhook/connection.py
5A11: From docstring of the method:

"Remote IP of client initiated HTTP request"

packages/fetchai/connections/webhook/connection.py
5A11: For now I take it out. It appears all of the operations of aiohttp is async by default and it has its own loops. I mean the module IS for "Asynchronous HTTP Client/Server for asyncio and Python". If upon testing we realise it needs that I run aiohttp in that loop.

PR #1085, opened 2020-04-10
AEA-531 fix for default agent loop timeout on contruction with AEABuild

solarw:
## Proposed changes
small refactoring over Agent to make tests easier: now it's possible to spin agents main loop manually

## Fixes

set default agent sleep timeout in AEABuilder to reduce cpu utilization

## Types of changes

## Checklist

### DIFF
--- a/aea/agent.py
+++ b/aea/agent.py
@@ -199,6 +199,18 @@ class Agent(ABC):
 
         :return: None
         """
+        self._start_setup()
+        self._run_main_loop()
+
+    def _start_setup(self) -> None:
+        """
+        setup Agent on start:
+        - up multiplexer
+        - call agent.setup
+        - set started
+
+        :return: None
+        """
         if not self.is_debug:
             self.multiplexer.connect()
 
@@ -206,7 +218,6 @@ class Agent(ABC):
         self.setup()
 
         self.liveness.start()
-        self._run_main_loop()
 
     def _run_main_loop(self) -> None:
         """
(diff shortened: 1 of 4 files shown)


aea/agent.py
```diff
@@ -199,14 +199,20 @@ def start(self) -> None:
 
         :return: None
         """
+        self._start_setup()
+        self._run_main_loop()
+
+    def _start_setup(self) -> None:
```
DavidMinarsch, line 205: Great idea! Please make sure the doc string below is clean (correct indentation)

tests/common/utils.py
```diff
@@ -0,0 +1,50 @@
...
+    """
+        context manager to measure execution time of code in context
+
+        :return TimeItResult
+
+        example:
+        with timeit_context() as result:
+            do_long_code()
+        print("Long code takes ", result.time_passed)
+    """
```
DavidMinarsch, line 43:
```suggestion
    """
    Context manager to measure execution time of code in context

    :return TimeItResult

    example:
    with timeit_context() as result:
        do_long_code()
    print("Long code takes ", result.time_passed)
    """
```
Note the indentation, for consistency I suggest changing this and always start block comments on 4 spaces in.

tests/test_aeabuilder.py
```diff
@@ -0,0 +1,82 @@
...
+from aea.crypto.fetchai import FETCHAI
+
+from tests.common.utils import timeit_context
+
+from .conftest import CUR_PATH
+
+
+def test_default_timeout_for_agent():
+    """
+        Tests agents loop sleep timeout
```
DavidMinarsch, line 32: See above re indentation

tests/test_aeabuilder.py
```diff
@@ -0,0 +1,82 @@
...
+    agent = builder.build()
+    assert agent._timeout == builder.DEFAULT_AGENT_LOOP_TIMEOUT
+
+    with timeit_context() as time_result:
+        agent._spin_main_loop()
+
+    assert time_result.time_passed > builder.DEFAULT_AGENT_LOOP_TIMEOUT
+    time_0_05 = time_result.time_passed
+
+    """ Timeout == 0.001 """
```
DavidMinarsch, line 52: I think it might be better to have a test class and multiple test methods for the different test cases.

solarw: @DavidMinarsch 
at the moment three cases are bound with `assert time_0 < time_0_001 and time_0_001 < time_0_05`
probably this check can be removed and just check that `agent._timeout` set on agent correctly, without measuring sleep time.

PR #1098, opened 2020-04-14
Skills end-to-end tests refactoring

panasevychol:
## Proposed changes

Skills tests switched to test_tools usage:
- weather skills test
- carpark skills test

## Types of changes

## Checklist


aea/test_tools/config.py
```diff
@@ -0,0 +1,31 @@
...
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+# ------------------------------------------------------------------------------
+"""This module contains the config of the AEA test framework."""
+
+from aea.cli_gui import DEFAULT_AUTHOR
```
DavidMinarsch, line 21: @panasevychol I removed this from your previous PR as it creates an unnecessary duplication of the config imports. check out what's on develop. The only constants not covered by my amendment are `DEFAULT_PRIVATE...` which you should import from wherever it is declared in the configs, same for FETCHAI.

aea/test_tools/test_cases.py
```diff
@@ -24,18 +24,23 @@
...
 from aea.test_tools.click_testing import CliRunner
+from aea.test_tools.config import (
+    AUTHOR,
+    CLI_LOG_OPTION,
+    DEFAULT_PRIVATE_KEY_FILE_NAME,
+    FETCHAI_NAME,
+)
 from aea.test_tools.exceptions import AEATestingException
 
-CLI_LOG_OPTION = ["-v", "OFF"]
```
DavidMinarsch, line 37: Throughout AEA we always declare constants in the relevant files, the config. file approach creates additional verbosity.

aea/test_tools/test_cases.py
```diff
@@ -224,12 +252,71 @@ def run_install(self):
...
+    def generate_private_key(self) -> None:
+        """
+        Generate AEA private key with CLI command.
+        Run from agent's directory.
+
+        :return: None
+        """
+        self.run_cli_command("generate-key", FETCHAI_NAME)
+
+    def add_private_key(self) -> None:
```
DavidMinarsch, line 264: add-key should have kwargs: identifier: str, key_file_path: str  rather than hard coded

aea/test_tools/test_cases.py
```diff
@@ -224,12 +252,71 @@ def run_install(self):
         """
         self.run_cli_command("install")
 
+    def generate_private_key(self) -> None:
```
DavidMinarsch, line 255: see below, should be kwarg: identifier: str rather than hard coded

aea/test_tools/test_cases.py
```diff
@@ -224,12 +252,71 @@ def run_install(self):
...
+    def add_private_key(self) -> None:
+        """
+        Add private key with CLI command.
+        Run from agent's directory.
+
+        :return: None
+        """
+        self.run_cli_command("add-key", FETCHAI_NAME, DEFAULT_PRIVATE_KEY_FILE_NAME)
+
+    def generate_wealth(self) -> None:
```
DavidMinarsch, line 273: see above, no hard coding please

tests/test_packages/test_skills/test_carpark.py
```diff
@@ -19,151 +19,49 @@
...
-        cls.runner = CliRunner()
-        cls.agent_name_one = "my_carpark_aea"
-        cls.agent_name_two = "my_carpark_client_aea"
-        cls.cwd = os.getcwd()
-        cls.t = tempfile.mkdtemp()
-        os.chdir(cls.t)
-
     def test_carpark(self, pytestconfig):
         """Run the weather skills sequence."""
         if pytestconfig.getoption("ci"):
```
DavidMinarsch, line 36: This should be moved to the parent class Oeftestcase

tests/test_packages/test_skills/test_carpark.py
```diff
@@ -19,151 +19,49 @@
...
-        result = self.runner.invoke(
-            cli,
-            [*CLI_LOG_OPTION, "create", "--local", self.agent_name_two],
-            standalone_mode=False,
-        )
-        assert result.exit_code == 0
+
+        self.initialize_aea()
+
+        agent_name_one = "my_carpark_aea"
```
DavidMinarsch, line 41: it would be much more readable if the variable had a name like `capark_aea_name`

tests/test_packages/test_skills/test_carpark.py
```diff
@@ -19,151 +19,49 @@
...
-            ],
-            standalone_mode=False,
-        )
-        assert result.exit_code == 0
-
-        result = self.runner.invoke(
-            cli, [*CLI_LOG_OPTION, "install"], standalone_mode=False
-        )
-        assert result.exit_code == 0
+        self.add_item("connection", "fetchai/oef:0.1.0")
```
DavidMinarsch, line 48: I think it is now oef:0.2.0

tests/test_packages/test_skills/test_carpark.py
```diff
@@ -19,151 +19,49 @@
...
-        )
-        whole_file = whole_file.replace(
-            "db_rel_dir: ../temp_files", "# db_rel_dir: ../temp_files"
-        )
+        # Load the agent yaml file and manually insert the things we need (ledger APIs)
 
-        # close the file
-        file.close()
+        # TODO: remove this block and replace with next two commented lines
+        # when "aea config set" will be able to handle dictionaties and non-existing keys
```
DavidMinarsch, line 59: just write a helper method on the test case which lets you do this by loading and dumping the file

tests/test_packages/test_skills/test_carpark.py
```diff
@@ -182,52 +80,23 @@ def test_carpark(self, pytestconfig):
...
-        )
-        assert result.exit_code == 0
-
-        result = self.runner.invoke(
-            cli, [*CLI_LOG_OPTION, "install"], standalone_mode=False
-        )
-        assert result.exit_code == 0
-
-        # Load the agent yaml file and manually insert the things we need
+        self.add_item("connection", "fetchai/oef:0.1.0")
```
DavidMinarsch, line 89: see above

tests/test_packages/test_skills/test_carpark.py
```diff
@@ -182,52 +80,23 @@ def test_carpark(self, pytestconfig):
...
+        self.add_item("skill", "fetchai/carpark_client:0.1.0")
+        self.run_install()
+
+        # TODO: same as above. Remove this block and replace with next two commented lines
+        # when "aea config set" will be able to handle dictionaties and non-existing keys
```
DavidMinarsch, line 94: see above

tests/test_packages/test_skills/test_carpark.py
```diff
@@ -182,52 +80,23 @@ def test_carpark(self, pytestconfig):
...
+
+        # setting_path = "agent.ledger_apis.fetchai.network"
+        # self.set_config(setting_path, "testnet")
+
+        # agent config update block start
         file = open("aea-config.yaml", mode="r")
```
DavidMinarsch, line 100: This logic from lines 100-118 seems suitable for extraction to the parent class

DavidMinarsch: with find_text, replace_text as the kwargs, so they stay here and are fed to the replace method

tests/test_packages/test_skills/test_weather.py
```diff
@@ -20,251 +20,59 @@
...
-                "add",
-                "--local",
-                "skill",
-                "fetchai/weather_client:0.1.0",
-            ],
-            standalone_mode=False,
-        )
-        assert result.exit_code == 0
-
-        # Load the agent yaml file and manually insert the things we need
```
DavidMinarsch, line 183: you're missing this..

panasevychol: It's here.
https://github.com/fetchai/agents-aea/pull/1098/files#diff-a9f80421196c9acba163cbf08ef4947fR60

tests/test_packages/test_skills/test_carpark.py
panasevychol: Any idea how can we move it (in parent class)?
IMO good way to optimize this - to write a decorator and use it on every test we need to skip in CI.

aea/test_tools/test_cases.py
panasevychol: Didn't have a real use cases in tests yet. But I can do that. Will add those args with default values.

DavidMinarsch: You'll need to think a bit ahead as well. The code in the `aea/` should be reusable. Your code is too specific for no good reason. providing the args from the start means its versatile and you do not have to go back to it once a slightly different test case comes up.

tests/test_packages/test_skills/test_weather.py
DavidMinarsch: Cannot see it there.

panasevychol: Line 60 this file. Don't know why Github leads to other place.
https://github.com/fetchai/agents-aea/pull/1098/files/039c2e6064a9f9195b812e8bd37cce12281343fa#diff-a9f80421196c9acba163cbf08ef4947fR60

aea/cli/config.py
```diff
@@ -51,6 +51,89 @@
 FALSE_EQUIVALENTS = ["f", "false", "False"]
 
 
+def handle_dotted_path(value):
```
DavidMinarsch, line 54: Function signature is incomplete!

aea/cli/config.py
```diff
@@ -66,79 +149,20 @@ def convert(self, value, param, ctx):
...
-                )
+        try:
+            (
+                json_path,
+                path_to_resource_configuration,
+                config_loader,
+            ) = handle_dotted_path(value)
+        except Exception as e:
+            self.fail(str(e))
         else:
```
DavidMinarsch, line 160: why the else?

aea/test_tools/generic.py
```diff
@@ -67,3 +71,47 @@ def read_envelope_from_file(file_path: str):
...
+
+
+def force_set_config(dotted_path: str, value: Any) -> None:
+    """
+    Set an AEA config without validation.
+    Run from agent's directory.
+
+    Allowed dotted_path:
+        'agent.an_attribute_name'
+        'protocols.my_protocol.an_attribute_name'
```
DavidMinarsch, line 98: also allowed `contracts. ...`

tests/test_packages/test_skills/test_weather.py
```diff
@@ -20,251 +20,56 @@
...
+        self.initialize_aea()
+        self.create_agents(weather_station_aea_name, weather_client_aea_name)
+
+        # prepare agent one (weather station)
+        weather_station_aea_dir_path = os.path.join(self.t, weather_station_aea_name)
+        os.chdir(weather_station_aea_dir_path)
+
+        self.add_item("connection", "fetchai/oef:0.2.0")
+        self.add_item("skill", "fetchai/weather_station:0.1.0")
+        self.disable_ledger_tx("fetchai", "skill", "weather_station")
```
DavidMinarsch, line 48: This method `disable_ledger_tx` needs to go from the test case, it is not generic at all (only works for one particular dotted path). Instead, use the config set functionality please.

tests/test_packages/test_skills/test_weather_ledger.py
```diff
@@ -19,300 +19,95 @@
...
-                *CLI_LOG_OPTION,
-                "config",
-                "set",
-                "agent.default_connection",
-                "fetchai/oef:0.2.0",
-            ],
-            standalone_mode=False,
-        )
-        assert result.exit_code == 0
+        weather_station_aea_config.write_text(weather_station_aea_config_content)
```
DavidMinarsch, line 60: You can now use config force set

aea/cli/config.py
panasevychol: To put in `try` statement only the call we catch exception from. And `else` block includes the correct result handling.

PR #1099, opened 2020-04-15
[AEA-532] Feature/limit execution time

solarw:
## Proposed changes

Tools for limiting python code execution time
To wrap act and handlers to prevent a long time blocking of agent's loop

## Types of changes

## Checklist


aea/helpers/exec_timeout.py
```diff
@@ -0,0 +1,154 @@
...
+    def is_cancelled_by_timeout(self):
+        return self.cancelled_by_timeout
+
+
+class TimeoutException(Exception):
+    pass
+
+
+class BaseExecTimeout(ABC):
+    def __init__(self, timeout: float = 0):
```
DavidMinarsch, line 51: Please add a docstring

DavidMinarsch: for all dunder methods and public methods

aea/helpers/exec_timeout.py
```diff
@@ -0,0 +1,154 @@
...
+    exception_class = TimeoutException  # type: Type[Exception]
+
+    _supervisor_thread = None  # type: threading.Thread
+    _loop = None  # type: AbstractEventLoop
+    _stopped_future = None  # type: asyncio.Future
+
+    @classmethod
+    def start(cls) -> None:
+        """
+        start supervisor thread to check timeouts
```
DavidMinarsch, line 105: Capitalize first word.

aea/helpers/exec_timeout.py
```diff
@@ -0,0 +1,154 @@
...
+from asyncio.events import AbstractEventLoop
+from types import TracebackType
+from typing import Type
+
+
+class TimeoutResult:
+    """ Result of ExecTimeout context manager """
+
+    def __init__(self):
+        self.cancelled_by_timeout = False
```
DavidMinarsch, line 37: could be made private (minor point though)

aea/helpers/exec_timeout.py
DavidMinarsch: default value of float should be `0.0`

DavidMinarsch: Please also add return type to each function signature.

aea/helpers/exec_timeout.py
```diff
@@ -0,0 +1,154 @@
...
+    def start(cls) -> None:
+        """
+        start supervisor thread to check timeouts
+
+        :return: None
+        """
+        if cls._supervisor_thread:
+            return
+
+        cls._loop = asyncio.new_event_loop()
```
DavidMinarsch, line 112: Does it need to be a *new* event loop?

aea/helpers/exec_timeout.py
solarw: done

aea/helpers/exec_timeout.py
solarw: done

aea/helpers/exec_timeout.py
solarw: done

aea/helpers/exec_timeout.py
solarw: I think so, cause it started in dedicated thread. It's better to have own event loop just for code execution time control and do not rely on other loops present in the process.

aea/helpers/exec_timeout.py
```diff
@@ -0,0 +1,246 @@
...
+        :return: None
+        """
+        raise NotImplementedError  # pragma: nocover
+
+
+class ExecTimeoutSigAlarm(BaseExecTimeout):
+    """
+        ExecTimeout context manager implementation usin signals and SIGALARM
+        Does not support threads, have to be used only in main thread.
+    """
```
DavidMinarsch, line 138:
```suggestion
    """
    ExecTimeout context manager implementation usin signals and SIGALARM
    Does not support threads, have to be used only in main thread.
    """
```

aea/helpers/exec_timeout.py
```diff
@@ -0,0 +1,246 @@
...
+        :return: None
+        """
+        signal.setitimer(signal.ITIMER_REAL, 0, 0)
+
+
+class ExecTimeoutThreadGuard(BaseExecTimeout):
+    """
+        ExecTimeout context manager implementation using threads and PyThreadState_SetAsyncExc
+        Support threads.
+        Requires supervisor thread start/stop to control execution time control
```
DavidMinarsch, line 162: Add comment on limitations here

aea/aea.py
```diff
@@ -223,7 +229,35 @@ def _handle(self, envelope: Envelope) -> None:
             return
 
         for handler in handlers:
-            handler.handle(msg)
+            self._handle_message_with_handler(msg, handler)
+
+    def _handle_message_with_handler(self, message: Message, handler: Handler) -> None:
+        """Handle one message with one predefined handler.
```
DavidMinarsch, line 235:
```suggestion
        """
        Handle one message with one predefined handler.
```
for consistency

aea/aea.py
```diff
@@ -223,7 +229,35 @@ def _handle(self, envelope: Envelope) -> None:
...
+
+        :param message: message to be handled.
+        :param handler: handler suitable for this message protocol.
+        """
+        with ExecTimeoutThreadGuard(self._execution_timeout) as timeout_result:
+            handler.handle(message)
+
+        if timeout_result.is_cancelled_by_timeout():
+            logger.warning(
+                f"Handler `{handler}` was terminated by timeout == {self._execution_timeout}"
```
DavidMinarsch, line 245:
```suggestion
                "Handler `{}` was terminated with timeout {} seconds".format(handler, self._execution_timeout)
```
For consistency.

aea/aea.py
```diff
@@ -223,7 +229,35 @@ def _handle(self, envelope: Envelope) -> None:
...
+            )
+
+    def _behaviour_act(self, behaviour: Behaviour) -> None:
+        """Call behaviour's act.
```
DavidMinarsch, line 249:
```suggestion
        """
        Call behaviour's act.
```
For consistency.

aea/aea.py
```diff
@@ -223,7 +229,35 @@ def _handle(self, envelope: Envelope) -> None:
...
+
+        :param behaviour: behaviour already defined
+        :return: None
+        """
+        with ExecTimeoutThreadGuard(self._execution_timeout) as timeout_result:
+            behaviour.act_wrapper()
+
+        if timeout_result.is_cancelled_by_timeout():
+            logger.warning(
+                f"Act of `{behaviour}` was terminated by timeout == {self._execution_timeout}"
```
DavidMinarsch, line 259:
```suggestion
                "Act of `{}` was terminated as its execution exceeded the timeout of {} seconds. Please refactor your code!".format(behaviour, self._execution_timeout)
```

tests/common/utils.py
```diff
@@ -56,16 +65,23 @@ def timeit_context():
 
 
 class AeaTool:
+    """AEA test wrapper tool.
```
DavidMinarsch, line 68:
```suggestion
    """
    AEA test wrapper tool.
```
for consistency

tests/common/utils.py
```diff
@@ -56,16 +65,23 @@ def timeit_context():
...
+
+    To make testing AEA instances easier
+    """
+
     def __init__(self, aea: AEA):
+        """:param aea: AEA instance to wrap for tests."""
```
DavidMinarsch, line 74:
```suggestion
        """
        Instantiate AeaTool.
       
        :param aea: AEA instance to wrap for tests.
        """
```

tests/test_aea_exectimeout.py
```diff
@@ -0,0 +1,178 @@
...
+
+        with timeit_context() as timeit:
+            self.aea_action()
+
+        assert function_sleep_time <= timeit.time_passed <= execution_timeout
+        assert self.function_finished
+        self.aea_tool.stop()
+
+    def test_no_timeout(self):
+        """Test function NOT terminated by timeout cause timeout == 0."""
```
DavidMinarsch, line 147: Soo timout 0 is a special case? shouldn't timeout 0 mean everything is timed out?

solarw: there is no meaning in set code already expired, it's easier to do not call it at all.

aea/aea.py
solarw: done

aea/aea.py
solarw: done. but curious why not f string?  it's supported from python3.6, in this case, is suitable for text formating

aea/aea.py
solarw: done

aea/aea.py
solarw: done

tests/common/utils.py
solarw: done

tests/common/utils.py
solarw: done

aea/aea.py
DavidMinarsch: Just a convention atm. Can discuss.

aea/aea.py
```diff
@@ -223,7 +229,41 @@ def _handle(self, envelope: Envelope) -> None:
...
+
+    def _behaviour_act(self, behaviour: Behaviour) -> None:
+        """
+        Call behaviour's act.
+
+        :param behaviour: behaviour already defined
+        :return: None
+        """
+        with ExecTimeoutThreadGuard(self._execution_timeout) as timeout_result:
+            behaviour.act_wrapper()
```
marcofavorito, line 259: what happens if someone in the `act` method does:
```python
def act(self):
    try:
        ...
    except TimeoutException:
        pass
    # do whatever I want

File: aea/aea.py [hunk 15]
```diff
@@ -53,9 +55,10 @@ def __init__(
         resources: Resources,
         loop: Optional[AbstractEventLoop] = None,
         timeout: float = 0.0,
+        execution_timeout: float = 0.1,
```
marcofavorito, line 58: TL;DR: what about `execution_timeout = 0.0` by default? Or a high value like `5.0` .

Thought: in `asyncio` programming, the library does not tell you that coroutine should be fast because otherwise they block the entire loop. Of course that is not part of "idiomatic asyncio programming" , but gives to the developer more flexibility (if I want to call `time.sleep(1.0)` in a coroutine I expect it to work despite it will block the loop).

The analogy between asyncio and the framework introduced in this PR is not strong since `aea` is not a standard library. However, the resulting experience of the developer might be worse because the `handle/act` always get blocked.

I see `execution_timeout` as a configuration for "production-ready" AEAs, whereas during development I would go for `execution_timeout=0.0` or setting it to a high value.

Also, many skills currently block the loop because of some interaction with the Decision Maker (@DavidMinarsch can you confirm?), so a default value too low or different than `0.0` would break those skills.

DavidMinarsch: @marcofavorito , last point first, interactions with the decision maker should be non-blocking (send a message to DM, pick it up via handler). Some skills might be affected though due to poor logic and it is something we need to test.

I think especially for development it would be helpful to know when you write blocking logic. But I see your point of view. Maybe we make the execution timeout configurable via the builder already?

aea/aea.py
solarw: @marcofavorito 
that exception will be handled. A possible solution is to make TimeoutException as a sublass of BaseException, so `except Exception:` will not catch it, but anyway user is able to catch it explicitly and make some shutdown code for this.

aea/aea.py
solarw: @marcofavorito 
python doc says: https://docs.python.org/3/library/asyncio-dev.html#running-blocking-code
so it assumed to not put a blocking code in a coroutine. anyway, a developer is free to do this.

the main idea of the execution timeout feature is to notify the developer that something wrong with his code and there can be side effects with performance.

actually we can warn it more gently, without interruption. just write a log.


in short: main proposal is just to make harder to write bad code. but it does not help to write a good one.

aea/aea.py
marcofavorito: yes, worth considering inheriting from `BaseException`, since it is a bad practice to do bare `except:` or `except BaseException`, and in that case the fault of bad behaviour would be of the developer.

aea/aea.py
marcofavorito: yes, there's room for improvement to customize the actual behaviour of this feature (e.g. only send a notification or raise an exception). This could be the jobby a later PR. 

I wonder what's a good default value for this feature until the other PR on customizability is in place.

solarw: better and more flexible config support is required. currently, the agent's timeout is not adjustable via config, so better agent build with config support is welcome

aea/aea.py
solarw: updated, thanks for the notice!
switched to BaseException

aea/aea.py
solarw: updated default timeout to 1 second.
suppose it's better to make a dedicated discussion for better control over code execution and possible ways to solve it

Issue #1062, opened 2020-04-06
Clearer error messages for agent builder.


Totoual:
**Is your feature request related to a problem? Please describe.**

When using `AgentBuilder.build`, if there's a missing dependency, say a protocol needed by a skill, 
you'll likely get an error like the following:

`Exception: An error occurred while loading skill author/package:0.1.0: No module named 'packages.author'; 'packages' is not a package`

It is not pointing to the problem. 

**Describe the solution you'd like**

Agent builder could provide better error messages both at "add time" and/or at "build time" regarding missing dependencies

PR #1100, opened 2020-04-15
Add AEAException and use it in exceptions of AEABuilder.add_component

marcofavorito:
## Proposed changes

This PR does the following:
- adds an user-defined `AEAException`. 
- improves the `AEABuilder.add_component` so it raises `AEAException`s in specific circumstances.

The `AEAException` class is supposed to be used whenever we need to describe exceptions at framework level.

For instance, trying to add a package in the AEABuilder class (`AEABuilder.add_component(type, path)`) might fail for many reasons:

1. the path to the package does not exist
2. the directory exists, but there isn't the expected configuration file.
3. the package already exists in the AEABuilder
4. there are some missing package dependencies
...

Despite we can't foresee all the exceptions, for some of them we can. For instance, case (3) and (4) are exceptions at framework level, so we can raise the right exception to the client of the class.

Needless to say, AEABuilder.add_component' is just one example. The same idea can be applied to other framework classes/functions.

## Fixes

First step toward fixing #1062 

## Types of changes

## Checklist

- [ ] I am making a pull request against the `develop` branch (left side). Also, you should start your branch off our `develop`.

## Further comments
N/A

### DIFF
--- a/aea/aea_builder.py
+++ b/aea/aea_builder.py
@@ -22,6 +22,7 @@ import inspect
 import itertools
 import logging
 import os
+import pprint
 import re
 from pathlib import Path
 from typing import Any, Collection, Dict, List, Optional, Set, Tuple, Union, cast
@@ -61,6 +62,7 @@ from aea.crypto.helpers import (
 )
 from aea.crypto.ledger_apis import LedgerApis
 from aea.crypto.wallet import SUPPORTED_CRYPTOS, Wallet
+from aea.exceptions import AEAException
 from aea.helpers.base import add_modules_to_sys_modules, load_all_modules, load_module
 from aea.identity.base import Identity
 from aea.mail.base import Address
@@ -190,19 +192,6 @@ class _DependenciesManager:
         for dependency in component.package_dependencies:
             self._inverse_dependency_graph[dependency].discard(component_id)
 
-    def check_package_dependencies(
-        self, component_configuration: ComponentConfiguration
-    ) -> bool:
-        """
-        Check that we have all the dependencies needed to the package.
-
-        return: True if all the dependencies are covered, False otherwise.
-        """
-        not_supported_packages = component_configuration.package_dependencies.difference(
-            self.all_dependencies
-        )
-        return len(not_supported_packages) == 0
-
     @property
     def pypi_dependencies(self) -> Dependencies:
         """Get all the PyPI dependencies."""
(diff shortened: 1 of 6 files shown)


aea/aea_builder.py
```diff
@@ -618,14 +607,33 @@ def _check_configuration_not_already_added(self, configuration) -> None:
...
-                "Component {} of type {} already added.".format(
+            raise AEAException(
+                "Component '{}' of type '{}' already added.".format(
                     configuration.public_id, configuration.component_type
                 )
             )
 
-    def _check_package_dependencies(self, configuration):
-        self._package_dependency_manager.check_package_dependencies(configuration)
+    def _check_package_dependencies(
```
DavidMinarsch, line 616: why private?

marcofavorito: because it is not supposed to be used from the outside. It does not return a `bool` but raises an exception if there's something wrong.

However, I agree it could be useful to have a public method to check if a package can be added. We should discuss how to expose such method (does it return a boolean? Or raises an exception?)

DavidMinarsch: A public method should probably return a `bool`. Can be done later though (YAGNI)

tests/test_aea.py
```diff
@@ -101,6 +101,9 @@ def test_react():
         builder = AEABuilder()
         builder.set_name(agent_name)
         builder.add_private_key(FETCHAI, private_key_path)
+        builder.add_protocol(
```
DavidMinarsch, line 104: Hmm not a fan of making these tests heavier than necessary.

tests/test_protocols/test_generator.py
```diff
@@ -262,6 +262,12 @@ def test_generated_protocol_end_to_end(self):
         builder_1.add_private_key(FETCHAI, FETCHAI_PRIVATE_KEY_FILE)
         builder_1.set_default_ledger(FETCHAI)
         builder_1.set_default_connection(PublicId.from_str("fetchai/oef:0.2.0"))
+        builder_1.add_protocol(
```
DavidMinarsch, line 265: why are they added?

tests/test_aea.py
marcofavorito: from 0231f54:

tests/test_protocols/test_generator.py
marcofavorito: see above

tests/test_aea.py
DavidMinarsch: Hmm ok.

PR #1102, opened 2020-04-15
AEA-542 Logger.error replaced with ClickException for correct handling.

panasevychol:
## Fixes

Logger.error replaced with ClickException for correct handling.

## Types of changes

## Checklist


aea/cli/create.py
```diff
@@ -160,13 +157,10 @@ def create(click_context, agent_name, author, local):
...
-        logger.error("Directory already exist. Aborting...")
-        sys.exit(1)
+        raise click.ClickException("Directory already exist. Aborting...")
     except ValidationError as e:
-        logger.error(str(e))
+        raise click.ClickException(str(e))
         shutil.rmtree(agent_name, ignore_errors=True)
-        sys.exit(1)
     except Exception as e:
         logger.exception(e)
```
DavidMinarsch, line 165: Missing!

aea/cli/create.py
```diff
@@ -160,13 +157,10 @@ def create(click_context, agent_name, author, local):
         _add_item(click_context, "skill", DEFAULT_SKILL)
 
     except OSError:
-        logger.error("Directory already exist. Aborting...")
-        sys.exit(1)
+        raise click.ClickException("Directory already exist. Aborting...")
     except ValidationError as e:
-        logger.error(str(e))
+        raise click.ClickException(str(e))
```
DavidMinarsch, line 162: wrong order

aea/cli/generate.py
```diff
@@ -131,30 +126,27 @@ def _generate_item(click_context, item_type, specification_path):
...
-        logger.error(
+        raise click.ClickException(
             "A {} with this name already exists. Please choose a different name and try again.".format(
                 item_type
             )
         )
-        sys.exit(1)
     except ProtocolSpecificationParseError as e:
-        logger.error(
+        raise click.ClickException(
```
DavidMinarsch, line 135: wrong order

aea/cli/generate.py
```diff
@@ -131,30 +126,27 @@ def _generate_item(click_context, item_type, specification_path):
...
     except Exception as e:
         logger.debug("Exception thrown: " + str(e))
-        logger.error(
-            "There was an error while generating the protocol. The protocol is NOT generated."
-        )
         shutil.rmtree(
             os.path.join(item_type_plural, protocol_spec.name), ignore_errors=True
         )
-        sys.exit(1)
+        raise click.ClickException(
```
DavidMinarsch, line 147: Remove debug and print exception message here

aea/cli/common.py
```diff
@@ -383,8 +378,10 @@ def convert(self, value, param, ctx):
             # everything ok - return the parameter to the command
             return value
         except Exception:
-            logger.error("The name provided is not a path to an AEA project.")
-            self.fail(value, param, ctx)
+            # self.fail(value, param, ctx)
```
DavidMinarsch, line 381: do we still need self.fail? if so, remove commenting, else remove comment altogether.

aea/cli/generate.py
DavidMinarsch: Sorry I think you misunderstood, the exception message should be moved inside the click exception, i.e.: `"There was an error while generating the protocol. The protocol is NOT generated. Exception thrown: " + str(e)`

panasevychol: Oh yes, now I see. Will do.

aea/cli/common.py
```diff
@@ -191,20 +190,18 @@ def try_to_load_agent_config(ctx: Context, is_exit_on_except: bool = True) -> No
             logging.config.dictConfig(ctx.agent_config.logging_config)
     except FileNotFoundError:
         if is_exit_on_except:
-            logger.error(
+            raise click.ClickException(
                 "Agent configuration file '{}' not found in the current directory.".format(
                     DEFAULT_AEA_CONFIG_FILE
                 )
             )
```
marcofavorito, line 197: I think this is a good use case for:
```
raise ExceptionX(...) from e
```

see https://docs.python.org/3/reference/simple_stmts.html#the-raise-statement

marcofavorito: why don't use `AEAException` which can then be re-raised later as a `ClickException`? So we can  distinguish between exception for our logic and exceptions outside it

aea/cli/generate.py
```diff
@@ -131,30 +126,27 @@ def _generate_item(click_context, item_type, specification_path):
...
-        logger.error(
+        shutil.rmtree(
+            os.path.join(item_type_plural, protocol_spec.name), ignore_errors=True
+        )
+        raise click.ClickException(
             "The following error happened while parsing the protocol specification: "
             + str(e)
         )
+    except Exception as e:
+        print("Exception thrown: " + str(e))
```
DavidMinarsch, line 143:
```suggestion
```

aea/cli/generate.py
```diff
@@ -131,30 +126,27 @@ def _generate_item(click_context, item_type, specification_path):
...
         shutil.rmtree(
             os.path.join(item_type_plural, protocol_spec.name), ignore_errors=True
         )
-        sys.exit(1)
-    except Exception as e:
-        logger.debug("Exception thrown: " + str(e))
-        logger.error(
+        raise click.ClickException(
             "There was an error while generating the protocol. The protocol is NOT generated."
```
DavidMinarsch, line 148:
```suggestion
            "There was an error while generating the protocol. The protocol is NOT generated. Exception: " + str(e)
```

aea/cli/common.py
DavidMinarsch, line 381:
```suggestion
```

Issue #890, opened 2020-03-03
Improve error messages when the loading of a skill fails.


marcofavorito:
**Is your feature request related to a problem? Please describe.**

The loading of a skill can fail due to a missing package. The error message should be more informative.
E.g. this is what we've got when running an example:
```
warning: A problem occurred while parsing the skill directory /Users/aristotelistriantafyllidis/Documents/agents-aea/deployer_aea/vendor/fetchai/skills/erc1155_deploy. Exception: No module named 'packages.fetchai'; 'packages' is not a package
```

However, the error was not about `packages.fetchai` but on what follows: `skills.generic_seller.dialogues`, which was missing.

**Describe the solution you'd like**

Better error messages. At least, provide the full import path.

**Describe alternatives you've considered**

**Additional context**

PR #1111, opened 2020-04-16
AEA-503 improve the import_module function in helpers.

marcofavorito:
## Proposed changes

This PR changes the `aea.helpers.base.import_module` helper function, called by `aea.helpers.base.add_modules_to_sys_modules`, which in turn is called by many `aea.aea_builder.load_and_add_<package_type>`.
Now it adds all the prefixes of the import path as namespace namespaces in `sys.modules`.

Also, rename the function to `import_aea_module` since it used to handle only import statements for an AEA package (i.e. of the form `packages.<author_name>.<package_type_plural>.<package_name>...`).

## Fixes

Fix (partially) #1062 and  #890

... why? 

Because now, when trying to import, say, `packages.fetchai but on what follows: skills.generic_seller.non_existing_module` (see #890), assuming `packages.fetchai but on what follows: skills.generic_seller` is loaded correctly, instead of:

    error: An error occurred while loading ...: No module named 'packages.fetchai'; 'packages' is not a package

It would likely give an error:

    error: An error occurred while loading ...: No module named 'packages.fetchai.skills.generic_seller.non_existing_module

Which is more explicative.

However, this is not a complete solution. If the module of an AEA package tries to import from another package that does not exists, e.g.

    import packages.fetchai.skills.non_existing_package

The output will be:

    No module named 'packages.fetchai.skills.non_existing_package'; 'packages.fetchai.skills' is not a package

which is quite cryptic, because `packages.fetchai.skills` is not a package but a namespace, and the error message is not very clear (it's that `non_existing_package` is not found).

Nevertheless, for some cases, the changes proposed in this PR are still helpful, and they don't harm.

## Types of changes

## Checklist

### DIFF
--- a/aea/helpers/base.py
+++ b/aea/helpers/base.py
@@ -201,21 +201,41 @@ def load_module(dotted_path: str, filepath: Path) -> types.ModuleType:
     return module
 
 
-def import_module(dotted_path: str, module_obj) -> None:
+def import_aea_module(dotted_path: str, module_obj) -> None:
     """
-    Add module to sys.modules.
+    Add an AEA module to sys.modules.
+
+    The parameter dotted_path has the form:
+
+        packages.<author_name>.<package_type>.<package_name>
+
+    If the closed-prefix packages are not present, add them to sys.modules.
+    This is done in order to emulate the behaviour of the true Python import system,
+    which in fact imports the packages recursively, for every prefix.
+
+    E.g. see https://docs.python.org/3/library/importlib.html#approximating-importlib-import-module
+    for an explanation on how the 'import' built-in function works.
 
     :param dotted_path: the dotted path to be used in the imports.
     :param module_obj: the module object. It is assumed it has been already executed.
     :return: None
     """
-    # if path is nested, and the root package is not present, add it to sys.modules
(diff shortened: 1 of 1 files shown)


aea/helpers/base.py
```diff
@@ -201,21 +201,41 @@ def load_module(dotted_path: str, filepath: Path) -> types.ModuleType:
...
-        sys.modules[root] = types.ModuleType(root)
 
-    # add the module at the specified path.
+    def add_namespace_to_sys_modules_if_not_present(dotted_path: str):
+        if dotted_path not in sys.modules:
+            sys.modules[dotted_path] = types.ModuleType(dotted_path)
+
+    # add all prefixes as 'namespaces', since they are not actual packages.
+    split = dotted_path.split(".")
+    assert len(split) > 3
```
DavidMinarsch, line 230: If this assert is not met I imagine the error message will not be very informative?

marcofavorito: Notice that the message of the `AssertionError` raised from this `assert` is only for the internals of the framework.  It is not related to the issue directly. But I agree with the comment, that assertion needs a message :-)

PR #1112, opened 2020-04-16
Wrap built-in __import__ for better error messages.

### LINKED ISSUE #890: Improve error messages when the loading of a skill fails.

marcofavorito:
## Proposed changes

(continuation of #1111)

This PR adds a wrapper to the built-in `__import__` function. It is automatically installed as soon as the framework is used.

It intercepts import statements whose import path start with `packages.`. Also, it filters out imports whose _level_ (see https://docs.python.org/3/library/functions.html#__import__) is `0`.

The right _level_ value has been found empirically. Unfortunately, it's not very reliable. And I couldn't find a better way to discriminate between proper AEA import statements and other import attempts e.g. `packages.six.moves.http_client` (`urllib3`), `packages.ssl_match_hostname` (`urllib3`), `packages.six.moves` etc.

An idea would be to add a 'blacklist' and to filter out import attempts whose `__package__` attribute in `globals`/`locals` is known. But still, there will always be corner cases for this.

## How to test the changes

From the repository root, execute these commands:
image

in the screenshot, I added a wrong import statement in `vendor/fetchai/protocols/fipa/serialization.py` (see line `31`):
image

## Fixes

Fixes #890 and #1062, but in my opinion, the changes increase instability and unpredictability. 
I suggest implementing a custom importer by relying on the `importlib` standard package (in particular, on [`importlib.machinery`](https://docs.python.org/3/library/importlib.html#module-importlib.machinery) and path hooks to be installed in `sys.path_hooks`).

Also the [Python docs on `__import__`](https://docs.python.org/3/library/functions.html#__import__) discourage patching the built-in `__import__`:

> It can be replaced (by importing the builtins module and assigning to builtins.__import__) in order to change the semantics of the import statement, but doing so is strongly discouraged as it is usually simpler to use import hooks (see PEP 302) to attain the same goals and does not cause issues with code which assumes the default import implementation is in use.

## Types of changes

## Further comments

Despite my bad recommendation on this PR, it has been useful to understand the limits of this approach. I'm investigating the use of `sys.meta_path` and `sys.path_hooks`, although it won't be trivial, as clearly explained [here](https://stackoverflow.com/questions/41990169/how-to-use-sys-path-hooks-for-customized-loading-of-modules).

Nevertheless, the PR addresses the issues, with a relatively small impact on the codebase..

### DIFF
--- a/aea/helpers/_import_wrapper.py
+++ b/aea/helpers/_import_wrapper.py
@@ -1,0 +1,122 @@
+# -*- coding: utf-8 -*-
+# ------------------------------------------------------------------------------
+#
+#   Copyright 2018-2019 Fetch.AI Limited
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+# ------------------------------------------------------------------------------
+
+"""Import patcher."""
+
+import builtins
+
+
+class ImportWrapper:
+    """This class patches the built-in __import__ function."""
+
+    _builtin_import = builtins.__import__
+
+    @staticmethod
+    def wrapped_import(name: str, globals=None, locals=None, fromlist=(), level=0):
+        """
+        The patched __import__ function.
+
+        It acts as a wrapper to the built-in __import__ function.
+
+        It only gets activated if the first part of the import path
+        is equal to "packages". Otherwise, it forwards the call to
+        the builtins.__import__ function.
+
+        For documentation on the parameters,
+        please look at https://docs.python.org/3/library/functions.html#__import__.
+        """
+        kwargs = dict(globals=globals, locals=locals, fromlist=fromlist, level=level)
+        # print(name, globals, locals, fromlist, level)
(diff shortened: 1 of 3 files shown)


aea/helpers/_import_wrapper.py
```diff
@@ -0,0 +1,122 @@
...
+
+        It only gets activated if the first part of the import path
+        is equal to "packages". Otherwise, it forwards the call to
+        the builtins.__import__ function.
+
+        For documentation on the parameters,
+        please look at https://docs.python.org/3/library/functions.html#__import__.
+        """
+        kwargs = dict(globals=globals, locals=locals, fromlist=fromlist, level=level)
+        # print(name, globals, locals, fromlist, level)
```
marcofavorito, line 45: to remove

aea/helpers/_import_wrapper.py
```diff
@@ -0,0 +1,122 @@
...
+        parts = name.split(".")
+        root = parts[0]
+        if root == "packages" and level == 0:
```
marcofavorito, line 48: without `level == 0`, many conflicting packages mentioned above get processed in our handler. However, this is only a workaround, since in some AEA package code users could import from those packages and make the patcher to break.

DavidMinarsch:
As discussed in person, I think the unpredictability and instability this brings don't warrant its inclusion at this point in time.

Issue #1095, opened 2020-04-11
Unexpected skill de-registration behaviour


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
If a skill is deregistered during initialization of one of its modules, it can lead to unexpected behaviour.

**Describe the solution you'd like**
Resolve below:
<img width="654" alt="Screenshot 2020-04-11 at 10 50 12" src="https://user-images.githubusercontent.com/19314578/79040715-68d77f00-7be2-11ea-8a2d-45a89a65c67b.png">

marcofavorito:
@DavidMinarsch please could you provide instructions on how to reproduce the issue?

Issue #1045, opened 2020-04-02
Add contracts automatically when adding a skill


marcofavorito:
**Is your feature request related to a problem? Please describe.**
Currently, when adding a skill, the protocols the skill depends on are added automatically.
However, that's not the case for contracts.

**Describe the solution you'd like**
Add contracts automatically as it is done with the protocols.

**Describe alternatives you've considered**
N/A

**Additional context**
It can be solved by the more general issue #960

PR #1119, opened 2020-04-19
AEA-537 allow for deregistration of skill in skill component __init__().

marcofavorito:
## Proposed changes

Changes:

1) instantiate skill object before skill component loading
2) in skill component registries, setup() now dispatches the call to only the skill components whose skill is active.

Change 1. The main motivation for this change is due to https://github.com/fetchai/agents-aea/issues/1095. In the tac_controller_contract skill, the Parameter model checks whether the skill should be deactivated or not in its __init__ method. But if the condition is met, it will try to set the attribute context.is_active,
which in turn requires the skill object being available to retrieve the skill id for a debug log message.

Hence, before this commit, trying to deactivate a skill from the setup() method of some skill component would have failed. 

The reason is that the current implementation of the loading of a skill follows this order:
```python
Skill.from_config(configuration, skill_context):
    # notice: skill_context._skill is None
    load handlers/behaviours/models
    # instantiate skill object
    skill = Skill(configuration, skill_context, handlers, behaviours, models)
    # set the skill instance to the skill context
    skill_context._skill = skill
    return skill
```
However, if a skill component needed to deactivate the skill, i.e.
```python
context.is_active = False
```

this would have caused an error because the setter of the attribute 'is_active' needs the skill_id (and so the skill object) in order to log a debug message, which was not available since the skill object was instantiated at the end of the loading of each component.

In order to fix that, move the instantiation before the loading of each skill, and add the skill components to the skill object at the end, i.e. 
```python 
Skill.from_config(configuration, skill_context):
    # this will instantiate empty indexes of skill components
    skill = Skill(configuration, skill_context)
    skill._skill = skill
    load handlers/behaviours/models
    skill.handlers.update(handlers)
    skill.behaviours.update(behaviours)
    skill.models.update(models)
    return skill
```

Change 2. Due to the previous change, it could happen that some skills are deactivated before the setup() methods of the skill components are called.

This change will make the component registry to first check whether a skill associated to a component is active before calling the setup() method of the component.

## Fixes

## Types of changes

## Checklist

## Further comments

More work needs to be done to handle re-activation of skill (and re-calling of setup(), if it should be done in the first place), as well as teardown of skill as soon as they are deactivated.


aea/registries/filter.py
```diff
@@ -103,7 +103,7 @@ def get_active_behaviours(self) -> List[Behaviour]:
         """
         behaviours = self.resources.get_all_behaviours()
         active_behaviour = list(
-            filter(lambda b: b.context.is_active and not b.is_done(), behaviours,)
+            filter(lambda b: b.context.is_active and not b.is_done(), behaviours)
```
marcofavorito, line 106: just removed trailing comma

docs/tac-skills.md
```diff
@@ -157,7 +157,6 @@ cd tac_controller_contract
 ``` bash
 aea add connection fetchai/oef:0.2.0
 aea add skill fetchai/tac_control_contract:0.1.0
-aea add contract fetchai/erc1155:0.2.0
```
marcofavorito, line 160: this is not needed anymore. Contracts are correctly added if a skill depends on them (see https://github.com/fetchai/agents-aea/issues/1045), and the skill `fetchai/tac_control_contract:0.1.0` depends on `fetchai/erc1155:0.2.0`.

aea/skills/base.py
```diff
@@ -643,12 +643,19 @@ def from_config(
         assert (
             configuration.directory is not None
         ), "Configuration must be associated with a directory."
+
+        # we put the initialization here because some skill components
+        # might need some info from the skill
+        # (e.g. see https://github.com/fetchai/agents-aea/issues/1095)
+        skill_context = SkillContext() if skill_context is None else skill_context
+        skill = Skill(configuration, skill_context)
+        skill_context._skill = skill
```
marcofavorito, line 652: in the specific case of the skill `fetchai/tac_control_contract:0.1.0`, we need this because the `__init__` method of the model `Parameters` calls the setter of `context.is_active = False`, which requires `skill.skill_id` in order to print a debug message.

aea/registries/base.py
```diff
@@ -341,7 +341,19 @@ def setup(self) -> None:
         :return: None
         """
         for item in self.fetch_all():
-            item.setup()
```
marcofavorito, line 344: we now need to skip the setup of skill components whose skill is deactivated, since a skill can be deactivated in the `__init__` (hence at load time).

aea/registries/base.py
```diff
@@ -47,7 +46,7 @@
 Item = TypeVar("Item")
 ItemId = TypeVar("ItemId")
 ComponentId = Tuple[SkillId, str]
-SkillComponentType = TypeVar("SkillComponentType", Handler, Behaviour, Task, Model)
```
DavidMinarsch, line 50: Why is task no longer considered one?

docs/tac-skills.md
DavidMinarsch: Yes it wasn't previously the case but now it is you are correct. Should be changed for the participants as well.

aea/registries/base.py
marcofavorito: Because it's not properly a skill component type. In fact, we don't include tasks in configuration files, nor in any registry.

PR #1139, opened 2020-04-23
AEA-522 Small doc updates based on user questions

### LINKED ISSUE #1071: Docs issues
A few issues with docs

- [x] explain how to disable consistency checks on aea cli
- [x] copy useful stuff from here: https://fetchai.github.io/oef-sdk-python/user/communication-protocols.html
- [x] add explanation of `--local` version to `init`.
- [x] point to https://docs.fetch.ai/aea/package-imports/ in quickstart
- [x] explain why own package management is used compared to python's default one. just some intro would be nice
- [x] explain local vs web based registry
- [x] explain only package with greatest version is activated
- [x] disambiguate "internal state"
- [x] explain callback design of handlers/behaviours
- [x] explain mix of configuration and code for modularity
- [x] explain `self.context.decision_maker_message_queue.put_nowait(tx_msg)`
- [x] problems persist mostly about onboarding: better to start by introducing the protocols & dialogues > main thing to learn
- [x] execution interruption: explain & configure (next release)
- [x] > describe dm: (can think of it like a wallet manager + “econ brain”); interface & implementation split; needs extension; Let developers mount their own decision maker by specifying it as a single python file in core of the project and then it gets mounted.; Implement a base interface class and then provide a common reference implementation.
- [x] in `config.md` it should be `loop_mode` not `agent_loop`
- [x] point to third party logos somewhere in `concepts` section to show we integrate with many different projects: community: ipfs, libp2p, ethereum, cosmos

DavidMinarsch:
## Proposed changes

Small docs updates

## Fixes

Partially fixes #1071 

## Types of changes

## Checklist

## Further comments

-


docs/quickstart.md
```diff
@@ -213,6 +219,80 @@ info: Echo Handler: teardown method called.
...
+from aea.protocols.default.serialization import DefaultSerializer
+from aea.test_tools.decorators import skip_test_ci
+from aea.test_tools.generic import (
+    read_envelope_from_file,
+    write_envelope_to_file,
+)
+from aea.test_tools.test_cases import AEATestCase
+
+
+class TestEchoSkill(AEATestCase):
```
DavidMinarsch, line 250: @marcofavorito and @panasevychol we would like to make this AEATestCase useful for developers. But actually, if you look at the below code taken from tests, we still lack a few things on the test case. So in the tests it makes sense to create the whole project from scratch. But here it would be preferable to lose lines 256-263 and instead point to the project directory which already exists and run the test. What do you think?

docs/quickstart.md
```diff
@@ -213,6 +219,80 @@ info: Echo Handler: teardown method called.
...
+        sent_envelope = Envelope(
+            to=agent_name,
+            sender="sender",
+            protocol_id=message.protocol_id,
+            message=DefaultSerializer().encode(message),
+        )
+
+        write_envelope_to_file(sent_envelope, DEFAULT_INPUT_FILE_NAME)
+
+        time.sleep(2.0)
```
marcofavorito, line 281: can't we use the "spin loop once" function?

docs/quickstart.md
```diff
@@ -213,6 +219,80 @@ info: Echo Handler: teardown method called.
...
+        received_envelope = read_envelope_from_file(DEFAULT_OUTPUT_FILE_NAME)
+
+        assert sent_envelope.to == received_envelope.sender
+        assert sent_envelope.sender == received_envelope.to
+        assert sent_envelope.protocol_id == received_envelope.protocol_id
+        assert sent_envelope.message == received_envelope.message
+
+        process.send_signal(signal.SIGINT)
+        process.wait(timeout=20)
+        assert process.returncode == 0
```
marcofavorito, line 291: Also this part could be improved. Put it in another method like `self.terminate_agent(signal=signal.SIGINT, timeout=20)`

docs/quickstart.md
```diff
@@ -213,6 +219,80 @@ info: Echo Handler: teardown method called.
...
+        agent_name = "my_first_agent"
+        self.create_agents(agent_name)
+
+        agent_dir_path = os.path.join(self.t, agent_name)
+        os.chdir(agent_dir_path)
+
+        self.add_item("skill", "fetchai/echo:0.1.0")
+
+        process = self.run_agent()
+        time.sleep(2.0)
```
marcofavorito, line 266: not in this PR, but ideally we could have `self.wait_until_running`. That would imply the agent can be "pinged" via some APIs (TCP-based? Pipe-based?), which we don't have atm :-)

docs/quickstart.md
marcofavorito: yes in my opinion. Maybe add a comment like:
```python
# we assume the AEA project has already been set up 
agent_dir_path = agent_name
os.chdir(agent_dir_path)
```

docs/quickstart.md
DavidMinarsch: @marcofavorito want to open an issue?

docs/quickstart.md
DavidMinarsch: Nice! combine these suggestions in issue below?

docs/quickstart.md
DavidMinarsch: The spin loop once is internals only atm. I am not sure whether it is wise to expose it to the test case at this state.

docs/quickstart.md
DavidMinarsch: The problem is, the test case does not support it atm, so we need to extend it.

DavidMinarsch: Have a look at the new implementation @marcofavorito

Issue #1136, opened 2020-04-22
Fix TaskManager resource consumption


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
The TaskManager uses the `multiprocessing` library and spawns 5 workers on instantiation.

**Describe the solution you'd like**
A quick fix:
when TaskManager.start is called, don't start the pool, just make it "accepting tasks". Then, on the first request, start the pool and limit it to 1 worker (configurable to more).

PR #1141, opened 2020-04-23
[AEA-552] TaskManager: lazy pool creation option

solarw:
## Proposed changes
TaskManager by default uses a lazy way to init workers pool on first enqueue instead of on start.
Can be controlled by TaskManager init argument: is_lazy_pool_start

default pool workers amount set to 1 instead of 5.

refers to https://github.com/fetchai/agents-aea/issues/1136
## Fixes
Saves some resources on agent start.

## Types of changes

## Checklist

### DIFF
--- a/aea/skills/tasks.py
+++ b/aea/skills/tasks.py
@@ -106,12 +106,14 @@ class Task:
         """
 
 
-def init_worker():
+def init_worker() -> None:
     """
     Initialize a worker.
 
     Disable the SIGINT handler.
     Related to a well-known bug: https://bugs.python.org/issue8296
+
+    :return: None
     """
     signal.signal(signal.SIGINT, signal.SIG_IGN)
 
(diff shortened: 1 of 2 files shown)


aea/skills/tasks.py
```diff
@@ -106,36 +106,53 @@ def teardown(self) -> None:
...
+    :return: None
     """
     signal.signal(signal.SIGINT, signal.SIG_IGN)
 
 
 class TaskManager:
     """A Task manager."""
 
-    def __init__(self, nb_workers: int = 5):
+    def __init__(self, nb_workers: int = 5, is_lazy_pool_start: bool = True):
```
DavidMinarsch, line 124: default nb_workers should be 1

DavidMinarsch, line 124:
```suggestion
    def __init__(self, nb_workers: int = 1, is_lazy_pool_start: bool = True):
```

Issue #1005, opened 2020-03-30
End-to-end tests: expand AeaTestCase and and use it in tests


panasevychol:
Currently most of end-to end tests extremely overweighted with repeatable code - file operations, CLI command calls, subprocesses, etc. We have AeaTestCase - a parent class with most usable operations (agents create, package settings update, install, etc.).

Currently we do not provide a test framework with helpful test classes for testing of AEAs

**Describe the solution you'd like**
Expand existing test case with more tools, use  it in other end-to-end tests. Develop such a framework and distribute it as part of the aea package

In particular:
- [x] create comprehensive test case class for end-to-end testing of (multiple) AEAs
- [x] add the test class as a core AEA module (`aea/test_tools`)
- [x] transition all existing end-to-end tests to use the new test case class (in `packages/tests_skills` and `packages/test_docs` as well as other end-to-end tests.)
- [x] add functionality to check for log output in test case
- [x] apply log output tests to all existing end-to-end tests (`tests/test_packages/skills` and `tests/test_docs/`)

DavidMinarsch:
The way I imagine the test:
- in a yaml file provide a list of expected logger statements
``` yaml
expected_logs:
- some long log here
- some other log here
- some more here
- this is the last line
```
- while some timeout is not hit, watch stdout for the expected logs, if all logs are detected within timeout then stop and exit code 0, otherwise 1

The tools need to be exposed in the `AEATestCase`.
Each skill test (`tests/test_packages/test_skills/*`) needs to have such a test.

@marcofavorito did I miss anything?

PR #1168, opened 2020-04-28
Check agent process output implemented AEA-488

panasevychol:
## Proposed changes

Check agent process output implemented and applied in echo test + some minor changes.
#1005 

## Types of changes

## Checklist

### DIFF
--- a/aea/test_tools/test_cases.py
+++ b/aea/test_tools/test_cases.py
@@ -25,10 +25,11 @@ import signal
 import subprocess  # nosec
 import sys
 import tempfile
+import time
 from io import TextIOWrapper
 from pathlib import Path
 from threading import Thread
-from typing import Any, Callable, List, Optional
+from typing import Any, Callable, Dict, List, Tuple
 
 import pytest
 
@@ -46,6 +47,8 @@ from aea.test_tools.exceptions import AEATestingException
 CLI_LOG_OPTION = ["-v", "OFF"]
 PROJECT_ROOT_DIR = "."
 
+DEFAULT_PROCESS_TIMEOUT = 60
+
 
 class AEATestCase:
     """Test case for AEA end-to-end tests."""
@@ -59,6 +62,8 @@ class AEATestCase:
     subprocesses: List  # list of launched subprocesses
     t: str  # temporary directory path
     threads: List  # list of started threads
+    stdout: Dict[int, str]  # dict of process.pid: string stdout
+    stderr: Dict[int, str]  # dict of process.pid: string stderr
 
     @classmethod
     def setup_class(cls, packages_dir_path: str = DEFAULT_REGISTRY_PATH):
@@ -86,6 +91,8 @@ class AEATestCase:
 
         cls.subprocesses = []
         cls.threads = []
+        cls.stdout = {}
+        cls.stderr = {}
 
         cls.author = AUTHOR
 
@@ -208,7 +215,7 @@ class AEATestCase:
 
     def terminate_agents(
         self,
-        subprocesses: Optional[List[subprocess.Popen]] = None,
+        *subprocesses: subprocess.Popen,
         signal: signal.Signals = signal.SIGINT,
         timeout: int = 10,
     ) -> None:
@@ -220,22 +227,22 @@ class AEATestCase:
         :param signal: the signal for interuption
         :timeout: the timeout for interuption
         """
-        if subprocesses is None:
-            subprocesses = self.subprocesses
+        if not subprocesses:
+            subprocesses = tuple(self.subprocesses)
+
         for process in subprocesses:
             process.send_signal(signal.SIGINT)
         for process in subprocesses:
             process.wait(timeout=timeout)
 
-    def is_successfully_terminated(
-        self, subprocesses: Optional[List[subprocess.Popen]] = None
-    ):
+    def is_successfully_terminated(self, *subprocesses: subprocess.Popen):
         """
         Check if all subprocesses terminated successfully
         """
-        if subprocesses is None:
-            subprocesses = self.subprocesses
-        all_terminated = [process.returncode == 0 for process in subprocesses]
+        if not subprocesses:
+            subprocesses = tuple(self.subprocesses)
+
+        all_terminated = all([process.returncode == 0 for process in subprocesses])
         return all_terminated
 
     def initialize_aea(self, author=None) -> None:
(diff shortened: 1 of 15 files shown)


aea/test_tools/test_cases.py
```diff
@@ -220,21 +245,19 @@ def terminate_agents(
         :param signal: the signal for interuption
         :timeout: the timeout for interuption
         """
-        if subprocesses is None:
-            subprocesses = self.subprocesses
+        if not subprocesses:
```
DavidMinarsch, line 248: I suggest reverting so you don't need the type ignore and there is also nothing wrong with the previous code.

aea/test_tools/test_cases.py
```diff
@@ -220,21 +245,19 @@ def terminate_agents(
...
-    def is_successfully_terminated(
-        self, subprocesses: Optional[List[subprocess.Popen]] = None
-    ):
+    def is_successfully_terminated(self, *subprocesses: subprocess.Popen):
         """
         Check if all subprocesses terminated successfully
         """
-        if subprocesses is None:
-            subprocesses = self.subprocesses
+        if not subprocesses:
```
DavidMinarsch, line 259: see above

aea/test_tools/test_cases.py
```diff
@@ -208,7 +233,7 @@ def run_agent(self, *args: str) -> subprocess.Popen:
 
     def terminate_agents(
         self,
-        subprocesses: Optional[List[subprocess.Popen]] = None,
+        *subprocesses: subprocess.Popen,
```
DavidMinarsch, line 236: The type declaration should remain unchanged I believe

DavidMinarsch, line 236: I think it should be:
```suggestion
        *subprocesses: Optional[subprocess.Popen] = None,
```

aea/test_tools/test_cases.py
```diff
@@ -182,6 +184,29 @@ def _run_python_subprocess(self, *args: str) -> subprocess.Popen:
...
+
+        :return: boolean is string in process output.
+        """
+        try:
+            output, _err = process.communicate(timeout=timeout)
+        except subprocess.TimeoutExpired:
+            self.terminate_agents(process)
+            output, _err = process.communicate()
+
+        return string in str(output)
```
DavidMinarsch, line 208: I think you'll need a loop here. The idea in pseudocode would be:
while time < timeout:
    # check if strings already present in output.
    if not:
        sleep for 0.5 sec
    else:
        break

aea/test_tools/test_cases.py
```diff
@@ -182,6 +184,29 @@ def _run_python_subprocess(self, *args: str) -> subprocess.Popen:
...
+        self,
+        process: subprocess.Popen,
+        string: str,
+        timeout: int = DEFAULT_PROCESS_TIMEOUT,
+    ) -> bool:
+        """
+        Check if string present in agent process output.
+
+        :param process: a subprocess object to communicate.
+        :param string: a substring expected to appear in process output.
```
DavidMinarsch, line 197: we want to check for a comma separated list of strings not just a single string.

tests/test_packages/test_skills/test_echo.py
```diff
@@ -76,6 +76,11 @@ def test_echo(self, pytestconfig):
         assert sent_envelope.protocol_id == received_envelope.protocol_id
         assert sent_envelope.message == received_envelope.message
 
-        self.terminate_agents([process])
+        check_string = "hello"
```
DavidMinarsch, line 79: Ok I can see this work for a single short string. How about a list of strings as mentioned in the issue?

aea/test_tools/test_cases.py
DavidMinarsch: The current implementation means you will always have to wait until the end of the timeout. But we want to succeed as early as we can to save time.

panasevychol: Tried that, not gonna work. `process.communicate()` stuck if trying to communicate to an active process.

aea/test_tools/test_cases.py
panasevychol: Will upgrade, makes sence.

aea/test_tools/test_cases.py
panasevychol: InvalidSyntax. We're unable to assign a default param to *args.

DavidMinarsch: Thanks

tests/test_packages/test_skills/test_thermometer.py
```diff
@@ -83,13 +83,16 @@ def test_thermometer(self, pytestconfig):
...
+        self._start_error_read_thread(thermometer_client_aea_process)
 
         time.sleep(20)
 
         self.terminate_agents()
 
         assert self.is_successfully_terminated(), "Thermometer test not successful."
+        import pdb
+
+        pdb.set_trace()
```
DavidMinarsch: please remove

aea/test_tools/test_cases.py
```diff
@@ -339,44 +346,77 @@ def replace_file_content(self, src: Path, dest: Path) -> None:
...
+        or timeout expired.
+
+        :param process: agent subprocess.
+        :param strings: tuple of strings expected to appear in output.
+        :param timeout: int amount of seconds before stopping check.
+        :param period: int period of checking.
+
+        :return: list of missed strings.
+        """
+        self._start_output_read_thread(process)
```
DavidMinarsch, line 399: You know, whilst we are at it. It might be helpful to start the error read also. And then in case of some missing strings, print it :)

DavidMinarsch: So after line 411 we'd print `stderr` if there are missing strings, otherwise not

DavidMinarsch: It's not guaranteed that a missing string coincides with an error, but it would be good to see the output

tests/test_docs/test_orm_integration/test_orm_integration.py
```diff
@@ -170,10 +170,10 @@ def test_orm_integration_docs_example(self, pytestconfig):
...
         process_two = self.run_agent("--connections", "fetchai/oef:0.2.0")
 
-        self.start_tty_read_thread(process_one)
-        self.start_error_read_thread(process_one)
-        self.start_tty_read_thread(process_two)
-        self.start_error_read_thread(process_two)
+        self._start_output_read_thread(process_one)
+        self._start_error_read_thread(process_one)
+        self._start_output_read_thread(process_two)
+        self._start_error_read_thread(process_two)
```
DavidMinarsch, line 176: Can be removed, we are not currently using the info here

tests/test_packages/test_skills/test_carpark.py
```diff
@@ -80,10 +80,10 @@ def test_carpark(self, pytestconfig):
...
         process_two = self.run_agent("--connections", "fetchai/oef:0.2.0")
 
-        self.start_tty_read_thread(process_one)
-        self.start_error_read_thread(process_one)
-        self.start_tty_read_thread(process_two)
-        self.start_error_read_thread(process_two)
+        self._start_output_read_thread(process_one)
+        self._start_error_read_thread(process_one)
+        self._start_output_read_thread(process_two)
+        self._start_error_read_thread(process_two)
```
DavidMinarsch, line 86: Can be removed, as we are not using it yet

tests/test_packages/test_skills/test_ml_skills.py
```diff
@@ -70,10 +70,10 @@ def test_ml_skills(self, pytestconfig):
...
         model_trainer_aea_process = self.run_agent("--connections", "fetchai/oef:0.2.0")
 
-        self.start_tty_read_thread(data_provider_aea_process)
-        self.start_error_read_thread(data_provider_aea_process)
-        self.start_tty_read_thread(model_trainer_aea_process)
-        self.start_error_read_thread(model_trainer_aea_process)
+        self._start_output_read_thread(data_provider_aea_process)
+        self._start_error_read_thread(data_provider_aea_process)
+        self._start_output_read_thread(model_trainer_aea_process)
+        self._start_error_read_thread(model_trainer_aea_process)
```
DavidMinarsch, line 76: not needed, we are not using it

tests/test_packages/test_skills/test_thermometer.py
```diff
@@ -83,10 +83,10 @@ def test_thermometer(self, pytestconfig):
...
         )
 
-        self.start_tty_read_thread(thermometer_aea_process)
-        self.start_error_read_thread(thermometer_aea_process)
-        self.start_tty_read_thread(thermometer_client_aea_process)
-        self.start_error_read_thread(thermometer_client_aea_process)
+        self._start_output_read_thread(thermometer_aea_process)
+        self._start_error_read_thread(thermometer_aea_process)
+        self._start_output_read_thread(thermometer_client_aea_process)
+        self._start_error_read_thread(thermometer_client_aea_process)
```
DavidMinarsch, line 89: not needed, we are not using it

tests/test_packages/test_skills/test_weather_ledger.py
```diff
@@ -75,10 +75,10 @@ def test_weather(self, pytestconfig):
...
         process_two = self.run_agent("--connections", "fetchai/oef:0.2.0")
 
-        self.start_tty_read_thread(process_one)
-        self.start_error_read_thread(process_one)
-        self.start_tty_read_thread(process_two)
-        self.start_error_read_thread(process_two)
+        self._start_output_read_thread(process_one)
+        self._start_error_read_thread(process_one)
+        self._start_output_read_thread(process_two)
+        self._start_error_read_thread(process_two)
```
DavidMinarsch, line 81: not needed, we are not using it

Issue #1106, opened 2020-04-16
Allow for fully programmatic usage via builder


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
Currently, adding packages to builder is done via file paths.

**Describe the solution you'd like**
Allow for programmatically adding packages too. And extend documentation accordingly.

DavidMinarsch:
In the words of @solarw:
```
One point I’m curious about is manual definition of skills inside a own written python module  and import in code with python import system: explicitly
Suppose it will be the most flexible way, may be I need to check some documentation to find it’s already possible

class Agent:
	Skills = [Skill(handlers=[Handler1])]
```

DavidMinarsch:
<img width="591" alt="Screenshot 2020-04-24 at 23 12 44" src="https://user-images.githubusercontent.com/19314578/80307598-ee058b00-87c1-11ea-93c9-50b65bff4a7d.png">

PR #1174, opened 2020-04-29
[AEA-563, AEA-569] better programmaticaly example. some updates on builder

solarw:
## Proposed changes
Better example to create AEA programmatically. AEABuilder.add_component_instance

## Fixes

- #1106

## Types of changes

## Checklist

### DIFF
--- a/aea/aea_builder.py
+++ b/aea/aea_builder.py
@@ -250,6 +250,7 @@ class AEABuilder:
         self._max_reactions: Optional[int] = None
 
         self._package_dependency_manager = _DependenciesManager()
+        self._connections: List[Connection] = []
 
         if with_default_packages:
             self._add_default_packages()
@@ -428,6 +429,19 @@ class AEABuilder:
         configuration._directory = directory
 
         return self
+
+    def add_component_instance(self, component: Component) -> None:
+        """
+        Add already initialized component object to resources or connections.
+
+        Please, pay attention, all dependencies have to be already loaded.
+
+        :params component: Component instance already initialized.
+        """
+        if component.component_type == ComponentType.CONNECTION:
+            self._connections.append(cast(Connection, component))
+        else:
+            self._add_component_to_resources(component)
 
     def set_context_namespace(self, context_namespace: Dict[str, Any]) -> None:
         """Set the context namespace."""
@@ -626,6 +640,8 @@ class AEABuilder:
         self._load_and_add_protocols()
         self._load_and_add_contracts()
         connections = self._load_connections(identity.address, connection_ids)
+        # add already loaded connections
+        connections += self._connections
         identity = self._update_identity(identity, wallet, connections)
         aea = AEA(
             identity,
@@ -781,7 +797,7 @@ class AEABuilder:
             with configuration_file_path.open(mode="r", encoding="utf-8") as fp:
                 loader = ConfigLoader.from_configuration_type(PackageType.AGENT)
                 agent_configuration = loader.load(fp)
-                logging.config.dictConfig(agent_configuration.logging_config)
+                logging.config.dictConfig(agent_configuration.logging_config)  # type: ignore
         except FileNotFoundError:
             raise Exception(
                 "Agent configuration file '{}' not found in the current directory.".format(
--- a/aea/aea_builder.py
+++ b/aea/aea_builder.py
@@ -250,12 +250,17 @@ class AEABuilder:
         self._max_reactions: Optional[int] = None
 
         self._package_dependency_manager = _DependenciesManager()
-        self._connections: List[Connection] = []
+        self._component_instances = {
+            ComponentType.CONNECTION: {},
+            ComponentType.CONTRACT: {},
+            ComponentType.PROTOCOL: {},
+            ComponentType.SKILL: {},
+        }  # type: Dict[ComponentType, Dict[ComponentConfiguration, Component]]
 
         if with_default_packages:
             self._add_default_packages()
 
-    def set_timeout(self, timeout: Optional[float]) -> None:
+    def set_timeout(self, timeout: Optional[float]) -> "AEABuilder":
         """
         Set agent loop idle timeout in seconds.
 
@@ -264,8 +269,9 @@ class AEABuilder:
         :return: None
         """
         self._timeout = timeout
+        return self
 
-    def set_execution_timeout(self, execution_timeout: Optional[float]) -> None:
+    def set_execution_timeout(self, execution_timeout: Optional[float]) -> "AEABuilder":
         """
         Set agent execution timeout in seconds.
 
@@ -274,8 +280,9 @@ class AEABuilder:
         :return: None
         """
         self._execution_timeout = execution_timeout
+        return self
 
-    def set_max_reactions(self, max_reactions: Optional[int]) -> None:
+    def set_max_reactions(self, max_reactions: Optional[int]) -> "AEABuilder":
         """
         Set agent max reaction in one react.
 
@@ -284,6 +291,7 @@ class AEABuilder:
         :return: None
         """
         self._max_reactions = max_reactions
+        return self
 
     def _add_default_packages(self) -> None:
         """Add default packages."""
@@ -430,7 +438,7 @@ class AEABuilder:
 
         return self
 
-    def add_component_instance(self, component: Component) -> None:
+    def add_component_instance(self, component: Component) -> "AEABuilder":
         """
         Add already initialized component object to resources or connections.
 
@@ -438,14 +446,18 @@ class AEABuilder:
 
         :params component: Component instance already initialized.
         """
-        if component.component_type == ComponentType.CONNECTION:
-            self._connections.append(cast(Connection, component))
-        else:
-            self._add_component_to_resources(component)
+        self._check_can_add(component.configuration)
+        # update dependency graph
+        self._package_dependency_manager.add_component(component.configuration)
+        self._component_instances[component.component_type][
+            component.configuration
+        ] = component
+        return self
 
-    def set_context_namespace(self, context_namespace: Dict[str, Any]) -> None:
+    def set_context_namespace(self, context_namespace: Dict[str, Any]) -> "AEABuilder":
         """Set the context namespace."""
         self._context_namespace = context_namespace
+        return self
 
     def _add_component_to_resources(self, component: Component) -> None:
         """Add component to the resources."""
@@ -640,8 +652,6 @@ class AEABuilder:
         self._load_and_add_protocols()
         self._load_and_add_contracts()
         connections = self._load_connections(identity.address, connection_ids)
-        # add already loaded connections
-        connections += self._connections
         identity = self._update_identity(identity, wallet, connections)
         aea = AEA(
             identity,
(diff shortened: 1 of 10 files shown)


aea/aea_builder.py
```diff
@@ -393,6 +394,19 @@ def add_component(
...
+        Add already initialized component object to resources or connections.
+
+        Please, pay attention, all dependencies have to be already loaded.
+
+        :params component: Component instance already initialized.
+        """
+        if component.component_type == ComponentType.CONNECTION:
+            self._connections.append(component)
+        else:
+            self._add_component_to_resources(component)
```
DavidMinarsch, line 408: @marcofavorito I sense this is missing some consistency checks but you are more familiar with this code. Please comment.

aea/aea_builder.py
```diff
@@ -590,6 +604,8 @@ def build(self, connection_ids: Optional[Collection[PublicId]] = None) -> AEA:
         self._load_and_add_protocols()
         self._load_and_add_contracts()
         connections = self._load_connections(identity.address, connection_ids)
+        # add alrady laoded connections
+        connections += self._connections
```
DavidMinarsch, line 608: what if a connection is in `self._connections` and also loaded above?

aea/aea_builder.py
DavidMinarsch: In any case, I think this would need to be called `load_` rather than `add_` to be in line with the rest of the builder conventions.

aea/skills/base.py
```diff
@@ -585,6 +585,9 @@ def __init__(
 
         self._contracts = {}  # type: Dict[str, Contract]
 
+        if self._skill_context:
```
DavidMinarsch, line 588: I prefer we keep the convention for explicit comparisons: `if self._skill_context is not None`:

docs/build-aea-programmatically.md
```diff
@@ -71,6 +73,40 @@ Next, we add the echo skill which will bounce our messages back to us. We first
...
+    from aea.protocols.default.message import DefaultMessage
+    from aea.skills.base import Handler
+
+    class DummyHandler(Handler):
+        """Dummy handler to handle messages."""
+
+        SUPPORTED_PROTOCOL = DefaultMessage.protocol_id
+
+        def setup(self) -> None:
+            """Noop setup."""
```
DavidMinarsch, line 89: What does `noop` mean?

docs/build-aea-programmatically.md
```diff
@@ -71,6 +73,40 @@ Next, we add the echo skill which will bounce our messages back to us. We first
...
+
+        def teardown(self) -> None:
+            """Noop teardown."""
+
+        def handle(self, message: Message) -> None:
+            """Handle incoming message."""
+            print("Hi, you got a ", message)
```
DavidMinarsch, line 96: please replace with `self.context.logger.info("You got a message: {}".format(str(message)))`

docs/build-aea-programmatically.md
```diff
@@ -71,6 +73,40 @@ Next, we add the echo skill which will bounce our messages back to us. We first
...
+
+    context = SkillContext()
+    config = SkillConfig()
```
DavidMinarsch, line 99: This is missing the public id elements

docs/build-aea-programmatically.md
```diff
@@ -71,6 +73,40 @@ Next, we add the echo skill which will bounce our messages back to us. We first
...
+    dummy_handler = DummyHandler(name="dummy_handler", skill_context=context)
```
DavidMinarsch, line 100: the name should be taken from the skill_config

docs/build-aea-programmatically.md
```diff
@@ -71,6 +73,40 @@ Next, we add the echo skill which will bounce our messages back to us. We first
...
+
+    skill_instance = Skill(
+        configuration=config,
+        skill_context=context,
+        handlers={"dummy_handler": dummy_handler},
```
DavidMinarsch, line 105: see above

DavidMinarsch: I need to check but I think the key needs to be the full skill id in str form...

docs/build-aea-programmatically.md
```diff
@@ -168,6 +206,37 @@ def run():
     # Add the echo skill (assuming it is present in the local directory 'packages')
     builder.add_skill("./packages/fetchai/skills/echo")
 
+    # create skill and handler manually
```
DavidMinarsch, line 209: as above

benchmark/framework/aea_test_wrapper.py
```diff
@@ -261,16 +260,22 @@ def set_fake_connection(
...
+
         envelope = envelope or self.dummy_envelope()
-        self.aea._connections.clear()
-        self.aea._connections.append(
-            FakeConnection(envelope, inbox_num, connection_id="fake_connection")
+        self._fake_connection = FakeConnection(
+            envelope, inbox_num, connection_id="fake_connection"
         )
+        # self.aea._connections.clear()
+        self.aea._connections.append(self._fake_connection)
```
DavidMinarsch, line 271: If you pass the connections to the AEA constructor you can avoid this

benchmark/framework/aea_test_wrapper.py
```diff
@@ -37,19 +37,18 @@
...
-    def __init__(self, name: str = "my_aea", skills: List[Dict[str, Dict]] = None):
+    def __init__(self, name: str = "my_aea", components: List[Component] = None):
         """
         Make an agency with optional name and skills.
 
         :param name: name of the agent
         :param skills: dict of skills to add to agent
         """
-        self.skills = skills or []
+        self.compmonents = components or []
```
DavidMinarsch, line 47: typo

benchmark/cases/react_speed_in_loop.py
```diff
@@ -35,8 +35,12 @@ def react_speed_in_loop(benchmark: BenchmarkControl, inbox_amount=1000) -> None:
 
     :return: None
     """
-    skill_definition = {"handlers": {"dummy_handler": DummyHandler}}
-    aea_test_wrapper = AEATestWrapper(name="dummy agent", skills=[skill_definition],)
+    aea_test_wrapper = AEATestWrapper(
+        name="dummy agent",
+        components=[
+            AEATestWrapper.make_skill(handlers={"dummy_handler": DummyHandler})
```
DavidMinarsch, line 41: I think the key needs to be the full public id in str form...but worth checking

aea/aea_builder.py
marcofavorito: I think we need to check if:
- we already have that component
- we have all the dependencies
- update the `_package_dependency_manager` object. 

basically, the same check we do on the `add_component` method.

marcofavorito: see https://github.com/fetchai/agents-aea/blob/v0.3.1/aea/aea_builder.py#L389-L391

marcofavorito: however, that means the user is responsible to provide good values in the configuration object attached to the component object. We didn't have that problem when we loaded them from the directory.

marcofavorito: > In any case, I think this would need to be called load_ rather than add_ to be in line with the rest of the builder conventions.

In my opinion `add_` is right, because the component is loaded outside the method. In other words, `component` is already loaded, so we just "add" it to the set of resources.

aea/aea_builder.py
```diff
@@ -590,6 +604,8 @@ def build(self, connection_ids: Optional[Collection[PublicId]] = None) -> AEA:
         self._load_and_add_protocols()
         self._load_and_add_contracts()
         connections = self._load_connections(identity.address, connection_ids)
+        # add alrady laoded connections
```
marcofavorito, line 607: fix typo: `already loaded` instead of `alrady laoded`

aea/aea_builder.py
```diff
@@ -244,6 +244,7 @@ def __init__(self, with_default_packages: bool = True):
         self._context_namespace = {}  # type: Dict[str, Any]
 
         self._package_dependency_manager = _DependenciesManager()
+        self._connections: List[Component] = []
```
marcofavorito, line 247: why not `List[Connection]`?

benchmark/cases/react_speed_in_loop.py
marcofavorito: no, we use just plain names for skill components.

docs/build-aea-programmatically.md
marcofavorito: I think is an abbreviation for [no operation](https://en.wikipedia.org/wiki/NOP_(code)).

docs/build-aea-programmatically.md
marcofavorito: yes, see my comment above.

I think that would be an argument to introduce heavy consistency checks on the fields of configurations object (and probably many others, like `Message`).

I we agree on that, worth considering in the future projects like Pydantic:  https://github.com/samuelcolvin/pydantic. Never used, but seems a good use case, since we have lots of data classes.

A nice standard package is `dataclasses`: https://docs.python.org/3/library/dataclasses.html. However, it's a pity that's only for Python 3.7+, where we also support Python 3.6.

benchmark/framework/aea_test_wrapper.py
```diff
@@ -261,16 +260,22 @@ def set_fake_connection(
 
         :return: None
         """
+        if self._fake_connection:
+            raise Exception("Fake connection is alrady set!")
```
marcofavorito, line 264: fix typo

docs/build-aea-programmatically.md
solarw: yes, it is. it is. it's empty, and nothing to write in comment.

aea/aea_builder.py
solarw: @marcofavorito 
yes, the main reason that there is no configuration provided as it done with packages.
so it differs from the standart dependency append.
this is why it's added directly to resources.

also I assume that developer knows what he does.

aea/aea_builder.py
solarw: the main idea that the developer is responsible for manual construction of the AEA.
so he had to control everything he does. and this also requires better configuration provided for components

aea/skills/base.py
solarw: done

docs/build-aea-programmatically.md
solarw: done, but suppose better to use f-strings

benchmark/framework/aea_test_wrapper.py
solarw: fixed, thanks

aea/aea_builder.py
solarw: will fix with cast, cause passed component and depending on it's type it's added as resource or connection

solarw: done

benchmark/framework/aea_test_wrapper.py
solarw: done

aea/aea_builder.py
DavidMinarsch: I don't think we can assume the developer knows what he does. It looks problematic to me that the checks are not applied. In particular, the dependency checks are needed to ensure the AEA can run successfully.

aea/aea_builder.py
DavidMinarsch: I don't think putting this responsibility on the developer will work. We should add consistency checks.

docs/build-aea-programmatically.md
DavidMinarsch: `Message` already has heavy consistency checks. SkillConfig is clearly missing them!

Further dependencies are not allowed at this stage. We are trying to reduce dependencies not add even more. This seems perfectly doable by ourselves.

docs/build-aea-programmatically.md
DavidMinarsch: Ok, confirmed, it only needs the name!

benchmark/cases/react_speed_in_loop.py
DavidMinarsch: Yes correct.

docs/build-aea-programmatically.md
DavidMinarsch: @solarw please, once this PR is merged (https://github.com/fetchai/agents-aea/pull/1180), merge develop and address the above by adding an `author` and a `name`.

aea/aea_builder.py
solarw: done.
agree with you, there is a difference in component construction.

benchmark/framework/aea_test_wrapper.py
solarw: it's a specific case for tests, and not assume will be used in production

solarw:
@DavidMinarsch 
latest develop  merged in

aea/aea_builder.py
```diff
@@ -429,9 +438,26 @@ def add_component(
...
+
+        :params component: Component instance already initialized.
+        """
+        self._check_can_add(component.configuration)
+        # update dependency graph
+        self._package_dependency_manager.add_component(component.configuration)
+        self._component_instances[component.component_type][
+            component.configuration
+        ] = component
+        return self
```
DavidMinarsch, line 455: We now mirror the checks we have when loading from dir.

aea/aea_builder.py
```diff
@@ -910,49 +936,142 @@ def get_connection_configuration(connection_id):
...
-            except Exception as e:
-                _handle_error_while_loading_component_generic_error(configuration, e)
+            if (
+                configuration
+                in self._component_instances[ComponentType.PROTOCOL].keys()
+            ):
+                protocol = self._component_instances[ComponentType.PROTOCOL][
+                    configuration
+                ]
+            else:
```
DavidMinarsch, line 953: we now first check if we have the component instance already in the builder, otherwise proceed to load from dir.

aea/aea_builder.py
```diff
@@ -910,49 +936,142 @@ def get_connection_configuration(connection_id):
...
+                    _handle_error_while_loading_component_generic_error(
+                        configuration, e
+                    )
             self._add_component_to_resources(skill)
 
+    def _load_connection(
+        self, address: Address, configuration: ConnectionConfig
+    ) -> Connection:
+        """
+        Load a connection from a directory.
```
DavidMinarsch, line 1023: we need access to self, so now exposed on builder

aea/aea_builder.py
```diff
@@ -910,49 +936,142 @@ def get_connection_configuration(connection_id):
...
+            connection = cast(
+                Connection,
+                self._component_instances[ComponentType.CONNECTION][configuration],
+            )
+            if connection.address != address:
+                logger.warning(
+                    "The address set on connection '{}' does not match the default address!".format(
+                        str(connection.connection_id)
+                    )
+                )
```
DavidMinarsch, line 1039: we give a warning if the addresses don't match (the developer has to take care to use the correct address to instantiate the connection)...

aea/skills/base.py
```diff
@@ -585,6 +587,8 @@ def __init__(
 
         self._contracts = {}  # type: Dict[str, Contract]
 
+        self._skill_context._skill = self
```
DavidMinarsch, line 590: We now take care of setting the skill context here, making usage more straightforward.

docs/build-aea-programmatically.md
```diff
@@ -71,6 +73,37 @@ Next, we add the echo skill which will bounce our messages back to us. We first
...
+            """Handle incoming message."""
+            self.context.logger.info("You got a message: {}".format(str(message)))
+
+    config = SkillConfig(name="test_skill", author="fetchai")
+    skill = Skill(configuration=config)
+    dummy_handler = DummyHandler(
+        name="dummy_handler", skill_context=skill.skill_context
+    )
+    skill.handlers.update({dummy_handler.name: dummy_handler})
+    builder.add_component_instance(skill)
```
DavidMinarsch, line 104: programmatic creation and adding is now straightforward.

PR #1217, opened 2020-05-11
Enable us to pass a Ledger Api object to the Agent Manager.

Totoual:
## Proposed changes

Enables the agent_builder to accept a ledger api object in the build function.

## Fixes

 - 

## Types of changes

## Checklist


aea/aea_builder.py
```diff
@@ -640,13 +640,23 @@ def _process_connection_ids(
...
 
+        :param ledger_api: the api ledger that we want to use.
         :param connection_ids: select only these connections to run the AEA.
         :return: the AEA object.
         """
+        _ledger_api = (
+            ledger_api
+            if ledger_api is not None
+            else LedgerApis(self.ledger_apis_config, self._default_ledger)
+        )
```
DavidMinarsch, line 659: We should add some assertions, that the ledger_apis_config and _default_ledger match with what has been set on `ledger_api`...otherwise this could lead to unexpected behaviour

Totoual: If you pass a `ledger_api` in the build function you are skipping this : 
`builder.add_ledger_api_config(FETCHAI, {"network": "testnet"})` 
So I am not sure where should I add the assertion. Possibly before we return the agent ?

Issue #1059, opened 2020-04-05
Decision maker API clean up


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
Currently, the decision maker API is not very accessible

**Describe the solution you'd like**
Improve the API, make it more accessible etc.

PR #1220, opened 2020-05-12
AEA-527, AEA-581 Refactor decision maker module, separating interface and implementation

DavidMinarsch:
## Proposed changes

This PR refactors the decision maker:
- it separates interface (base) from implementation (scaffold/default)
- it prepares for user defined dm

## Fixes

It addresses #1059

## Types of changes

## Checklist

## Further comments

-


aea/aea.py
```diff
@@ -28,7 +28,7 @@
 from aea.context.base import AgentContext
 from aea.crypto.ledger_apis import LedgerApis
 from aea.crypto.wallet import Wallet
-from aea.decision_maker.base import DecisionMaker
+from aea.decision_maker.active import DecisionMaker
```
DavidMinarsch, line 31: The active decision maker will be available in `.active`

aea/decision_maker/active.py
```diff
@@ -0,0 +1,26 @@
...
+#
+# ------------------------------------------------------------------------------
+
+"""This module contains the pointers to the active decision maker class and auxilliary classes."""
+
+from aea.decision_maker.default import DecisionMaker  # noqa
+from aea.decision_maker.default import GoalPursuitReadiness  # noqa
+from aea.decision_maker.default import LedgerStateProxy  # noqa
+from aea.decision_maker.default import OwnershipState  # noqa
+from aea.decision_maker.default import Preferences  # noqa
```
DavidMinarsch, line 26: For now we just import the default dm. The idea is that the user can provide a path to a user defined dm which then gets loaded here and is therefore available for import from here.

aea/decision_maker/base.py
```diff
@@ -100,133 +84,44 @@ def update(self, new_status: Status) -> None:
...
-        self._amount_by_currency_id = None  # type: Optional[CurrencyHoldings]
-        self._quantities_by_good_id = None  # type: Optional[GoodHoldings]
-
-    def _set(
-        self,
-        amount_by_currency_id: CurrencyHoldings,
-        quantities_by_good_id: GoodHoldings,
-    ) -> None:
+    @abstractmethod
+    def set(self, **kwargs) -> None:
```
DavidMinarsch, line 91: Here we have a problem: set should only be accessible to the dm..several ways to address it but non sticks out to me (e.g. could separate it into the Data class and access interface class, only access interface shared with skill context)

aea/decision_maker/base.py
```diff
@@ -100,133 +84,44 @@ def update(self, new_status: Status) -> None:
...
-        self._amount_by_currency_id = copy.copy(amount_by_currency_id)
-        self._quantities_by_good_id = copy.copy(quantities_by_good_id)
 
-    def _apply_delta(
-        self,
-        delta_amount_by_currency_id: Dict[str, int],
-        delta_quantities_by_good_id: Dict[str, int],
-    ) -> None:
+    @abstractmethod
+    def apply_delta(self, **kwargs) -> None:
```
DavidMinarsch, line 100: see above

aea/decision_maker/base.py
```diff
@@ -236,220 +131,61 @@ def apply_transactions(
...
-        self._quantity_shift = QUANTITY_SHIFT
-
-    def _set(
-        self,
-        exchange_params_by_currency_id: ExchangeParams,
-        utility_params_by_good_id: UtilityParams,
-        tx_fee: int,
-    ) -> None:
+    @abstractmethod
+    def set(self, **kwargs,) -> None:
```
DavidMinarsch, line 162: see above

aea/decision_maker/default.py
```diff
@@ -0,0 +1,837 @@
...
+#
+# ------------------------------------------------------------------------------
+
+"""This module contains the decision maker class."""
+
+import copy
+import logging
+import math
+from typing import Any, Dict, List, Optional, cast
+
```
DavidMinarsch, line 26: This is the same implementation we previously had in base

aea/decision_maker/scaffold.py
```diff
@@ -0,0 +1,189 @@
...
+from typing import List
+
+from aea.decision_maker.base import DecisionMaker as BaseDecisionMaker
+from aea.decision_maker.base import LedgerStateProxy as BaseLedgerStateProxy
+from aea.decision_maker.base import OwnershipState as BaseOwnershipState
+from aea.decision_maker.base import Preferences as BasePreferences
+from aea.decision_maker.messages.base import InternalMessage
+from aea.decision_maker.messages.transaction import TransactionMessage
+
+
```
DavidMinarsch, line 31: We can let the user scaffold a dm as a single python file in the root of the project. Scaffold will copy the file there and then add the filepath in aea config for loading to `.active`

aea/decision_maker/scaffold.py
```diff
@@ -0,0 +1,189 @@
...
+    def apply_delta(self, **kwargs,) -> None:
+        """
+        Apply a state update to the ownership state.
+
+        This method is used to apply a raw state update without a transaction.
+
+        :param
+        :return: None
+        """
+        raise NotImplementedError
```
DavidMinarsch, line 61: Note, these two methods should only be used from the decision maker not the skills. Current interface does not make that clear at all.

aea/decision_maker/scaffold.py
```diff
@@ -0,0 +1,189 @@
...
+
+    def set(self, **kwargs,) -> None:
+        """
+        Set values on the preferences.
+
+        :param kwargs: the relevant keyword arguments
+        """
+        raise NotImplementedError
```
DavidMinarsch, line 128: see above

aea/decision_maker/scaffold.py
```diff
@@ -0,0 +1,189 @@
...
+
+    def is_affordable_transaction(self, tx_message: TransactionMessage) -> bool:
+        """
+        Check if the transaction is affordable on the default ledger.
+
+        :param tx_message: the transaction message
+        :return: whether the transaction is affordable on the ledger
+        """
+        raise NotImplementedError
```
DavidMinarsch, line 110: With moving the ledger apis to connections, we will have to rethink this..it is not currently used so we can just keep it as is

aea/skills/base.py
```diff
@@ -43,7 +43,7 @@
 from aea.context.base import AgentContext
 from aea.contracts.base import Contract
 from aea.crypto.ledger_apis import LedgerApis
-from aea.decision_maker.base import GoalPursuitReadiness, OwnershipState, Preferences
+from aea.decision_maker.active import GoalPursuitReadiness, OwnershipState, Preferences
```
DavidMinarsch, line 46: note, here we load the implementation rather than base and then cast appropirately.

DavidMinarsch:
The scaffold command can be tested as such: `aea create ma && aea scaffold decision-maker`

aea/decision_maker/active.py
marcofavorito: see above on using `aea/decision_maker/__init__.py` to achieve that

marcofavorito: Another idea: rename `aea.decision_maker.default.DecisionMaker` with `DefaultDecisionMaker` and then in in the `__init__.py`:

```python
from aea.decision_maker.default import DefaultDecisionMaker as DecisionMaker
```

marcofavorito: It should be clear however that if you only need the interface you have to import from `aea.decision_maker.base`, whereas if you need the concrete class you have to import from `aea.decision_maker.some_module_name.(Concrete)DecisionMaker`.

Not prepending some word like `Default` could cause some clashes with the `base` name. 
So either rename `.base.DecisionMaker` (`AbstractDecisionMaker`, `IDecisionMaker`, `ADecisionMaker` are some ideas) or rename the concrete classes

aea/decision_maker/scaffold.py
marcofavorito: re. using `.active` please see above. In general, not sure I like the fact that the user should modify modules by hand. We should expose an API that let him doing that at runtime (maybe declaring it in the `aea-config.yaml` so the framework knows which class to replace)

marcofavorito: Also, what about using the `Strategy` pattern? The `DecisionMaker` class remains the same, whereas the user can control the decision maker behaviour via composition:

https://en.wikipedia.org/wiki/Strategy_pattern

aea/skills/base.py
marcofavorito: see above

aea/aea.py
```diff
@@ -87,16 +88,19 @@ def __init__(
 
         self.max_reactions = max_reactions
         self._task_manager = TaskManager()
-        self._decision_maker = DecisionMaker(identity, wallet, ledger_apis)
+        decision_maker_handler = DecisionMakerHandler(
+            identity=identity, wallet=wallet, ledger_apis=ledger_apis
+        )
+        self._decision_maker = DecisionMaker(
+            decision_maker_handler=decision_maker_handler
+        )
```
DavidMinarsch, line 96: @marcofavorito here it would now be possible to provide a user provided DMHandler

aea/context/base.py
```diff
@@ -126,19 +121,9 @@ def decision_maker_message_queue(self) -> Queue:
...
-        """Get the preferences of the agent."""
-        return self._preferences
-
-    @property
-    def goal_pursuit_readiness(self) -> GoalPursuitReadiness:
-        """Get the goal pursuit readiness."""
-        return self._goal_pursuit_readiness
+    def decision_maker_handler_context(self) -> SimpleNamespace:
+        """Get the decision maker handler context."""
+        return self._decision_maker_handler_context
```
DavidMinarsch, line 126: Very verbose. @marcofavorito any better naming suggestion?

aea/decision_maker/base.py
```diff
@@ -583,34 +261,77 @@ def protected_get(
...
+    def handle(self, message: InternalMessage) -> None:
+        """
+        Handle an internal message from the skills.
+
+        :param message: the internal message
+        :return: None
+        """
+
+
 class DecisionMaker:
```
DavidMinarsch, line 314: DM is now framework side only. DMHandler can be user defined

aea/decision_maker/scaffold.py
DavidMinarsch: Followed your advice and went for strategy pattern. We now create a handler for the decision maker, similar to the handler in the skills. The difference is, that objects passed to `kwargs` will be exposed to each skill context, whilst identity and crucially wallet are only accessible to dm handler. the ledger_apis will eventually be removed here.

tests/test_decision_maker/test_default.py
```diff
@@ -304,12 +309,22 @@ def test_decision_maker_handle_state_update_initialize_and_apply(self):
             tx_fee=tx_fee,
         )
         self.decision_maker.handle(state_update_message)
-        assert self.decision_maker.ownership_state.amount_by_currency_id is not None
-        assert self.decision_maker.ownership_state.quantities_by_good_id is not None
         assert (
-            self.decision_maker.preferences.exchange_params_by_currency_id is not None
+            self.decision_maker_handler.context.ownership_state.amount_by_currency_id
```
DavidMinarsch, line 313: I don't like the access pattern here..

aea/cli/scaffold.py
```diff
@@ -155,3 +163,44 @@ def _scaffold_item(click_context, item_type, item_name):
...
+        shutil.copyfile(src, dest)
+
+        # add the item to the configurations.
+        logger.debug(
+            "Registering the decision_maker into {}".format(DEFAULT_AEA_CONFIG_FILE)
+        )
+        import pdb
+
+        pdb.set_trace()
+        ctx.agent_config.decision_maker_path = str(dest)
```
marcofavorito, line 199: I suggest using the following format:

```
path.to.module:ClassName
```

It would give much more flexibility to the developer. That would point to a precise class.

marcofavorito: this is extendible to other components like protocols, connections, handlers, behaviours, models etc.

It imposes far less restrictions to the name of the classes. Of course, by convention, they could be called `XXXHandler`, but IMHO it shouldn't be a requirement.

marcofavorito: that being said, I think it's out of the scope of this PR, since it is a more general issue. Thoughts?

aea/context/base.py
marcofavorito: `dm_context`? We could drop `handler` because conceptually you can identify the handler with the decision-maker, since is the class that defines the DM behaviour the most.

Re. `dm`, maybe it is a reasonable abbreviation? If you ever care about the decision making... well you know what `dm` refers to.

aea/test_tools/test_cases.py
```diff
@@ -213,6 +231,55 @@ def fetch_agent(cls, public_id: str, agent_name: str) -> None:
         cls.run_cli_command("fetch", "--local", public_id, "--alias", agent_name)
         cls.agents.add(agent_name)
 
+    @classmethod
+    def difference_to_fetched_agent(cls, public_id: str, agent_name: str) -> List[str]:
```
DavidMinarsch, line 235: This can be improved in the future...

aea/cli/scaffold.py
```diff
@@ -155,3 +163,44 @@ def _scaffold_item(click_context, item_type, item_name):
...
+        logger.debug("Copying decision maker. src={} dst={}".format(src, dest))
+        shutil.copyfile(src, dest)
+
+        # add the item to the configurations.
+        logger.debug(
+            "Registering the decision_maker into {}".format(DEFAULT_AEA_CONFIG_FILE)
+        )
+        import pdb
+
+        pdb.set_trace()
```
marcofavorito, line 198: remove?

aea/test_tools/test_cases.py
```diff
@@ -170,6 +174,20 @@ def _run_python_subprocess(cls, *args: str, cwd: str = ".") -> subprocess.Popen:
         cls.subprocesses.append(process)
         return process
 
+    @classmethod
+    def start_subprocess(cls, *args: str, cwd: str = ".") -> subprocess.Popen:
```
marcofavorito, line 178: I wonder, why not make it more general? The `args` could just be forwarded to `Popen`.
(Maybe it already exists, but not sure, worth checking).

Then, you can define a new function `call_python_interpreter`, or something like that, that prepend the `sys.executable` and calls the generic `start_subprocess(sys.executable, *args)`

DavidMinarsch: Fair, but tbh I don't want to work any more on this class atm. We can do it when we need it.

aea/aea.py
```diff
@@ -87,16 +92,24 @@ def __init__(
...
+            DefaultDecisionMakerHandler
+            if decision_maker_handler_class is None
+            else decision_maker_handler_class
+        )
+        decision_maker_handler = decision_maker_handler_class(
+            identity=identity, wallet=wallet, ledger_apis=ledger_apis
+        )
+        self._decision_maker = DecisionMaker(
+            decision_maker_handler=decision_maker_handler
+        )
```
DavidMinarsch, line 105: @marcofavorito  if no decision maker class is provided we use the default...

aea/aea_builder.py
```diff
@@ -284,11 +292,36 @@ def set_max_reactions(self, max_reactions: Optional[int]) -> "AEABuilder":
...
+        :param file_path: the file path to the file which contains the decision maker handler
+
+        :return: self
+        """
+        dotted_path, class_name = decision_maker_handler_dotted_path.split(":")
+        module = load_module(dotted_path, file_path)
+        try:
+            import pdb; pdb.set_trace()
+            _class = getattr(module, class_name)
+            self._decision_maker_handler_class = _class
```
DavidMinarsch, line 316: @marcofavorito not currently working. Try: `aea create my_agent && cd my_agent && aea scaffold decision-maker-handler && aea run`. What we want is a `NotImplementedError` to be raised as the class is instantiated..but we cannot locate the class..

aea/aea_builder.py
```diff
@@ -875,6 +921,14 @@ def _set_from_configuration(
...
         self.set_execution_timeout(agent_configuration.execution_timeout)
         self.set_max_reactions(agent_configuration.max_reactions)
+        if agent_configuration.decision_maker_handler != {}:
+            dotted_path_full = (
+                agent_configuration.decision_maker_handler["dotted_path"]
+                + ":"
+                + agent_configuration.decision_maker_handler["class_name"]
+            )
+            file_path = agent_configuration.decision_maker_handler["file_path"]
+            self.set_decision_maker_handler(dotted_path_full, file_path)
```
DavidMinarsch, line 931: can be cleaned up when we got it to work

aea/cli/scaffold.py
```diff
@@ -155,3 +163,49 @@ def _scaffold_item(click_context, item_type, item_name):
...
+    try:
+        agent_name = ctx.agent_config.agent_name
+        click.echo(
+            "Adding decision maker scaffold to the agent '{}'...".format(agent_name)
+        )
+
+        # create the file name
+        dest = Path("decision_maker.py")
+        dotted_path = ".decision_maker"
+        class_name = "DecisonMakerHandler"
```
DavidMinarsch, line 189: loading not working like that...

aea/aea.py
marcofavorito: I think it can be simplified by putting `DefaultDecisionMakerHandler` as a default parameter:

```python
        ...
        is_debug: bool = False,
        max_reactions: int = 20,
        decision_maker_handler_class: Type[DecisionMakerHandler] = DefaultDecisionMakerHandler,
        **kwargs,
    ) -> None:
```

aea/cli/scaffold.py
marcofavorito: typo on `DecisonMakerHandler`, it should be "DecisionMakerHandler"

aea/aea_builder.py
marcofavorito: I think we could use the `dotted_path:ClassName` format in the configuration file, rather than building it in the code.

Also, worth exploring the `pytest` approach: `path/to/python/module.py::SomeClassName`

aea/aea_builder.py
marcofavorito: after the fix, it seems to work

PR #1221, opened 2020-05-12
Feature/asyncio agent event loop

solarw:
## Proposed changes
Switchable main loop for AEA. Sync and async on top of asyncio.

## Fixes
Should improve responsibility and decrease CPU usage.

## Types of changes

## Checklist


aea/agent_loop.py
```diff
@@ -0,0 +1,184 @@
...
+    def _stop_all_behaviours(self):
+        for scheduled_behaviour in self._registered_behaviours.values():
+            scheduled_behaviour.stop()
+        self._registered_behaviours = {}
+
+    def _unregister_behaviour(self, behaviour):
+        scheduled_behaviour = self._registered_behaviours.pop(behaviour, None)
+        if not scheduled_behaviour:
+            return
+        scheduled_behaviour.stop()
```
DavidMinarsch, line 144: All this happens outside the agent loop, the loop just accesses the resources

aea/agent_loop.py
```diff
@@ -0,0 +1,184 @@
...
+
+    def start(self):
+        self._state.set("started")
+        self._loop.run_until_complete(self._run())
+
+    async def _run(self):
+        for behaviour in self._behaviours:
+            self._register_behaviour(behaviour)
```
DavidMinarsch, line 152: registration happens during setup()

aea/agent_loop.py
```diff
@@ -0,0 +1,184 @@
...
+
+        tasks = self._processing_tasks()
+        await self._state.wait("stopping")
```
DavidMinarsch, line 155: we maintain task processing outside the main event loop for now (it was missed out in the Jira ticket, will add)

aea/agent_loop.py
```diff
@@ -0,0 +1,184 @@
...
+        self._registered_behaviours = {}
+
+    def _unregister_behaviour(self, behaviour):
+        scheduled_behaviour = self._registered_behaviours.pop(behaviour, None)
+        if not scheduled_behaviour:
+            return
+        scheduled_behaviour.stop()
+
+    def start(self):
+        self._state.set("started")
```
DavidMinarsch, line 147: This is missing the setup here

aea/agent_loop.py
```diff
@@ -0,0 +1,184 @@
...
+    @property
+    def is_running(self):
+        return self._state.state == "started"
+
+    async def _process_inbox(self):
+        # do we need process inbox if no handlers?
+        # probably yes, to avoid inbox overflow
+        while self.is_running:
+            msg = await self._inbox.get()
+            await self._handle_incoming_envelope(msg)
```
DavidMinarsch, line 180: Yes, in `_handle` of AEA we deal with the case of no handler present, so we need to call it for each envelope

aea/agent_loop.py
```diff
@@ -0,0 +1,184 @@
...
+            )
+
+    def start(self):
+        if self._timerhandle:
+            return
+        self._schedule_call()
+
+    def stop(self):
+        if not self._timerhandle:
+            return
```
DavidMinarsch, line 115: It would be more relevant to see this in the context of the existing behaviour class.

aea/agent_loop.py
```diff
@@ -0,0 +1,184 @@
...
+from asyncio.tasks import create_task
+
+
+def ensure_list(value):
+    if not isinstance(value, list):
+        value = [value]
+    return value
+
+
+class AsyncState:
```
DavidMinarsch, line 34: Can you provide motivation for this..

aea/agent.py
```diff
@@ -226,22 +250,10 @@ def _run_main_loop(self) -> None:
         :return: None
         """
         logger.info("[{}]: Start processing messages...".format(self.name))
-        while not self.liveness.is_stopped:
-            self._tick += 1
-            self._spin_main_loop()
+        self._main_loop = SyncAgentLoop(self)
```
DavidMinarsch, line 253: `self._main_loop` should already be set at this point ... so remove 253

aea/agent.py
```diff
@@ -226,22 +250,10 @@ def _run_main_loop(self) -> None:
...
+        self._main_loop.start()
```
DavidMinarsch, line 254: extract as its own method similar to AEA for symmetry?

aea/agent.py
DavidMinarsch:
```
def _run_main_loop(self) -> None:
        """
        Run the main loop of the agent.
        :return: None
        """
        logger.info("[{}]: Start processing messages...".format(self.name))
        self._main_loop.start()  # type: ignore
        logger.debug("[{}]: Exiting main loop...".format(self.name))
```

aea/agent.py
DavidMinarsch:
```
def _run_main_loop(self) -> None:
        """
        Run the main loop of the agent.
        :return: None
        """
        logger.info("[{}]: Start processing messages...".format(self.name))
        self._main_loop.start()  # type: ignore
        logger.debug("[{}]: Exiting main loop...".format(self.name))
```

aea/aea.py
```diff
@@ -291,3 +301,13 @@ def teardown(self) -> None:
...
         ExecTimeoutThreadGuard.stop()
+
+    def _run_main_loop(self) -> None:
+        """
+        Run the main loop of the agent.
+
+        :return: None
+        """
+        logger.info("[{}]: Start processing messages...".format(self.name))
+        self._main_loop.start()  # type: ignore
```
DavidMinarsch, line 312: why type ignore?

aea/agent_loop.py
```diff
@@ -0,0 +1,402 @@
...
+
+    def _unregister_behaviour(self, behaviour: TickerBehaviour) -> None:
+        """
+        Unregister periodic execution of the behaviour.
+
+        :param behaviour: TockerBehaviour to schedule periodic execution.
+        :return: None
+        """
+        periodic_caller = self._behaviours_registry.pop(behaviour, None)
+        if not periodic_caller:
```
DavidMinarsch, line 282: prefer explicit `if periodic_caller is None:`

aea/agent_loop.py
```diff
@@ -0,0 +1,402 @@
...
+                return
+
+            self._agent.react()
+
+    async def _task_process_internal_messages(self) -> None:
+        """Process decision maker's internal messages."""
+        queue = self._agent.decision_maker.message_out_queue
+        while self.is_running:
+            msg = await queue.async_get()
+            self._agent._filter._process_internal_message(msg)  # type: ignore
```
DavidMinarsch, line 365: why not create a property for `_agent` and I also think the `AEA` now  exposes a method which subsumes `_filter._process_internal_message(msg)`

DavidMinarsch: Otherwise it should to avoid daisy chaining the private methods/objects

aea/agent_loop.py
```diff
@@ -0,0 +1,402 @@
...
+
+    async def _task_process_new_behaviours(self) -> None:
+        """Process new behaviours added to skills in runtime."""
+        while self.is_running:
+            self._agent._filter._handle_new_behaviours()  # type: ignore
```
DavidMinarsch, line 370: see above

aea/agent_loop.py
DavidMinarsch: `type ignore` should not be needed

aea/agent_loop.py
```diff
@@ -0,0 +1,402 @@
...
+import datetime
+import logging
+import time
+from abc import ABC, abstractmethod
+from asyncio.events import AbstractEventLoop, TimerHandle
+from asyncio.futures import Future
+from asyncio.tasks import ALL_COMPLETED, FIRST_COMPLETED, Task
+from typing import Any, Callable, Dict, List, Optional, Set, Tuple, Union, cast
+
+from aea.skills.behaviours import TickerBehaviour
```
DavidMinarsch, line 30: Can we create a `TickerBehaviour` class here which avoids this import?

Then in `aea.skills.behaviours` we should import this class and have `Behaviour(TickerBehaviour, ABC)`

aea/connections/stub/connection.py
```diff
@@ -253,7 +253,7 @@ async def disconnect(self) -> None:
 
         assert self.in_queue is not None, "Input queue not initialized."
         self._observer.stop()
-        self._observer.join()
+        self._observer.join()  # TODO: blocking code?
```
DavidMinarsch, line 256: @solarw please open an issue and remove the comment, otherwise the hash will be invalidated

aea/connections/stub/connection.yaml
```diff
@@ -7,7 +7,7 @@ license: Apache-2.0
 aea_version: '>=0.3.0, <0.4.0'
 fingerprint:
   __init__.py: QmWwepN9Fy9gHAp39vUGFSLdnB9JZjdyE3STnbowSUhJkC
-  connection.py: QmNxHnzoRvmHAAjuUPFF8zLQXuEJrXVbAcm8fogXS2kKNh
+  connection.py: QmUgjVWhuDC8vgZCSgNVxMAqZ22GChJ1AZiYxxiwUTwrCQ
```
DavidMinarsch, line 10: see above

aea/mail/base.py
```diff
@@ -762,6 +762,25 @@ def get(
...
+    async def async_get(self) -> Envelope:
+        """
+        Get an envelope async way.
+
+        :return: the envelope
+        """
+        try:
+            return await self.in_queue.async_get()
+        except queue.Empty:
+            raise Empty
```
DavidMinarsch, line 774: Are we sure we want to label all coroutines with `async` in the method name?

benchmark/cases/react_speed.py
```diff
@@ -1,53 +0,0 @@
...
-#
-# ------------------------------------------------------------------------------
-"""Example performance test using benchmark framework. test react speed on amount of incoming messages."""
-from benchmark.cases.helpers.dummy_handler import DummyHandler
-from benchmark.framework.aea_test_wrapper import AEATestWrapper
-from benchmark.framework.benchmark import BenchmarkControl
-from benchmark.framework.cli import TestCli
-
-
-def react_speed(benchmark: BenchmarkControl, amount: int = 1000) -> None:
```
DavidMinarsch, line 27: Why removed?

packages/fetchai/skills/carpark_detection/behaviours.py
```diff
@@ -39,7 +38,7 @@
 DEFAULT_IMAGE_CAPTURE_INTERVAL = 300
 
 
-class CarParkDetectionAndGUIBehaviour(Behaviour):
+class CarParkDetectionAndGUIBehaviour(TickerBehaviour):
```
DavidMinarsch, line 41: See above, this needs to be reverted

packages/fetchai/skills/tac_control/behaviours.py
```diff
@@ -38,7 +37,7 @@
 )
 
 
-class TACBehaviour(Behaviour):
+class TACBehaviour(TickerBehaviour):
```
DavidMinarsch, line 40: See above, this needs to be reverted

packages/fetchai/skills/tac_control_contract/behaviours.py
```diff
@@ -52,7 +50,7 @@
 )
 
 
-class TACBehaviour(SimpleBehaviour):
+class TACBehaviour(TickerBehaviour):
```
DavidMinarsch, line 53: See above, this needs to be reverted

packages/fetchai/skills/tac_negotiation/behaviours.py
```diff
@@ -32,7 +31,7 @@
 from packages.fetchai.skills.tac_negotiation.transactions import Transactions
 
 
-class GoodsRegisterAndSearchBehaviour(Behaviour):
+class GoodsRegisterAndSearchBehaviour(TickerBehaviour):
```
DavidMinarsch, line 34: See above, this needs to be reverted

tests/data/dummy_skill/behaviours.py
```diff
@@ -16,13 +16,11 @@
...
 # ------------------------------------------------------------------------------
-
 """This module contains the behaviours for the 'dummy' skill."""
-
-from aea.skills.base import Behaviour
+from aea.skills.behaviours import TickerBehaviour
 
 
-class DummyBehaviour(Behaviour):
+class DummyBehaviour(TickerBehaviour):
```
DavidMinarsch, line 23: See above, this needs to be reverted

tests/data/exception_skill/behaviours.py
```diff
@@ -16,12 +16,11 @@
...
 #
 # ------------------------------------------------------------------------------
-
 """This module contains the behaviours for the 'exception' skill."""
-from aea.skills.base import Behaviour
+from aea.skills.behaviours import TickerBehaviour
 
 
-class ExceptionBehaviour(Behaviour):
+class ExceptionBehaviour(TickerBehaviour):
```
DavidMinarsch, line 23: See above, this needs to be reverted

tests/test_aea.py
```diff
@@ -268,221 +262,198 @@ def setup_class(cls):
...
+            sender=aea.identity.address,
             protocol_id=DefaultMessage.protocol_id,
-            message=DefaultSerializer().encode(cls.expected_message),
+            message=DefaultSerializer().encode(expected_message),
         )
-        cls.aea._start_setup()
-        cls.aea.outbox.put(envelope)
-        AeaTool(cls.aea).wait_inbox().spin_main_loop()
-
-    def test_initialize_aea_programmatically(self):
```
DavidMinarsch, line 294: What happened to this test?

aea/agent.py
solarw: sorry, removed.

aea/agent_loop.py
solarw: done

aea/mail/base.py
solarw: @DavidMinarsch it's about to discuss, cause get,put already present.

benchmark/cases/react_speed.py
solarw: the same test is made on a higher level and covers more part of the agent tested.

this was created earlier as for framework testing

.github/workflows/workflow.yml
```diff
@@ -35,8 +74,8 @@ jobs:
         run: tox -e bandit
       - name: Safety Check
         run: tox -e safety
-      - name: License Check
-        run: tox -e liccheck
+      #- name: License Check
+      #  run: tox -e liccheck
```
marcofavorito, line 78: Why has this been commented?

tests/test_helpers/test_async_friendly_queue.py
```diff
@@ -0,0 +1,83 @@
...
+        async def wait_msg(sq, timeout, results):
+            await asyncio.wait_for(sq.async_get(), timeout)
+            results.append("done")
+
+        loop.run_until_complete(wait_msg(sq, timeout, results))
+        loop.close()
+
+    for _ in range(num_threads):
+        t = Thread(target=test, args=(sq, 5, results))
+        t.setDaemon(True)
```
marcofavorito, line 72: use `t.daemon = True` instead:
https://docs.python.org/3.8/library/threading.html#threading.Thread.setDaemon

aea/agent_loop.py
solarw: +

.github/workflows/workflow.yml
solarw: @marcofavorito  for testing needs, cause liccheck fails on the new version of packaging module

aea/connections/stub/connection.py
solarw: done

aea/aea.py
solarw: fixed. caused by Optional type hint

aea/agent.py
solarw: method already extracted

tests/test_helpers/test_async_friendly_queue.py
solarw: done

aea/agent_loop.py
```diff
@@ -0,0 +1,400 @@
...
+        return list(map(create_task, tasks))
+
+    @property
+    def is_running(self) -> bool:
+        """Get running state of the loop."""
+        return self._state.state == "started"
+
+    async def _task_process_inbox(self) -> None:
+        """Process incoming messages."""
+        queue = self._agent._inbox
```
marcofavorito, line 349: minor, optional, for readability: rename it as `inbox`, because it's an `InBox` instance. Maybe also add some type annotation.

aea/helpers/async_friendly_queue.py
```diff
@@ -0,0 +1,84 @@
...
+        self._non_empty_waiters.append(waiter)
+        try:
+            await waiter
+        finally:
+            try:
+                self._non_empty_waiters.remove(waiter)
+            except ValueError:
+                pass
+
+    async def async_get(self) -> Any:
```
marcofavorito, line 73: maybe not now, but eventually: support for `block` and `timeout` parameters?

aea/helpers/async_friendly_queue.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+class AsyncFriendlyQueue(queue.Queue):
+    """queue.Queue with async_get method."""
+
+    def __init__(self, *args, **kwargs):
+        """Init queue."""
+        super().__init__(*args, **kwargs)
+        self._lock = threading.Lock()  # to cover _waiters access
+        self._non_empty_waiters = []
```
marcofavorito, line 33: use `deque` instead?
https://docs.python.org/3.8/library/collections.html#collections.deque

solarw: good point

aea/helpers/async_friendly_queue.py
```diff
@@ -0,0 +1,84 @@
...
+    def put(self, item: Any, *args, **kwargs) -> None:
+        """
+        Put an item into the queue.
+
+        :param item: item to put in the queue
+        :param args, kwargs: similar to queue.Queue.put
+        """
+        super().put(item, *args, **kwargs)
+        if self._non_empty_waiters:
+            waiter = self._non_empty_waiters.pop(0)
```
marcofavorito, line 44: if `deque` is used, this becomes 
```
waiter = self._non_empty_waiters.popleft(0) 
```

marcofavorito: In particular:
>Though list objects support similar operations, they are optimized for fast fixed-length operations and incur O(n) memory movement costs for pop(0) and insert(0, v) operations which change both the size and position of the underlying data representation.

aea/helpers/async_friendly_queue.py
solarw: `block` and `timeout` is redundant for coroutines:
* non block should be done by sync implementation and should be called instead of async block=False,
* timeout is done by `wait_for` with timeout option

aea/agent_loop.py
solarw: make sense, thanks

aea/helpers/async_friendly_queue.py
solarw: done

aea/helpers/async_friendly_queue.py
```diff
@@ -0,0 +1,85 @@
...
+from typing import Any
+
+
+class AsyncFriendlyQueue(queue.Queue):
+    """queue.Queue with async_get method."""
+
+    def __init__(self, *args, **kwargs):
+        """Init queue."""
+        super().__init__(*args, **kwargs)
+        self._lock = threading.Lock()  # to cover _waiters access
```
marcofavorito, line 33: is this needed? I haven't spotted usage

solarw: @marcofavorito  thanks, lock usage was removed.
fixed

aea/helpers/async_friendly_queue.py
```diff
@@ -0,0 +1,83 @@
...
+        self._non_empty_waiters = deque()
+
+    def put(self, item: Any, *args, **kwargs) -> None:
+        """
+        Put an item into the queue.
+
+        :param item: item to put in the queue
+        :param args, kwargs: similar to queue.Queue.put
+        """
+        super().put(item, *args, **kwargs)
```
marcofavorito, line 41: maybe we can do something slightly better (correct me if I'm mistaken). Basically, we want always to wake up only one thread that is waiting for messages.

With the current implementation, it seems to me we always wake up a coroutine, if any. But it might happen that there were already other threads waiting, and `super().put()` wakes up only one of them.

If that's the case, then we don't need to wake up any coroutine, because the waken thread will consume the message. This could save some time at scale.

That being said, I'm happy with the current implementation, and extra-work might be an overkill today. Thoughts?

solarw: It's async friendly queue and the main idea is to wake up coroutine on the item put.
It assumed to be used with coroutines not threads.
I see no reason to create a solution for one consumer only.

marcofavorito: > It assumed to be used with coroutines not threads.

You mean only coroutines can consume the items in the queue?
However, that's not the case since the parent `get()` is still callable (it just forwards the call to super().get()`)

solarw: I mean async_get created to support coroutines as well, queue can be consumed by get.
the main idea is to create quue to be consumable by synchronous and asynchronous code but not the same time.

solarw: I agree solution I provided is not perfect, but the main problem is to create queue thread safe and consumable by coroutines running in a different thread

solarw: @marcofavorito probably it's better to use  https://github.com/aio-libs/janus

marcofavorito: Yes, janus is an interesting project. Anyway as I said I'm happy with the current implementation and we can iterate over this if needed

.github/workflows/workflow.yml
marcofavorito: are we going to reintroduce it at some point? Maybe we can merge this PR and open an issue to keep track of it

aea/agent_loop.py
```diff
@@ -0,0 +1,403 @@
...
+        """Start agent loop."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def stop(self) -> None:
+        """Stop agent loop."""
+        raise NotImplementedError
+
+
+class AgentLoopException(Exception):
```
marcofavorito, line 193: inherit from `aea.exceptions.AEAException` ?

aea/agent_loop.py
```diff
@@ -0,0 +1,403 @@
...
+        self._state = initial_state
+        self._watchers: Set[Future] = set()
+        self._loop = loop or asyncio.get_event_loop()
+
+    @property
+    def state(self) -> Any:
+        """Return current state."""
+        return self._state
+
+    def set(self, state: Any) -> None:
```
marcofavorito, line 73: what about using a property setter?:
```python

@state.setter
def state(self, state: Any) -> None:
    # copy the implementation
```

aea/agent.py
DavidMinarsch: Yes in AEA, but not here. I am saying, let's do the same here..

aea/agent.py
```diff
@@ -101,6 +107,7 @@ def __init__(
         self._timeout = timeout
 
         self._tick = 0
+        self._main_loop: BaseAgentLoop = None  # type: ignore
```
DavidMinarsch, line 110: It should be `Optional[BaseAgentLoop] = None` and then the `type ignore` should be removed

solarw: done

aea/agent_loop.py
solarw: done

aea/agent_loop.py
solarw: done

aea/agent_loop.py
```diff
@@ -0,0 +1,406 @@
...
+
+    def _state_changed(self, state: Any) -> None:
+        """Fulfill watchers for state."""
+        for watcher in list(self._watchers):
+            if state in watcher._states:  # type: ignore
+                self._loop.call_soon_threadsafe(
+                    watcher.set_result, (self._state, state)
+                )
+
+    async def wait(self, state_or_states: Union[Any, List[Any]]) -> Tuple[Any, Any]:
```
marcofavorito, line 91: consider using `Sequence` instead of `List` if it doesn't matter whether it's a tuple or a list

aea/agent_loop.py
```diff
@@ -0,0 +1,406 @@
...
+        self._exceptions: List[Exception] = []
+
+    def start(self):
+        """Start agent loop."""
+        self._state.state = "started"
+        self._loop.run_until_complete(self._run())
+
+    def stop(self):
+        """Stop agent loop."""
+        self._state.state = "stopping"
```
marcofavorito, line 234: we could use an `enum.Enum` for the possible state values?

marcofavorito: I'm aware it requires non-trivial extra work, so take it as a suggestion, and if we have no time then create an issue if you consider it worthwhile.

aea/agent_loop.py
```diff
@@ -0,0 +1,406 @@
...
+    def _stop_all_behaviours(self) -> None:
+        """Unregister periodic execution of all registered behaviours."""
+        for periodic_caller in self._behaviours_registry.values():
+            periodic_caller.stop()
+        self._behaviours_registry = {}
+
+    def _create_tasks(self):
+        """Create tasks to execute and wait."""
+        tasks = self._create_processing_tasks()
+        stopping_task = create_task(self._state.wait(["stopping", "error"]))
```
marcofavorito, line 297: see above on using enums

aea/agent_loop.py
```diff
@@ -0,0 +1,406 @@
...
+            self._periodic_callable()
+        except Exception as exception:
+            if not self._exception_callback:
+                raise
+            self._exception_callback(self._periodic_callable, exception)
+
+    def _schedule_call(self) -> None:
+        """Set schedule for call."""
+        if self._period is None:
+            return
```
marcofavorito, line 151: can `self._period` be `None`? When?

From the constructor, I can see `self._period` is suggested to be of type `float`, not `Optional[float]`.

aea/agent_loop.py
solarw: it does not require extra work, but it's not assumed somebody will set state manually.
so it's internal states and I'm not sure enum is really required here.

aea/agent_loop.py
```diff
@@ -0,0 +1,406 @@
...
+
+        if self._timerhandle is None:
+            ts = time.mktime(self._start_at.timetuple())
+            delay = max(0, ts - time.time())
+            self._timerhandle = self._loop.call_later(delay, self._callback)
+        else:
+            self._timerhandle = self._loop.call_later(self._period, self._callback)
+
+    def start(self) -> None:
+        """Activate periodi calls."""
```
marcofavorito, line 161: fix typo

aea/agent_loop.py
solarw: done

aea/agent_loop.py
solarw: done

aea/agent_loop.py
solarw: done

aea/agent_loop.py
solarw: done

aea/agent_loop.py
solarw: done

aea/agent_loop.py
```diff
@@ -0,0 +1,406 @@
...
+        """
+        self._loop = loop or asyncio.get_event_loop()
+        self._periodic_callable = callback
+        self._start_at = start_at or datetime.datetime.now()
+        self._period = period
+        self._timerhandle: Optional[TimerHandle] = None
+        self._exception_callback = exception_callback
+
+    def _callback(self) -> None:
+        """Call on each schduled call."""
```
marcofavorito, line 139: fix typo

aea/agent_loop.py
```diff
@@ -0,0 +1,406 @@
...
+
+class PeriodicCaller:
+    """Schedule a periodic call of callable using event loop."""
+
+    def __init__(
+        self,
+        callback: Callable,
+        period: float,
+        start_at: Optional[datetime.datetime] = None,
+        exception_callback: Optional[Callable] = None,
```
marcofavorito, line 130: maybe specify which arguments the callable expects?

From the code below I can see it expects:
- the callback (hence, of type `Callable`
- the exception (type `Exception`)

so the full type should be:
```
exception_callback: Optional[Callable[[Callable], Exception]] = None,
```

do you agree?

aea/agent_loop.py
```diff
@@ -0,0 +1,406 @@
...
+        self._exception_callback = exception_callback
+
+    def _callback(self) -> None:
+        """Call on each schduled call."""
+        self._schedule_call()
+        try:
+            self._periodic_callable()
+        except Exception as exception:
+            if not self._exception_callback:
+                raise
```
marcofavorito, line 156: is it a good use case for `raise from exception`?

aea/agent_loop.py
solarw: done

aea/agent_loop.py
```diff
@@ -0,0 +1,406 @@
...
+        self._loop = loop or asyncio.get_event_loop()
+        self._periodic_callable = callback
+        self._start_at = start_at or datetime.datetime.now()
+        self._period = period
+        self._timerhandle: Optional[TimerHandle] = None
+        self._exception_callback = exception_callback
+
+    def _callback(self) -> None:
+        """Call on each schduled call."""
+        self._schedule_call()
```
marcofavorito, line 151: Question: shouldn't `self._schedule_call()` be called after `_periodic_callable()`? Does it make any difference for you? Also, what if it is called only if the call to the periodic_callable is successful?

I think it's a question also related to behaviours because this class is going to be used mainly by them. To rephrase it: should the `act()` of a behaviour be rescheduled even though previous act calls were failing (i.e. raise exception)?

aea/agent_loop.py
solarw: it reraises the last handled exception.
https://docs.python.org/3/tutorial/errors.html

aea/agent_loop.py
solarw: yes, it's a nice point
updated.

aea/agent_loop.py
solarw: done

aea/agent_loop.py
solarw: no, it should not be called after. otherwise, there will be a shift on time of execution of `_periodic_callable()`.
example: execution takes 1 second, the period is 2 seconds, if we schedule it after execution, the next call will be in 3 seconds, not in 2.

failing of any act will be handled by agent, and the agent has to make a decision: keep going or stop running and cancel all scheduled tasks.

aea/agent_loop.py
marcofavorito: yes, but I was thinking that we could give some more information to the user on what happened: which was the periodic task, etc.
In other words, to enrich the exception, but raising from the same exception context so the stack trace won't be lost.

Anyway, not a very crucial feature to have at the moment. It was just a thought.

solarw: but it's already done, cause it's reraised in case no exception_callback provided, in exception callback there is a logging and agent termination code

aea/agent_loop.py
```diff
@@ -0,0 +1,426 @@
...
+
+        :return: None
+        """
+        if behaviour in self._behaviours_registry:
+            # already registered
+            return
+
+        periodic_caller = PeriodicCaller(
+            behaviour.act,
+            behaviour._tick_interval,
```
marcofavorito, line 281: with the current implementation, behaviour code is allowed to change its tick interval. I'm aware that this was the case also in the previous implementation. However, I wonder: is it our desired property?

marcofavorito: Moreover, do you see any problems in letting the user set `tick_interval` to `0.0`?
Maybe we could still read from `behaviour._tick_interval`, but we could do some check like:
```
max(behaviour._tick_interval, 0.001)
```

solarw: runtime change of interval was discussed and decided that it should be immutable in runtime, but can check with David.

if user set tick interval to 0, or -1  or None or string -it's a user's problem.
otherwise we can make strict parameters check and raise value error

aea/agent_loop.py
```diff
@@ -0,0 +1,426 @@
...
+    async def _task_process_inbox(self) -> None:
+        """Process incoming messages."""
+        inbox: InBox = self._agent._inbox
+        while self.is_running:
+            await inbox.async_wait()
+
+            if not self.is_running:  # make it close faster
+                return
+
+            self._agent.react()
```
marcofavorito, line 380: Whilst all behaviours are run concurrently, it seems that's not the case for handlers, because the messages are processed serially. 

I wonder if it could be better to run them concurrently. Probably it doesn't make much sense now because `Handler.handle` is not an `async` method, so all the `handle()` will be executed serially anyway. However, if it wasn't the case, there could have been some issues with side-effects over the handler objects.

Nothing required to do, just a thought I wanted to share...

solarw: it's a nice point actually.
the best way is to develop runtime truly async and just keep using sync callbacks for compatibility. but currently, e have two kinds of loops: sync and async, and we have to use sync approach as well. I think we can switch to better async approach next time with possible parallel act/handle in mind

.github/workflows/workflow.yml
marcofavorito: I think @DavidMinarsch solved this problem in another PR

aea/agent_loop.py
marcofavorito: >runtime change of interval was discussed and decided that it should be immutable in runtime, but can check with David.

However, it seems to me it is mutable because I can write something like:

```python

class MyBehaviour(Behaviour):

    def act(self):
        """Change tick interval at runtime"""
        self._tick_interval = self._tick_interval * 0.9

```

So the tick interval gets smaller and smaller, and if I deactivate and re-activate it, the new tick interval will be loaded. Anyway, it's a corner case, we can ignore it.

aea/agent_loop.py
```diff
@@ -0,0 +1,426 @@
...
+        )
+        self._behaviours_registry[behaviour] = periodic_caller
+        periodic_caller.start()
+
+    def _register_all_behaviours(self) -> None:
+        """Register all AEA behaviours to run periodically."""
+        for behaviour in self._agent._get_active_behaviours():
+            self._register_behaviour(behaviour)
+
+    def _unregister_behaviour(self, behaviour: Behaviour) -> None:
```
marcofavorito, line 294: it seems it's not used anywhere

solarw: moved to stop all behaviours

PR #1224, opened 2020-05-13
AEA-589, AEA-513 Dialogue api improvement

5A11:
## Proposed changes

Moved `is_valid_next_message` from Fipa to base dialogue
Added it as a precondition for extending a dialogue with a new message
Added empty_dialogue() method that returns a new empty dialogue
Removed validation of message from the get_dialogue() method
Changed the body of is_permited_for_new_dialogue to use the new empty_dialogue()

## Fixes

- #913 

## Types of changes

## Checklist


aea/helpers/dialogue/base.py
```diff
@@ -178,16 +178,26 @@ def outgoing_extend(self, message: "Message") -> None:
         :param message: a message to be added
         :return: None
         """
-        self._outgoing_messages.extend([message])
+        if self.is_valid_next_message(message):
```
DavidMinarsch, line 181: what if it is not valid? presumably we should raise a `ValueError`

aea/helpers/dialogue/base.py
```diff
@@ -178,16 +178,26 @@ def outgoing_extend(self, message: "Message") -> None:
...
     def incoming_extend(self, message: "Message") -> None:
         """
         Extend the list of messages which keeps track of incoming messages.
 
-        :param messages: a message to be added
+        :param message: a message to be added
         :return: None
         """
-        self._incoming_messages.extend([message])
+        if self.is_valid_next_message(message):
```
DavidMinarsch, line 191: See above

aea/helpers/dialogue/base.py
DavidMinarsch: Otherwise the failure mode is silent...

aea/helpers/dialogue/base.py
```diff
@@ -178,16 +178,26 @@ def outgoing_extend(self, message: "Message") -> None:
...
+            self._incoming_messages.extend([message])
+
+    def is_valid_next_message(self, message: "Message") -> bool:
```
DavidMinarsch, line 194: missing `@abstractmethod` decorator..remove `pass`

aea/helpers/dialogue/base.py
```diff
@@ -241,6 +251,25 @@ def get_dialogue(self, msg: Message, agent_addr: Address) -> Dialogue:
         :return: the dialogue
         """
 
+    @staticmethod
+    def empty_dialogue(
+        opponent_address: Address = "", starter_address: Address = ""
+    ) -> Dialogue:
+        """
+        Create an empty dialogue, optionally with the supplied opponent address and dialogue starter address.
+
```
DavidMinarsch, line 260: What is the purpose of an empty dialogue?

packages/fetchai/protocols/fipa/dialogues.py
```diff
@@ -335,18 +327,66 @@ def get_dialogue(self, fipa_msg: Message, agent_addr: Address) -> Dialogue:
...
-            if other_initiated_dialogue.is_valid_next_message(fipa_msg):
-                result = other_initiated_dialogue
+            result = other_initiated_dialogue
         if self_initiated_dialogue_label in self.dialogues:
             self_initiated_dialogue = cast(
                 FipaDialogue, self.dialogues[self_initiated_dialogue_label]
             )
-            if self_initiated_dialogue.is_valid_next_message(fipa_msg):
-                result = self_initiated_dialogue
+            result = self_initiated_dialogue
```
DavidMinarsch, line 335: Note, we most likely need to perform the is valid next message checks elsewhere then...otherwise some current implementations might fail..

packages/fetchai/protocols/fipa/dialogues.py
```diff
@@ -244,20 +244,12 @@ def is_permitted_for_new_dialogue(self, fipa_msg: Message) -> bool:
...
-        this_target = fipa_msg.target
-        this_performative = fipa_msg.performative
-
-        result = (
-            this_message_id == FipaDialogue.STARTING_MESSAGE_ID
-            and this_target == FipaDialogue.STARTING_TARGET
-            and this_performative == FipaMessage.Performative.CFP
-        )
+        empty_dialogue = self.empty_dialogue()
+        result = empty_dialogue.is_valid_next_message(fipa_msg)
```
DavidMinarsch, line 252: Does the empty dialogue perform checks equivalent to the LHS?

packages/fetchai/protocols/fipa/dialogues.py
```diff
@@ -335,18 +327,66 @@ def get_dialogue(self, fipa_msg: Message, agent_addr: Address) -> Dialogue:
...
+        :param agent_addr: the agent address
+        :return: the new or existing dialogue the message is intended for, or None in case of any errors.
+        """
+        fipa_msg = cast(FipaMessage, message)
+        dialogue_reference = fipa_msg.dialogue_reference
+
+        # new dialogue by other
+        if (
+            dialogue_reference[0] != ""
+            and dialogue_reference[1] == ""
```
DavidMinarsch, line 364: The dialogue reference object should expose a property `is_other_initiated` which provides this check. Same for `is_self_initiated`...then we hide this complexity and can alter it on the dialogue reference without affecting this implementation here.

aea/helpers/dialogue/base.py
DavidMinarsch: The problem is, if it can fail, we need to change the public api..so I would revert and instead add a second public api which does the check and raises on error and is called `try_outgoing_extend`

packages/fetchai/protocols/fipa/dialogues.py
```diff
@@ -335,18 +327,66 @@ def get_dialogue(self, fipa_msg: Message, agent_addr: Address) -> Dialogue:
...
+            and not is_self_message
+        ):
+            dialogue = self.create_opponent_initiated(
+                message.counterparty, fipa_msg.dialogue_reference, is_seller
+            )
+            dialogue.incoming_extend(message)
+        # new dialogue by self
+        elif (
+            dialogue_reference[0] != ""
+            and dialogue_reference[1] == ""
```
DavidMinarsch, line 374: see above

packages/fetchai/protocols/fipa/dialogues.py
```diff
@@ -335,18 +327,66 @@ def get_dialogue(self, fipa_msg: Message, agent_addr: Address) -> Dialogue:
...
-                result = self_initiated_dialogue
+            result = self_initiated_dialogue
         if result is None:
             raise ValueError("Should have found dialogue.")
         return result
 
+    def update(
+        self,
+        message: Message,
+        is_self_message: bool,
```
DavidMinarsch, line 343: is_self_message? why is this needed

DavidMinarsch: We can infer it from the state of the dialogue..

aea/helpers/dialogue/base.py
5A11: I like the idea of creating a second pair of extend. 
The new ones could extend the dialogue with the new message if the message is valid. Returns true if it does, false if it doesn't.

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: Used in the `is_permitted_for_new_dialogue` method

packages/fetchai/protocols/fipa/dialogues.py
5A11: The call to `is_valid_next_message` must do that for the empty dialogue.

packages/fetchai/protocols/fipa/dialogues.py
5A11: The new way, the checks should be done in the extend calls. 
result = incoming/outoing_extend(msg)
if not result:
     # failed
     # error handling
else:
     # succeeded

packages/fetchai/protocols/fipa/dialogues.py
5A11: There is no dialogue here. 
The idea is, you've got a message here, either created by yourself or received from another agent. 
Then you want to update your dialogue models with this new message. You first have to check if it is a new message for a new dialogue. if it is, then you have to check whether the message is by you or your opponent and create the new dialogue accordingly (call `self_initiated` vs `other_initiated`). That's where we need the is_self_message, since we cannot infer it from the message. If messages had the sender address, we could retrieve this info and call the appropriate `create_dialogue` method.

packages/fetchai/protocols/fipa/dialogues.py
5A11: How to know, from `dialogue_reference` alone, if the dialogue is self_initiated vs other_initiated? Am I wrong in assuming that In both cases, the agent first creating the dialogue just fills in the first element of `dialogue_reference` and the receiver fills in the second element?

5A11: I agree that If there is a way, then it should be done in the dialogue label class and hidden in the update method.

aea/helpers/dialogue/base.py
5A11: Done

packages/fetchai/protocols/fipa/dialogues.py
5A11: Its call to `is_valid_next_message` does the checks

aea/helpers/dialogue/base.py
```diff
@@ -182,63 +238,92 @@ def outgoing_extend(self, message: "Message") -> None:
...
 
+    def outgoing_safe_extend(self, message: "Message") -> bool:
+        """
+        Extend the list of outgoing messages with 'message', if 'message' is valid
+
+        :param message: a message to be added
+        :return: True if message successfully added, false otherwise
+        """
+        if self.is_valid_next_message(message):
+            self._outgoing_messages.extend([message])
```
DavidMinarsch, line 256: self.outgoing_extend(self, message: "Message")

aea/helpers/dialogue/base.py
```diff
@@ -182,63 +238,92 @@ def outgoing_extend(self, message: "Message") -> None:
...
+
+    def incoming_safe_extend(self, message: "Message") -> bool:
+        """
+        Extend the list of incoming messages with 'message', if 'message' is valid
+
+        :param message: a message to be added
+        :return: True if message successfully added, false otherwise
+        """
+        if self.is_valid_next_message(message):
+            self._incoming_messages.extend([message])
```
DavidMinarsch, line 269: see above

aea/helpers/dialogue/base.py
```diff
@@ -182,63 +238,92 @@ def outgoing_extend(self, message: "Message") -> None:
...
-        Check whether an agent message is permitted for a new dialogue.
-
-        :param msg: the agent message
-
-        :return: a boolean indicating whether the message is permitted for a new dialogue
-        """
+    @property
+    def agent_address(self) -> Address:
+        """Get the address of the agent for whom dialogues are maintained."""
+        return self._agent_address
```
DavidMinarsch, line 306: Add setter and getter for counterparty here as well.

aea/helpers/dialogue/base.py
```diff
@@ -182,63 +238,92 @@ def outgoing_extend(self, message: "Message") -> None:
...
-        :param agent_addr: the address of the agent
-
-        :return: boolean indicating whether the message belongs to a registered dialogue
+        :param message: the agent message
+        :return: the dialogue if such a dialogue is found, None otherwise
         """
 
-    @abstractmethod
-    def get_dialogue(self, msg: Message, agent_addr: Address) -> Dialogue:
+    def update(self, message: Message,) -> Optional[Dialogue]:
```
DavidMinarsch, line 317: add @abstractmethod

aea/protocols/base.py
```diff
@@ -78,6 +79,21 @@ def counterparty(self, counterparty: Address) -> None:
...
         self._counterparty = counterparty
 
+    @property
+    def is_incoming(self) -> bool:
+        """
+        Get the is_incoming value of the message.
+
+        :return whether the message is incoming or is out going
+        """
+        assert self._is_incoming is not None, "is_incoming must not be None."
```
DavidMinarsch, line 89: can drop assert

packages/fetchai/skills/weather_client/dialogues.py
```diff
@@ -48,6 +50,15 @@ def __init__(self, dialogue_label: DialogueLabel, is_seller: bool) -> None:
         FipaDialogue.__init__(self, dialogue_label=dialogue_label, is_seller=is_seller)
         self.proposal = None  # type: Optional[Description]
 
+    @staticmethod
+    def role_from_first_message(message: Message) -> BaseDialogue.Role:
```
DavidMinarsch, line 54: think

packages/fetchai/skills/weather_client/handlers.py
```diff
@@ -63,26 +65,23 @@ def handle(self, message: Message) -> None:
...
-        # handle message
-        if fipa_msg.performative == FipaMessage.Performative.PROPOSE:
-            self._handle_propose(fipa_msg, dialogue)
-        elif fipa_msg.performative == FipaMessage.Performative.DECLINE:
-            self._handle_decline(fipa_msg, dialogue)
-        elif fipa_msg.performative == FipaMessage.Performative.MATCH_ACCEPT_W_INFORM:
-            self._handle_match_accept(fipa_msg, dialogue)
-        elif fipa_msg.performative == FipaMessage.Performative.INFORM:
-            self._handle_inform(fipa_msg, dialogue)
+        else:
```
DavidMinarsch, line 73: drop else

packages/fetchai/skills/weather_client/handlers.py
```diff
@@ -150,7 +150,8 @@ def _handle_propose(self, msg: FipaMessage, dialogue: Dialogue) -> None:
                 target=new_target_id,
                 performative=FipaMessage.Performative.ACCEPT,
             )
-            dialogue.outgoing_extend(accept_msg)
+            accept_msg.counterparty = msg.counterparty
+            dialogues.update(accept_msg)
```
DavidMinarsch, line 154: I'd prefer if we also allowed for `update` on the dialogue module

packages/fetchai/skills/weather_client/handlers.py
```diff
@@ -347,9 +351,6 @@ def _handle_search(self, agents: Tuple[str, ...]) -> None:
             # pick first agent found
             opponent_addr = agents[0]
             dialogues = cast(Dialogues, self.context.dialogues)
-            dialogue = dialogues.create_self_initiated(
```
DavidMinarsch, line 350: missing dialogue creation

packages/fetchai/skills/weather_station/handlers.py
```diff
@@ -52,40 +52,24 @@ def handle(self, message: Message) -> None:
...
-        # handle message
-        if fipa_msg.performative == FipaMessage.Performative.CFP:
-            self._handle_cfp(fipa_msg, dialogue)
-        elif fipa_msg.performative == FipaMessage.Performative.DECLINE:
-            self._handle_decline(fipa_msg, dialogue)
-        elif fipa_msg.performative == FipaMessage.Performative.ACCEPT:
-            self._handle_accept(fipa_msg, dialogue)
-        elif fipa_msg.performative == FipaMessage.Performative.INFORM:
-            self._handle_inform(fipa_msg, dialogue)
+        else:
```
DavidMinarsch, line 62: see above

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: On second thought, counterparts here does not work, as different dialogues may be with different counterparts. 

We can add agent_address and counterparts properties for dialogue, but I suggest we hold that off until these field are needed.

packages/fetchai/skills/weather_client/handlers.py
5A11: Done

aea/protocols/base.py
5A11: Done

aea/helpers/dialogue/base.py
```diff
@@ -132,48 +134,153 @@ def __str__(self):
...
+            return self.value
+
+    class EndState(Enum):
+        """This class defines the end states of a dialogue."""
+
+        def __str__(self):
+            """Get the string representation."""
+            return self.value
+
+    def __init__(self, dialogue_label: DialogueLabel, role: Role) -> None:
```
DavidMinarsch, line 157: For backwards compatibility the role needs to have a default value set. If we do `role: Optional[Role] = None` we will have to take care of the optionality of this field ...alternatively, you could have a dummy role, like `EmptyRole` which is declared here, so `role: Role = EmptyRole(None)` where 
```
class EmptyRole(Role):
    EMPTY = None
```

Not sure what is best. Maybe there is a neater approach as well...

aea/helpers/dialogue/base.py
```diff
@@ -132,48 +134,153 @@ def __str__(self):
...
+
+        :return: the last message if it exists, None otherwise
+        """
+        if (
+            self.last_incoming_message is not None
+            and self.last_outgoing_message is not None
+        ):
+            last_incoming_message_id = cast(
+                int, self.last_incoming_message.get("message_id")
+            )
```
DavidMinarsch, line 245:
```suggestion
            last_incoming_message_id = self.last_incoming_message.message_id
```
is much cleaner

aea/helpers/dialogue/base.py
```diff
@@ -132,48 +134,153 @@ def __str__(self):
...
+            last_outgoing_message_id = cast(
+                int, self.last_outgoing_message.get("message_id")
+            )
```
DavidMinarsch, line 248: see above

aea/helpers/dialogue/base.py
```diff
@@ -132,48 +134,153 @@ def __str__(self):
...
+                if last_outgoing_message_id > last_incoming_message_id
+                else self.last_incoming_message
+            )
+        elif (
+            self.last_incoming_message is not None
+            and self.last_outgoing_message is None
+        ):
+            last_incoming_message_id = cast(
+                int, self.last_incoming_message.get("message_id")
+            )
```
DavidMinarsch, line 260: see above

aea/helpers/dialogue/base.py
```diff
@@ -132,48 +134,153 @@ def __str__(self):
...
+            return self.last_incoming_message if last_incoming_message_id == 1 else None
```
DavidMinarsch, line 261: That seems wrong...if there is a last incoming message, then this should be returned. The feasibility check (`if last_incoming_message_id == 1`) should always return `True` here as only feasible messages should be permitted into the dialogue

aea/helpers/dialogue/base.py
```diff
@@ -132,48 +134,153 @@ def __str__(self):
...
+        elif (
+            self.last_incoming_message is None
+            and self.last_outgoing_message is not None
+        ):
+            last_outgoing_message_id = cast(
+                int, self.last_outgoing_message.get("message_id")
+            )
+            return self.last_outgoing_message if last_outgoing_message_id == 1 else None
```
DavidMinarsch, line 269: see above

aea/helpers/dialogue/base.py
```diff
@@ -132,48 +134,153 @@ def __str__(self):
...
+        else:
+            return None
```
DavidMinarsch, line 271: It would be better to have one return statement at the end, like `return last_message` and do the assignment in the conditionals...

aea/helpers/dialogue/base.py
```diff
@@ -132,48 +134,153 @@ def __str__(self):
...
+
+    def is_empty(self) -> bool:
```
DavidMinarsch, line 273: Should be a property

aea/helpers/dialogue/base.py
```diff
@@ -182,70 +289,169 @@ def outgoing_extend(self, message: "Message") -> None:
...
+        Extend the list of outgoing messages with 'message', if 'message' is valid
+
+        :param message: a message to be added
+        :return: True if message successfully added, false otherwise
+        """
+        if self.is_valid_next_message(message):
+            self.outgoing_extend(message)
+            return True
+        else:
+            return False
```
DavidMinarsch, line 310:
```suggestion
        is_extended = False
        if self.is_valid_next_message(message):
            self.outgoing_extend(message)
            is_extended =  True
        return is_extended
```

aea/helpers/dialogue/base.py
```diff
@@ -182,70 +289,169 @@ def outgoing_extend(self, message: "Message") -> None:
...
+        Extend the list of incoming messages with 'message', if 'message' is valid
+
+        :param message: a message to be added
+        :return: True if message successfully added, false otherwise
+        """
+        if self.is_valid_next_message(message):
+            self.incoming_extend(message)
+            return True
+        else:
+            return False
```
DavidMinarsch, line 323: see above

aea/helpers/dialogue/base.py
```diff
@@ -182,70 +289,169 @@ def outgoing_extend(self, message: "Message") -> None:
...
+        :return: True if message successfully added, false otherwise
+        """
+        if self.is_valid_next_message(message):
+            if message.is_incoming:
+                self.incoming_extend(message)
+            else:
+                self.outgoing_extend(message)
+            return True
+        else:
+            return False
```
DavidMinarsch, line 339: why not use your safe extend methods to avoid two levels of nesting?

aea/helpers/dialogue/base.py
```diff
@@ -182,70 +289,169 @@ def outgoing_extend(self, message: "Message") -> None:
...
+        else:
+            all_messages = [
+                msg
+                for msg in itertools.chain(
+                    *itertools.zip_longest(
+                        self._incoming_messages, self._outgoing_messages
+                    )
+                )
+                if msg is not None
+            ]
```
DavidMinarsch, line 387: almost identical logic, create private helper method and call it.

aea/helpers/dialogue/base.py
```diff
@@ -182,70 +289,169 @@ def outgoing_extend(self, message: "Message") -> None:
...
+        :return: The role of the agent
+        """
+
+    def __str__(self) -> str:
+        """
+        Get the string representation.
+
+        :return: The string representation of the dialogue
+        """
+        rep = "Dialogue Label: " + str(self.dialogue_label) + "\n"
```
DavidMinarsch, line 366: we prefer full variable names and no abbreviations

aea/helpers/dialogue/base.py
```diff
@@ -182,70 +289,169 @@ def outgoing_extend(self, message: "Message") -> None:
...
+        rep = rep[:-1]
+        return rep
+
 
 class Dialogues:
-    """The dialogues class keeps track of all dialogues."""
+    """The dialogues class keeps track of all dialogues for an agent."""
 
-    def __init__(self) -> None:
+    def __init__(self, agent_address: Address) -> None:
```
DavidMinarsch, line 399: see above, the default address needs to be provided for backwards compatibility. Here you could do `agent_address: Address = ""` and then in the `agent_address` property assert that `assert self._agent_address != "", "Address not set!"`

aea/helpers/dialogue/base.py
DavidMinarsch: I think my comment was referring to dialogue not dialogues...

aea/helpers/dialogue/base.py
```diff
@@ -182,70 +289,169 @@ def outgoing_extend(self, message: "Message") -> None:
...
         """
         self._dialogues = {}  # type: Dict[DialogueLabel, Dialogue]
+        self._agent_address = agent_address
         self._dialogue_nonce = 0
 
     @property
     def dialogues(self) -> Dict[DialogueLabel, Dialogue]:
-        """Get dictionary of dialogues in which the agent is engaged in."""
+        """Get dictionary of dialogues in which the agent engages."""
         return self._dialogues
```
DavidMinarsch, line 414: We need to maintain all the historic APIs, although you can remove the decorator `@abstractmethod` so we no longer require them to be implemented in the derived class

packages/fetchai/protocols/fipa/dialogues.py
```diff
@@ -60,53 +59,48 @@
     ],
 }  # type: Dict[FipaMessage.Performative, List[Union[None, FipaMessage.Performative]]]
 
+SUPPLY_DATAMODEL_NAME = "supply"
+DEMAND_DATAMODEL_NAME = "demand"
```
DavidMinarsch, line 63: This needs to be defined elsewhere (`custom_types.py` for instance). It can not be generated...

packages/fetchai/protocols/fipa/dialogues.py
```diff
@@ -60,53 +59,48 @@
...
+    class AgentRole(Dialogue.Role):
+        """This class defines the agent's role in a fipa dialogue."""
 
         SELLER = "seller"
         BUYER = "buyer"
 
     def __init__(
-        self, dialogue_label: DialogueLabel, is_seller: bool, **kwargs
+        self, dialogue_label: DialogueLabel, role: AgentRole, **kwargs
     ) -> None:
```
DavidMinarsch, line 85: `**kwargs` can be dropped, we do not allow for them here.

packages/fetchai/protocols/fipa/dialogues.py
```diff
@@ -60,53 +59,48 @@
...
+        """Infer the role of the agent from an incoming/outgoing first message
 
-    @property
-    def role(self) -> "FipaDialogue.AgentRole":
-        """Get role of agent in dialogue."""
-        return self._role
+        :param message: an incoming/outgoing first message
+        :return: The role of the agent
+        """
+        pass
```
DavidMinarsch, line 103:
```suggestion
        role = role_from_first_message(message)
        self.role = role
```
Then `role_from_first_message` can be imported from `custom_types.py`

packages/fetchai/protocols/fipa/dialogues.py
```diff
@@ -208,180 +202,171 @@ def add_dialogue_endstate(
...
         """
         Initialize dialogues.
 
         :return: None
         """
-        Dialogues.__init__(self)
-        self._initiated_dialogues = {}  # type: Dict[DialogueLabel, FipaDialogue]
-        self._dialogues_as_seller = {}  # type: Dict[DialogueLabel, FipaDialogue]
-        self._dialogues_as_buyer = {}  # type: Dict[DialogueLabel, FipaDialogue]
+        Dialogues.__init__(self, agent_address)
```
DavidMinarsch, line 211: `agent_address=agent_address` as we will make it a kwarg

packages/fetchai/protocols/fipa/dialogues.py
```diff
@@ -208,180 +202,171 @@ def add_dialogue_endstate(
...
-
-    @property
-    def dialogues_as_buyer(self) -> Dict[DialogueLabel, FipaDialogue]:
-        """Get dictionary of dialogues in which the agent acts as a buyer."""
-        return self._dialogues_as_buyer
-
     @property
     def dialogue_stats(self) -> FipaDialogueStats:
         """Get the dialogue statistics."""
         return self._dialogue_stats
```
DavidMinarsch, line 218: Should be defined in base imo

packages/fetchai/protocols/fipa/dialogues.py
```diff
@@ -208,180 +202,171 @@ def add_dialogue_endstate(
...
             dialogue_reference, fipa_msg.counterparty, fipa_msg.counterparty
         )
-        result = False
         if other_initiated_dialogue_label in self.dialogues:
             other_initiated_dialogue = cast(
                 FipaDialogue, self.dialogues[other_initiated_dialogue_label]
             )
-            result = other_initiated_dialogue.is_valid_next_message(fipa_msg)
+            result = other_initiated_dialogue
         if self_initiated_dialogue_label in self.dialogues:
```
DavidMinarsch, line 241: Should it be if elif?

packages/fetchai/skills/weather_client/dialogues.py
```diff
@@ -36,18 +38,26 @@
...
         """
         Initialize a dialogue label.
 
         :param dialogue_label: the identifier of the dialogue
-        :param is_seller: indicates whether the agent associated with the dialogue is a seller or buyer
 
         :return: None
         """
-        FipaDialogue.__init__(self, dialogue_label=dialogue_label, is_seller=is_seller)
+        FipaDialogue.__init__(self, dialogue_label, FipaDialogue.AgentRole.BUYER)
```
DavidMinarsch, line 49: Why hard code it? I'd prefer we implement `role_from_first_message` properly below

packages/fetchai/skills/weather_client/handlers.py
```diff
@@ -63,14 +65,9 @@ def handle(self, message: Message) -> None:
...
-        if dialogues.is_belonging_to_registered_dialogue(
-            fipa_msg, self.context.agent_address
-        ):
-            dialogue = cast(
-                Dialogue, dialogues.get_dialogue(fipa_msg, self.context.agent_address)
-            )
-            dialogue.incoming_extend(fipa_msg)
-        else:
+        fipa_dialogue = dialogues.update(fipa_msg)
+        dialogue = cast(Dialogue, fipa_dialogue)
```
DavidMinarsch, line 69: very confusing with the naming, how about we just do: `dialogue = cast(Dialogue, dialogues.update(fipa_msg))`

packages/fetchai/skills/weather_client/handlers.py
```diff
@@ -125,6 +122,7 @@ def _handle_propose(self, msg: FipaMessage, dialogue: Dialogue) -> None:
         :param dialogue: the dialogue object
         :return: None
         """
+        dialogues = cast(Dialogues, self.context.dialogues)
```
DavidMinarsch, line 125: remove

packages/fetchai/skills/weather_client/handlers.py
DavidMinarsch: Change to `dialogue.update(accept_msg)` so no need to import in line 125

packages/fetchai/skills/weather_client/handlers.py
```diff
@@ -169,7 +168,8 @@ def _handle_propose(self, msg: FipaMessage, dialogue: Dialogue) -> None:
                 target=new_target_id,
                 performative=FipaMessage.Performative.DECLINE,
             )
-            dialogue.outgoing_extend(decline_msg)
+            decline_msg.counterparty = msg.counterparty
+            dialogues.update(decline_msg)
```
DavidMinarsch, line 172: see above

packages/fetchai/skills/weather_client/handlers.py
```diff
@@ -253,7 +254,8 @@ def _handle_match_accept(self, msg: FipaMessage, dialogue: Dialogue) -> None:
                 performative=FipaMessage.Performative.INFORM,
                 info={"Done": "Sending payment via bank transfer"},
             )
-            dialogue.outgoing_extend(inform_msg)
+            inform_msg.counterparty = msg.counterparty
+            dialogues.update(inform_msg)
```
DavidMinarsch, line 258: see above

packages/fetchai/skills/weather_client/handlers.py
```diff
@@ -358,12 +357,13 @@ def _handle_search(self, agents: Tuple[str, ...]) -> None:
...
                 message_id=Dialogue.STARTING_MESSAGE_ID,
-                dialogue_reference=dialogue.dialogue_label.dialogue_reference,
+                dialogue_reference=dialogues.new_self_initiated_dialogue_reference(),
                 performative=FipaMessage.Performative.CFP,
                 target=Dialogue.STARTING_TARGET,
                 query=query,
             )
-            dialogue.outgoing_extend(cfp_msg)
+            cfp_msg.counterparty = opponent_addr
+            dialogues.update(cfp_msg)
```
DavidMinarsch, line 366: see above

packages/fetchai/skills/weather_station/dialogues.py
```diff
@@ -36,19 +38,26 @@
...
         """
         Initialize a dialogue label.
 
         :param dialogue_label: the identifier of the dialogue
-        :param is_seller: indicates whether the agent associated with the dialogue is a seller or buyer
-
         :return: None
         """
-        FipaDialogue.__init__(self, dialogue_label=dialogue_label, is_seller=is_seller)
+        FipaDialogue.__init__(self, dialogue_label, FipaDialogue.AgentRole.SELLER)
```
DavidMinarsch, line 48: see above

tests/test_packages/test_protocols/test_fipa.py
```diff
@@ -614,3 +614,38 @@ def test_dialogues_self_initiated_is_seller(self):
...
+        )
+        proposal_msg.counterparty = self.buyer_addr
+
+        self.seller_dialogues.update(proposal_msg)
+        proposal_msg.counterparty = self.seller_addr
+        proposal_msg.is_incoming = True
+        self.buyer_dialogues.update(proposal_msg)
+
+        print(buyer_dialogue)
+        print(seller_dialogue)
```
DavidMinarsch, line 651: what are you actually testing? no assertions...

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: Remove if not needed.

aea/helpers/dialogue/base.py
```diff
@@ -132,48 +134,153 @@ def __str__(self):
...
-        Initialize a dialogue label.
+        Initialize a dialogue.
 
         :param dialogue_label: the identifier of the dialogue
+        :param role: the role of the agent this dialogue is maintained for
 
         :return: None
         """
         self._dialogue_label = dialogue_label
+        self._role = role
```
DavidMinarsch, line 167: `self._role = None  # type: Optional[Role]`

aea/helpers/dialogue/base.py
```diff
@@ -132,48 +134,153 @@ def __str__(self):
...
+        """
         return self._dialogue_label
 
+    @property
+    def role(self) -> "Role":
+        """
+        Get the agent's role in the dialogue.
+
+        :return: the agent's role
+        """
```
DavidMinarsch, line 192: `assert self._role is not None, "Role not set!"`

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: It is needed. It replaces `is_seller` in the fipa dialogue constructor which was moved up to the base class. 

Done

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: Done

packages/fetchai/protocols/fipa/dialogues.py
5A11: Done

packages/fetchai/protocols/fipa/dialogues.py
5A11: Done

packages/fetchai/protocols/fipa/dialogues.py
5A11: Done

packages/fetchai/skills/weather_client/handlers.py
5A11: Done. line 69 was unnecessary

aea/helpers/dialogue/base.py
5A11: Done

packages/fetchai/protocols/fipa/dialogues.py
5A11: Remove the whole method

Done

packages/fetchai/protocols/fipa/dialogues.py
5A11: I changed it to that. But it there is no kwarg in the base constructor and I don't see why we might add it there.

packages/fetchai/protocols/fipa/dialogues.py
5A11: Next PR where I move stuff to the base.

packages/fetchai/skills/weather_client/dialogues.py
5A11: Done

packages/fetchai/skills/weather_client/dialogues.py
5A11: This method is needed for now. If I remove it the skill wouldn't work. It can be removed once a generic implementation for `role_from_first_message` is provided in the `custom_types` file for fipa. 

Note this implementation should work across all skills.

packages/fetchai/skills/weather_client/handlers.py
5A11: Need a small change for this to work. The `dialogue.update` now misses the `update_self_initiated_dialogue_label_on_second_message` functionality in `dialogues.update()` 

I will add this in the next PR where stuff are moved into base.

packages/fetchai/skills/weather_client/handlers.py
5A11: See above

packages/fetchai/skills/weather_client/handlers.py
5A11: See above

packages/fetchai/skills/weather_client/handlers.py
5A11: See above

packages/fetchai/skills/weather_client/handlers.py
5A11: See above

packages/fetchai/skills/weather_station/dialogues.py
5A11: Done

aea/helpers/dialogue/base.py
```diff
@@ -132,120 +134,655 @@ def __str__(self):
...
+        if self.is_empty:
+            result = True
+        else:
+            target = cast(int, message.get("target"))
+            last_target = self.last_message.target  # type: ignore
+            result = target == last_target + 1
+        return result
+
+    @abstractmethod
+    def initial_performative(self):
```
DavidMinarsch, line 438: @5A11 missing abstract performative class which can be used for typing here.

DavidMinarsch: We can add this in `aea/protocols/base.py`

packages/fetchai/protocols/fipa/custom_types.py
```diff
@@ -22,8 +22,37 @@
...
 from aea.helpers.search.models import Query as BaseQuery
-
+from aea.protocols.base import Message
 
 Description = BaseDescription
 
 Query = BaseQuery
+
+
+def role_from_first_message(message: Message) -> str:
```
DavidMinarsch, line 32: Remove it here

packages/fetchai/skills/weather_station/handlers.py
```diff
@@ -270,6 +256,7 @@ def _handle_inform(self, msg: FipaMessage, dialogue: Dialogue) -> None:
 
         strategy = cast(Strategy, self.context.strategy)
         if strategy.is_ledger_tx and ("transaction_digest" in msg.info.keys()):
+            is_valid = False
```
DavidMinarsch, line 258: ?

packages/fetchai/skills/thermometer/handlers.py
```diff
@@ -153,7 +136,7 @@ def _handle_cfp(self, msg: FipaMessage, dialogue: Dialogue) -> None:
             dialogue.proposal = proposal
             self.context.logger.info(
                 "[{}]: sending a PROPOSE with proposal={} to sender={}".format(
-                    self.context.agent_name, proposal.values, msg.counterparty[-5:],
+                    self.context.agent_name, msg.counterparty[-5:], proposal.values
```
DavidMinarsch: wrong order, revert please

packages/fetchai/skills/carpark_detection/handlers.py
```diff
@@ -149,18 +134,19 @@ def _handle_cfp(self, msg: FipaMessage, dialogue: Dialogue) -> None:
             dialogue.carpark_data = carpark_data
             dialogue.proposal = proposal
             self.context.logger.info(
-                "[{}]: sending sender={} a PROPOSE with proposal={}".format(
+                "[{}]: sending a PROPOSE with proposal={} to sender={}".format(
                     self.context.agent_name, msg.counterparty[-5:], proposal.values
```
DavidMinarsch, line 137: wrong order

packages/fetchai/protocols/fipa/dialogues.py
```diff
@@ -25,139 +25,97 @@
...
-                this_message_id == last_message_id + 1
-                and this_target == last_target + 1
-                and last_performative in VALID_PREVIOUS_PERFORMATIVES[this_performative]
-            )
-        return result
-
-    def assign_final_dialogue_label(self, final_dialogue_label: DialogueLabel) -> None:
+        return FipaMessage.Performative.CFP
+
+    def get_replies(self, performative) -> FrozenSet:
```
DavidMinarsch, line 108: missing type hints

packages/fetchai/skills/tac_negotiation/strategy.py
```diff
@@ -209,16 +210,21 @@ def _get_proposal_for_query(
...
         :param query: the query for which to build the proposal
-        :is_seller: whether the agent making the proposal is a seller or not
+        :param role: the role of the agent making the proposal (seller or buyer)
 
         :return: a description
         """
+        if role == Dialogue.AgentRole.SELLER:
+            is_seller = True
+        else:
+            is_seller = False
```
DavidMinarsch, line 226: how about `is_seller = role == Dialogue.AgentRole.SELLER` :)?

packages/fetchai/skills/tac_negotiation/strategy.py
```diff
@@ -322,10 +328,15 @@ def is_profitable_transaction(
...
         :param transaction_msg: the transaction_msg
-        :param is_seller: the bool indicating whether the agent is a seller.
+        :param role: the role of the agent (seller or buyer)
 
         :return: True if the transaction is good (as stated above), False otherwise.
         """
+        if role == Dialogue.AgentRole.SELLER:
+            is_seller = True
+        else:
+            is_seller = False
```
DavidMinarsch, line 338: see above

packages/fetchai/skills/tac_negotiation/transactions.py
```diff
@@ -96,18 +97,23 @@ def generate_transaction_message(
...
         :param dialogue_label: the dialogue label
-        :param is_seller: the agent is a seller
+        :param role: the role of the agent (seller or buyer)
         :param agent_addr: the address of the agent
         :return: a transaction message
         """
+        if role == Dialogue.AgentRole.SELLER:
+            is_seller = True
+        else:
+            is_seller = False
```
DavidMinarsch, line 115: see above

packages/fetchai/skills/tac_negotiation/transactions.py
```diff
@@ -307,17 +313,22 @@ def _register_transaction_with_time(self, transaction_id: TransactionId) -> None
...
-        :param as_seller: whether the agent is a seller or not
+        :param role: the role of the agent (seller or buyer)
         :raise AssertionError: if the transaction is already present.
 
         :return: None
         """
+        if role == Dialogue.AgentRole.SELLER:
+            as_seller = True
+        else:
+            as_seller = False
```
DavidMinarsch, line 330: see above

PR #1226, opened 2020-05-13
AEA-583 Add part of ledger api cosmos

DavidMinarsch:
## Proposed changes

This PR adds the ledger api for cosmos interaction. It does so in an adhoc fashion. The interface is maintained but the internals are hacky. Since this is for the testnet only we can start using it and improve it.

This PR increases test coverage of crypto module.

Useful resources:
- https://cosmos.network/rpc/v0.38.3

## Fixes

-

## Types of changes

## Checklist

## Further comments

-

### DIFF
--- a/aea/crypto/base.py
+++ b/aea/crypto/base.py
@@ -20,7 +20,7 @@
 """Abstract module wrapping the public and private key cryptography and ledger api."""
 
 from abc import ABC, abstractmethod
-from typing import Any, BinaryIO, Optional
+from typing import Any, BinaryIO, Optional, Tuple
 
 from aea.mail.base import Address
 
@@ -89,14 +89,14 @@ class Crypto(ABC):
     @abstractmethod
     def recover_message(
         self, message: bytes, signature: str, is_deprecated_mode: bool = False
-    ) -> Address:
+    ) -> Tuple[Address, ...]:
         """
-        Recover the address from the hash.
+        Recover the addresses from the hash.
 
         :param message: the message we expect
         :param signature: the transaction signature
         :param is_deprecated_mode: if the deprecated signing was used
-        :return: the recovered address
+        :return: the recovered addresses
         """
 
     @classmethod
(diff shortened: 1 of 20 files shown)


aea/crypto/base.py
```diff
@@ -89,14 +89,14 @@ def sign_transaction(self, transaction: Any) -> Any:
     @abstractmethod
     def recover_message(
         self, message: bytes, signature: str, is_deprecated_mode: bool = False
-    ) -> Address:
+    ) -> Tuple[Address, ...]:
```
DavidMinarsch, line 92: @marcofavorito a necessary breaking change (though not used anywhere in packages). Due to the properties of elliptic curves multiple addresses can match

aea/crypto/cosmos.py
```diff
@@ -236,7 +268,57 @@ def transfer(
         :param chain_id: the Chain ID of the Ethereum transaction. Default is 1 (i.e. mainnet).
         :return: tx digest if present, otherwise None
         """
-        raise NotImplementedError
+        result = self._try_get_account_number_and_sequence(crypto.address)
```
DavidMinarsch, line 271: Very adhoc, we can improve this in the weeks to come. But at least it works now and can be used.

aea/crypto/cosmos.py
```diff
@@ -126,30 +128,34 @@ def sign_transaction(self, transaction: Any) -> Any:
...
-        address = self.get_address_from_public_key(public_key)
-        return address
+        public_keys = [
+            verifying_key.to_string("compressed").hex()
+            for verifying_key in verifying_keys
+        ]
+        addresses = [
+            self.get_address_from_public_key(public_key) for public_key in public_keys
+        ]
+        return tuple(addresses)
```
DavidMinarsch, line 158: Unlike in ethereum, in cosmos any signature+message can be matched to two public keys/addresses since the `s` is not restricted.

aea/crypto/base.py
marcofavorito: ok let's keep it in mind!

aea/crypto/cosmos.py
marcofavorito: ok

Issue #881, opened 2020-03-02
Crypto.load and Crypto.dump methods


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
Currently we use
_create_fetchai_private_key()
_create_ethereum_private_key()

**Describe the solution you'd like**
Replace with Crypto.load and Crypto.dump methods

DavidMinarsch:
everything below this could be refactored by having a map from ledger id to crypto classes.
Pseudocode

```python
_from_ledger_id_to_crytpo_class = {
    FETCHAI: FetchAICrypto,
    ETHEREUM: EthereumCrypto
}

for ledger_id, crypto_class in _from_ledger_id_to_crytpo_class.items()
    private_key_path = aea_conf.private_key_paths.read(ledger_id)
    if private_key_path is None:
        crypto_class().dump(file_pointer)
    else:
        # use crypto_class.load somehow
```

Of course the code can be different, this is just to give the idea.
We could also find a smarter way to define the mapping. Maybe built dynamically from the `*Crypto` classes in `aea.crypto` package.

PR #1230, opened 2020-05-13
AEA-505, AEA-578 Implement crypto registry

### LINKED ISSUE #1183: Crypto API refactor
TBD

TBD

marcofavorito:
## Proposed changes

This PR adds the Crypto registry, inspired by the OpenAI Gym registry.

It lets you register new Crypto components at runtime by using `aea.crypto.register()`, and to create objects of the registered class with `aea.crypto.make()`

In particular:

- `register(id, entry_point, **kwargs)` takes in input:
    - `id`, that has to match `[^\d\W]\w*` (any Python identifier wiht ASCII characters), e.g. `fetchai`, `ethereum`
    - `entry_point`: a string of the form `path.to.module:className`
    - `kwargs`: additional arguments to be forwarded to the class constructor (WARNING: we are not actually using it atm)

- `make(id, module, **kwargs)`:
    - `id`: see above
    - `module`: an optional dotted path to a Python module that will be loaded before checking for registered cryptos. Hence, the loading of the module will trigger a `register` call, so the following check for the crypto associated to `id` is safe. The use case is that you know that some crypto class belongs to some PyPI package `aea-crypto-x` and, instead of explicitly do `import aea_crypto_x` to trigger the `register` call, you provide it as an argument. By convention, the calls to `register` are triggered when the top-level module is imported. (Notice: this is a common pattern: both in the library https://github.com/openai/gym/blob/master/gym/envs/__init__.py and in user-defined items https://github.com/openai/gym-soccer/blob/master/gym_soccer/__init__.py)
    - `kwargs`: keyword arguments to be forwarded to the crypto class constructor.

The PR includes also some checks on the parameters provided. It could be improved.

Another way to improve it is to let `entry_point` to be a callable, or even the class itself.

Related resources:
https://github.com/openai/gym/wiki/Environments
https://github.com/openai/gym/blob/master/docs/creating-environments.md

## Fixes

Fixes #1183 , #881

## Types of changes

## Checklist


aea/crypto/__init__.py
```diff
@@ -18,3 +18,9 @@
 # ------------------------------------------------------------------------------
 
 """This module contains the crypto modules."""
+
+from .registry import register, make  # noqa
+
+register(id="fetchai", entry_point="aea.crypto.fetchai:FetchAICrypto")
+
+register(id="ethereum", entry_point="aea.crypto.ethereum:EthereumCrypto")
```
DavidMinarsch, line 26: @marcofavorito will merge cosmos so you can add that too.

aea/crypto/registry.py
```diff
@@ -0,0 +1,218 @@
...
+        self.id = CryptoId(id)
+        self.entry_point = EntryPoint(entry_point)
+        self._kwargs = {} if kwargs is None else kwargs
+
+    def make(self, **kwargs) -> Crypto:
+        """Instantiates an instance of the crypto object with appropriate arguments."""
+        _kwargs = self._kwargs.copy()
+        _kwargs.update(kwargs)
+        cls = self.entry_point.load()
+        crypto = cls()
```
DavidMinarsch, line 126: I think we do have one optional kwarg at the moment: private_key_path

packages/fetchai/skills/erc1155_client/handlers.py
```diff
@@ -150,7 +151,7 @@ def _handle_propose(self, msg: FipaMessage, dialogue: Dialogue) -> None:
             )
             contract = cast(ERC1155Contract, self.context.contracts.erc1155)
             contract.set_address(
-                ledger_api=self.context.ledger_apis.ethereum_api,
+                ledger_api=self.context.ledger_apis.get_api(ETHEREUM),  # type: ignore
```
marcofavorito, line 154: I had to add `type: ignore` here and in analogous places because `contract.set_address` required `EthereumApi`. I preferred this apporach to `cast(EthereumApi, obj)` because we would have to import the class `EthereumApi`, hence increasing coupling.

Ideally, the contract should work with any ledger api, at design level.

aea/aea_builder.py
```diff
@@ -58,7 +58,7 @@
 from aea.contracts.base import Contract
 from aea.crypto.helpers import (
     IDENTIFIER_TO_KEY_FILES,
-    _create_private_key,
```
marcofavorito, line 61: now this is a public method

aea/crypto/base.py
```diff
@@ -30,6 +30,13 @@ class Crypto(ABC):
 
     identifier = "base"
 
+    def __init__(self, **kwargs):
```
marcofavorito, line 33: we need this so `aea.crypto.make` can forward the keyword argument.

I can see an opportunity to apply a further refactoring. Every subclass, so far, requires the `private_key_path` input. If `None`, _generate_ a key, o/w _load_ it. These two could be two different methods, and the constructor would implement the check.

aea/crypto/base.py
```diff
@@ -99,16 +106,6 @@ def recover_message(
         :return: the recovered addresses
         """
 
-    @classmethod
-    @abstractmethod
-    def load(cls, fp: BinaryIO) -> "Crypto":
```
marcofavorito, line 104: Notice: we do the loading in the constructor, if `private_key_path is None`.

Thoughts? O/w we have to change the constructor so to accept an already loaded sequence of bytes, aka private key.

aea/crypto/cosmos.py
```diff
@@ -179,16 +179,6 @@ def get_address_from_public_key(cls, public_key: str) -> str:
         address = bech32_encode("cosmos", five_bit_r)
         return address
 
-    @classmethod
-    def load(cls, fp: BinaryIO):
```
marcofavorito, line 183: see above

aea/crypto/ethereum.py
```diff
@@ -175,16 +175,6 @@ def get_address_from_public_key(cls, public_key: str) -> str:
         address = Web3.toChecksumAddress(raw_address)
         return address
 
-    @classmethod
-    def load(cls, fp: BinaryIO):
```
marcofavorito, line 179: see above

aea/crypto/fetchai.py
```diff
@@ -153,16 +154,6 @@ def get_address_from_public_key(cls, public_key: str) -> Address:
         address = str(FetchaiAddress(identity))
         return address
 
-    @classmethod
-    def load(cls, fp: BinaryIO):
```
marcofavorito, line 157: see above

aea/crypto/helpers.py
```diff
@@ -51,72 +46,21 @@
...
-    :raises: an exception if the private key is invalid.
+    :raises: ValueError if the identifier is invalid.
     """
     try:
-        with open(private_key_path, "r") as key:
-            data = key.read()
-            SigningKey.from_string(bytes.fromhex(data), curve=SECP256k1)
+        # to validate the file, we just try to create a crypto object
+        # with private_key_path as parameter
+        aea.crypto.make(ledger_id, private_key_path=private_key_path)
```
marcofavorito, line 63: validate <-> make a Crypto given the ledger id and a private key path

aea/crypto/helpers.py
```diff
@@ -129,94 +73,21 @@ def _try_validate_cosmos_private_key_path(
...
-    elif ledger_id == COSMOS:
-        _try_validate_cosmos_private_key_path(private_key_path, exit_on_error)
-    else:
-        raise ValueError(
-            "Ledger id {} is not valid.".format(repr(ledger_id))
-        )  # pragma: no cover
-
-
-def _create_private_key(ledger_id: str, private_key_file: Optional[str] = None) -> None:
+def create_private_key(ledger_id: str, private_key_file: Optional[str] = None) -> None:
```
marcofavorito, line 76: now this is a public "helper" method

aea/crypto/base.py
DavidMinarsch: Ok sure, can you add this?

aea/crypto/helpers.py
DavidMinarsch: Why do we not expose this on the crypto object instead?

docs/api/crypto/ledger_apis.md
```diff
@@ -49,6 +49,8 @@ Get the apis.
 <a name=".aea.crypto.ledger_apis.LedgerApis.has_fetchai"></a>
 #### has`_`fetchai
 
+TODO to remove!
```
DavidMinarsch, line 52: Can remove comment, will be auto generated anyway

aea/crypto/helpers.py
marcofavorito: You mean for instance why we don't implement a class method in the `Crypto` interface, like `Crypto.load(ledger_id, private_key_file)`>

It is surely doable. However, I think that would break the single-responsibility principle for the `Crypto` interface. It would add a couping between the `Crypto` interface and the `aea.crypto.(register|make)` APIs. If we later change the `aea.crypto.registry`, we have to change `Crypto.load` as well.

PR #1238, opened 2020-05-15
AEA-505, AEA-578 update packages for compatibility with any crypto/ledger

DavidMinarsch:
## Proposed changes

Makes skills compatible with any crypto/ledger api.

## Fixes

-

## Types of changes

## Checklist

## Further comments

-


aea/crypto/base.py
```diff
@@ -224,3 +224,18 @@ def generate_tx_nonce(self, seller: Address, client: Address) -> str:
         :param client: the address of the client.
         :return: return the hash in hex.
         """
+
+
+class FaucetApi(ABC):
```
DavidMinarsch, line 229: @marcofavorito now we have a common interface here too

aea/crypto/cosmos.py
```diff
@@ -33,19 +33,20 @@
...
 import requests
 
-from aea.crypto.base import Crypto, LedgerApi
+from aea.crypto.base import Crypto, FaucetApi, LedgerApi
 from aea.mail.base import Address
 
 logger = logging.getLogger(__name__)
 
-COSMOS = "cosmos"
+_COSMOS = "cosmos"
```
DavidMinarsch, line 41: @marcofavorito I think the use of this constant was dangerous as it could lead to this being out of sync with what is set as `identifier` class var on `Crypto` (especially now with user defined types). I replaced it throughout.

aea/crypto/helpers.py
```diff
@@ -87,58 +91,14 @@ def create_private_key(ledger_id: str, private_key_file: Optional[str] = None) -
...
-                    "Response: {}\nMessage: {}".format(
-                        response.status_code, response_dict.get("message")
-                    )
-                )  # pragma: no cover
-    except Exception as e:
-        logger.warning(
-            "An error occured while attempting to generate wealth:\n{}".format(e)
-        )
-        sys.exit(1)
+    faucet_api = IDENTIFIER_TO_FAUCET_APIS.get(identifier, None)
```
DavidMinarsch, line 102: @marcofavorito thanks to common interface now this is generic...we can also make it user extendable in the future

aea/crypto/base.py
marcofavorito: shouldn't this be incorporated into the Ledger APIs? Hence, be part of a connection?

DavidMinarsch: Not really, a ledger api and a faucet are different. Faucets only exist for test nets and are usually not part of third party ledger apis

Issue #997, opened 2020-03-28
Remove business logic from CLI


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
The CLI often contains business logic which should reside in other modules

**Describe the solution you'd like**
 In particular, any business logic not related to file operations or registry interaction should be moved to the relevant core AEA modules to allow for programmatic usage of the functionality as well.

Check & resolve:
- [ ] add.py
- [ ] common.py
- [ ] config.py
- [ ] core.py
- [ ] create.py
- [ ] fetch.py
- [ ] fingerprint.py
- [ ] generate.py
- [ ] init.py
- [ ] install.py
- [ ] launch.py
- [ ] list.py
- [ ] loggers.py
- [ ] login.py
- [ ] logout.py
- [ ] publish.py
- [ ] push.py
- [ ] register.py
- [ ] remove.py
- [ ] run.py
- [ ] scaffold.py
- [ ] search.py
- [ ] registry/*

PR #1244, opened 2020-05-18
WIP CLI add refactoring AEA-490

panasevychol:
## Proposed changes

CLI add refactoring (wip)
#997 

## Types of changes

## Checklist


aea/cli/add.py
```diff
@@ -181,58 +155,124 @@ def _add_item(click_context, item_type, item_public_id) -> None:
...
+
+    :return: None
+    """
     if item_type in {"connection", "skill"}:
-        _add_protocols(click_context, item_configuration.protocols)
+        _add_protocols(click_context, item_config.protocols)
 
     if item_type == "skill":
-        for contract_public_id in item_configuration.contracts:
-            if contract_public_id not in ctx.agent_config.contracts:
```
DavidMinarsch, line 196: why has this line been removed?

aea/cli/add.py
```diff
@@ -181,58 +155,124 @@ def _add_item(click_context, item_type, item_public_id) -> None:
...
+
+    :return: None.
+    """
+    fingerprint = _compute_fingerprint(
+        package_path, ignore_patterns=item_config.fingerprint_ignore_patterns
+    )
+    return item_config.fingerprint == fingerprint
+
+
+def _load_item_config(item_type: str, package_path: Path) -> ConfigLoader:
```
DavidMinarsch, line 226: The return type should be `Configuration`
This should be a class method of `Configuration` imo. Then it is more likely to be reused.

aea/cli/add.py
```diff
@@ -181,58 +155,124 @@ def _add_item(click_context, item_type, item_public_id) -> None:
...
+    ctx = cast(Context, click_context.obj)
+    for protocol_public_id in protocols:
+        if protocol_public_id not in ctx.agent_config.protocols:
+            logger.debug(
+                "Adding protocol '{}' to the agent...".format(protocol_public_id)
+            )
+            _add_item(click_context, "protocol", protocol_public_id)
+
+
+def _is_fingerprint_correct(package_path: Path, item_config) -> bool:
```
DavidMinarsch, line 211: This should be a generic helper. So not private, and in helper/ folder

aea/cli/add.py
```diff
@@ -181,58 +155,124 @@ def _add_item(click_context, item_type, item_public_id) -> None:
...
+    Add protocols to AEA by list of public IDs.
+
+    :param click_context: click context object.
+    :param protocols: a Collection of protocol public IDs to be added.
+
+    :return: None
+    """
+    ctx = cast(Context, click_context.obj)
+    for protocol_public_id in protocols:
+        if protocol_public_id not in ctx.agent_config.protocols:
```
DavidMinarsch, line 204: Just move these two lines to 253, no need for an extra method..confuses more than it helps

aea/cli/add.py
```diff
@@ -181,58 +155,124 @@ def _add_item(click_context, item_type, item_public_id) -> None:
...
+    item_config = _load_item_config(item_type, package_path)
+
+    if not _is_fingerprint_correct(package_path, item_config):
+        raise click.ClickException("Failed to add an item with incorrect fingerprint.")
+
+    _add_item_deps(click_context, item_type, item_config)
+    _register_item(ctx, item_type, item_public_id)
+
+
+def _is_item_present(ctx: Context, item_type: str, item_public_id: PublicId) -> bool:
```
DavidMinarsch, line 167: Seems like the `Configuration` should expose a method to check if any item of item type is present in it. Then this method is no longer needed.

aea/cli/add.py
```diff
@@ -181,58 +155,124 @@ def _add_item(click_context, item_type, item_public_id) -> None:
...
 
     if item_type == "skill":
-        for contract_public_id in item_configuration.contracts:
-            if contract_public_id not in ctx.agent_config.contracts:
-                _add_item(click_context, "contract", contract_public_id)
+        for contract_public_id in item_config.contracts:
+            _add_item(click_context, "contract", contract_public_id)
+
+
+def _register_item(ctx: Context, item_type: str, item_public_id: PublicId) -> None:
```
DavidMinarsch, line 260: `Configuration` should expose this as a method

aea/cli/add.py
panasevychol: Agree. Basically no need as we already check that in _is_item_present.

aea/cli/add.py
panasevychol: Requires Configuration upgrade. Currently AgentConfig (I suppose this method should belong to this class) doesn't have a property for cwd, well it seems it has `directory` property but in it's code I see that is never assigned and always None. So currently we can't check a package presence in agent folder.
I propose to leave it for now here and remove after AgentConfig upgrade so we can use a cwd from it's props (worth to create a ticket for that).
Correct me if I'm wrong.

PR #1251, opened 2020-05-20
CLI common module reorganized to a package AEA-490

panasevychol:
## Proposed changes

CLI common module reorganized to a package
#997 

## Types of changes

## Checklist


aea/cli/utils/config.py
```diff
@@ -0,0 +1,113 @@
...
+    except jsonschema.exceptions.ValidationError:
+        if is_exit_on_except:
+            raise click.ClickException(
+                "Agent configuration file '{}' is invalid. Please check the documentation.".format(
+                    DEFAULT_AEA_CONFIG_FILE
+                )
+            )
+
+
+def _init_cli_config() -> None:
```
DavidMinarsch, line 76: All functions imported in other modules should be public for consistency imo.

aea/cli/utils/constants.py
```diff
@@ -0,0 +1,39 @@
...
+    "connections",
+    "protocols",
+    "contracts",
+    "vendor",
+    "packages",
+    "aea",
+]
+
+
+from_string_to_type = dict(str=str, int=int, bool=bool, float=float)
```
DavidMinarsch, line 39: Make constant, i.e `FROM_STRING_TO_TYPE`?

aea/cli/utils/config.py
```diff
@@ -0,0 +1,113 @@
...
+
+import yaml
+
+from aea.cli.utils.constants import CLI_CONFIG_PATH
+from aea.cli.utils.context import Context
+from aea.cli.utils.package_utils import _load_yaml
+from aea.configurations.base import DEFAULT_AEA_CONFIG_FILE
+
+
+def try_to_load_agent_config(
```
DavidMinarsch, line 40: seems a prime candidate for making this a helper method on the `Context` object in `../utils/context.py` rather than a helper function.

aea/cli/utils/decorators.py
```diff
@@ -0,0 +1,172 @@
...
+    _compare_fingerprints,
+    _get_default_configuration_file_name_from_type,
+)
+from aea.configurations.loader import ConfigLoaders
+
+
+pass_ctx = click.make_pass_decorator(Context)
+
+
+def _validate_config_consistency(ctx: Context):
```
DavidMinarsch, line 47: Same here, prime candidate for method on `ctx` rather than a helper function?

aea/cli/utils/exceptions.py
```diff
@@ -0,0 +1,24 @@
...
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+# ------------------------------------------------------------------------------
+
+"""Module with exceptions of the aea cli."""
+
+
+class AEAConfigException(Exception):
```
DavidMinarsch, line 23: Should this inherit from a base exception we create in `aea/exceptions.py`?

aea/cli/utils/formatting.py
```diff
@@ -0,0 +1,62 @@
...
+
+"""Module with formatting utils of the aea cli."""
+
+from typing import Dict
+
+from aea.configurations.base import AgentConfig
+from aea.configurations.loader import ConfigLoader
+
+
+def _format_items(items):
```
DavidMinarsch, line 28: Would make both functions "public".

aea/cli/utils/package_utils.py
```diff
@@ -0,0 +1,377 @@
...
+from aea.configurations.loader import ConfigLoader
+from aea.crypto.helpers import (
+    IDENTIFIER_TO_KEY_FILES,
+    _try_validate_private_key_path,
+    create_private_key,
+)
+from aea.crypto.registry import registry
+
+
+def _verify_or_create_private_keys(ctx: Context) -> None:
```
DavidMinarsch, line 54: method on context?

aea/cli/utils/package_utils.py
```diff
@@ -0,0 +1,377 @@
...
+    try:
+        shutil.copytree(src, dest)
+    except Exception as e:
+        raise click.ClickException(str(e))
+
+    Path(ctx.cwd, "vendor", author_name, item_type_plural, "__init__.py").touch()
+    return Path(dest)
+
+
+def _find_item_locally(ctx, item_type, item_public_id) -> Path:
```
DavidMinarsch, line 225: this and next one could be methods on context?

tests/test_cli/test_list.py
```diff
@@ -174,20 +174,36 @@ def teardown_class(cls):
             pass
 
 
-@mock.patch("aea.cli.common.try_to_load_agent_config")
 class ListContractsCommandTestCase(TestCase):
```
cyenyxe, line 177: I don't know if it has anything to do with the changes applied to this class, but the test `TestListSkills::test_correct_output` is now failing in CI.

aea/cli/utils/click_utils.py
```diff
@@ -0,0 +1,119 @@
...
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+# ------------------------------------------------------------------------------
+
+"""Module with click utils of the aea cli."""
```
cyenyxe, line 20: I feel like the classes in this file would better belong in other parts:
- `ConnectionsOption` with any other code about option parsing
- `PublicIdParameter` and `AgentDirectory` with other code about context handling

aea/cli/utils/package_utils.py
```diff
@@ -0,0 +1,377 @@
...
+                )
+
+    # update aea config
+    path = Path(DEFAULT_AEA_CONFIG_FILE)
+    fp = path.open(mode="w", encoding="utf-8")
+    agent_loader.dump(aea_conf, fp)
+    ctx.agent_config = aea_conf
+
+
+def _validate_package_name(package_name: str):
```
cyenyxe, line 91: Are any of these functions called outside `packages`? In that case, even if they are standalone functions it may make more sense to include them in that folder, because they are not shared utilities per se.

aea/cli/utils/package_utils.py
```diff
@@ -0,0 +1,377 @@
...
+            aea_conf.private_key_paths.update(identifier, private_key_path)
+        else:
+            try:
+                _try_validate_private_key_path(identifier, private_key_path)
+            except FileNotFoundError:  # pragma: no cover
+                raise click.ClickException(
+                    "File {} for private key {} not found.".format(
+                        repr(private_key_path), identifier,
+                    )
+                )
```
cyenyxe, line 82: Probably not something to address in this PR, but this structure mixes internal logic exceptions with CLI ones. The `FileNotFoundError` should be raised up to the point where the top level CLI component has to handle it. This would isolate both as much as possible.

When working on this type of code I always think how to support both CLI and GUI, and that usually helps to come up with a modular design.

aea/cli/add.py
```diff
@@ -26,18 +26,17 @@
...
 
-from aea.cli.common import (
-    Context,
-    PublicIdParameter,
+from aea.cli.registry.utils import fetch_package
+from aea.cli.utils.click_utils import PublicIdParameter
+from aea.cli.utils.context import Context
+from aea.cli.utils.decorators import check_aea_project, clean_after
+from aea.cli.utils.loggers import logger
+from aea.cli.utils.package_utils import (
```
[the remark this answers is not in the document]
panasevychol: Partially I agree. But we have a module `click` so I think better to have a different name. As for `package` or `packages` I don't like how it sounds. Could be suggested that there's or 're some packages inside which is not true.

aea/cli/utils/__init__.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# ------------------------------------------------------------------------------
+#
+#   Copyright 2018-2019 Fetch.AI Limited
```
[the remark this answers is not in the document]
panasevychol: Haha true) Will change.

tests/test_cli/test_list.py
panasevychol: Weird issue, I see that. Previous time `test_search` failed and this passed. And locally everything passes. Will find out.

aea/cli/utils/click_utils.py
panasevychol: It looks like that. But for the purpose of usage (expanding the click commands functionality with custom things) all of them playing a similar role. We don't use it directly for parsing or validating parameter, `click` tool uses.

aea/cli/utils/package_utils.py
panasevychol: Yes! That's the next goal - to leave private only really private methods. ToDo in next PR cause it will also require many lines changes.

aea/cli/utils/package_utils.py
panasevychol: Well, in general I agree, it's better to raise ClickException on a first level. But 2 moments: if we gonna raise another exception anyway we have to catch that so usage will grow up from 1 to at least 4 lines.
And the second moment generally it will never be used outside the CLI command. So the changes in p.1 don't worth it.

aea/cli/utils/package_utils.py
cyenyxe: Just for clarity, what I mean here is that these methods could be moved to `packages` instead of being in `utils`, if they are only used from `packages`.

aea/cli/utils/package_utils.py
cyenyxe: Capturing more exception types improves the semantics. Working only with `ClickException` means that at a higher level you can only log the error but not do any cleanup (if needed) because you don't know anything about the underlying cause.

Regarding the fact it is only used from the CLI command, fair but if the Click API changes or you ever decide to use something else, a lot of files containing business logic (not UI code!) will need to be modified because all layers are coupled.

aea/cli/utils/package_utils.py
panasevychol: No. These methods purpose not to be a part of packages. These are the helpers methods for CLI to work with packages - copy, paste, move, etc.

aea/cli/utils/package_utils.py
panasevychol: Directly opposite actually - we raise ClickException only if it should be displayed as user-friendly error message. `click` lib catches that and translates to a CLI error message. Otherwise, if it should fail with other exception - it's not caught, it will failed hardly as it should.
And for the cleanup - we have it exactly for ClickException. Take a look at `aea.cli.utils.decorators.clean_after`. This one catches ClickException, cleans up and then re-raise so the error message could reach the user interface.

aea/cli/utils/config.py
panasevychol: Sure. I'll do it in next PR. This one is already overweighted.

aea/cli/utils/config.py
panasevychol: Also worth to do, totally agree. But too many code rewriting for this PR. I'll have to manually re-write an every usage of those methods (and that are the most commonly used funcs). That take a lot of time and line changes.

aea/cli/utils/exceptions.py
panasevychol: Yes, I think so. Will do.

PR #1254, opened 2020-05-21
AEA-587 AEA-588 Feature/aea skill exception handling

solarw:
## Proposed changes

Better exception handling/logging and AEA exception handle policy: log, stop, propagate.

## Types of changes

## Checklist

### DIFF
--- a/aea/helpers/exception_policy.py
+++ b/aea/helpers/exception_policy.py
@@ -1,0 +1,29 @@
+# -*- coding: utf-8 -*-
+# ------------------------------------------------------------------------------
+#
+#   Copyright 2018-2019 Fetch.AI Limited
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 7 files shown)


aea/helpers/exception_policy.py
```diff
@@ -0,0 +1,29 @@
...
+
+from enum import Enum
+
+
+class ExceptionPolicyEnum(Enum):
+    """AEA Exception policies."""
+
+    propagate = "propagate"
+    just_log = "just_log"
+    stop_and_exit = "stop_and_exit"
```
DavidMinarsch, line 29: Would be good to provide a short description of each policy in the docstring

solarw: some comments added

solarw:
> LGTM, we also need some way to set the policy on the aea-config.yaml

I think, config variables can be refactored in a dedicated task.

PR #1257, opened 2020-05-21
AEA-589 AEA-579 Updating generator with latest dialogue api

5A11:
## Proposed changes

This PR updates the protocol generator so it reflects the latest dialogue API

## Types of changes

## Checklist


packages/fetchai/protocols/fipa/dialogues.py
```diff
@@ -60,32 +50,43 @@
...
-    FipaMessage.Performative.DECLINE: frozenset(),
+    FipaMessage.Performative.PROPOSE: frozenset(
+        [
+            FipaMessage.Performative.PROPOSE,
+            FipaMessage.Performative.ACCEPT,
+            FipaMessage.Performative.ACCEPT_W_INFORM,
+            FipaMessage.Performative.DECLINE,
+        ]
+    ),
 }  # type: Dict[FipaMessage.Performative, FrozenSet[FipaMessage.Performative]]
```
5A11, line 70: The order is different because originally there was no order but the generator now uses sorting to ensure the order is fixed across repeated generations.

packages/fetchai/protocols/fipa/dialogues.py
```diff
@@ -60,32 +50,43 @@
...
+        BUYER = "buyer"
 
     class EndState(Dialogue.EndState):
-        """This class defines the end states of a dialogue."""
+        """This class defines the end states of a fipa dialogue."""
 
         SUCCESSFUL = 0
         DECLINED_CFP = 1
         DECLINED_PROPOSE = 2
         DECLINED_ACCEPT = 3
```
5A11, line 89: Again, alphabetically, `AgentRole` appears before `EndStates`.

packages/fetchai/protocols/fipa/dialogues.py
```diff
@@ -60,32 +50,43 @@
...
 
 class FipaDialogue(Dialogue, ABC):
-    """The FIPA dialogue class maintains state of a dialogue and manages it."""
+    """The fipa dialogue class maintains state of a dialogue and manages it."""
+
+    class AgentRole(Dialogue.Role):
+        """This class defines the agent's role in a fipa dialogue."""
+
+        SELLER = "seller"
+        BUYER = "buyer"
```
5A11, line 80: Again, alphabetically, AgentRole appears before EndStates.

packages/fetchai/protocols/fipa/dialogues.py
```diff
@@ -60,32 +50,43 @@
...
-    FipaMessage.Performative.INFORM: frozenset([FipaMessage.Performative.INFORM]),
-    FipaMessage.Performative.DECLINE: frozenset(),
+    FipaMessage.Performative.PROPOSE: frozenset(
+        [
+            FipaMessage.Performative.PROPOSE,
+            FipaMessage.Performative.ACCEPT,
+            FipaMessage.Performative.ACCEPT_W_INFORM,
+            FipaMessage.Performative.DECLINE,
+        ]
+    ),
```
5A11, line 69: The order is different because originally there was no order but the generator now uses sorting to ensure the order is fixed across repeated generations.

examples/protocol_specification_ex/fipa.yaml
```diff
@@ -36,8 +36,8 @@ ct:Description: |
 reply:
   cfp: [propose, decline]
   propose: [propose, accept, accept_w_inform, decline]
-  accept: [decline, match_accept, match_accept_w_inform]
-  accept_w_inform: [decline, match_accept, match_accept_w_inform]
+  accept: [match_accept, match_accept_w_inform, decline]
```
DavidMinarsch, line 39: All list entries should be alphabetically sorted

aea/protocols/generator.py
```diff
@@ -550,12 +552,38 @@ def _setup(self) -> None:
...
+            initial_performative_set = set_of_all_performatives.difference(
+                set_of_all_replies
+            )
+            initial_performative_list = list(initial_performative_set)
+            if len(initial_performative_list) != 1:
+                raise ProtocolSpecificationParseError(
+                    "Invalid reply structure. There must be a single speech-act which is not a valid reply to any other speech-acts so it can be designated as the inital speech-act. Found {} of such speech-acts in the specification".format(
+                        len(initial_performative_list),
+                    )
+                )
```
DavidMinarsch, line 581: interesting and makes sense given current setup, can we think of a dialogue where this restriction might be a problem?

examples/protocol_specification_ex/fipa.yaml
5A11: Done

aea/protocols/generator.py
5A11: Well, on second thought, the above is kinda restrictive (works with fipa case though). Imaging a simple generic dialogue where the two agents just send `inform(msg)` to each other then: 

```
speech-acts:
  inform: msg: ct:Message
```

and 

```
reply:
inform: {inform}
```

In this case, the above condition fails since `len(initial_performative_list) == 0`.

DavidMinarsch: In the first case, `initial_performative_list=[inform]`..although I agree, your current implementation would raise an error.

In general, we should probably allow for more than one initial performative. So this could make the list of potential additions in future.

Issue #1241, opened 2020-05-18
Exceptions in skills are not properly displayed to developer


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
Since the introduction of the execution manager for skills exceptions in skills no longer show correctly.

**Describe the solution you'd like**
Fix, so the developer can identify where the exception is thrown.

Issue #1232, opened 2020-05-14
Skills exception handle policy


solarw:
**Is your feature request related to a problem? Please describe.**
Current implementation supposes that the agent will be terminated on any exception raised during act, handle, or another operation.

Would be nice to review this behaviour and decide do we need agent termination, or it can keep going regardless of some exceptions

PR #1260, opened 2020-05-21
AEA-587, AEA-596, AEA-597 - expose exception handling and default routing on aea config

DavidMinarsch:
## Proposed changes

This PR lets the developer optionally specify two more fields in aea-config:
- skill exception policy (one of "propagate" (default), "just_log", "log_and_exit")
- default routing (a map from protocol ids to connection ids) which defines a default for outgoing envelopes routing

Both can also be set on AEA/Multiplexer respectively.

WIP: validations, tests

## Fixes

-#1232, #1241

## Types of changes

## Checklist

## Further comments

To test the exception policy:
1. `aea create my_agent && cd my_agent && aea add --local skill fetchai/simple_service_registration:0.2.0 && aea add --local connection fetchai/oef:0.3.0 && aea config set agent.default_connection fetchai/oef:0.3.0`
2. Run oef node locally
3. Place `assert False` in the behaviour code of service registration skill
4. `aea run --connections fetchai/oef:0.3.0`

You should see something like:
Screenshot 2020-05-21 at 21 04 22

Now modify the aea-config:
`aea config set agent.skill_exception_policy just_log`
Run again...you should see something like:

Screenshot 2020-05-21 at 21 02 54

Finally, modify the aea-config:
`aea config set agent.skill_exception_policy log_and_exit`
Run again...you should see something like:

Screenshot 2020-05-21 at 21 03 57


aea/configurations/schemas/aea-config_schema.json
```diff
@@ -110,6 +110,12 @@
     },
     "decision_maker_handler": {
       "type": "object"
+    },
+    "skill_exception_policy": {
+      "$ref": "definitions.json#/definitions/skill_exception_policy"
+    },
+    "default_routing": {
+      "type": "object"
```
DavidMinarsch, line 118: This could do with more typing/constraints..not my area.. @marcofavorito ?

aea/configurations/schemas/definitions.json
```diff
@@ -113,6 +113,10 @@
     "execution_timeout": {
       "type": ["number", "null"],
       "minimum": 0
+    },
+    "skill_exception_policy": {
+      "type": ["string", "null"],
+      "minimum": 0
```
DavidMinarsch, line 119: could restrict to 3 available policies

aea/mail/base.py
```diff
@@ -715,20 +726,27 @@ async def _send(self, envelope: Envelope) -> None:
...
-            envelope_context.connection_id if envelope_context is not None else None
-        )
+        # first, try to route by context
+        if envelope_context is not None:
+            connection_id = envelope_context.connection_id
+
+        # second, try to route by default routing
+        if connection_id is None and envelope.protocol_id in self.default_routing:
+            connection_id = self.default_routing[envelope.protocol_id]
+            logger.debug("Using default routing: {}".format(connection_id))
```
DavidMinarsch, line 738: Idea is, that optionally the user can define a default mapping, so they don't have to set context on each envelope. For instance, `{PublicId.from_str("fetchai/oef_search:0.1.0"):  PublicId.from_str("fetchai/soef:0.1.0")}` could be set when I want to run the agent with the p2p_p2plib and soef connections. Then all outgoing traffic will go to `p2p_libp2p` unless it is a search envelope, which will go to `soef`.

aea/configurations/schemas/definitions.json
marcofavorito: Shall we use enums?
https://json-schema.org/understanding-json-schema/reference/generic.html#enumerated-values

aea/configurations/schemas/aea-config_schema.json
marcofavorito: yes. We could use:

- `patternProperties` to validate the keys (they must be public ids)
- the usual `$ref` to the public id definition.

For the first bit: I can't see an elegant way. We have to copy the regular expression in `definitions.json#public_id` and put it there.

You can see an example of `patternProperties` in `aea-config_schema.json` when we define `ledger_apis`.

aea/mail/base.py
marcofavorito: OK, sounds good.
An idea for the future. Could a protocol be associated with multiple connections? Meaning that the message of a protocol must be sent to each connection. Basically, the routing relationship would become a many-to-many association. The same should be possible via envelope context. The use cases would be quite rare though.

docs/config.md
```diff
@@ -0,0 +1,147 @@
+This document describes the configuration files of the different packages.
+
+## AEA config yaml
```
DavidMinarsch, line 3: @marcofavorito added full list of available configs..can still be improved but this is a start

aea/configurations/schemas/aea-config_schema.json
DavidMinarsch: @marcofavorito check out what I did, hope this is correct

aea/mail/base.py
DavidMinarsch: Yes it definitely could. But I think, as you say, it is likely a rare case, and already supported if you use the `EnvelopeContext` for routing (which takes precedence !). I think this is a good first iteration as it covers the current issues we have without making it a massive feature.

aea/aea_builder.py
```diff
@@ -713,8 +757,12 @@ def build(
             is_debug=False,
             max_reactions=self._get_max_reactions(),
             decision_maker_handler_class=self._get_decision_maker_handler_class(),
+            skill_exception_policy=self._get_skill_exception_policy(),
+            default_routing=self._get_default_routing(),
             **self._context_namespace,
         )
+        aea.DEFAULT_RUN_LOOP = self._get_loop_mode()
```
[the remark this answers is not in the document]
solarw: pr prepared for this update

docs/config.md
marcofavorito: Excellent!

PR #1279, opened 2020-05-26
Add Two DHT Nodes Relay Routing tests

lrahmani:
## Proposed changes

Add relay routing tests for libp2p clients connected to different dht nodes.


packages/fetchai/connections/p2p_libp2p/connection.py
```diff
@@ -333,6 +333,9 @@ async def _connect(self) -> None:
         :return: None
         """
         if self._connection_attempts == 1:
+            with open(self.log_file, "r") as f:
```
DavidMinarsch, line 336: Since you're changing the p2p connection and this is now released on registry, please bump its version to `0.2.0`

tests/test_packages/test_connections/test_p2p_libp2p/test_communication.py
```diff
@@ -123,11 +123,10 @@ def setup_class(cls):
         cls.multiplexer1 = Multiplexer([cls.connection1])
         cls.multiplexer1.connect()
```
DavidMinarsch, line 124: Mentioned last week, but we didn't have time then: the tests should be running in a test dir (see `setup_class` where this is commented out).

lrahmani: right, I can look into it as well

PR #1277, opened 2020-05-25
Refactor `aea/aea_builder.py` code re. component loading

marcofavorito:
## Proposed changes

This PR aims to simplify the code in the `aea/aea_builder.py` module regarding the loading of the AEA components.

## Fixes

Fixes #1190 

## Types of changes

What types of changes does your code introduce to agents-aea?
_Put an `x` in the boxes that apply_

- [X] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

_Put an `x` in the boxes that apply._

- [ ] I have read the [CONTRIBUTING](../CONTRIBUTING.md) doc
- [ ] I am making a pull request against the `develop` branch (left side). Also you should start your branch off our `develop`.
- [ ] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have checked that code coverage does not decrease.
- [ ] I have checked that the documentation about the `aea cli` tool works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

## Further comments
n/a

PR #1280, opened 2020-05-26
AEABuilder reentrancy

### LINKED ISSUE #1190: AEABuilder refactoring
TBD

TBD

marcofavorito:
## Proposed changes

**Merge #1277 before this PR**

Make `AEABuilder.build` reentrant, but only for components loaded from directories.

The reason is that manually added component cannot be deep-copied due to unpredictable attributes that user-define classes could have. More importantly, objects like `queue.Queue` cannot be deep-copied.

Notice also that `copy.copy` is not enough because the component objects are complex.

## Fixes

Fixes #1190 

## Types of changes

## Checklist

## Further comments

n/a

### DIFF
--- a/aea/aea_builder.py
+++ b/aea/aea_builder.py
@@ -1163,8 +1163,11 @@ class AEABuilder:
                 skill_context = SkillContext()
                 skill_context.set_agent_context(context)
                 skill_context.logger = logging.getLogger(logger_name)
-                skill = load_component_from_config(  # type: ignore
-                    ComponentType.SKILL, configuration, skill_context=skill_context
+                skill = cast(
+                    Skill,
+                    load_component_from_config(
+                        ComponentType.SKILL, configuration, skill_context=skill_context
+                    ),
                 )
             resources.add_component(skill)
 
--- a/aea/aea_builder.py
+++ b/aea/aea_builder.py
@@ -22,7 +22,7 @@ import itertools
 import logging
 import os
 import pprint
-from copy import deepcopy, copy
+from copy import deepcopy
 from pathlib import Path
 from typing import Any, Collection, Dict, List, Optional, Set, Tuple, Type, Union, cast
 
--- a/aea/aea_builder.py
+++ b/aea/aea_builder.py
@@ -22,7 +22,7 @@ import itertools
 import logging
 import os
 import pprint
-from copy import deepcopy
+from copy import copy, deepcopy
 from pathlib import Path
 from typing import Any, Collection, Dict, List, Optional, Set, Tuple, Type, Union, cast
 
@@ -725,9 +725,9 @@ class AEABuilder:
         :return: the AEA object.
         """
         resources = Resources()
-        wallet = Wallet(deepcopy(self.private_key_paths))
+        wallet = Wallet(copy(self.private_key_paths))
         identity = self._build_identity_from_wallet(wallet)
-        ledger_apis = deepcopy(self._load_ledger_apis(ledger_apis))
+        ledger_apis = self._load_ledger_apis(ledger_apis)
         self._load_and_add_components(ComponentType.PROTOCOL, resources)
         self._load_and_add_components(ComponentType.CONTRACT, resources)
         connections = self._load_connections(identity.address, connection_ids)
@@ -762,6 +762,7 @@ class AEABuilder:
         """
         if ledger_apis is not None:
             self._check_consistent(ledger_apis)
+            ledger_apis = deepcopy(ledger_apis)
         else:
             ledger_apis = LedgerApis(self.ledger_apis_config, self._default_ledger)
         return ledger_apis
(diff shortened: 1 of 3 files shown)


tests/test_aea_builder.py
```diff
@@ -93,3 +93,42 @@ def test_when_package_has_missing_dependency():
...
+    aea2 = builder.build()
+
+    aea1_skills = aea1.resources.get_all_skills()
+    aea2_skills = aea2.resources.get_all_skills()
+
+    assert aea1_skills != aea2_skills
+
+    aea1_skills_configs = [s.configuration for s in aea1_skills]
+    aea2_skills_configs = [s.configuration for s in aea2_skills]
+    assert aea1_skills_configs != aea2_skills_configs
```
DavidMinarsch, line 134: can we extend this to include protocols, connections and contracts?

aea/aea_builder.py
```diff
@@ -256,10 +257,8 @@ def __init__(self, with_default_packages: bool = True):
         :param with_default_packages: add the default packages.
         """
         self._name = None  # type: Optional[str]
-        self._resources = Resources()
         self._private_key_paths = {}  # type: Dict[str, Optional[str]]
         self._ledger_apis_configs = {}  # type: Dict[str, Dict[str, Union[str, int]]]
-        self._default_key = None  # set by the user, or instantiate a default one.
```
DavidMinarsch, line 262: why has this been removed?

aea/aea_builder.py
```diff
@@ -743,19 +724,20 @@ def build(
...
         :return: the AEA object.
         """
-        wallet = Wallet(self.private_key_paths)
+        resources = Resources()
+        wallet = Wallet(deepcopy(self.private_key_paths))
         identity = self._build_identity_from_wallet(wallet)
-        ledger_apis = self._load_ledger_apis(ledger_apis)
-        self._load_and_add_components(ComponentType.PROTOCOL)
-        self._load_and_add_components(ComponentType.CONTRACT)
+        ledger_apis = deepcopy(self._load_ledger_apis(ledger_apis))
```
DavidMinarsch, line 730: Could move 'deepcopy' into method

DavidMinarsch:
A few questions @marcofavorito:
- how to we expect the user to know they cannot use the reentrancy on the custom components?
- would it help if we use pickle for the custom components?
- are you saying if any component, even when loaded from directories, contains `queue.Queue` the building will fail?

tests/test_aea_builder.py
marcofavorito: yes.

aea/aea_builder.py
marcofavorito: not used.

aea/aea_builder.py
marcofavorito: what do you mean? A wrapper to `deepcopy`? I don't see much benefits.

DavidMinarsch: No, I mean use deepcopy in the method being called rather than here, just for symmetry with the rest, not crucial though.

marcofavorito:
> how do we expect the user to know they cannot use the reentrancy on the custom components?

From the documentation. Or, we could add a check somewhere, although it increases the complexity of the implementation.

> would it help if we use pickle for the custom components?
It could help, but standard `pickle` has several issues in serializing some type of objects like callables. A `lambda` attribute would cause the pickle to fail. And that would be another surprise for the user if the `add_component_instance` does not work for such reasons.
There exist other pickling libraries like [dill](https://pypi.org/project/dill/) that don't have such limitations.

> are you saying if any component, even when loaded from directories, contains queue.Queue the building will fail?
No. The loading from directories works well, because a new instance is created, and the configuration object is cloned, so we don't share any memory with the new component instance.
The issue with `queue.Queue` happened when I tried to deep-copy an instance of `Skill`, which contains a queue object.

aea/aea_builder.py
marcofavorito: ah! Sorry. Yes, good point.

marcofavorito:
A solution:

`AEABuilder.add_component` accepts an abstract `ComponentLoader` instance. Its purpose is to create component instances. For instance, concrete classes of such abstract class would be:

- `FileSystemComponentLoader`: load from directory
- `ObjectComponentLoader`: take the Python class object and the parameter for `__init__`, and upon invokation just initialize the object
- `RegistryComponentLoader`: load a component from registry.
- `SomeCustomComponentLoader`: specified by the user.
These APIs might not be public, so the builder could wrap these features somehow. However, it might be useful to have public as well, we should discuss in case we accept this proposal.

DavidMinarsch:
Thanks for the comprehensive response. We should then extend the documentation as part of this PR and ensure the docstring in `.build(..` is sufficiently clear on the limitation to the reentrancy. Can we have an issue for the proposed solution? Seems worthy of a separate PR.

PR #1283, opened 2020-05-27
fix loop mode display and docs, fix test tool

DavidMinarsch:
## Proposed changes

This PR fixes two issues:
- loop_mode was not displayed correctly on `aea run` and wrongly defined in docs
- test case did not work for aea projects as `packages` was a requirement when it is not

## Fixes

-

## Types of changes

## Checklist

## Further comments

-

### DIFF
--- a/aea/test_tools/test_cases.py
+++ b/aea/test_tools/test_cases.py
@@ -77,6 +77,7 @@ class BaseAEATestCase(ABC):
     subprocesses: List[subprocess.Popen] = []  # list of launched subprocesses
     threads: List[Thread] = []  # list of started threads
     packages_dir_path: Path = Path("packages")
+    use_packages_dir: bool = True
     package_registry_src: Path = Path()
     old_cwd: Path  # current working directory path
     t: Path  # temporary directory path
@@ -630,9 +631,10 @@ class BaseAEATestCase(ABC):
         cls.t = Path(tempfile.mkdtemp())
         cls.change_directory(cls.t)
 
-        registry_tmp_dir = cls.t / cls.packages_dir_path
-        cls.package_registry_src = cls.old_cwd / cls.packages_dir_path
-        shutil.copytree(str(cls.package_registry_src), str(registry_tmp_dir))
+        if cls.use_packages_dir:
+            registry_tmp_dir = cls.t / cls.packages_dir_path
+            cls.package_registry_src = cls.old_cwd / cls.packages_dir_path
+            shutil.copytree(str(cls.package_registry_src), str(registry_tmp_dir))
 
         cls.initialize_aea(cls.author)
         cls.stdout = {}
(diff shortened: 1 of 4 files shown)


aea/test_tools/test_cases.py
```diff
@@ -734,6 +738,6 @@ def setup_class(cls):
     def teardown_class(cls):
         """Teardown the test class."""
         cls.path_to_aea = Path(".")
-        cls.packages_dir_path = Path("..", "packages")
+        # cls.packages_dir_path = Path("..", "packages")
```
solarw, line 741: do we need a commented line here?

DavidMinarsch: For now, because this is only a quick fix. I did not want to remove this, can add todo

aea/test_tools/test_cases.py
```diff
@@ -712,7 +715,7 @@ def setup_class(cls):
         """Set up the test class."""
         # make paths absolute
         cls.path_to_aea = cls.path_to_aea.absolute()
-        cls.packages_dir_path = cls.packages_dir_path.absolute()
+        # cls.packages_dir_path = cls.packages_dir_path.absolute()
```
DavidMinarsch, line 718:
```suggestion
        # TODO: decide whether to keep optionally: cls.packages_dir_path = cls.packages_dir_path.absolute()
```

aea/test_tools/test_cases.py
```diff
@@ -723,7 +726,8 @@ def setup_class(cls):
         cls.agent_name = agent_configuration.agent_name
 
         # this will create a temporary directory and move into it
-        BaseAEATestCase.packages_dir_path = cls.packages_dir_path
+        # BaseAEATestCase.packages_dir_path = cls.packages_dir_path
```
DavidMinarsch, line 729:
```suggestion
        # TODO: decide whether to keep optionally:  BaseAEATestCase.packages_dir_path = cls.packages_dir_path
```

aea/test_tools/test_cases.py
DavidMinarsch, line 741:
```suggestion
        # TODO: decide whether to keep optionally:  cls.packages_dir_path = Path("..", "packages")
```

aea/test_tools/test_cases.py
```diff
@@ -712,7 +715,7 @@ def setup_class(cls):
         """Set up the test class."""
         # make paths absolute
         cls.path_to_aea = cls.path_to_aea.absolute()
-        cls.packages_dir_path = cls.packages_dir_path.absolute()
+        # TODO: decide whether to keep optionally: cls.packages_dir_path = cls.packages_dir_path.absolute()
```
marcofavorito, line 718: why not just `None` if `use_packages_dir = False`? It's responsibility of the user to know whether he can use it or not.

DavidMinarsch: Well there is already quite a lot for the user to know to use this test case anyway. We need to improve them, but the quick fix is as I did so not to run into any edge cases where an optional field might cause issues.

PR #1284, opened 2020-05-27
[WIP] Feature/runtime

solarw:
## Proposed changes

runtime option to select multithreaded or one async loop mode to run agent and multiplexer.

## Types of changes

## Checklist

### DIFF
--- a/aea/agent_loop.py
+++ b/aea/agent_loop.py
@@ -409,21 +409,36 @@ class AsyncAgentLoop(BaseAgentLoop):
 class SyncAgentLoop(BaseAgentLoop):
     """Synchronous agent loop."""
 
-    def __init__(self, agent: "Agent") -> None:
+    def __init__(self, agent: "Agent", loop: AbstractEventLoop = None):
         """
         Init agent loop.
 
-        :param agent: agent or AEA instance.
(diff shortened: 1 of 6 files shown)


solarw:
@marcofavorito  @DavidMinarsch 

this is an intermediate PR, it's not completed.

Multiplexer refactoring review is required!
Reason for refactoring: create one async entry point to set up multiplexer to run instead of several independent tasks. this will be useful to start multiplexer with asyncio loop already set instead of using its own thread with its own loop.

I kept in mind that multiplexer provides only two public methods widely used: connect and disconnect.

so the main idea is to set `_connect` task, to perform all connection and message processing stuff and disconnect terminates it with cancel, and cleanup made inside the same coroutine.

possible benefits are: 
* only one task spawn and cancel
* all states of all inner tasks can be handled inside this coroutine. (to be done)
* better error handling can be performed.

disadvantages:
* need to pay more attention to cleanup.
* probably two main states are possible: connected and disconnected completely
* not possible to disconnect twice


future development:
* need to decide about the exception handling in multiplexer workflow. what is critical and what is not.
* better refactoring and overall design overview in terms of full async approach

solarw:
not actuall at the moment

PR #1289, opened 2020-05-28
[WIP] Async friendly multiplexer

solarw:
## Proposed changes

multiplexer with asyncio friendly interface

## Types of changes

## Checklist

### DIFF
--- a/aea/agent_loop.py
+++ b/aea/agent_loop.py
@@ -409,21 +409,36 @@ class AsyncAgentLoop(BaseAgentLoop):
 class SyncAgentLoop(BaseAgentLoop):
     """Synchronous agent loop."""
 
-    def __init__(self, agent: "Agent") -> None:
+    def __init__(self, agent: "Agent", loop: AbstractEventLoop = None):
         """
         Init agent loop.
 
-        :param agent: agent or AEA instance.
+        :param agent: AEA instance
+        :param loop: asyncio loop to use. optional
         """
         super().__init__(agent)
+        self._agent: "AEA" = self._agent
+
+        try:
+            self._loop = loop or asyncio.get_event_loop()
+            assert not self._loop.is_closed()
+            assert not self._loop.is_running()
+        except (RuntimeError, AssertionError):
+            self._loop = asyncio.new_event_loop()
+            asyncio.set_event_loop(self._loop)
+
         self.is_running = False
 
     def start(self) -> None:
         """Start agent loop."""
         self.is_running = True
+        self._loop.run_until_complete(self._run())
+
+    async def _run(self) -> None:
+        """Run loop inside coroutine but call synchronous callbacks from agent."""
         while self.is_running:
             self._spin_main_loop()
-            time.sleep(self._agent._timeout)
+            await asyncio.sleep(self._agent._timeout)
 
     def _spin_main_loop(self):
         """Run one spin of agent loop: act, react, update."""
--- a/aea/agent_loop.py
+++ b/aea/agent_loop.py
@@ -18,38 +18,28 @@
 # ------------------------------------------------------------------------------
 """This module contains the implementation of an agent loop using asyncio."""
 import asyncio
-import datetime
 import logging
-import time
 from abc import ABC, abstractmethod
-from asyncio.events import AbstractEventLoop, TimerHandle
-from asyncio.futures import Future
-from asyncio.tasks import ALL_COMPLETED, FIRST_COMPLETED, Task
+from asyncio.events import AbstractEventLoop
+from asyncio.tasks import Task
 from enum import Enum
 from functools import partial
 from typing import (
-    Any,
     Callable,
     Dict,
     List,
-    Optional,
-    Sequence,
-    Set,
-    Tuple,
-    Union,
-    cast,
 )
 
 from aea.exceptions import AEAException
+from aea.helpers.async_utils import (
+    AsyncState,
+    PeriodicCaller,
+    create_task,
+    ensure_loop,
+    wait_and_cancel,
+)
 from aea.mail.base import InBox
 from aea.skills.base import Behaviour
-
-
-try:
-    from asyncio import create_task
-except ImportError:  # pragma: no cover
-    # for python3.6!
-    from asyncio import ensure_future as create_task  # type: ignore
 
 
 logger = logging.getLogger(__file__)
@@ -57,128 +47,6 @@ logger = logging.getLogger(__file__)
 if False:  # MYPY compatible for types definitions
     from aea.aea import AEA  # pragma: no cover
     from aea.agent import Agent  # pragma: no cover
-
-
-def ensure_list(value: Any) -> List:
-    """Return [value] or list(value) if value is a sequence."""
-    if not isinstance(value, (list, tuple)):
-        value = [value]
-    return list(value)
-
-
-class AsyncState:
-    """Awaitable state."""
-
-    def __init__(self, initial_state: Any = None, loop: AbstractEventLoop = None):
-        """Init async state.
-
-        :param initial_state: state to set on start.
-        :param loop: optional asyncio event loop.
-        """
-        self._state = initial_state
-        self._watchers: Set[Future] = set()
-        self._loop = loop or asyncio.get_event_loop()
-
-    @property
-    def state(self) -> Any:
-        """Return current state."""
-        return self._state
-
-    @state.setter
-    def state(self, state: Any) -> None:
-        """Set state."""
-        if self._state == state:  # pragma: no cover
-            return
-        self._state_changed(state)
-        self._state = state
-
-    def _state_changed(self, state: Any) -> None:
-        """Fulfill watchers for state."""
-        for watcher in list(self._watchers):
-            if state in watcher._states:  # type: ignore
-                self._loop.call_soon_threadsafe(
-                    watcher.set_result, (self._state, state)
-                )
-
-    async def wait(self, state_or_states: Union[Any, Sequence[Any]]) -> Tuple[Any, Any]:
-        """Wait state to be set.
-
-        :params state_or_states: state or list of states.
-        :return: tuple of previous state and new state.
-        """
-        states = ensure_list(state_or_states)
-
-        if self._state in states:
-            return (None, self._state)
-
-        watcher: Future = Future()
-        watcher._states = states  # type: ignore
-        self._watchers.add(watcher)
-        try:
-            return await watcher
-        finally:
-            self._watchers.remove(watcher)
-
-
-class PeriodicCaller:
-    """Schedule a periodic call of callable using event loop."""
-
-    def __init__(
-        self,
-        callback: Callable,
-        period: float,
-        start_at: Optional[datetime.datetime] = None,
-        exception_callback: Optional[Callable[[Callable, Exception], None]] = None,
-        loop: Optional[AbstractEventLoop] = None,
-    ):
-        """
-        Init periodic caller.
-
-        :param callback: function to call periodically
-        :param period: period in seconds.
-        :param start_at: optional first call datetime
-        :param exception_callback: optional handler to call on exception raised.
-        :param loop: optional asyncio event loop
-        """
-        self._loop = loop or asyncio.get_event_loop()
-        self._periodic_callable = callback
-        self._start_at = start_at or datetime.datetime.now()
-        self._period = period
-        self._timerhandle: Optional[TimerHandle] = None
-        self._exception_callback = exception_callback
-
-    def _callback(self) -> None:
-        """Call on each scheduled call."""
-        self._schedule_call()
-        try:
-            self._periodic_callable()
-        except Exception as exception:
-            if not self._exception_callback:
-                raise
-            self._exception_callback(self._periodic_callable, exception)
-
-    def _schedule_call(self) -> None:
-        """Set schedule for call."""
-        if self._timerhandle is None:
-            ts = time.mktime(self._start_at.timetuple())
-            delay = max(0, ts - time.time())
-            self._timerhandle = self._loop.call_later(delay, self._callback)
-        else:
-            self._timerhandle = self._loop.call_later(self._period, self._callback)
-
-    def start(self) -> None:
-        """Activate period calls."""
-        if self._timerhandle:
-            return
-        self._schedule_call()
-
-    def stop(self) -> None:
-        """Remove from schedule."""
-        if not self._timerhandle:
-            return
-
-        self._timerhandle.cancel()
-        self._timerhandle = None
 
 
 class BaseAgentLoop(ABC):
@@ -231,13 +99,7 @@ class AsyncAgentLoop(BaseAgentLoop):
         super().__init__(agent)
         self._agent: "AEA" = self._agent
 
-        try:
-            self._loop = loop or asyncio.get_event_loop()
-            assert not self._loop.is_closed()
-            assert not self._loop.is_running()
-        except (RuntimeError, AssertionError):
-            self._loop = asyncio.new_event_loop()
-            asyncio.set_event_loop(self._loop)
+        self._loop: AbstractEventLoop = ensure_loop(loop)
 
         self._behaviours_registry: Dict[Behaviour, PeriodicCaller] = {}
         self._state: AsyncState = AsyncState()
@@ -309,40 +171,20 @@ class AsyncAgentLoop(BaseAgentLoop):
         for behaviour in list(self._behaviours_registry.keys()):
             self._unregister_behaviour(behaviour)
 
-    def _create_tasks(self):
-        """Create tasks to execute and wait."""
-        tasks = self._create_processing_tasks()
-        stopping_task = create_task(
-            self._state.wait([AgentLoopStates.stopping, AgentLoopStates.error])
-        )
-
-        tasks.append(stopping_task)
-        return tasks
-
-    async def _cancel_and_wait_tasks(self, tasks: List[Task]) -> None:
-        """Cancel all tasks and wait they completed."""
-        for t in tasks:
-            t.cancel()
-
-        await asyncio.wait(tasks, return_when=ALL_COMPLETED)
-
-        for t in tasks:
-            if t.cancelled():
-                continue
-            exc = t.exception()
-            if exc:
-                self._exceptions.append(cast(Exception, exc))
+    async def _task_wait_for_stop(self) -> None:
+        """Wait for stop and unregister all behaviours on exit."""
+        try:
+            await self._state.wait([AgentLoopStates.stopping, AgentLoopStates.error])
+        finally:
+            # stop all behaviours on cancel or stop
+            self._stop_all_behaviours()
 
     async def _run(self) -> None:
         """Run all tasks and wait for stopping state."""
         tasks = self._create_tasks()
 
-        await asyncio.wait(tasks, return_when=FIRST_COMPLETED)
-
-        # one task completed by some reason: error or state == stopped.
-        # clean up
-        self._stop_all_behaviours()
-        await self._cancel_and_wait_tasks(tasks)
+        exceptions = await wait_and_cancel(tasks)
+        self._exceptions.extend(exceptions)
 
         if self._exceptions:
             # check exception raised during run
@@ -360,9 +202,9 @@ class AsyncAgentLoop(BaseAgentLoop):
 
         raise self._exceptions[0]
 
-    def _create_processing_tasks(self) -> List[Task]:
+    def _create_tasks(self) -> List[Task]:
         """
-        Create processing tasks.
+        Create tasks.
 
         :return: list of asyncio Tasks
         """
@@ -370,6 +212,7 @@ class AsyncAgentLoop(BaseAgentLoop):
             self._task_process_inbox(),
             self._task_process_internal_messages(),
             self._task_process_new_behaviours(),
+            self._task_wait_for_stop(),
         ]
         return list(map(create_task, tasks))
 
--- a/aea/helpers/async_utils.py
+++ b/aea/helpers/async_utils.py
@@ -1,0 +1,199 @@
+# -*- coding: utf-8 -*-
+# ------------------------------------------------------------------------------
+#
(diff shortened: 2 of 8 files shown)


aea/mail/base.py
```diff
@@ -395,7 +394,7 @@ def __str__(self):
         )
 
 
-class Multiplexer:
```
DavidMinarsch, line 398: Please do not relabel this class. This would break backwards compatibility and is also not desirable, the multiplexer should by definition be async so it does not need it in the class name.

solarw: @DavidMinarsch 
it's done for compatibility on the first stage.
Multiplexer turned completely async and incompatible with all current code and tests
for this purpose it's renamed to AsyncMultiplexer, Multiplexeit is a sublaclass of async version with full compatible interface for current code.
after switching to runtime, only async version will stay and  named Multiplexer

aea/agent_loop.py
```diff
@@ -231,13 +99,7 @@ def __init__(self, agent: "AEA", loop: AbstractEventLoop = None):
...
         self._agent: "AEA" = self._agent
 
-        try:
-            self._loop = loop or asyncio.get_event_loop()
-            assert not self._loop.is_closed()
-            assert not self._loop.is_running()
-        except (RuntimeError, AssertionError):
-            self._loop = asyncio.new_event_loop()
-            asyncio.set_event_loop(self._loop)
+        self._loop: AbstractEventLoop = ensure_loop(loop)
```
DavidMinarsch, line 102: `get_loop` would maybe be more explicit?

aea/helpers/async_utils.py
```diff
@@ -0,0 +1,324 @@
...
+    def stop(self) -> None:
+        """Remove from schedule."""
+        if not self._timerhandle:
+            return
+
+        self._timerhandle.cancel()
+        self._timerhandle = None
+
+
+def ensure_loop(loop: AbstractEventLoop = None) -> AbstractEventLoop:
```
DavidMinarsch, line 174: Please provide complete docstring, including args.return

aea/helpers/async_utils.py
```diff
@@ -0,0 +1,324 @@
...
+        loop = loop or asyncio.get_event_loop()
+        assert not loop.is_closed()
+        assert not loop.is_running()
+    except (RuntimeError, AssertionError):
+        loop = asyncio.new_event_loop()
+        asyncio.set_event_loop(loop)
+    return loop
+
+
+async def wait_and_cancel(
```
DavidMinarsch, line 186: Docstring incomplete

aea/helpers/async_utils.py
```diff
@@ -0,0 +1,324 @@
...
+    exceptions: List[Exception] = []
+
+    _, pending = await asyncio.wait(tasks, return_when=FIRST_COMPLETED)
+
+    for task in pending:
+        task.cancel()
+
+    completed, pending = await asyncio.wait(tasks, return_when=ALL_COMPLETED)
+
+    assert not pending
```
DavidMinarsch, line 207: Provide error message for when assert not met.

aea/agent_loop.py
```diff
@@ -409,21 +252,36 @@ async def _task_process_new_behaviours(self) -> None:
...
         super().__init__(agent)
+        self._agent: "AEA" = self._agent
+
+        try:
+            self._loop = loop or asyncio.get_event_loop()
+            assert not self._loop.is_closed()
+            assert not self._loop.is_running()
+        except (RuntimeError, AssertionError):
+            self._loop = asyncio.new_event_loop()
+            asyncio.set_event_loop(self._loop)
```
DavidMinarsch, line 271: use `ensure_loop`?

aea/agent_loop.py
```diff
@@ -409,21 +252,36 @@ async def _task_process_new_behaviours(self) -> None:
...
+    def __init__(self, agent: "Agent", loop: AbstractEventLoop = None):
         """
         Init agent loop.
 
-        :param agent: agent or AEA instance.
+        :param agent: AEA instance
+        :param loop: asyncio loop to use. optional
         """
         super().__init__(agent)
+        self._agent: "AEA" = self._agent
```
DavidMinarsch, line 263: ?

aea/agent_loop.py
```diff
@@ -409,21 +252,36 @@ async def _task_process_new_behaviours(self) -> None:
...
+            asyncio.set_event_loop(self._loop)
+
         self.is_running = False
 
     def start(self) -> None:
         """Start agent loop."""
         self.is_running = True
+        self._loop.run_until_complete(self._run())
+
+    async def _run(self) -> None:
```
DavidMinarsch, line 280: So we use asyncio just for consistency in the interface?

aea/helpers/async_utils.py
```diff
@@ -0,0 +1,324 @@
...
+        if not include_cancelled and task.cancelled():
+            continue
+        exc = task.exception()
+        if exc:
+            exceptions.append(cast(Exception, exc))
+
+    return exceptions
+
+
+class AnotherThreadTask:
```
DavidMinarsch, line 219: That is not an appropriate name for a class imo. Can we be more specific?

aea/mail/base.py
```diff
@@ -463,9 +461,7 @@ def connections(self) -> Tuple["Connection"]:
     @property
     def is_connected(self) -> bool:
         """Check whether the multiplexer is processing envelopes."""
-        return self._loop.is_running() and all(
```
DavidMinarsch, line 466: Loop is running is ensured at `ensure_loop` and hence we remove, right?

aea/mail/base.py
```diff
@@ -482,104 +478,67 @@ def connection_status(self) -> ConnectionStatus:
         """Get the connection status."""
         return self._connection_status
 
-    def connect(self) -> None:
+    async def connect(self) -> None:
         """Connect the multiplexer."""
-        with self._lock:
+        self._out_queue = asyncio.Queue()
```
DavidMinarsch, line 483: Where did we previously initialise the outqueue?

aea/mail/base.py
```diff
@@ -482,104 +478,67 @@ def connection_status(self) -> ConnectionStatus:
         """Get the connection status."""
         return self._connection_status
 
-    def connect(self) -> None:
+    async def connect(self) -> None:
         """Connect the multiplexer."""
```
DavidMinarsch, line 482: For symmetry, can add logger debug statement: `"Connect called."`

aea/mail/base.py
```diff
@@ -799,7 +758,74 @@ async def async_wait(self) -> None:
...
+        Connect the multiplexer.
+
+        Synchronously in thread spawned if new loop created.
+        """
+        with self._sync_lock:
+            if not self._loop.is_running():
+                self._thread_runner.start()
+                self._thread_was_started = True
+
+            self._thread_runner.call(super().connect()).result(240)
```
DavidMinarsch, line 804: not clear what 240 is, a timeout I imagine, make constant with explicit name for better readability

aea/mail/base.py
DavidMinarsch: Thanks for explanation. I suggest we maintain the two classes. We want to be able to run the AEA in two modes: one single threaded mode were the event loop is shared between multiplexer and agent loop, another mode with two threads, one for the multiplexer, one for the agent loop. It seems to me the `Multiplexer` class can then in mode one directly instantiate the `AsyncMultiplexer` and in mode two implement its current functionality.

aea/helpers/async_utils.py
```diff
@@ -0,0 +1,324 @@
...
+    Tuple,
+    Union,
+    cast,
+)
+
+try:
+    from asyncio import create_task
+except ImportError:  # pragma: no cover
+    # for python3.6!
+    from asyncio import ensure_future as create_task  # type: ignore # noqa: F401
```
DavidMinarsch, line 46: I think it might be better to explicitly check for the python version using `sys`?

aea/helpers/async_utils.py
```diff
@@ -0,0 +1,324 @@
...
+        watcher._states = states  # type: ignore
+        self._watchers.add(watcher)
+        try:
+            return await watcher
+        finally:
+            self._watchers.remove(watcher)
+
+
+class PeriodicCaller:
+    """Schedule a periodic call of callable using event loop."""
```
DavidMinarsch, line 114: It would be good if you extended this docstring a bit to give an example of how the class is used and for what.

aea/helpers/async_utils.py
```diff
@@ -0,0 +1,324 @@
...
+        """
+        self._loop = loop or asyncio.get_event_loop()
+        self._periodic_callable = callback
+        self._start_at = start_at or datetime.datetime.now()
+        self._period = period
+        self._timerhandle: Optional[TimerHandle] = None
+        self._exception_callback = exception_callback
+
+    def _callback(self) -> None:
+        """Call on each scheduled call."""
```
DavidMinarsch, line 141: All docstrings missing `return: None`

aea/helpers/async_utils.py
```diff
@@ -0,0 +1,324 @@
...
+        """Remove from schedule."""
+        if not self._timerhandle:
+            return
+
+        self._timerhandle.cancel()
+        self._timerhandle = None
+
+
+def ensure_loop(loop: AbstractEventLoop = None) -> AbstractEventLoop:
+    """Use loop provided or create new if not provided or closed."""
```
DavidMinarsch, line 175: incomplete docstring

aea/helpers/async_utils.py
```diff
@@ -0,0 +1,324 @@
...
+    return loop
+
+
+async def wait_and_cancel(
+    tasks: Sequence[Task],
+    include_cancelled: bool = False,
+    loop: Optional[AbstractEventLoop] = None,
+) -> List[Exception]:
+    """
+    Wait first task completed or exception raised and cancel other tasks.
```
DavidMinarsch, line 192: incomplete docstring

aea/helpers/async_utils.py
DavidMinarsch, line 207: missing assertion failure message

aea/helpers/async_utils.py
DavidMinarsch: In particular `Another` should be replaced

aea/helpers/async_utils.py
```diff
@@ -0,0 +1,324 @@
...
+        Init the task.
+
+        :param coro: coroutine to schedule
+        :param loop: an event loop to schedule on.
+        """
+        self._task = loop.create_task(coro)
+        self._loop = loop
+        self._future = asyncio.run_coroutine_threadsafe(self._get_task_result(), loop)
+
+    async def _get_task_result(self) -> Any:
```
DavidMinarsch, line 237: missing return value in docstrings

aea/helpers/async_utils.py
```diff
@@ -0,0 +1,324 @@
...
+            return
+        if self._loop.is_running():
+            logger.debug("Stopping loop...")
+            self._loop.call_soon_threadsafe(lambda: self._loop.stop())
+        logger.debug("Wait thread to join...")
+        self.join(10)
+        logger.debug("Stopped.")
+
+
+async def cancel_and_wait(task: Optional[Task]) -> Any:
```
DavidMinarsch, line 314: We have `wait_and_cancel` and `cancel_and_wait`. It would help if these functions where next to each other so the reader spots the difference immediately.

tests/test_multiplexer.py
```diff
@@ -75,14 +75,25 @@ def test_connect_twice():
...
 
+def test_disconnect_twice():
+    """Test that connecting twice the multiplexer behaves correctly."""
+    multiplexer = Multiplexer([_make_dummy_connection()])
+
+    assert not multiplexer.connection_status.is_connected
+    multiplexer.connect()
+    assert multiplexer.connection_status.is_connected
+    multiplexer.disconnect()
+    multiplexer.disconnect()
```
DavidMinarsch, line 86: What are we testing here exactly? Maybe it can be clarified by adding 
``` python
assert not multiplexer.connection_status.is_connected
```
below first disconnect and then some logger output assertion on the second disconnect

tests/test_packages/test_connections/test_oef/test_communication.py
```diff
@@ -1093,10 +1093,11 @@ async def test_connecting_twice_is_ok(self, pytestconfig):
 async def test_cannot_connect_to_oef():
     """Test the case when we can't connect to the OEF."""
     oef_connection = _make_oef_connection(
-        address=FETCHAI_ADDRESS_ONE, oef_addr="a_fake_address", oef_port=10000,
+        address=FETCHAI_ADDRESS_ONE,
+        oef_addr="127.0.0.1",
+        oef_port=61234,  # use addr instead of hostname to avoid name resolution
```
DavidMinarsch, line 1098: I believe the point was to have an invalid address, now we seem to have a valid one?

aea/mail/base.py
solarw: @DavidMinarsch 
the difference is pretty small between to classes, just three interface functions, so need to find how to use those two interfaces and select it. Suppose it will be solved on making runtimes switchable

aea/agent_loop.py
solarw: I will mark as TODO for this part, it's unclean a bit how to pass loop and who will construct it.

aea/helpers/async_utils.py
marcofavorito: I agree with David, the behaviour is more predictable (e.g. the import could fail for weird reasons). However, I'm not against this approach, I've seen it around quite often.

aea/helpers/async_utils.py
marcofavorito: it's a private method, maybe we can omit it

tests/test_packages/test_connections/test_oef/test_communication.py
marcofavorito: maybe we could use mocking in this case

aea/helpers/async_utils.py
```diff
@@ -0,0 +1,324 @@
...
+    except (RuntimeError, AssertionError):
+        loop = asyncio.new_event_loop()
+        asyncio.set_event_loop(loop)
+    return loop
+
+
+async def wait_and_cancel(
+    tasks: Sequence[Task],
+    include_cancelled: bool = False,
+    loop: Optional[AbstractEventLoop] = None,
```
marcofavorito, line 189: `loop` is not used

aea/helpers/async_utils.py
```diff
@@ -0,0 +1,324 @@
...
+        self._loop = loop or asyncio.new_event_loop()
+        assert not self._loop.is_closed()
+        super().__init__(daemon=True)
+
+    def start(self) -> None:
+        """Start event loop in dedicated thread."""
+        if self.is_alive() or self._loop.is_running():
+            return
+        super().start()
+        self.call(asyncio.sleep(0.001)).result(1)  # type: ignore # to ensure loop is started
```
marcofavorito, line 284: my IDE's linter gives: 

image

aea/helpers/async_utils.py
marcofavorito: Looking at the implementation, there's no constraint on whether the `loop` provided in `__init__` belongs to another thread or to the same thread (@solarw please correct me if I am mistaken).

If the above is correct, I'd suggest `ThreadSafeTask`, or `ThreadSafeAsyncTask`.

aea/helpers/async_utils.py
```diff
@@ -0,0 +1,324 @@
...
+
+    def done(self) -> bool:
+        """Check task is done."""
+        return self._future.done()
+
+
+class ThreadedAsyncRunner(Thread):
+    """Util to run thread with event loop and execute coroutines inside."""
+
+    def __init__(self, loop=None) -> None:
```
marcofavorito, line 269: for consistency with the above, I'd use type hinting here as well:  `loop: Optional[AbstractEventLoop] = None`

aea/helpers/async_utils.py
marcofavorito: Another idea (maybe not today because might require too much extra work): what about inheriting from `asyncio.Future`, or `asyncio.Task`? So we are compliant with those interfaces. We could add a `TODO` for that.

aea/helpers/async_utils.py
```diff
@@ -0,0 +1,324 @@
...
+    """Return [value] or list(value) if value is a sequence."""
+    if not isinstance(value, (list, tuple)):
+        value = [value]
+    return list(value)
+
+
+class AsyncState:
+    """Awaitable state."""
+
+    def __init__(self, initial_state: Any = None, loop: AbstractEventLoop = None):
```
marcofavorito, line 62: `initial_state: Optional[AgentLoopStates] = None`?

aea/helpers/async_utils.py
```diff
@@ -0,0 +1,324 @@
...
+
+        :param initial_state: state to set on start.
+        :param loop: optional asyncio event loop.
+        """
+        self._state = initial_state
+        self._watchers: Set[Future] = set()
+        self._loop = loop or asyncio.get_event_loop()
+
+    @property
+    def state(self) -> Any:
```
marcofavorito, line 73: `-> Optional[AgentLoopStates]`?

aea/helpers/async_utils.py
```diff
@@ -0,0 +1,324 @@
...
+        """Return current state."""
+        return self._state
+
+    @state.setter
+    def state(self, state: Any) -> None:
```
marcofavorito, line 78: `state: Optional[AgentLoopStates]`?

aea/helpers/async_utils.py
marcofavorito: I know, it refers to old code, but if you consider the suggestion correct then I'd update it anyway

aea/helpers/async_utils.py
```diff
@@ -0,0 +1,324 @@
...
+        """Set state."""
+        if self._state == state:  # pragma: no cover
+            return
+        self._state_changed(state)
+        self._state = state
+
+    def _state_changed(self, state: Any) -> None:
```
marcofavorito, line 85: see above

aea/helpers/async_utils.py
```diff
@@ -0,0 +1,324 @@
...
+class AsyncState:
+    """Awaitable state."""
+
+    def __init__(self, initial_state: Any = None, loop: AbstractEventLoop = None):
+        """Init async state.
+
+        :param initial_state: state to set on start.
+        :param loop: optional asyncio event loop.
+        """
+        self._state = initial_state
```
marcofavorito, line 68: Actually, if the above is true, this could be:
```python
self._state = AgentLoopStates(initial_state)
```

That is:
- if `initial_state` is an instance of `AgentLoopStates`, then nothing bad happens: it just returns the same enum value.
- if `initial_state` is a valid enum value (that is, the values defined on the right of the enumeration names, e.g. `None`, `"started"`, `"stopped"`, etc.) then the return value is the corresponding `AgentLoopStates` instance.

aea/mail/base.py
```diff
@@ -799,7 +758,74 @@ async def async_wait(self) -> None:
...
+
+        Also stops a dedicated thread for event loop if spawned on connect.
+        """
+        logger.debug("Disconnect called")
+        with self._sync_lock:
+            if not self._loop.is_running():
+                return
+
+            if self._is_connected:
+                self._thread_runner.call(super().disconnect()).result(240)
```
marcofavorito, line 819: same thing here re. `240`, see above

aea/mail/base.py
```diff
@@ -482,104 +478,67 @@ def connection_status(self) -> ConnectionStatus:
...
-            self._loop.call_soon_threadsafe(self._loop.stop)
-            self._thread.join()
+            await connection.disconnect()
         logger.debug("Multiplexer stopped.")
 
-    async def _connect_all(self):
+    async def _connect_all(self) -> None:
         """Set all the connection up."""
         logger.debug("Start multiplexer connections.")
         connected = []  # type: List[PublicId]
```
marcofavorito, line 544: notice: below we are connecting each connection _serially_, and not concurrently.

Do you see any pros/cons in making it concurrently connecting? One thing is that it's harder to debug in case some connection does not connect.

aea/mail/base.py
```diff
@@ -619,7 +578,7 @@ async def _connect_one(self, connection_id: PublicId) -> None:
                 )
             )
 
-    async def _disconnect_all(self):
+    async def _disconnect_all(self) -> None:
         """Tear all the connections down."""
         logger.debug("Tear the multiplexer connections down.")
         for connection_id, connection in self._id_to_connection.items():
```
marcofavorito, line 584: analogous ideas to concurrent connections apply here.

aea/agent_loop.py
solarw: cause async event loop uses some methods from AEA and incompatible with agent.
so it's just a type redifinition here.

aea/mail/base.py
solarw: it's better to set queue whet event loop is already set.

aea/mail/base.py
solarw: i think it's another task. current refactoring made only for better asyncio interface, not for internal optimizations

but yes, it can be done in parallel, but have to concern about better exception info provided.

aea/helpers/async_utils.py
solarw: no, it can return any state, it's not limited only by agent loop usage

aea/helpers/async_utils.py
solarw: state can be any, it's not bound to agent loop

aea/helpers/async_utils.py
solarw: same for me, this is why it ignored in comments, cause code is valid and works as expected

the reason can be: different implementation in python3.8 :
```
async def sleep(delay, result=None, *, loop=None):
    """Coroutine that completes after a given time (in seconds)."""
    if delay <= 0:
        await __sleep0()
        return result
```

solarw: updated type hints to Awaitable. should help

PR #1293, opened 2020-05-29
[MOBI-59, AEA-627] Adds disconnect functionality to the soef connection

Totoual:
## Proposed changes

Adds a disconnect functionality in the soef connection package.

## Fixes

## Types of changes

## Checklist


packages/fetchai/connections/soef/connection.py
```diff
@@ -165,6 +165,13 @@ def register_service(self, service_description: Description) -> None:
                 )
             )
 
+    def disconnect(self):
```
DavidMinarsch, line 168: Only the service referenced in the data model should be disconnected.

packages/fetchai/connections/soef/connection.py
```diff
@@ -165,6 +165,13 @@ def register_service(self, service_description: Description) -> None:
...
+        for values in self.service_name_to_page_address.values():
+            url = self.base_url + "/" + values
+            params = {"command": "unregister"}
+            response = requests.get(url=url, params=params)
```
DavidMinarsch, line 172: Let's wrap in try, except to avoid this taking down the AEA

packages/fetchai/connections/soef/connection.py
```diff
@@ -475,6 +482,7 @@ async def disconnect(self) -> None:
             self.connection_status.is_connected or self.connection_status.is_connecting
         ), "Call connect before disconnect."
         assert self.in_queue is not None
+        self.channel.disconnect()
```
DavidMinarsch, line 485: The channel has no concept of disconnect, it is a http client only

packages/fetchai/connections/soef/connection.yaml
```diff
@@ -6,7 +6,7 @@ license: Apache-2.0
 aea_version: '>=0.3.0, <0.4.0'
 fingerprint:
   __init__.py: Qmd5VBGFJHXFe1H45XoUh5mMSYBwvLSViJuGFeMgbPdQts
-  connection.py: QmRTUx1SSuKGrD77MKPYV8oc4J2G6vW92cJMcEXaijDXUc
+  connection.py: QmePFpSeVgKnyi5kprB6dQz8fg9kgJpxeJ59PaxJYuKPd4
```
DavidMinarsch, line 9: When we update code, for the first time following a release we need to bump the version number of the package.

packages/fetchai/connections/soef/connection.py
Totoual: So the functionality you would like it to have is that the skill calls the disconnect part ? Otherwise how I would know the service name if we don't store it ? How do you get reference of the data_model inside the connection?

packages/fetchai/connections/soef/connection.py
Totoual: I can move the code in the un-register function that has the `NotImplementedError` but the `TODO` there says to add the ping functionality there.

packages/fetchai/connections/soef/connection.py
DavidMinarsch: We do store the service name, that's the point. Have a look at the remaining implementation.

packages/fetchai/connections/soef/connection.py
DavidMinarsch: What you can do is implement a `disconnect` method in the channel which looks like your current unregister, in that it unregisters all services of the agent.

packages/fetchai/connections/soef/connection.py
```diff
@@ -187,7 +187,7 @@ def _register_service(self, service_name: str) -> Optional[str]:
         :return: the unique page address
         """
         logger.debug("Applying to SOEF lobby with address={}".format(self.address))
-        url = self.base_url + "/register"
+        url = "".join([self.base_url, "/register"])
```
DavidMinarsch, line 190: This is not how urls should be joined, please use:
```
from urllib import parse
parse.urljoin(url1, url2)
```

packages/fetchai/connections/soef/connection.py
```diff
@@ -213,7 +213,7 @@ def _register_service(self, service_name: str) -> Optional[str]:
                     unique_token = child.text
             if len(unique_page_address) > 0 and len(unique_token) > 0:
                 logger.debug("Registering service {}".format(service_name))
-                url = self.base_url + "/" + unique_page_address
+                url = "".join([self.base_url, "/", unique_page_address])
```
DavidMinarsch, line 216: see above

packages/fetchai/connections/soef/connection.py
```diff
@@ -250,7 +250,7 @@ def _set_location(
             logger.debug(
                 "Registering position lat={}, long={}".format(latitude, longitude)
             )
-            url = self.base_url + "/" + unique_page_address
+            url = "".join([self.base_url, "/", unique_page_address])
```
DavidMinarsch, line 253: see above

packages/fetchai/connections/soef/connection.py
```diff
@@ -351,7 +379,7 @@ def _search_range(
             logger.debug(
                 "Searching in radius={} of service={}".format(radius, service_name)
             )
-            url = self.base_url + "/" + unique_page_address
+            url = "".join([self.base_url, "/", unique_page_address])
```
DavidMinarsch, line 382: see above

packages/fetchai/connections/soef/connection.py
```diff
@@ -266,7 +266,35 @@ def _set_location(
...
+                url = "".join([self.base_url, "/", unique_page_address])
+                params = {"command": "unregister"}
+                try:
+                    response = requests.get(url=url, params=params)
+                    if "<response><message>Goodbye!</message></response>" in response.text:
+                        logger.info("Successfully unregistered from the s-oef.")
+                except Exception as e:
+                    logger.error(
+                        "Something went wrong cannot unregister the service! {}".format(e)
+                    )
```
DavidMinarsch, line 282: This block should be extracted in a private helper method

packages/fetchai/connections/soef/connection.py
```diff
@@ -266,7 +266,35 @@ def _set_location(
...
+            for values in self.service_name_to_page_address.values():
+                url = "".join([self.base_url, "/", values])
+                params = {"command": "unregister"}
+                response = requests.get(url=url, params=params)
+                if "<response><message>Goodbye!</message></response>" in response.text:
+                    logger.info("Successfully unregistered from the s-oef.")
+        except Exception as e:
+            logger.error(
+                "Something went wrong cannot unregister the service! {}".format(e)
+            )
```
DavidMinarsch, line 297: reuse `_try_unregister` here

PR #1308, opened 2020-06-02
AEA-598 Feature/async runtime

solarw:
## Proposed changes

async runtime for an agent

## Types of changes

## Checklist


aea/aea.py
```diff
@@ -349,6 +349,8 @@ def teardown(self) -> None:
 
         :return: None
         """
+        logger.debug("[{}]: Calling teardown method...".format(self.name))
+        self.liveness.stop()
```
DavidMinarsch, line 353: `self.liveness.stop()` is already called in `stop` from which `teardown` is called. Please remove it here.

aea/agent.py
```diff
@@ -86,6 +86,7 @@ def __init__(
         timeout: float = 1.0,
         is_debug: bool = False,
         loop_mode: Optional[str] = None,
+        runtime: Optional[BaseRuntime] = None,
```
DavidMinarsch, line 89: why not `runtime = BaseRuntime`?

aea/aea.py
solarw: I think about remove liveness object and use state of runtime.
i think runtime better represents the current state of the agent

aea/agent.py
DavidMinarsch: So to maintain the consistency to current implementation?

aea/agent.py
```diff
@@ -96,23 +97,31 @@ def __init__(
...
         :param is_debug: if True, run the agent in debug mode (does not connect the multiplexer).
         :param loop_mode: loop_mode to choose agent run loop.
+        :param runtime: runtime to up agent.
 
         :return: None
         """
         self._identity = identity
         self._connections = connections
 
+        self._runtime = runtime or AsyncRuntime(agent=self, loop=loop)
```
DavidMinarsch, line 107: See above; I think the default should be `ThreadedRuntime`

DavidMinarsch: Not sure how we could provide `AsyncRuntime` as an argument to `agent if it requires agent in its constructor...circularity.

DavidMinarsch: I also think `BaseRuntime` needs to have a constructor which allows the following here: `runtime(agent=self, loop=loop)`

DavidMinarsch, line 107:
```suggestion
        self._runtime = runtime(agent=self, loop=loop)
```

aea/agent.py
DavidMinarsch, line 89:
```suggestion
        runtime: BaseRuntime = ThreadedRuntime,
```

aea/agent.py
```diff
@@ -276,15 +265,7 @@ def stop(self) -> None:
...
         """
-        self.liveness.stop()
-        if self._main_loop is not None:
-            self._main_loop.stop()
-        logger.debug("[{}]: Calling teardown method...".format(self.name))
-        self.teardown()
-
-        logger.debug("[{}]: Stopping message processing...".format(self.name))
-        self.multiplexer.disconnect()
-        logger.debug("[{}]: Stopped".format(self.name))
```
DavidMinarsch, line 287: probably still worth defining these in a methods here and let the runtime call that method

aea/agent.py
```diff
@@ -216,37 +225,18 @@ def start(self) -> None:
...
-
-        :return: None
-        """
-        if self._loop_mode not in self.RUN_LOOPS:
-            raise ValueError(
-                f"Loop `{self._loop_mode} is not supported. valid are: `{list(self.RUN_LOOPS.keys())}`"
-            )
-
-        loop_cls = self.RUN_LOOPS[self._loop_mode]
-        self._main_loop = loop_cls(self)
```
DavidMinarsch, line 236: see below

aea/agent_loop.py
```diff
@@ -409,28 +284,29 @@ async def _task_process_new_behaviours(self) -> None:
...
         """
-        super().__init__(agent)
-        self.is_running = False
+        super().__init__(agent=agent, loop=loop)
+        self._agent: "AEA" = self._agent
+        asyncio.set_event_loop(self._loop)
 
-    def start(self) -> None:
-        """Start agent loop."""
-        self.is_running = True
```
DavidMinarsch, line 423: what happened to `self.is_running`?

aea/runtime.py
```diff
@@ -0,0 +1,238 @@
...
+logger = logging.getLogger(__name__)
+
+
+class RuntimeStates(Enum):
+    """Runtime states."""
+
+    initial = "not_started"
+    starting = "starting"
+    started = "started"
+    stopped = "stopped"
```
DavidMinarsch, line 46: what about "stopping"? Is there no use for it?

aea/runtime.py
```diff
@@ -0,0 +1,238 @@
...
+        logger.debug("Runtime stop called!")
+        self.teardown()
+        self._stop()
+        logger.debug("[{}]: Stopped".format(self._agent.name))
+
+    def teardown(self) -> None:
+        """Tear down agent."""
+        logger.debug("Runtime teardown ...")
+        self._agent.teardown()
+        logger.debug("[{}]: Teardown completed".format(self._agent.name))
```
DavidMinarsch, line 86: `Runtime teardown completed.`

aea/runtime.py
```diff
@@ -0,0 +1,238 @@
...
+    def stop(self) -> None:
+        """Stop agent and runtime."""
+        logger.debug("Runtime stop called!")
+        self.teardown()
+        self._stop()
+        logger.debug("[{}]: Stopped".format(self._agent.name))
+
+    def teardown(self) -> None:
+        """Tear down agent."""
+        logger.debug("Runtime teardown ...")
```
DavidMinarsch, line 84: add agent name?

aea/runtime.py
```diff
@@ -0,0 +1,238 @@
...
+    def start(self) -> None:
+        """Start agent using runtime."""
+        if self._state.get() is RuntimeStates.started:
+            logger.error("Already started!")
+            return
+        self._start()
+
+    def stop(self) -> None:
+        """Stop agent and runtime."""
+        logger.debug("Runtime stop called!")
```
DavidMinarsch, line 77: `"Runtime stopping ...".format(self._agent.name)`

aea/runtime.py
```diff
@@ -0,0 +1,238 @@
...
+        self.teardown()
+        self._stop()
+        logger.debug("[{}]: Stopped".format(self._agent.name))
```
DavidMinarsch, line 80: `Runtime stopped.`

aea/runtime.py
```diff
@@ -0,0 +1,238 @@
...
+        self._agent.teardown()
+        logger.debug("[{}]: Teardown completed".format(self._agent.name))
+
+    @abstractmethod
+    def _start(self) -> None:
+        """Implement runtime start function here."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def _stop(self) -> None:
```
DavidMinarsch, line 94: Not sure why they are labelled as private.. Either all private or none

aea/runtime.py
```diff
@@ -0,0 +1,238 @@
...
+        self._loop = ensure_loop(
+            loop
+        )  # TODO: decide who constructs loop: agent, runtime, multiplexer.
+        self._state: AsyncState = AsyncState(RuntimeStates.initial)
+
+    def start(self) -> None:
+        """Start agent using runtime."""
+        if self._state.get() is RuntimeStates.started:
+            logger.error("Already started!")
+            return
```
DavidMinarsch, line 72: Add logger statement?

aea/runtime.py
```diff
@@ -0,0 +1,238 @@
...
+        """
+        super().__init__(agent=agent, loop=loop)
+        self._stopping_task: Optional[asyncio.Task] = None
+        self._async_stop_lock: Optional[asyncio.Lock] = None
+
+    def _start(self) -> None:
+        """Start runtime."""
+        try:
+            if self._state.get() is RuntimeStates.started:
+                raise ValueError("Runtime alrady started!")
```
DavidMinarsch, line 120: Before try?

aea/runtime.py
```diff
@@ -0,0 +1,238 @@
...
+
+            asyncio.set_event_loop(self._loop)
+            self._agent._multiplexer.set_loop(self._loop)
+            self._agent._main_loop.set_loop(self._loop)
```
DavidMinarsch, line 124: Outside try?

aea/runtime.py
```diff
@@ -0,0 +1,238 @@
...
+            logger.debug(f"Start runtime event loop {self._loop}: {id(self._loop)}")
+            self._task = self._loop.create_task(self._start_coro())
+            self._loop.run_until_complete(self._task)
+            logger.debug("runtime loop stopped!")
+            self._stopping_task = None
+        except Exception:
+            logger.exception("During runtime processing")
+            raise
+
+    async def _start_coro(self) -> None:
```
DavidMinarsch, line 139: `_start_coro` is not a great name; the method name should reflect what the coroutine does.

aea/runtime.py
```diff
@@ -0,0 +1,238 @@
...
+
+    async def _start_multiplexer(self) -> None:
+        """Call multiplexer connect asynchronous way."""
+        await AsyncMultiplexer.connect(self._agent._multiplexer)
+
+    async def _start_agent(self) -> None:
+        """Start agent main loop asynchronous way."""
+        await self._agent._main_loop._start_coro()
+
+    async def _stop_coro(self) -> None:
```
DavidMinarsch, line 164: see above, re naming

aea/runtime.py
```diff
@@ -0,0 +1,238 @@
...
+        :param agent: Agent to run.
+        :param loop: optional event loop. if not provided a new one will be created.
+        :return: None
+        """
+        super().__init__(agent=agent, loop=loop)
+        self._stopping_task: Optional[asyncio.Task] = None
+        self._async_stop_lock: Optional[asyncio.Lock] = None
+
+    def _start(self) -> None:
+        """Start runtime."""
```
DavidMinarsch, line 117: Provide some details on what this does in the docstring.

aea/runtime.py
DavidMinarsch: Provide some details on what it does in the docstring

aea/runtime.py
```diff
@@ -0,0 +1,238 @@
...
+
+    async def _start_coro(self) -> None:
+        """Implmement main loop of runtime."""
+        try:
+            await self._start_multiplexer()
+            self._agent._start_setup()
+            await self._agent._main_loop._start_coro()
+        except Exception:
+            logger.exception("AsyncRuntime exception during run:")
+            raise
```
DavidMinarsch, line 147: should we re-raise a specific exception here, like `AsyncRuntimeException`?

tests/test_skills/test_error.py
```diff
@@ -63,40 +66,40 @@ class TestSkillError:
 
     def setup(self):
         """Test the initialisation of the AEA."""
-        cls = self
-        cls.node = LocalNode()
```
DavidMinarsch, line 67: was the node not needed?

tests/test_multiplexer.py
```diff
@@ -75,14 +75,25 @@ def test_connect_twice():
...
 
+def test_disconnect_twice():
+    """Test that connecting twice the multiplexer behaves correctly."""
+    multiplexer = Multiplexer([_make_dummy_connection()])
+
+    assert not multiplexer.connection_status.is_connected
+    multiplexer.connect()
+    assert multiplexer.connection_status.is_connected
+    multiplexer.disconnect()
+    multiplexer.disconnect()
```
DavidMinarsch, line 86: disconnect twice

aea/agent.py
solarw: changed to `runtime_mode` similar to `loop_mode`

aea/agent.py
solarw: changed to same behaviour  as for `loop_mode`

aea/agent.py
solarw: updated

aea/runtime.py
solarw: will add and review all states and how it changed

aea/runtime.py
solarw: nice point!

tests/test_multiplexer.py
solarw: this is the idea, check double disconnect. test named disconnect twice)

tests/test_skills/test_error.py
solarw: I did not found any reference to node in test code. also tests work without node, so looks node definition just copy/pasted.

aea/agent.py
solarw: updated

aea/agent.py
solarw: it's too runtime dependent: sync/async so no need to stay these methods here.
it will be duplicated and mostly will be unused.

a possible solution is to create standalone agent class which will support old-style behavior

aea/agent_loop.py
solarw: it moved  to is_running property depends on state

aea/runtime.py
solarw: updated

aea/runtime.py
solarw: updated

aea/runtime.py
solarw: updated

aea/runtime.py
solarw: done

aea/runtime.py
solarw: nice point. updated

aea/runtime.py
solarw: nice point. moved outside

aea/runtime.py
solarw: renamed, docstring slightly updated

aea/runtime.py
solarw: renamed

aea/runtime.py
solarw: updated

aea/runtime.py
solarw: not think so, I expect `AEAException` here, or general exception if something goes really wrong. there is no specific exception related to runtime should be raised during runtime execution.

aea/aea.py
solarw: will be done in separate task

Issue #1191, opened 2020-05-04
Wallet refactoring (split)


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
- Currently, the wallet considers all keys to be used by the agent
- Some are used by connection; these should not be the same used for econ/finanical exchange

**Describe the solution you'd like**
- Allow for the wallet to be appropriately configured to have key-pairs for dm and key-pairs for connections.
- Extend loading mechanisms of connections accordingly

marcofavorito:
We agreed on having an optional dictionary in `aea-config.yaml` with identifier `connection_private_key_paths` that will be exposed to some kind of "connection context" to each connection. Maybe `AgentContext` might fit the purpose.

PR #1318, opened 2020-06-02
Add implementation of connection loading in the abstract base class 

marcofavorito:
## Proposed changes

Add implementation of connection loading in the abstract base class 

## Fixes

Addresses #1191 

## Types of changes

## Checklist


aea/connections/base.py
```diff
@@ -193,11 +199,37 @@ def from_config(
...
+        connection_classes = list(
+            filter(lambda x: re.match(connection_class_name, x[0]), classes)
+        )
+        name_to_class = dict(connection_classes)
+        logger.debug("Processing connection {}".format(connection_class_name))
+        connection_class = name_to_class.get(connection_class_name, None)
+        assert connection_class is not None, "Connection class '{}' not found.".format(
+            connection_class_name
+        )
+        return connection_class.from_config(
```
DavidMinarsch, line 233: this should not be `from_config`, should call constructor instead.

packages/fetchai/connections/local/connection.py
```diff
@@ -377,19 +375,3 @@ async def receive(self, *args, **kwargs) -> Optional["Envelope"]:
...
-        """
-        Get the local OEF connection from the connection configuration.
-
-        :param configuration: the connection configuration.
-        :param identity: the identity object.
-        :param cryptos: object to access the connection crypto objects.
-        :return: the connection object
-        """
-        return OEFLocalConnection(
-            LocalNode(), configuration=configuration, identity=identity, cryptos=cryptos
```
DavidMinarsch, line 394: missing update to constructor

aea/configurations/base.py
```diff
@@ -843,9 +843,27 @@ def __init__(
         excluded_protocols: Optional[Set[PublicId]] = None,
         dependencies: Optional[Dependencies] = None,
         description: str = "",
+        connection_id: Optional[PublicId] = None,
         **config,
     ):
         """Initialize a connection configuration object."""
+        if connection_id is None:
```
DavidMinarsch, line 850: @marcofavorito we need to have some consistency between the args

aea/connections/base.py
```diff
@@ -53,14 +58,15 @@ def __init__(self):
 class Connection(Component, ABC):
     """Abstract definition of a connection."""
 
+    connection_id = None  # type: PublicId
+
     def __init__(
         self,
-        configuration: Optional[ConnectionConfig] = None,
+        configuration: ConnectionConfig,
```
DavidMinarsch, line 65: @marcofavorito I think this needs to be a required argument

aea/connections/base.py
```diff
@@ -73,25 +79,24 @@ def __init__(
         :param cryptos: the crypto store for encrypted communication.
         :param restricted_to_protocols: the set of protocols ids of the only supported protocols for this connection.
         :param excluded_protocols: the set of protocols ids that we want to exclude for this connection.
-        :param connection_id: the connection identifier.
         """
+        assert configuration is not None, "The configuration must be provided."
         super().__init__(configuration)
+        assert (
+            super().public_id == self.connection_id
+        ), "Connection ids in configuration and class not matching."
```
DavidMinarsch, line 87: configuration and object id must match...a consistency check

aea/connections/base.py
```diff
@@ -122,21 +127,14 @@ def address(self) -> "Address":
...
         """Get the component type."""
         return ComponentType.CONNECTION
 
-    @property
-    def connection_id(self) -> PublicId:
-        """Get the id of the connection."""
-        if self._configuration is None:
-            return cast(PublicId, self._connection_id)
-        else:
-            return super().public_id
```
DavidMinarsch, line 138: moved to class var

aea/connections/stub/connection.py
```diff
@@ -201,23 +199,21 @@ class StubConnection(Connection):
...
-        Initialize a stub connection.
+    connection_id = PUBLIC_ID
 
-        :param input_file_path: the input file for the incoming messages.
-        :param output_file_path: the output file for the outgoing messages.
-        """
+    def __init__(self, **kwargs):
+        """Initialize a stub connection."""
         if kwargs.get("configuration") is None and kwargs.get("connection_id") is None:
             kwargs["connection_id"] = PUBLIC_ID
```
DavidMinarsch, line 207: I don't think. we need this anymore now.

aea/connections/base.py
marcofavorito: yes that's why we check in the body of `__init__`. The problem is that we want it a keyword argument because we use `__init__(**kwargs)` extensively in other parts of the framework

PR #1321, opened 2020-06-03
Proposal to simplify serialization of messages and envelopes

### LINKED ISSUE #473: Remove need to import MessageSerializer
Currently, we always need to import the serializer for a message.

Make the serializer or its methods attributes or methods of the message object.

DavidMinarsch:
## Proposed changes

This PR makes a proposal to simplify the serialization of envelopes and messages. In particular:
- remove requirement for unnecessary serialization for internal transport
- simplify usage of messages
- rearrange protocols 
- applies simplified `put_message` to code base
- adjusts protocol generator
- aligns versions

## Fixes

#473 

## Types of changes

## Checklist

## Further comments

-


aea/aea.py
```diff
@@ -239,7 +239,10 @@ def _handle(self, envelope: Envelope) -> None:
             return
 
         try:
-            msg = protocol.serializer.decode(envelope.message)
+            if isinstance(envelope.message, Message):
+                msg = envelope.message
+            else:
+                msg = protocol.serializer.decode(envelope.message)
```
DavidMinarsch, line 245: Incoming messages from connecition might or might not be serialized

aea/mail/base.py
```diff
@@ -267,7 +268,7 @@ def __init__(
         to: Address,
         sender: Address,
         protocol_id: ProtocolId,
-        message: bytes,
+        message: Union[Message, bytes],
```
DavidMinarsch, line 271: We accept both; because envelopes on the connection side do not have access to protocol serializer objects

aea/mail/base.py
```diff
@@ -971,10 +988,24 @@ def put_message(
         :param to: the recipient of the envelope.
         :param sender: the sender of the envelope.
         :param protocol_id: the protocol id.
-        :param message: the content of the message.
+        :param message: the message.
+        :param context: the envelope context
```
DavidMinarsch, line 992: add missing context field

aea/protocols/base.py
```diff
@@ -165,12 +167,17 @@ def __str__(self):
...
 
+    def encode(self) -> bytes:
+        """Encode the message."""
+        return self.serializer.encode(self)
+
 
 class Encoder(ABC):
     """Encoder interface."""
 
+    @staticmethod
```
DavidMinarsch, line 178: these are static methods; fixed now

aea/protocols/default/__init__.py
```diff
@@ -18,3 +18,10 @@
 # ------------------------------------------------------------------------------
 
 """This module contains the support resources for the default protocol."""
+
+from aea.protocols.default.message import DefaultMessage as ImportedDefaultMessage
+from aea.protocols.default.serialization import DefaultSerializer
+
+
+class DefaultMessage(ImportedDefaultMessage):
+    serializer = DefaultSerializer
```
DavidMinarsch, line 27: To avoid circular imports we can do this. I didn;'t find a neater solution yet.

aea/protocols/scaffold/message.py
```diff
@@ -23,12 +23,14 @@
...
 from aea.configurations.base import PublicId
 from aea.protocols.base import Message
+from aea.protocols.scaffold.serialization import MyScaffoldSerializer
 
 
 class MyScaffoldMessage(Message):
     """The scaffold message class."""
 
     protocol_id = PublicId("fetchai", "scaffold", "0.1.0")
+    serializer = MyScaffoldSerializer
```
DavidMinarsch, line 33: Unfortunately, this causes a circular import issue

aea/skills/error/handlers.py
```diff
@@ -151,5 +150,5 @@ def send_unsupported_skill(self, envelope: Envelope) -> None:
             to=envelope.sender,
             sender=self.context.agent_address,
             protocol_id=DefaultMessage.protocol_id,
-            message=DefaultSerializer().encode(reply),
+            message=reply,
```
DavidMinarsch, line 153: Note, skill side we no longer need to deal with serialization

packages/fetchai/connections/gym/connection.py
```diff
@@ -96,8 +95,10 @@ def handle_gym_message(self, envelope: Envelope) -> None:
         :param envelope: the envelope
         :return: None
         """
-        gym_message = GymSerializer().decode(envelope.message)
-        gym_message = cast(GymMessage, gym_message)
+        assert isinstance(
+            envelope.message, GymMessage
+        ), "Message not of type GymMessage"
+        gym_message = cast(GymMessage, envelope.message)
```
DavidMinarsch, line 101: Note, connections no longer have to encode/decode if they don't use envelopes for transport or are never leaving the process

packages/fetchai/connections/http_client/connection.py
```diff
@@ -154,7 +154,7 @@ def to_envelope(
             sender="HTTP Server",
             protocol_id=PublicId.from_str("fetchai/http:0.1.0"),
             context=context,
-            message=HttpSerializer().encode(http_message),
+            message=http_message,
```
DavidMinarsch, line 157: we save entirely unecessary decoding/encoding steps here

aea/protocols/base.py
marcofavorito: the reason why they were non-static is that the serializer developer might want to be stateful on the serialization side, for some reason. Very rare use case though.

aea/aea.py
marcofavorito: Small detail: I'd prefer the other way around; check for `bytes` first; then in the `else` branch save in `msg` whatever is there. `Message` can change in the future, `bytes` won't.

marcofavorito: > Incoming messages from connection might or might not be serialized

Why not giving the connection objects access to the serializers? It would give us "design symmetry": for outgoing envelopes we will have `isinstance(envelope.message, Message)`, and for incoming messages we will have the same, rather than "maybe bytes".

aea/mail/base.py
marcofavorito: why not just `Message`? After the change suggested above.

aea/aea.py
marcofavorito: It also connects to the "restricted_protocols" and "excluded_protocols" fields: the connection will only have/won't have access (respectively) to those protocols.

I know, it's not an easy one... Maybe in the next releases.

packages/fetchai/protocols/oef_search/message.py
```diff
@@ -43,6 +44,7 @@ class OefSearchMessage(Message):
     """A protocol for interacting with an OEF search service."""
 
     protocol_id = ProtocolId("fetchai", "oef_search", "0.1.0")
+    serializer = OefSearchSerializer
```
marcofavorito, line 47: From here I can see the issue:

`oef_search.serializer` requires the definition of `OefSearchMessage`  (it was already there)
`oef_search.message` requires the definition of `OefSearchSerializer` (added in this PR)

Some ideas:
- move the import of `OefSearchMessage` into the `OefSearchSerializer.encode|decode`. 
- populate the `OefSearchMessage.serializer` dynamically (at decode time is easy, but if the message is not from a decoding?)
- retrieve the serializer from the protocol registry via public id. I think what you are trying to achieve is that the connection knows how to serialize the message, given only the message. The public id is available, it just has to load the serializer associated with that public id.
- ... do we need a serializer? `encode/decode` could be abstract methods of `Message`... It is against the single responsibility principle, but in theory that could solve the problem. It's just not doable easily...

aea/aea.py
DavidMinarsch: If I invert the check, then mypy will complain due to it not knowing that it is of `Message` type. Giving the connections access to the serialisers means a significant rewrite. We would have to provide each connection with the dictionary of all serialisers which means resources bleed into the multiplexer. I tried and could not find a clean enough approach. I think this approach, although asymmetric is neater; as all the connection does is deal with an envelope on a per instance basis.

aea/mail/base.py
DavidMinarsch: Because the change above is not straightforward and has its own downsides.

aea/protocols/base.py
DavidMinarsch: Well they cannot now. It seemed like an obvious design flaw before and we do not support statefulness of the Serializer framework side anyway. We were creating a new instance of the Serializer on each occasion; very strange!

packages/fetchai/protocols/oef_search/message.py
DavidMinarsch: Note, it has been solved above in the default protocol. The other protocols have not yet been finished. Re your suggestions:
-import in serializer code: not a fan
- dynamic not a fan either
- the problem with relying on public id is that we then need to maintain a protocol registry in the multiplexer (which for reasons highlighted above is potentially not ideal)
- yes i think we should maintain the separation into two classes.

The solution is actually surprisingly simple since we are dealing with a class variable. So a single assignment in the code base will satisfy our requirements. I now do this assignment in init:
``` python

DefaultMessage.serializer = DefaultSerializer
```

aea/protocols/base.py
marcofavorito: "Statefulness" means they could have had an initializer with custom attributes that use in the serialization. Now this is not possible at all, since the `staticmethod`. Probably `classmethod` would be a good compromise.

packages/fetchai/protocols/oef_search/message.py
marcofavorito: I don't think this it is a good solution. The user has to take into account yet another convention. If he forgets to update the serializer attribute, some code of some remote connection might fail quite unpredictably. This is another drop to the tons of other things users and developers have to take into account worsening user-experience and hence traction.

aea/aea.py
```diff
@@ -255,7 +259,7 @@ def _handle(self, envelope: Envelope) -> None:
             return
 
         for handler in handlers:
-            self._handle_message_with_handler(msg, handler)
+            self._handle_message_with_handler(deepcopy(msg), handler)
```
DavidMinarsch, line 262: @marcofavorito we had previously not done that; we need to in order to avoid the same object being accessed by different skills. Do you agree?

aea/agent.py
```diff
@@ -104,7 +104,7 @@ def __init__(
 
         self._multiplexer = Multiplexer(self._connections, loop=loop)
         self._inbox = InBox(self._multiplexer)
-        self._outbox = OutBox(self._multiplexer)
+        self._outbox = OutBox(self._multiplexer, identity.address)
```
DavidMinarsch, line 107: We provide the default address to the outbox to simplify the `put_message` api (unless a non-default address is used; the `sender` field can be populated framework side)

aea/mail/base.py
```diff
@@ -971,10 +993,27 @@ def put_message(
...
+            protocol_id == message.protocol_id
+        ), "Inconsistent protocol_id and message."
         envelope = Envelope(
-            to=to, sender=sender, protocol_id=protocol_id, message=message
+            to=to or message.counterparty,
+            sender=sender or self._default_address,
+            protocol_id=protocol_id or message.protocol_id,
+            message=message,
+            context=context,
         )
```
DavidMinarsch, line 1018: We now enable the developer to simply do `put_message(message)` which is a large gain in terms of less LOC.

aea/protocols/default/__init__.py
```diff
@@ -18,3 +18,8 @@
 # ------------------------------------------------------------------------------
 
 """This module contains the support resources for the default protocol."""
+
+from aea.protocols.default.message import DefaultMessage
+from aea.protocols.default.serialization import DefaultSerializer
+
+DefaultMessage.serializer = DefaultSerializer
```
DavidMinarsch, line 25: Since the `serializer` is a class var; this is sufficient to expose the serializer everywhere.

aea/aea.py
marcofavorito: No, I think we can save the time required to copy the object many times. It could slow down the handling phase due to several copies. Moreover, the public APIs of the objects are already read-only because we use the `@property` annotation.

The message is Python has reference semantics regarding argument passing, Python developers should know that and avoid side effects on the message instance. If we don't want that, we should change the programming model.

examples/gym_ex/proxy/env.py
```diff
@@ -200,7 +180,7 @@ def _decode_percept(self, envelope: Envelope, expected_step_id: int) -> Message:
         """
         if envelope is not None:
             if envelope.protocol_id == PublicId.from_str("fetchai/gym:0.1.0"):
-                gym_msg = GymSerializer().decode(envelope.message)
+                gym_msg = GymMessage.serializer.decode(envelope.message)
```
DavidMinarsch, line 183: Note, the new usage is much more concise! No need to import the serializer individually; no need to instantiate anything.

aea/protocols/base.py
DavidMinarsch: It also means a much more complicated design. Previously we had a separate instance of serializer for each message; that is not optimal by any means. I still don't see what the usecase would be where the serializer is stateful and in particular why we need an initializer. All messages of a given protocol should be serialized identially, so I think the class is the right level; not the instance.

packages/fetchai/protocols/oef_search/message.py
DavidMinarsch: This is done by the generator. The user does not need to do anything here. I think you need to look at how protocols are used in skills. The previous implementation was were convoluted to say the least; now we save a substantial amount of LOC! I think requiring more caution from the protocol designer (although it is auto generated, remember) than from the protocol user is preferable. The new approach makes using protocols MUCH simpler and it only adds one small requirement for the designer (which the generator takes care of).

aea/aea.py
DavidMinarsch: Ok, so revert this one then.

aea/protocols/base.py
marcofavorito: With classmethod you don't need the initializer.

PR #1322, opened 2020-06-04
Add libp2p client connection

lrahmani:
## Proposed changes

- Add libp2p client connection.
- Update libp2p connection (and node) to act as delegate for client connection
- Add libp3p client unit tests

## Missing

- [x] Address ci 
- [x] Fix flaky unit test
- [ ] Update client connection yaml file to support multiple delegate nodes
- [ ] Test with aea cli
- [ ] Update client connection yaml file with public dht addresses as default delegate nodes
- [ ] Redeploy public DHT with new changes


packages/fetchai/connections/p2p_libp2p/aea/api.go
```diff
@@ -25,20 +25,22 @@ import (
...
+	host_public   string
+	port_public   uint16
+	host_delegate string
+	port_delegate uint16
+	msgin         *os.File
+	msgout        *os.File
+	out_queue     chan *Envelope
+	closing       bool
+	sandbox       bool
 }
```
DavidMinarsch, line 44: Can we have comments somewhere here describing each attribute in more detail?

packages/fetchai/connections/p2p_libp2p/aea/api.go
```diff
@@ -136,7 +144,7 @@ func (aea *AeaApi) Init() error {
 	if uri_public != "" {
 		parts = strings.SplitN(uri_public, ":", -1)
 		if len(parts) < 2 {
-			fmt.Println("[aea-api  ][error] malformed Uri:", uri)
+			fmt.Println("[aea-api  ][error] malformed Uri:", uri_public)
```
DavidMinarsch, line 147: Was this change intentional? If so, comment should be `malformed Public URI:`

packages/fetchai/connections/p2p_libp2p/aea/api.go
```diff
@@ -147,6 +155,21 @@ func (aea *AeaApi) Init() error {
 		aea.port_public = 0
 	}
 
+	// parse delegate uri
+	if uri_delegate != "" {
+		parts = strings.SplitN(uri_delegate, ":", -1)
+		if len(parts) < 2 {
+			fmt.Println("[aea-api  ][error] malformed Uri:", uri_delegate)
```
DavidMinarsch, line 162: `malformed Delegate URI` or similar

packages/fetchai/connections/p2p_libp2p/aea/api.go
DavidMinarsch: You can reuse the ones from the python side

packages/fetchai/connections/p2p_libp2p_client/connection.py
```diff
@@ -0,0 +1,314 @@
...
+            self._process_messages_task.cancel()
+            # TOFIX(LR) mypy issue https://github.com/python/mypy/issues/8546
+            # self._process_messages_task = None
+
+        logger.debug("disconnecting libp2p client connection...")
+        self._writer.write_eof()
+        await self._writer.drain()
+        self._writer.close()
+        # TOFIX(LR) requires python 3.7 minimum
+        # await self._writer.wait_closed()
```
DavidMinarsch, line 195: Can you change it to a python 3.6 compatible api please

packages/fetchai/connections/p2p_libp2p/aea/api.go
lrahmani: will refactor that

packages/fetchai/connections/p2p_libp2p/aea/api.go
lrahmani: it'll be redundant, as the uri will be printed.
there is a test in python class to catch user-input malformed uris

packages/fetchai/connections/p2p_libp2p_client/connection.py
lrahmani: It is. Commented code requires python 3.7

PR #1307, opened 2020-06-02
AEA-622 refactor docs and add new sections

DavidMinarsch:
## Proposed changes

This PR updates the docs in line with user and other requests.

## Fixes

#1276, #1071 

## Types of changes

What types of changes does your code introduce to agents-aea?
_Put an `x` in the boxes that apply_

- [x] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

_Put an `x` in the boxes that apply._

- [x] I have read the [CONTRIBUTING](../CONTRIBUTING.md) doc
- [x] I am making a pull request against the `develop` branch (left side). Also you should start your branch off our `develop`.
- [x] Lint and unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have checked that code coverage does not decrease.
- [x] I have checked that the documentation about the `aea cli` tool works
- [x] I have added necessary documentation (if appropriate)
- [x] Any dependent changes have been merged and published in downstream modules

## Further comments

-

PR #1332, opened 2020-06-05
Continuing PR #1307

5A11:
## Proposed changes

This PR makes updates to docs by continuing the work in this PR #1307.

## Types of changes

## Checklist

### DIFF
--- a/docs/language-agnostic-definition.md
+++ b/docs/language-agnostic-definition.md
@@ -24,9 +24,9 @@ message Envelope{
 
 The format for the above fields, except `message`, is specified below. For those with `regexp`, the format is described in <a href="https://docs.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-language-quick-reference" target=_blank>regular expression.</a>.
 
- * to: (`regexp`) `something`
- * sender: (`regexp`) `something` 
- * protocol_id: (`regexp`) `[a-zA-Z0-9_]+/[a-zA-Z0-9_]+:[0-9]+.[0-9]+.[0-9]+`
+ * to: any string
+ * sender: any string 
+ * protocol_id: (`regexp`) `^[a-zA-Z0-9_]*/[a-zA-Z_][a-zA-Z0-9_]*:(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$`
  * URI: <a href="https://tools.ietf.org/html/rfc3986" target=_blank>this syntax</a>
 
 - It MUST implement each protocol with the required meta-fields:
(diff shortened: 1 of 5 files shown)


docs/language-agnostic-definition.md
```diff
@@ -15,11 +15,19 @@ message Envelope{
...
+<!--
 <div class="admonition note">
   <p class="admonition-title">Note</p>
   <p>Additional constraints on envelope fields will be added soon!</p>
 </div>
+-->
+
+The format for the above fields, except `message`, is specified below. For those with `regexp`, the format is described in <a href="https://docs.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-language-quick-reference" target=_blank>regular expression.</a>.
+
+ * to: (`regexp`) `something`
```
5A11, line 27: What is the regex for `to` and `sender`?

Also check the regex for `protocol_id` is right.

docs/orm-integration.md
```diff
@@ -272,6 +272,7 @@ In the generic buyer skill config (`my_buyer_aea/vendor/fetchai/skills/generic_b
 |ledgers: ['fetchai']               |ledgers: ['ethereum']             |
 |----------------------------------------------------------------------|
 ```
+
```
5A11, line 275: I can change the format to the html table, but it would take.a bit of time. 
So wanted to double check, would that be what you want?
I don't know if that would solve the selection issue, will have to see.

docs/language-agnostic-definition.md
DavidMinarsch: We have not yet implemented the regex for `to` and `sender`. We need a regex which encapsulates valid addresses.

docs/language-agnostic-definition.md
```diff
@@ -15,11 +15,19 @@ message Envelope{
...
+ * sender: (`regexp`) `something`
```
DavidMinarsch, line 28: Rather than `something`; just say `any string` for now and add an admonition note that this is incomplete.

docs/language-agnostic-definition.md
```diff
@@ -89,10 +97,12 @@ message DefaultMessage{
 }
 ```
 
-- It MUST process `Envelopes` asynchronously.
+- It is recommended that it processes `Envelopes` asynchronously. Note, the specification regarding the processing of messages does not impose any particular implementation choice/constraint; for example, the AEA can process envelopes either synchronously and asynchronously. However, due to the high level of activity that an AEA might be subject to, other AEAs expect a certain minimum level of responsiveness and reactivity of an AEA's implementation, especially in the case of many concurrent dialogues with other peers. That could imply the need for asynchronous programming to make the AEA's implementation scalable.
 
 - It MUST have an identity in the form of, at a minimum, an address derived from a public key and its associated private key.
 
+- It SHOULD implement handling of errors using the `default` protocol.
```
DavidMinarsch, line 104: Please add the `proto` schema from `default` protocol package here

docs/orm-integration.md
DavidMinarsch: Don't change the table atm. We will deal with this another time.

docs/language-agnostic-definition.md
5A11: ok

docs/orm-integration.md
5A11: ok

docs/language-agnostic-definition.md
5A11: The schema is just above in the document. I mentioned it in text.

PR #1333, opened 2020-06-05
Fix AEA package loading

marcofavorito:
## Proposed changes

make proper use of importlib and module specs before executing the module.

## Fixes

n/a

## Types of changes

## Checklist

### DIFF
--- a/aea/helpers/base.py
+++ b/aea/helpers/base.py
@@ -39,6 +39,7 @@ from dotenv import load_dotenv
 
 import yaml
 
+from aea.configurations.base import ComponentConfiguration
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


aea/protocols/base.py
```diff
@@ -309,21 +309,13 @@ def from_config(cls, configuration: ProtocolConfig) -> "Protocol":
         assert (
             configuration.directory is not None
         ), "Configuration must be associated with a directory."
-        directory = configuration.directory
-        package_modules = load_all_modules(
-            directory, glob="__init__.py", prefix=configuration.prefix_import_path
+        load_aea_package(configuration)
+        class_module = importlib.import_module(
```
marcofavorito, line 313: this is important: `importlib.import_module` will check the `sys.modules` cache, which has been populated in the line above with `load_aea_package`.

Notice that it is different from the previous approach. We re-loaded the module, hence we lose the changes done in other modules (e.g. in __init__.py when we set the `message.serialzier` field.)

marcofavorito: `load_aea_package` will guarantee that in `sys.modules` there are the modules we want, and not other modules loaded earlier. 

However, that might cause previously loaded modules to not work correctly. It is a known issue.

aea/helpers/base.py
```diff
@@ -119,6 +120,42 @@ def locate(path):
...
+    assert dir is not None
+
+    # patch sys.modules with dummy modules
+    prefix_root = "packages"
+    prefix_author = prefix_root + f".{configuration.author}"
+    prefix_pkg_type = prefix_author + f".{configuration.component_type.to_plural()}"
+    prefix_pkg = prefix_pkg_type + f".{configuration.name}"
+    sys.modules[prefix_root] = types.ModuleType(prefix_root)
+    sys.modules[prefix_author] = types.ModuleType(prefix_author)
+    sys.modules[prefix_pkg_type] = types.ModuleType(prefix_pkg_type)
```
marcofavorito, line 142: basically, for each prefix of the import path we add a dummy module in `sys.modules`, e.g.
```
sys.modules["packages"] = dummy
sys.modules["packages.fetchai"] = dummy
sys.modules["packages.fetchai.protocols"] = dummy
```
So import statements that use these as prefix of the import path will not fail.

PR #1342, opened 2020-06-07
libp2p tests print node log after timeout

lrahmani:
## Proposed changes

Add a decorator to libp2p-related tests that prints libp2p nodes logs to standard output in case of test failure. Practical for debugging ci test failures.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -793,6 +793,25 @@ def _make_libp2p_client_connection(
     return P2PLibp2pClientConnection(configuration=configuration, identity=identity)
 
 
+def libp2p_log_on_failure(fn: Callable) -> Callable:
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -793,6 +793,28 @@ def _make_libp2p_client_connection(
...
+    :return: decorated method.
+    """
+
+    @wraps(fn)
+    def wrapper(self, *args, **kwargs):
+        try:
+            fn(self, *args, **kwargs)
+        except Exception as e:
+            for log_file in self.log_files:
+                print("libp2p log file ======================= {}".format(log_file))
```
DavidMinarsch, line 809: Use `logger` instead please.

PR #1347, opened 2020-06-08
Update component package module loading

marcofavorito:
## Proposed changes

In #1333, a new approach to loading was introduced, but only for protocol. Extend it to connections, contracts and skills.

## Fixes

n/a

## Types of changes

## Checklist

### DIFF
--- a/aea/helpers/base.py
+++ b/aea/helpers/base.py
@@ -30,10 +30,8 @@ import subprocess  # nosec
 import sys
 import types
 from collections import OrderedDict, UserString
-from contextlib import contextmanager
 from pathlib import Path
-from threading import RLock
-from typing import Any, Dict, Sequence, TextIO, Tuple
+from typing import Any, Dict, TextIO
 
 from dotenv import load_dotenv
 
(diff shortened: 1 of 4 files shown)


aea/helpers/base.py
```diff
@@ -174,77 +172,6 @@ def load_aea_package(configuration: ComponentConfiguration) -> None:
         spec.loader.exec_module(module)  # type: ignore
 
 
-def load_all_modules(
```
DavidMinarsch, line 177: Interesting how much code we no longer need ...

marcofavorito: a lot of it was just not used anymore, and it has not been removed.

marcofavorito: moreover, `load_all_modules` was more generic, hence it needed other code to work.

marcofavorito: and, btw, we are still far from the ideal loading mechanism. We still don't handle correctly the conflicting versions for packages of different AEA projects in the same process...

DavidMinarsch: I think we decided we won't handle that at all. We have the requirement that there is one version per process; that's it. Totally ok as its also the case with any other python package.

PR #1350, opened 2020-06-08
Skill.from_config accepts agent context instead of skill context.

marcofavorito:
## Proposed changes

make Skill.from_config and from_dir accept agent context instead of skill context.

## Fixes

related to  #1178 and  #1290

## Types of changes

What types of changes does your code introduce to agents-aea?
_Put an `x` in the boxes that apply_

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

_Put an `x` in the boxes that apply._

- [ ] I have read the [CONTRIBUTING](../CONTRIBUTING.md) doc
- [ ] I am making a pull request against the `develop` branch (left side). Also you should start your branch off our `develop`.
- [ ] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have checked that code coverage does not decrease.
- [ ] I have checked that the documentation about the `aea cli` tool works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

## Further comments

n/a

Issue #1290, opened 2020-05-28
Solve AEABuilder.build non re-entrancy when components are loaded manually


marcofavorito:
**Is your feature request related to a problem? Please describe.**

`AEABuilder.build` is not re-entrant when some components are loaded manually

**Describe the solution you'd like**
Make it re-entrant.

**Describe alternatives you've considered**
n/a 

**Additional context**

The docstrings documenting this problem are here:
https://github.com/fetchai/agents-aea/pull/1280/commits/c905aa2b43af199407e648f20fa6458430a5f5ed

Please update the docstrings when this issue will be completed.
This issue is related to this PR: #1280

Issue #1178, opened 2020-04-29
Clean up code duplication in AEABuilder


marcofavorito:
**Is your feature request related to a problem? Please describe.**

Currently, some code in `AEABuilder` is a bit repetitive.
https://github.com/fetchai/agents-aea/blob/v0.3.1/aea/aea_builder.py#L813-L865

Also, the `load_connection` part is an exception since the loading of that kind of component is slightly different. 

**Describe the solution you'd like**
- resolve the connection loading exception
- squash all the `_load_and_add_*` into a single method.

The code will be much neater and maintainable.

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

PR #1351, opened 2020-06-09
Connection registry

marcofavorito:
# Merge  #1350 before this PR
## Proposed changes

Introduce the connection registry, so we treat the connection as any other component.

## Fixes

addresses #1290, #1190 and #1178 

## Types of changes

## Checklist

## Further comments
n/a


aea/aea_builder.py
```diff
@@ -819,16 +817,36 @@ def build(
...
+        # TODO check address is the same of identity address
+        # TODO filter connections with connection_ids
+        self._load_and_add_components(
+            ComponentType.CONNECTION,
+            resources,
+            identity=identity,
+            crypto_store=wallet.connection_cryptos,
+        )
+        connection_ids = self._process_connection_ids(connection_ids)
+        self._unregister_non_selected_connections(resources, connection_ids)
```
DavidMinarsch, line 835: Should not happen here I think...feels wrong

aea/agent.py
```diff
@@ -229,6 +229,10 @@ def loop_mode(self) -> str:
         """Get the agent loop mode."""
         return self._loop_mode
 
+    def _connect(self) -> None:
```
DavidMinarsch, line 232: Should not be private.

aea/context/base.py
```diff
@@ -45,6 +46,8 @@ def __init__(
         decision_maker_message_queue: Queue,
         decision_maker_handler_context: SimpleNamespace,
         task_manager: TaskManager,
+        default_connection: Optional[PublicId],
+        default_routing: Dict[PublicId, PublicId],
```
DavidMinarsch, line 50: In the past the agent context was not used for connection related stuff...I am not sure this is a good idea adding it here. But worth discussing.

aea/runtime.py
```diff
@@ -258,7 +258,11 @@ def _start(self) -> None:
 
         self._agent.multiplexer.set_loop(asyncio.new_event_loop())
 
-        self._agent.multiplexer.connect()
+        # it is important that 'connect' is before 'start setup'.
+        # some skills might start using the multiplexer queue (e.g. gym skill)
+        # and the interleaving with the multiplexer connection set up
+        # might cause concurrency issues.
+        self._agent._connect()
```
DavidMinarsch, line 265: So both connect and start setup should not be private;...

aea/context/base.py
marcofavorito: we were doing it anyway, because `default_routing` is passes in the constructor of `AEA` as keyword, and in turn forwarded to the `AgentContext` constructor.

aea/aea.py
```diff
@@ -157,6 +164,16 @@ def task_manager(self) -> TaskManager:
         """Get the task manager."""
         return self._task_manager
 
+    def _connect(self) -> None:
```
DavidMinarsch, line 167: Public rather than private I believe

aea/runtime.py
marcofavorito: I think yes, instead. It is not supposed to be a public API.

marcofavorito: I think yes, instead. It is not supposed to be a public API.

marcofavorito: I suggest leaving as it is, so we avoid to add too much stuff in one PR, and given that we were already using that approach, although I agree that it is a design smell (as many other things... one step at a time :-) )

aea/aea_builder.py
marcofavorito: I agree. We should add a "connection filter" so that we retrieve only the active connection. Again, that's for a later PR.

marcofavorito:
@DavidMinarsch this PR wasn't supposed to be merged on `develop`, but on `fix/1290`. Hence, I'll close this one.

PR #1352, opened 2020-06-09
Add method refactored: passing ctx instead of click_context.

panasevychol:
## Proposed changes

Add method refactored: passing ctx instead of click_context.
That makes `add_item` programmatically callable without `click_context` object that is CLI command specific. Also it wasn't really needed as we operate our custom `ctx: aea.ci.utils.context.Context` object together with `@pass_ctx` decorator in CLI commands.
Also public method `add_item` name fixed.

## Types of changes

## Checklist

### DIFF
--- a/aea/cli/add.py
+++ b/aea/cli/add.py
@@ -22,13 +22,12 @@
 from typing import cast
 
 import click
-from click.core import Context as ClickContext
 
 from aea.cli.registry.add import fetch_package
 from aea.cli.utils.click_utils import PublicIdParameter
 from aea.cli.utils.config import load_item_config
 from aea.cli.utils.context import Context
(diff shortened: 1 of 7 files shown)


aea/cli/add.py
```diff
@@ -136,28 +132,27 @@ def _add_item(
         raise click.ClickException("Failed to add an item with incorrect fingerprint.")
 
     register_item(ctx, item_type, item_public_id)
-    _add_item_deps(click_context, item_type, item_config)
+    add_item_deps(ctx, item_type, item_config)
```
DavidMinarsch, line 135: I think this should remain private?

panasevychol: Yes, pardon, will change now.

PR #1353, opened 2020-06-09
Connection registry

marcofavorito:
# Merge  #1350 before this PR
## Proposed changes

Introduce the connection registry, so we treat the connection as any other component.

## Fixes

## Types of changes

## Checklist

### DIFF
--- a/aea/aea.py
+++ b/aea/aea.py
@@ -164,6 +164,16 @@ class AEA(Agent):
         """Get the task manager."""
         return self._task_manager
 
+    def _connect(self) -> None:
+        """Connect the multiplexer."""
+        self.multiplexer.default_routing = self.context.default_routing
+        self.multiplexer._connections = []
+        for c in self.resources.get_all_connections():
+            self.multiplexer.add_connection(
+                c, c.public_id == self.context.default_connection
+            )
+        super()._connect()
+
     def setup(self) -> None:
         """
         Set up the agent.
@@ -181,13 +191,6 @@ class AEA(Agent):
         self.decision_maker.start()
         self.resources.setup()
         ExecTimeoutThreadGuard.start()
-
-        self.multiplexer.default_routing = self.context.default_routing
-        self.multiplexer._connections = []
-        for c in self.resources.get_all_connections():
-            self.multiplexer.add_connection(
-                c, c.public_id == self.context.default_connection
-            )
 
     def act(self) -> None:
         """
--- a/aea/aea.py
+++ b/aea/aea.py
@@ -23,6 +23,7 @@ from typing import Any, Callable, Dict, List, Optional, Sequence, Type, cast
 
 from aea.agent import Agent
 from aea.agent_loop import AsyncAgentLoop, BaseAgentLoop, SyncAgentLoop
+from aea.configurations.base import PublicId
 from aea.configurations.constants import DEFAULT_SKILL
 from aea.connections.base import Connection
 from aea.context.base import AgentContext
@@ -75,6 +76,8 @@ class AEA(Agent):
         skill_exception_policy: ExceptionPolicyEnum = ExceptionPolicyEnum.propagate,
         loop_mode: Optional[str] = None,
         runtime_mode: Optional[str] = None,
+        default_connection: Optional[PublicId] = None,
+        default_routing: Dict[PublicId, PublicId] = None,
         **kwargs,
     ) -> None:
         """
@@ -94,6 +97,8 @@ class AEA(Agent):
         :param skill_exception_policy: the skill exception policy enum
         :param loop_mode: loop_mode to choose agent run loop.
         :param runtime_mode: runtime mode (async, threaded) to run AEA in.
+        :param default_connection: public id to the default connection
+        :param default_routing: dictionary for default routing.
         :param kwargs: keyword arguments to be attached in the agent context namespace.
 
         :return: None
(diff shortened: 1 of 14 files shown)


aea/aea.py
```diff
@@ -157,6 +164,16 @@ def task_manager(self) -> TaskManager:
         """Get the task manager."""
         return self._task_manager
 
+    def _connect(self) -> None:
```
DavidMinarsch, line 167: This should be public not private imo. Also; it seems to me this should call a method in multiplexer like so 
```
connections = self.resources.get_all_connections()
self.multiplexer.setup(connections, default_routing)
```
as all this is scoped to the multiplexer

aea/aea_builder.py
```diff
@@ -1211,78 +1225,10 @@ def _load_and_add_components(
...
-            connection = cast(
-                Connection,
-                self._component_instances[ComponentType.CONNECTION][configuration],
-            )
-            if connection.address != identity.address:
-                logger.warning(
-                    "The address set on connection '{}' does not match the default address!".format(
-                        str(connection.connection_id)
-                    )
-                )
```
marcofavorito, line 1263: notice, we are not checking this anywhere anymore.

aea/aea.py
marcofavorito: done

aea/aea_builder.py
```diff
@@ -819,16 +816,37 @@ def build(
...
+            default_connection=self._get_default_connection(),
             loop_mode=self._get_loop_mode(),
             runtime_mode=self._get_runtime_mode(),
             **deepcopy(self._context_namespace),
         )
         # load connection
-        self._load_and_add_connections(aea, wallet, connection_ids=connection_ids)
-        aea.multiplexer.default_routing = self._get_default_routing()
-        self._load_and_add_skills(aea.context, resources)
+        # TODO check address is the same of identity address
```
DavidMinarsch, line 825: What do you mean `# TODO check address is the same of identity address`? How could it not be with the new loading mechanism?

aea/aea_builder.py
```diff
@@ -819,16 +816,37 @@ def build(
             decision_maker_handler_class=self._get_decision_maker_handler_class(),
             skill_exception_policy=self._get_skill_exception_policy(),
             default_routing=self._get_default_routing(),
+            default_connection=self._get_default_connection(),
             loop_mode=self._get_loop_mode(),
             runtime_mode=self._get_runtime_mode(),
             **deepcopy(self._context_namespace),
         )
         # load connection
```
DavidMinarsch, line 824: unnecessary comment

aea/aea_builder.py
```diff
@@ -819,16 +816,37 @@ def build(
...
-        self._load_and_add_connections(aea, wallet, connection_ids=connection_ids)
-        aea.multiplexer.default_routing = self._get_default_routing()
-        self._load_and_add_skills(aea.context, resources)
+        # TODO check address is the same of identity address
+        # TODO filter connections with connection_ids
```
DavidMinarsch, line 826: I think filtering is now done inside the AEA (call to `connect`) rather than outside...What am I missing?

aea/aea_builder.py
```diff
@@ -819,16 +816,37 @@ def build(
...
+        self._load_and_add_components(
+            ComponentType.CONNECTION,
+            resources,
+            identity=identity,
+            crypto_store=wallet.connection_cryptos,
+        )
+        connection_ids = self._process_connection_ids(connection_ids)
+        # TODO this must be corrected before merging fix/1290
+        self._unregister_non_selected_connections(resources, connection_ids)
```
DavidMinarsch, line 835: I thought we are now doing this at the `connect` call..

aea/aea_builder.py
DavidMinarsch: I don't think we have to as the loading mechanism ensures the address comes from the identity (i.e. we pass the identity into the connection!)

aea/multiplexer.py
```diff
@@ -73,6 +73,15 @@ def __init__(
...
         self.set_loop(loop if loop is not None else asyncio.new_event_loop())
 
+    @property
+    def default_connection(self) -> Connection:
+        """Get the default connection."""
+        if self._default_connection is not None:
+            return self._default_connection
+        else:
+            # return the first
+            return self._connections[0]
```
DavidMinarsch, line 83: Would it not be better to by default set the default connection. It is a bit confusing otherwise..

docs/cli-vs-programmatic-aeas.md
```diff
@@ -105,9 +105,7 @@ def run():
     resources = Resources()
 
     # create the AEA
-    my_aea = AEA(
-        identity, [oef_connection], wallet, ledger_apis, resources,  # stub_connection,
-    )
+    my_aea = AEA(identity, [], wallet, ledger_apis, resources,)  # stub_connection,
```
DavidMinarsch, line 108: I think it is time to remove the `connections` field from the AEA constructor.

tests/test_aea.py
```diff
@@ -345,27 +347,33 @@ def test_initialize_aea_programmatically_build_resources():
             ledger_apis = LedgerApis({}, FETCHAI)
             identity = Identity(agent_name, address=wallet.addresses[FETCHAI])
             connection = _make_local_connection(agent_name, node)
-            connections = [connection]
 
             resources = Resources()
+            # TODO remove connection argument from AEA.__init__
```
DavidMinarsch, line 352: Yes!

PR #1360, opened 2020-06-12
Feature/refactor

DavidMinarsch:
## Proposed changes

This PR fixes some docstrings and public access patterns.

## Fixes

-

## Types of changes

## Checklist

## Further comments

-

### DIFF
--- a/aea/aea_builder.py
+++ b/aea/aea_builder.py
@@ -589,7 +589,7 @@ class AEABuilder:
         self._check_can_add(configuration)
         # update dependency graph
         self._package_dependency_manager.add_component(configuration)
-        configuration._directory = directory
+        configuration._directory = directory  # pylint: disable=protected-access
 
         return self
 
@@ -948,7 +948,9 @@ class AEABuilder:
             else self.DEFAULT_RUNTIME_MODE
         )
 
-    def _check_configuration_not_already_added(self, configuration: ComponentConfiguration) -> None:
+    def _check_configuration_not_already_added(
+        self, configuration: ComponentConfiguration
+    ) -> None:
         """
         Check the component configuration has not already been added.
 
@@ -1056,7 +1058,7 @@ class AEABuilder:
                 )
             )
 
-    def _set_from_configuration(
+    def set_from_configuration(
         self,
         agent_configuration: AgentConfig,
         aea_project_path: Path,
@@ -1175,7 +1177,7 @@ class AEABuilder:
         loader = ConfigLoader.from_configuration_type(PackageType.AGENT)
         agent_configuration = loader.load(configuration_file.open())
 
-        builder._set_from_configuration(
+        builder.set_from_configuration(
             agent_configuration, aea_project_path, skip_consistency_check
         )
         return builder
--- a/aea/aea_builder.py
+++ b/aea/aea_builder.py
@@ -845,7 +845,7 @@ class AEABuilder:
 
     def _check_consistent(self, ledger_apis: LedgerApis) -> None:
         """
-        Check the ledger apis are consistent with the configs
+        Check the ledger apis are consistent with the configs.
 
         :param ledger_apis: the ledger apis provided
         :return: None
@@ -896,7 +896,7 @@ class AEABuilder:
 
     def _get_decision_maker_handler_class(self) -> Type[DecisionMakerHandler]:
         """
-        Return the decision maker handler class
+        Return the decision maker handler class.
 
         :return: decision maker handler class
         """
@@ -920,7 +920,7 @@ class AEABuilder:
 
     def _get_default_routing(self) -> Dict[PublicId, PublicId]:
         """
-        Return the default routing
+        Return the default routing.
 
         :return: the policy
         """
@@ -928,7 +928,7 @@ class AEABuilder:
 
     def _get_loop_mode(self) -> str:
         """
-        Return the loop mode name
+        Return the loop mode name.
 
         :return: the loop mode name
         """
@@ -938,7 +938,7 @@ class AEABuilder:
 
     def _get_runtime_mode(self) -> str:
         """
-        Return the runtime mode name
+        Return the runtime mode name.
 
         :return: the runtime mode name
         """
@@ -948,7 +948,14 @@ class AEABuilder:
             else self.DEFAULT_RUNTIME_MODE
         )
 
-    def _check_configuration_not_already_added(self, configuration) -> None:
+    def _check_configuration_not_already_added(self, configuration: ComponentConfiguration) -> None:
+        """
+        Check the component configuration has not already been added.
+
+        :param configuration: the configuration being added
+        :return: None
+        :raises AEAException: if the component is already present.
+        """
         if (
             configuration.component_id
             in self._package_dependency_manager.all_dependencies
@@ -983,8 +990,7 @@ class AEABuilder:
 
     def _check_pypi_dependencies(self, configuration: ComponentConfiguration):
         """
-        Check that PyPI dependencies of a package don't conflict with
-        the existing ones.
+        Check that PyPI dependencies of a package don't conflict with the existing ones.
 
         :param configuration: the component configuration.
         :return: None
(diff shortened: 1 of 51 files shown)


aea/helpers/dialogue/base.py
```diff
@@ -434,18 +433,30 @@ def _additional_rules(self, message: Message) -> bool:
             result = target == last_target + 1
         return result
 
+    def update_dialogue_label(self, final_dialogue_label: DialogueLabel) -> None:
```
DavidMinarsch, line 436: Write some tests for this

aea/helpers/dialogue/base.py
```diff
@@ -495,27 +502,6 @@ def __str__(self) -> str:
         representation = representation[:-1]
         return representation
 
-    # ToDo the following methods are left for backwards compatibility reasons and are unsafe to use. They will be removed in the future
```
DavidMinarsch, line 498: Removed as no longer required

aea/helpers/dialogue/base.py
```diff
@@ -741,76 +725,3 @@ def _next_dialogue_nonce(self) -> int:
         """
         self._dialogue_nonce += 1
         return self._dialogue_nonce
-
-    # TODO the following methods are left for backwards compatibility reasons and will be removed in the future
```
DavidMinarsch, line 745: No longer required

aea/protocols/base.py
```diff
@@ -114,12 +115,30 @@ def body(self, body: Dict) -> None:
...
+        """Get the performative of the message."""
+        assert self.is_set("performative"), "performative is not set."
+        return cast(Enum, self.get("performative"))
+
+    @property
+    def target(self) -> int:
+        """Get the target of the message."""
+        assert self.is_set("target"), "target is not set."
+        return cast(int, self.get("target"))
+
```
DavidMinarsch, line 141: The base dialogue requires these properties. There is no harm of setting them in the base class as we expect them to be present in all messages anyway as per current generator.

aea/aea_builder.py
```diff
@@ -589,7 +589,7 @@ def add_component(
         self._check_can_add(configuration)
         # update dependency graph
         self._package_dependency_manager.add_component(configuration)
-        configuration._directory = directory
+        configuration._directory = directory  # pylint: disable=protected-access
```
marcofavorito, line 592: Optional, minor comment: what about a one-time setter?

```python
@directory.setter
def directory(self, directory: str) -> None:
    assert self._directory is None, "Directory already set"
    self._directory = directory
```

DavidMinarsch: Sounds good!

aea/helpers/async_friendly_queue.py
```diff
@@ -41,7 +41,9 @@ def put(self, item: Any, *args, **kwargs) -> None:
         super().put(item, *args, **kwargs)
         if self._non_empty_waiters:
             waiter = self._non_empty_waiters.popleft()
-            waiter._loop.call_soon_threadsafe(waiter.set_result, True)
+            waiter._loop.call_soon_threadsafe(  # pylint: disable=protected-access
```
marcofavorito, line 44: If we didn't have to be compatible with Python 3.6, we could have used `get_loop()`:
https://docs.python.org/3/library/asyncio-future.html#asyncio.Future.get_loop

aea/decision_maker/default.py
```diff
@@ -246,7 +246,7 @@ def apply_transactions(
         """
         new_state = copy.copy(self)
         for tx_message in transactions:
-            new_state._update(tx_message)
+            new_state._update(tx_message)  # pylint: disable=protected-access
```
marcofavorito, line 249: does `_update` has to be private? O/w make it public

aea/helpers/async_utils.py
```diff
@@ -127,7 +127,7 @@ async def wait(self, state_or_states: Union[Any, Sequence[Any]]) -> Tuple[Any, A
             return (None, self._state)
 
         watcher: Future = Future()
-        watcher._states = states  # type: ignore
+        watcher._states = states  # type: ignore  # pylint: disable=protected-access
```
marcofavorito, line 130: I don't think there are counterarguments in making the `_states` attribute to be public. It is a custom attribute, so it won't be used by the `asyncio` library (I would be very surprised if it is).

Moreover, an ideal approach would be to extend the `Future` class to include such custom attributes.

marcofavorito: This, of course, will affect the above code, which must be changed accordingly. But that would save some Pylint `disable `directives.

aea/helpers/dialogue/base.py
```diff
@@ -391,20 +386,20 @@ def _basic_rules(self, message: Message) -> bool:
         :return: True if valid, False otherwise.
         """
         message_id = message.message_id
-        target = cast(int, message.get("target"))
-        performative = cast(Enum, message.get("performative"))
+        target = message.target
+        performative = message.performative
 
-        if self.is_empty:
+        if self.last_message is None:
```
marcofavorito, line 392: why did you change this?

DavidMinarsch: Because the implementation meant one had to then check again if the last message was present for pypi. It makes more sense this way.

aea/helpers/async_utils.py
DavidMinarsch: Good points; although I'd rather not address them now as I don't want to open a can of worms.

aea/decision_maker/default.py
DavidMinarsch: It can be public. Will fix.

aea/helpers/async_friendly_queue.py
DavidMinarsch: But we do have to be compatible :)

aea/protocols/base.py
marcofavorito: OK, so why in the future we don't make the class `Message` less generic? That is, the constructor requires the `performative`, `target` etc.?

Sometimes for a framework, it is good to "take decisions" and force the user to think in the framework's terms.

aea/helpers/search/models.py
```diff
@@ -605,7 +609,7 @@ def _check_validity(self):
                 "subexpression must be at least 2.".format(type(self).__name__)
             )
         for constraint in self.constraints:
-            constraint._check_validity()
+            constraint._check_validity()  # pylint: disable=protected-access
```
marcofavorito, line 612: maybe also this one could be made public. As far as I can see, there are no side effects in the implementation of that method.

aea/protocols/base.py
DavidMinarsch: Yes that can be done; although it's probably best to postpone this until later when we have more certainty on the required fields.

aea/helpers/search/models.py
DavidMinarsch: Yeah we can do

benchmark/framework/aea_test_wrapper.py
```diff
@@ -168,15 +167,15 @@ def set_loop_timeout(self, timeout: float) -> None:
 
         :return: None
         """
-        self.aea._timeout = timeout
+        self.aea._timeout = timeout  # pylint: disable=protected-access
```
marcofavorito, line 170: setter? Although we shouldn't allow undesired change of timeout... So probably better leave it for now.

tests/test_configurations/test_aea_config.py
```diff
@@ -164,7 +164,7 @@ def test_builder_applies_default_value_to_aea(self, mock: Mock) -> None:
         """Test AEABuilder applies default value to AEA instance when option is not specified in config."""
         configuration = self._make_configuration(NotSet)
         builder = AEABuilder()
-        builder._set_from_configuration(configuration, aea_project_path=Path("."))
+        builder.set_from_configuration(configuration, aea_project_path=Path("."))
```
marcofavorito, line 167: If public, this might lead to undesired side-effects. 
The implementation "adds" stuff (components/private-keys etc) to the builder, doesn't replace it.

marcofavorito: An easy solution is to make it private again.

benchmark/framework/aea_test_wrapper.py
DavidMinarsch: Yes I think it should be left as is (this is a unique exception imo)

Issue #1270, opened 2020-05-24
Dialogue models for remaining protocols which need one (tac; oef; ml; default)


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
Some protocols have no dialogue models although their design implies one.

**Describe the solution you'd like**
Add the missing dialogue models and generate them.

PR #1368, opened 2020-06-15
AEA-662 proposal for dialogue models for existing protocols

DavidMinarsch:
## Proposed changes

This PR adds dialogue models for the protocols currently missing them.

## Fixes

- #1270 

## Types of changes

## Checklist

## Further comments

-


examples/protocol_specification_ex/default.yaml
```diff
@@ -24,3 +24,10 @@ ct:ErrorCode: |
     }
   ErrorCodeEnum error_code = 1;
 ...
+---
+reply:
+  bytes: [bytes, error]
+  error: []
+roles: {agent}
+end_states: [open, closed]
```
marcofavorito, line 32: when is the dialogue "closed" in case of only-`bytes` conversation?

5A11, line 32: `end_states` here should list the states in which _terminated_ default dialogues can be. 

I don't think terminated default dialogues can be _open_ or _closed_, they are probably always _closed_.

I think we should first see how a default dialogues can terminate. From the reply structure, it appears that a default dialogue is a continues stream of `bytes` messages until a `bytes` message is replied to with an `error` message. In which case the dialogue is then marked as terminated. If so, an `error` message terminates default dialogues. As long as no `error` message, the dialogue is open and agents can send either `bytes` or `error` messages. As soon as an `error` message is sent, the dialogue is marked as terminated, in which case the end_state is always `error_happend` (or something with a better name).

Probably a better name for `end_states` is `outcome`, (also consistent with the literature), which probably better conveys that we are talking about the state of terminated dialogues of X protocol.

examples/protocol_specification_ex/ml_trade.yaml
```diff
@@ -29,3 +29,12 @@ ct:Query: |
...
   bytes description = 1;
 ...
+---
+reply:
+  cfp: [terms]
+  terms: [accept]
+  accept: [data]
+  data: [accept, terms]
+roles: {seller, buyer}
+end_states: [successful, open]
```
5A11, line 39: As above. A _terminated_ `ml_trade` cannot be `open`, right? Is it always successful once terminated? Also when does it terminate? From the reply structure it seems it can be infinite. How do we know when an `ml_trade` dialogue is terminated?

marcofavorito: I think it should terminate with `data`. You can't receive/send other messages after `data`.

examples/protocol_specification_ex/oef_search.yaml
```diff
@@ -37,3 +37,13 @@ ct:OefErrorOperation: |
...
 ...
+---
+reply:
+  register_service: [oef_error]
+  unregister_service: [oef_error]
+  search_services: [search_result,oef_error]
+  search_result: []
+  oef_error: []
+roles: {agent, oef_node}
+end_states: [open, closed]
```
5A11, line 48: I think this should be `successful, failed`. This dialogue seems clearly structured. An `oef_search` dialogue terminates with either `search_result` or `oef_error` message. In the former case, it is `successful`, in the latter `failed`.

examples/protocol_specification_ex/default.yaml
marcofavorito: I start thinking that `bytes` cannot be answered with other `bytes`. In other words, a default dialogue should have only one message in case of bytes, optionally followed by an error message from the other end.

With that change, the possible sequences of messages are:
```
a) bytes
b) bytes, error
```

Can we express such optionality? In this case, dialogue `(a)` would be in a "maybe closed" state.

Shouldn't we also prevent a dialogue from starting with `error`?

examples/protocol_specification_ex/tac.yaml
```diff
@@ -57,3 +57,15 @@ ct:ErrorCode: |
...
+  register: [tac_error]
+  unregister: [tac_error]
+  transaction: [transaction_confirmation,tac_error]
+  cancelled: []
+  game_data: []
+  transaction_confirmation: []
+  tac_error: []
+roles: {agent, controller}
+end_states: [open, closed]
+...
```
5A11, line 71: So from reply structure, `cancelled`, `game_data`, `transaction_confirmation`, `tac_error`, all terminate a dialogue. The latter two are just one move dialogues initiated by the controller. Right?

In which case, I suggest the following: `game_cancelled`, `game_data_sent`, `transaction_confirmed`, and `tac_error` for end_states of the different dialogues which essentially convey the different states of the game. 

Note, this messaging structure isn't the best because it includes 4 different types of dialogues between agent, controller (some one message dialogues). Probably a better structure would be the following:

register: [tac_error,  game_data, unregister, cancelled]
unregister: [tac_error]
game_data: [transaction]
transaction: [transaction_confirmation, tac_error]
cancelled: []
transaction_confirmation: [transaction]
tac_error: []

examples/protocol_specification_ex/default.yaml
DavidMinarsch: Can I remind you both that `end_states` simply defines an enum which then can be used by the user in setting `end_states` on `dialogue_stats` once they are reached. Hence, the `end_states` do not get automatically set by the dialogue state. Therefore, all that's required here is to determine sensible states a dialogue can be in. So in this light and @5A11's comment I would suggest we use only `end_states = {closed}`. Note though, that not all dialogues are closed. Only those specified as such by the user. By default they are in an undefined endstate. I disagree that `end_states` can only be reached once an error happens. It is up to the user to define (e.g. based on message content) when the dialogue has ended. This PR is not about changing anything on the dialogue generator/naming of fields/optionality of fields. So any suggestions wrt to altering this should be moved to an issue and discussed there. I don;t see the value in not permitting `bytes` to be responded with `bytes`.

examples/protocol_specification_ex/gym.yaml
```diff
@@ -22,3 +22,12 @@ speech_acts:
...
   bytes any = 1;
 ...
+---
+reply:
+  act: [percept]
+  percept: []
+  reset: []
+  close: []
+roles: {agent, environment}
+end_states: [open, closed]
```
DavidMinarsch, line 32: only `closed` in line with above

examples/protocol_specification_ex/http.yaml
```diff
@@ -19,3 +19,10 @@ speech_acts:
     headers: pt:str
     bodyy: pt:bytes
 ...
+---
+reply:
+  request: [response]
+  response: []
+roles: {client, server}
+end_states: []
```
DavidMinarsch, line 27: `closed` in line with above

examples/protocol_specification_ex/ml_trade.yaml
DavidMinarsch: I think `data` needs to be fixed to be `data: []`. And yes, `end_states: [closed]` is the only one.

examples/protocol_specification_ex/oef_search.yaml
DavidMinarsch: In case of registration it is neither successful nor failed as we never get a response...happy to have `end_states: [successful, failed]` though

examples/protocol_specification_ex/tac.yaml
DavidMinarsch: In light of your comments, I think it should actually be :
```
  register: [tac_error, game_data, cancelled]
  unregister: [tac_error]
  transaction: [transaction_confirmation,tac_error]
  cancelled: []
  game_data: [transaction]
  transaction_confirmation: [transaction]
  tac_error: []
```
`unregister` can be sent by the user at any time, they themselves would have had to send a register before but this cannot be expressed in our reply structure.

5A11: If `unregister` is always after `register` we can make it so it replies to `register`. This prevents single message dialogues (just `unregister`).

examples/protocol_specification_ex/oef_search.yaml
5A11: Again, if we want to avoid multiple "dialogue sequences" under this dialogue definition, we can make `unregister` a reply to `register`. This makes it a longer running interaction between oef and agent. That's if `unregister` always follows `register`.

5A11:
An overall comment: I think multiple dialogue sequences under one dialogue definition isn't desirable. We should try to update these dialogue definitions at some point, as this PR's aim doesn't seem to be that. I will add it as an idea in GH Issues.

examples/protocol_specification_ex/default.yaml
5A11: `closed` isn't really a good name for it, as per definition, `end_states` are always about "closed" dialogues. If default dialogues always look like a potentially infinite `bytes` stream, optionally at some point an `error` which terminates the dialogue, I think either leave `end_states` empty (signifying there no different end states for this dialogue), or name it `failed`/`erratic`/`error_happened`/... to convey such a dialogue's actual state.

examples/protocol_specification_ex/gym.yaml
5A11: I am not quite sure if I understood the gym interaction here. Is it that it allows three dialogue streams, the first being maximum two messages and the others single message dialogues:

`act, percept`
`reset`
`close`

If so, I suggest:
`end_states: [act-percept_loop, reset, close]`

examples/protocol_specification_ex/http.yaml
5A11: Again, I don't think close is the right name, either leave empty or name it something like
`end_states: [request_response_exchange]`.

examples/protocol_specification_ex/ml_trade.yaml
5A11: I agree with the change to replies for `data`. In line with my comments above, I suggest something else over `close`, maybe `data_exchanged` or something.

Incidentally, this dialogue seems it always succeeds (ends with data being exchanged). So with this dialogue definition, agents can't really decline terms. It is fine if its purpose is only to demonstrate a use-case.

examples/protocol_specification_ex/default.yaml
DavidMinarsch: So we can go with `[successful, failed]`. The key is, you want to distinguish dialogues which are finished and not failed from those which are still ongoing. So we need two end states.

examples/protocol_specification_ex/gym.yaml
DavidMinarsch: We could narrow it further:
```
  act: [percept]
  percept: [act, close, reset]
  close: []
  reset: []
```
End states should be only `closed` I suppose.

DavidMinarsch: Alternatively, `success` or `finished`..`act_percept_exchange` is very verbose and not very informative.

examples/protocol_specification_ex/default.yaml
DavidMinarsch: The wider issue here is that it should probably be `states` rather than `end_states` because then we could have `[successful, failed, open]` which is really much more useful.

5A11: Exactly, the problem is we are dealing with two concepts: whether the dialogue is open/terminated, and what the outcome is once it is terminated. 

For now, we can combine `open/closed` and `end-states` in a single `states` and have `[successful, failed, open]`. Though, based on the definition, a default dialogue can never be in the `successful` state, only `open` or `failed`.

examples/protocol_specification_ex/gym.yaml
5A11: Ok with the new reply structure, there are two ways a dialogue may end, either with `close` or `reset`, so the `end_states: [reset, closed]`.

examples/protocol_specification_ex/tac.yaml
DavidMinarsch: Its not a reply though; it is just another message initiated by the same side which sends register.

DavidMinarsch: So this points at scenarios where you have a dialogue but another type of message can be sent "in reply" to an own message (unregister "in reply" to register, both sent by the agent).

examples/protocol_specification_ex/default.yaml
```diff
@@ -24,3 +24,12 @@ ct:ErrorCode: |
     }
   ErrorCodeEnum error_code = 1;
 ...
+---
+entry: [bytes, error]
```
DavidMinarsch, line 28: I have added dialogue entry and exit points now for the purpose of our discussion. Exit can be both determined by the reply structure AND the message content, hence the reply structure alone is not sufficient to derive exit. I.e. here, the user might receive a bytes message and label the dialogue as `successful` based on its content.

examples/protocol_specification_ex/gym.yaml
```diff
@@ -22,3 +22,15 @@ speech_acts:
 ct:AnyObject: |
   bytes any = 1;
 ...
+---
+entry: [reset]
+reply:
+  reset: [status]
+  status: [act, close, reset]
```
DavidMinarsch, line 29: Adding a `status` message field seems necessary here in order for us to be able to specify a proper dialogue.

examples/protocol_specification_ex/default.yaml
5A11: It's now much cleaner. Only I would propose different names for these:

`initial-messages` and `final-messages` (or `terminating-messages` or `terminal-messages`)
or
`initiation` and `termination`

examples/protocol_specification_ex/http.yaml
```diff
@@ -19,3 +19,12 @@ speech_acts:
...
     bodyy: pt:bytes
 ...
+---
+entry: [request]
+reply:
+  request: [response]
+  response: []
+exit: [response]
+roles: {client, server}
+end_states: [successful]
```
5A11, line 29: Looks fine. Always a two message dialogue: `request`, `response`

examples/protocol_specification_ex/oef_search.yaml
```diff
@@ -37,3 +37,15 @@ ct:OefErrorOperation: |
       }
   OefErrorEnum oef_error = 1;
 ...
+---
+entry: [register_service, unregister_service, search_services]
```
5A11, line 41: I would make it so `unregister` is a reply to `register` and avoid the second type of dialogue sequence. Then we would only have two sequences:
`register`, (`error`), `unregister`, (`error`)
`search`, `results`/`error`

examples/protocol_specification_ex/tac.yaml
5A11: That is precisely what the notion of reply is. It's not about the sender/receiver of messages, but it is a relation over messages defining how one message relates to another message.

5A11: With our current setup, replying to own message is already possible. There are no restriction when replying that you should reply to someone else's message. So we do not have to do anything to enable this.

5A11: If one wants to reply, regardless of the sender of the message being replied to, the process is the same.

examples/protocol_specification_ex/oef_search.yaml
5A11: Shouldn't be a problem replying to your own message. See below.

examples/protocol_specification_ex/http.yaml
DavidMinarsch: Yes I think the former makes more sense in light of the RESTfullness (https://restfulapi.net/statelessness/)

examples/protocol_specification_ex/tac.yaml
DavidMinarsch: I agree with what you say.

Issue #1376, opened 2020-06-16
Require "aea init" before GUI launch and take author name from config.


panasevychol:
Require "aea init" before GUI launch and take author name from config.

Issue #1057, opened 2020-04-05
Test and fix GUI


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
The GUI current functionality has not been tested by hand for a while.

**Describe the solution you'd like**
- [x] Test all functionality and fix any bugs arising.
- [x] Refactor the GUI directory to be clean and easy to navigate.
- [x] Ensure it works with remote registry by default, and local registry if configured (tests should run against local registry)

DavidMinarsch:
Also address problematic integration tests

DavidMinarsch:
Also refactor the GUI directory to be clean and easy to navigate

Issue #579, opened 2019-12-31
Update gui for new search and comms


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
The GUI depends on OEF node

**Describe the solution you'd like**
Update it to not depend on it

PR #1386, opened 2020-06-17
GUI upgrades AEA-630, AEA-631

panasevychol:
## Proposed changes

- GUI oef node removed
- init before GUI launch implemented
- using real author in GUI implemented
- remove item in GUI fixed
- CLI remove unregister item fixed
- some smaller changes
#579 #1057 #1376 

## Types of changes

## Checklist


aea/cli_gui/__init__.py
```diff
@@ -222,10 +220,9 @@ def search_registered_items(item_type: str, search_term: str):
 def create_agent(agent_id: str):
     """Create a new AEA project."""
     ctx = Context(cwd=app_context.agents_dir)
+    author = app_context.author if app_context.local else None
```
DavidMinarsch, line 223: why only `if app_context.local`

aea/test_tools/constants.py
```diff
@@ -0,0 +1,23 @@
...
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+# ------------------------------------------------------------------------------
+
+"""This is a module with constants for test tools."""
+
+
+DEFAULT_AUTHOR = "default_author"
```
DavidMinarsch, line 23: Please remove this; that constant should not exist in the aea distribution

aea/cli/utils/constants.py
```diff
@@ -66,3 +66,5 @@
     "contracts": DEFAULT_CONTRACT_CONFIG_FILE,
 }  # type: Dict[str, str]
 FALSE_EQUIVALENTS = ["f", "false", "False"]
+
+DEFAULT_AUTHOR = "default_author"
```
DavidMinarsch, line 70: please remove this constant; it is not needed and should not exist in aea distribution

aea/cli_gui/__init__.py
panasevychol: Because of the logic in create method:
```
    if author is not None:
        if local:
            do_init(author, False, False)
        else:
            raise click.ClickException(
                "Author is not set up. Please use 'aea init' to initialize."
            )
```
That means that initializing default author is for local usage only (which makes sense).

aea/cli/core.py
```diff
@@ -85,12 +87,26 @@ def cli(click_context, skip_consistency_check: bool) -> None:
...
+def _init_gui(local: bool) -> None:
+    """
+    Initialize GUI before start.
+
+    :param local: a boolean flag for local author.
+
+    :return: None
+    """
+    config = get_or_create_cli_config()
+    author = config.get(AUTHOR_KEY, None)
```
DavidMinarsch, line 106: Make fail if none;

aea/cli_gui/__init__.py
DavidMinarsch: See above and remove;

Issue #1127, opened 2020-04-19
Subclassing of classes in packages


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
Currently, whilst possible, we do not support it very well.

**Describe the solution you'd like**
Support it, in particular, provide package dependencies on own type

Issue #939, opened 2020-03-17
Import a skill from another skill


marcofavorito:
**Is your feature request related to a problem? Please describe.**
Composition of skills is currently limited since a skill cannot explicitly depend on another skill.
Also, there's no mechanism that skills can require a connection to run.

**Describe the solution you'd like**
Add a section in the `skill.yaml` both for `skills` and `connections`, to be interpreted as required dependencies.

**Describe alternatives you've considered**

**Additional context**

Issue #362, opened 2019-10-31
Need ability to create "Abstract" skills


dishmop:
**Is your feature request related to a problem? Please describe.**
The carpark agent and weather station agent share 90% of the same code- but currently this needs to be duplicated and is a bit of a maintenance nightmare. 

In addition there is a LOT of code in these agents deal with the logic of negotiation using a FIPA protocol - rather than business logic.

**Describe the solution you'd like**
Approach 1: Ability to have abstract skills with "virtual functions" so both carpark and weather station would "inherit" from a FIPANegotation skill and just fill in their own logic. When you "scaffold" a skill, you could optionally pass in a base class.

Approach 2: Have a FipaNegotation skill which is not abstract, but is actually instantiated, however, then have a carpark skill that is able to set it up (perhaps passing in functions where bespoke logic is needed). 

**Describe alternatives you've considered**
There are probably other way to achieve this aim. I haven't looked into it much.

PR #1403, opened 2020-06-22
AEA-496 Support skill to be dependent on other skills.

### LINKED ISSUE #960: Revise component dependencies assumptions

Currently, the following holds:
- A protocol cannot depend on other protocols/connections/skills/contracts
- A connection can depend on protocols, but not on other connections/skills/contracts
- A skill can depend on protocols/connections/contracts, but not on other skills
- A contract cannot depend on other protocols/connections/skills/contracts

Also, an agent can depend on any component, but not on other agents.

Relax these assumptions and make them more generic: a package can depend on any other package.

Maybe keep some constraints (e.g. protocols and contracts cannot have dependencies.)

**Additional context**

marcofavorito:
## Proposed changes

Make a skill to have other skills in its dependencies.

## Fixes

Fixes #960, #939, #362, #1127

## Types of changes

## Checklist


aea/cli/add.py
```diff
@@ -131,8 +131,8 @@ def add_item(ctx: Context, item_type: str, item_public_id: PublicId) -> None:
     if not is_fingerprint_correct(package_path, item_config):  # pragma: no cover
         raise click.ClickException("Failed to add an item with incorrect fingerprint.")
 
-    register_item(ctx, item_type, item_public_id)
     _add_item_deps(ctx, item_type, item_config)
+    register_item(ctx, item_type, item_public_id)
```
marcofavorito, line 135: this is a bit hacky. We first try to add the sub dependencies, and then register the current package in `aea-config.yaml`.

Consider `skill_a` with a self-loop. Up to line `134`, the package is copied in `vendor` thanks to `copy_package_directory` or `fetch_package`. When line `134` is executed, the `add` command is executed with `skill_a` as argument. However, the second time the `copy*` function will fail because it is already present.

marcofavorito: if approved, will open an issue to keep track of this workaround, and find a proper solution.

aea/configurations/base.py
```diff
@@ -1118,15 +1116,26 @@ def component_type(self) -> ComponentType:
     @property
     def package_dependencies(self) -> Set[ComponentId]:
         """Get the connection dependencies."""
```
DavidMinarsch, line 1118: docstring is incorrect

PR #1408, opened 2020-06-23
Abstract skills

### LINKED ISSUE #362: Need ability to create "Abstract" skills

marcofavorito:
## Proposed changes

Introduce 'abstract skills': skills that are not instantiated by the framework, but importable from other skills.

## Fixes

Fixes #362 

## Types of changes

## Checklist


docs/config.md
```diff
@@ -25,7 +25,8 @@ connections:                                    # The list of connection public
 contracts: []                                   # The list of contract public ids the AEA project depends on (each public id must satisfy PUBLIC_ID_REGEX).
 protocols:                                      # The list of protocol public ids the AEA project depends on (each public id must satisfy PUBLIC_ID_REGEX).
 - fetchai/default:0.3.0
-skills:                                         # The list of skill public ids the AEA project depends on (each public id must satisfy PUBLIC_ID_REGEX).
+skills: []                                      # The list of skill public ids the AEA project depends on (each public id must satisfy PUBLIC_ID_REGEX).
+is_abstract: false                              # A boolean that if `true` makes the skill abstract, i.e. not instantiated by the framework but importable from other skills. Defaults to `false`.
```
DavidMinarsch, line 29: This should be declared as an optional field here

marcofavorito: 'Defaults to `false`" already suggests it is optional. I will make more clear.

aea/aea_builder.py
```diff
@@ -1300,6 +1300,11 @@ def _load_and_add_components(
             if configuration in self._component_instances[component_type].keys():
                 component = self._component_instances[component_type][configuration]
             else:
+                if component_type == ComponentType.SKILL:
+                    is_skill_abstract = cast(SkillConfig, configuration).is_abstract
+                    if is_skill_abstract:
+                        load_aea_package(configuration)
+                        continue
```
marcofavorito, line 1307: very ugly, but I think it's the best place where to do this check. `Skill.from_config` should still return a component instance, even if abstract. Thoughts?

DavidMinarsch: At least extract it as a generic helper function so we can then extend it for other components, something like:
```
if ...
elif is_abstract_component(configuration):
    load_aea_package(configuration)
else:
    ....
```
where `is_abstract_component` is a helper function in helpers

PR #1410, opened 2020-06-23
Generator refactoring 

5A11:
## Proposed changes

This PR breaks the code in the generator module into the following modules:
 - `validate`: containing code used to validate a protocol specification
 - `extract_specification`: containing code to extract fields from a protocol specification into pythonic objects (all fields are encapsulated in `PythonicProtocolSpecification`)
 - `common`: containing constants and functions used in more than one module
 - `generator`: containing the old generator code minus the above separated codes

The PR doesn't do anything else other than breaking existing functionality into separate modules (except some minor necessary tweaks to make that happen).

Some notes:

`validate` is now supposed to just take a protocol specification and indicates whether it is a valid specification or not. Crucially it no longer makes any conversion to Python code/objects. This means it can be reused before conversions of specification to other languages. Also, other format checking logic would later go into this module.

`extract_specification` is placed into its own module mainly because the extraction logic has many dependencies (private functions, constants, etc) which cluttered up the old generator and they could logically be grouped together in an independent way. `extract()` automatically calls `validate()` because it always needs a valid specification to do its work.

All the above are now placed in a `generator` package under `aea.protocols`

## Types of changes

## Checklist

### DIFF
--- a/aea/cli/generate.py
+++ b/aea/cli/generate.py
@@ -38,7 +38,7 @@ from aea.configurations.base import (
     PublicId,
 )
 from aea.configurations.loader import ConfigLoader
-from aea.protocols.generator.generator import ProtocolGenerator
+from aea.protocols.generator.base import ProtocolGenerator
 
 
 @click.group()
--- a/aea/cli/generate.py
+++ b/aea/cli/generate.py
@@ -38,7 +38,7 @@ from aea.configurations.base import (
     PublicId,
 )
 from aea.configurations.loader import ConfigLoader
-from aea.protocols.generator import ProtocolGenerator
+from aea.protocols.generator.generator import ProtocolGenerator
 
 
 @click.group()
--- a/aea/protocols/generator/common.py
+++ b/aea/protocols/generator/common.py
@@ -1,0 +1,110 @@
+# -*- coding: utf-8 -*-
+# ------------------------------------------------------------------------------
+#
+#   Copyright 2018-2019 Fetch.AI Limited
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+# ------------------------------------------------------------------------------
+"""This module contains the protocol generator."""
+# pylint: skip-file
+
+SPECIFICATION_PRIMITIVE_TYPES = ["pt:bytes", "pt:int", "pt:float", "pt:bool", "pt:str"]
+
+
+def _get_sub_types_of_compositional_types(compositional_type: str) -> tuple:
+    """
+    Extract the sub-types of compositional types.
+
+    This method handles both specification types (e.g. pt:set[], pt:dict[]) as well as python types (e.g. FrozenSet[], Union[]).
+
+    :param compositional_type: the compositional type string whose sub-types are to be extracted.
+    :return: tuple containing all extracted sub-types.
+    """
+    sub_types_list = list()
+    if compositional_type.startswith("Optional") or compositional_type.startswith(
+        "pt:optional"
+    ):
+        sub_type1 = compositional_type[
+            compositional_type.index("[") + 1 : compositional_type.rindex("]")
+        ].strip()
+        sub_types_list.append(sub_type1)
+    if (
+        compositional_type.startswith("FrozenSet")
+        or compositional_type.startswith("pt:set")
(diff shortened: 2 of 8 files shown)


aea/protocols/generator/common.py
```diff
@@ -0,0 +1,110 @@
+# -*- coding: utf-8 -*-
```
5A11, line 1: This module has code common to all generator modules. Currently only one function and one constant.

aea/protocols/generator/extract_specification.py
```diff
@@ -0,0 +1,262 @@
+# -*- coding: utf-8 -*-
```
5A11, line 1: This module now has the logic to extract protocol specification fields into pythonic equivalences. 

It has one function `extract()` with all its private sub-functions. It extracts all specification fields into python objects and encapsulates them into a `PythonicProtocolSpecification` class.

I separated these code into its own module mainly to declutter the main generator and because they have a nice logical grouping.

aea/protocols/generator/generator.py
```diff
@@ -2405,50 +2012,61 @@ def _generate_file(self, file_name: str, file_content: str) -> None:
         with open(pathname, "w") as file:
             file.write(file_content)
 
-    def generate(self) -> None:
+    def generate(self, protobuf_only: bool = False) -> None:
         """
-        Create the protocol package with Message, Serialization, __init__, protocol.yaml files.
+        Run the generator. If in "full" mode (protobuf_only is False), it:
+        a) validates the protocol specification.
```
DavidMinarsch, line 2018: Validation seems to be happening elsewhere, I cannot see it below.

aea/protocols/generator/validate.py
```diff
@@ -0,0 +1,101 @@
+# -*- coding: utf-8 -*-
```
5A11, line 1: This module has code to validate a protocol specification. It crucially no longer does validation on Pythonic conversion of specification fields. All validation is done purely on the specification. 

This means it can be reused just before converting specification to other languages.

5A11: Currently it has the existing validation logic. Other format checking logic (an outstanding generator issue) will be added in this module.

aea/protocols/generator/generator.py
```diff
@@ -2405,50 +2012,61 @@ def _generate_file(self, file_name: str, file_content: str) -> None:
...
+                self._generate_file(
+                    DIALOGUE_DOT_PY_FILE_NAME, self._dialogue_class_str()
+                )
+            if len(self.spec.all_custom_types) > 0:
+                self._generate_file(
+                    CUSTOM_TYPES_DOT_PY_FILE_NAME, self._custom_types_module_str()
+                )
+            self._generate_file(
+                SERIALIZATION_DOT_PY_FILE_NAME, self._serialization_class_str()
             )
```
DavidMinarsch, line 2056: This should be wrapped in a public method `generate_python_files` so it becomes a single line here

aea/protocols/generator/generator.py
```diff
@@ -2405,50 +2012,61 @@ def _generate_file(self, file_name: str, file_content: str) -> None:
...
         self._generate_file(
             "{}.proto".format(self.protocol_specification.name),
             self._protocol_buffer_schema_str(),
         )
 
-        # Warn if specification has custom types
-        if len(self._all_custom_types) > 0:
-            incomplete_generation_warning_msg = "The generated protocol is incomplete, because the protocol specification contains the following custom types: {}. Update the generated '{}' file with the appropriate implementations of these custom types.".format(
-                self._all_custom_types, CUSTOM_TYPES_DOT_PY_FILE_NAME
+        if not protobuf_only:
```
DavidMinarsch, line 2038: Have a guard clause here:
``` python
if protobuf_only:
    return
```

aea/protocols/generator/generator.py
```diff
@@ -26,12 +26,11 @@
 from datetime import date
```
DavidMinarsch, line 26: Better to call it `base.py` to align with our naming conventions.

aea/cli/generate.py
```diff
@@ -38,7 +38,7 @@
     PublicId,
 )
 from aea.configurations.loader import ConfigLoader
-from aea.protocols.generator import ProtocolGenerator
+from aea.protocols.generator.generator import ProtocolGenerator
```
DavidMinarsch, line 41: Your init file below is missing header info

aea/protocols/generator/generator.py
5A11: Validation is a prerequisite for extraction so `extract()` already calls `validate()`. 

The docstring gives an overall explanation of major things that happen in the two modes.

aea/protocols/generator/generator.py
5A11: Done

aea/protocols/generator/generator.py
5A11: Done

aea/protocols/generator/generator.py
5A11: Done

aea/cli/generate.py
5A11: Done

PR #1442, opened 2020-07-01
Minor changes to dialogue API

5A11:
This is a minor PR with the following changes:

- changed `AgentRole` --> `Role` in FipaDialogue.
- `counterparty` field is automatically set for all messages except the initial one. This can be done for the initial message once a new `create_dialogue` method is added which gets the `counterparty` as a field.
- minor formatting

Due to the changed names (item 1 and 2 above), skill codes no longer work and linters fail. Those were not changed because skills are being heavily changed in another branch.

## Types of changes

## Checklist


aea/helpers/dialogue/base.py
```diff
@@ -697,6 +696,7 @@ def update(self, message: Message) -> Optional[Dialogue]:
             dialogue = self.get_dialogue(message)
 
         if dialogue is not None:
+            message.counterparty = dialogue.dialogue_label.dialogue_opponent_addr
```
DavidMinarsch, line 699: Do we have a check in here that guarantees the counterparty is not overwritten and throws an error on inconsistencies?

packages/fetchai/protocols/fipa/dialogues.py
```diff
@@ -82,7 +82,7 @@ class FipaDialogue(Dialogue):
         ),
     }
 
-    class AgentRole(Dialogue.Role):
+    class Role(Dialogue.Role):
```
DavidMinarsch, line 85: Please regenerate all protocols with the current generator (but don't bump the versions). Otherwise the hashes won't be correct.

aea/helpers/dialogue/base.py
5A11: Done

packages/fetchai/protocols/tac/dialogues.py
```diff
@@ -68,11 +68,11 @@ class TacDialogue(Dialogue):
         ),
     }
 
-    class AgentRole(Dialogue.Role):
+    class Role(Dialogue.Role):
```
DavidMinarsch, line 71: The roles are not  auto sorted by alphabet as you can see below

DavidMinarsch: Please amend generator to do so

aea/protocols/generator/base.py
```diff
@@ -2017,14 +2017,14 @@ def generate_full_mode(self) -> None:
...
-        try_run_black_formatting(self.path_to_generated_protocol_package)
-
         # Run protocol buffer compiler
         try_run_protoc(
             self.path_to_generated_protocol_package, self.protocol_specification.name
         )
 
+        # Run black formatting
+        try_run_black_formatting(self.path_to_generated_protocol_package)
+
```
5A11, line 2027: Running `black` after `protoc` so it formats `_pb2.py` files too.

PR #1446, opened 2020-07-02
Fix registry tests

marcofavorito:
## Proposed changes

n/a

## Fixes

n/a

## Types of changes

## Checklist

## Further comments
n/a


tests/test_packages/test_contracts/test_erc1155/test_contract.py
```diff
@@ -64,21 +64,15 @@
 @pytest.fixture(params=ledger)
 def ledger_api(request):
     ledger_id, config = request.param
-    import aea  # noqa # ensures registries are populated!
```
marcofavorito, line 67: this is ensured by the above change

aea/__init__.py
```diff
@@ -21,6 +21,7 @@
 
 import inspect
 import os
+import aea.crypto  # triggers registry population
```
marcofavorito, line 24: this calls `aea.crypto.__init__.py`, which contains the instructions for the population of the registry

tests/test_packages/test_contracts/test_erc1155/test_contract.py
```diff
@@ -64,21 +64,15 @@
...
-
     api = ledger_apis_registry.make(ledger_id, **config)
     yield api
-    ledger_apis_registry.specs.pop(ledger_id)
```
marcofavorito, line 71: This line was simply wrong.

tests/test_packages/test_contracts/test_erc1155/test_contract.py
```diff
@@ -64,21 +64,15 @@
...
 
 
 @pytest.fixture(params=crypto)
 def crypto_api(request):
     crypto_id = request.param[0]
-    import aea  # noqa # ensures registries are populated!
-
     api = crypto_registry.make(crypto_id)
     yield api
-    crypto_registry.specs.pop(crypto_id)
```
marcofavorito, line 81: see above

tests/test_packages/test_contracts/test_erc1155/test_contract.py
marcofavorito: Ledger APIs are registered only once; they shouldn't be deregistered

tests/test_packages/test_connections/test_ledger/test_contract_api.py
```diff
@@ -70,13 +70,18 @@ def load_erc1155_contract():
     contract = Contract.from_config(configuration)
     assert contract.contract_interface is not None
 
+    # TODO some other tests don't deregister contracts from the registry.
+    #   find a neater solution.
+    if configuration.public_id in contract_registry.specs.keys():
+        contract_registry.specs.pop(str(configuration.public_id))
```
marcofavorito, line 76: I think we need this because the contract may be already registered by other tests, not necessarily the ones using this fixture

marcofavorito: (In particular, any in-process code that runs an AEA with the ERC1155 contract!)

tests/test_packages/test_contracts/test_erc1155/test_contract.py
marcofavorito: Ledger APIs are registered only once, and they should not be deregistered.

setup.py
```diff
@@ -17,50 +17,15 @@
...
-    skills_dependencies = {
-        k.split("_")[0] + "-skill": v
-        for k, v in vars(skills_module).items()
-        if re.match(".+_dependencies", k)
-    }
-    result.update(skills_dependencies)
-
-    return result
-
-
```
marcofavorito, line 63: quite obsolete now

aea/connections/__init__.py
```diff
@@ -17,7 +17,4 @@
 #
 # ------------------------------------------------------------------------------
 
-"""This module contains the channel modules."""
-from typing import List
-
-stub_dependencies = ["watchdog"]  # type: List[str]
```
marcofavorito, line 23: stub connection does not depend on watchdog anymore, hence I removed it

aea/__init__.py
marcofavorito: Interestingly, this line caused the `setup.py build` to fail on CI.
The reason is that we import from `aea` to retrieve the extras, but that causes other third party packages to be imported, and those attempt fails because they are not installed yet.

Solved by removing the obsolete retrieval of the extras (see the changes in `setup.py`).

setup.py
DavidMinarsch: are we sure we don't want to keep it around?

marcofavorito: don't know; Anyway, if we will really need many default packages, I think it is better to retrieve the dependencies from their `yaml` file instead of this approach.

PR #1470, opened 2020-07-06
Fix check of receiving task list in Ledger API connection

marcofavorito:
## Proposed changes

Add a while loop to check until the receiving task list is empty.

## Fixes

n/a

## Types of changes

## Checklist


aea/helpers/base.py
```diff
@@ -295,7 +295,6 @@ def wrapper(*args, **kwargs):
             try:
                 return fn(*args, **kwargs)
             except Exception as e:  # pylint: disable=broad-except  # pragma: no cover  # generic code
-                logger.exception(e)
```
marcofavorito, line 298: it was added for debugging purposes; no need to have it (although it could still be useful to have the exception trace in the output)

packages/fetchai/connections/ledger/connection.py
```diff
@@ -97,12 +97,12 @@ async def send(self, envelope: "Envelope") -> None:
...
         :return: None
         """
-        task = await self._schedule_request(envelope)
+        task = self._schedule_request(envelope)
         self.receiving_tasks.append(task)
         self.task_to_request[task] = envelope
         self.event_new_receiving_task.set()
 
-    async def _schedule_request(self, envelope: Envelope) -> Task:
+    def _schedule_request(self, envelope: Envelope) -> Task:
```
marcofavorito, line 105: `_schedule_request` wasn't needed to be an asynchronous method. It doesn't change the behavior because we weren't using `await` anyway. But I think it simplifies the analysis.

packages/fetchai/connections/ledger/connection.py
```diff
@@ -133,9 +133,13 @@ async def receive(self, *args, **kwargs) -> Optional["Envelope"]:
...
 
-        if len(self.receiving_tasks) == 0:
-            await self.event_new_receiving_task.wait()
+        while len(self.receiving_tasks) == 0:
             self.event_new_receiving_task.clear()
+            try:
+                await self.event_new_receiving_task.wait()
+            except Exception:
+                return None
+        self.event_new_receiving_task.clear()
```
marcofavorito, line 142: Check it the list is empty. If so, clear the flag (that could be inadvertently set?).
Then, wait until the flag has been set (by the `send()` method). If something bad happens (e.g. the receive task is cancelled), return None.
Then, clear the flag again, because we know the receiving task list is not empty.

scripts/generate_ipfs_hashes.py
```diff
@@ -406,8 +406,6 @@ def update_hashes(arguments: argparse.Namespace) -> int:
                         package_path.name, package_type
                     )
                 )
-                if package_path.name == "dummy_aea":
-                    print("help")
```
marcofavorito, line 410: added for debugging, not needed at all

aea/helpers/base.py
DavidMinarsch: Yes very anoying; it needs to be removed otherwise it will confuse the user.

packages/fetchai/connections/ledger/connection.py
```diff
@@ -133,9 +133,13 @@ async def receive(self, *args, **kwargs) -> Optional["Envelope"]:
             done_task = self.done_tasks.pop()
             return self._handle_done_task(done_task)
 
-        if len(self.receiving_tasks) == 0:
-            await self.event_new_receiving_task.wait()
+        while len(self.receiving_tasks) == 0:
```
DavidMinarsch, line 136: why the check to while?

marcofavorito: please check the new version

packages/fetchai/connections/ledger/connection.py
```diff
@@ -134,8 +134,8 @@ async def receive(self, *args, **kwargs) -> Optional["Envelope"]:
             return self._handle_done_task(done_task)
 
         if len(self.receiving_tasks) == 0:
-            await self.event_new_receiving_task.wait()
             self.event_new_receiving_task.clear()
+            await self.event_new_receiving_task.wait()
```
marcofavorito, line 138: no need to clear after, because we wouldn't wait for the event to be set if `len(self.receiving_tasks) > 0`

Also, clear the event before, because if the list is empty we are sure we need to wait.

PR #1471, opened 2020-07-06
AEA-608 Experimental Dialogue API

5A11:
This is an experimental PR. It shows the effects of the following changes in dialogue's API:

- Added `dialogues.create(...)` method,
- Added `dialogue.reply(...)` method,
- Removed `create_dialpogue(...)` abstract method

on the following:

- base dialogue module, 
- Fipa dialogue module, 
- Fipa dialogue's usage in generic buyer skill.

The ideas behind these changes are the following:
- A motivation behind both  `dialogues.create(...)` and `dialogue.reply(...)` is that if you are creating new messages (either when it is an initial message for a new dialogue, or replying to an existing one) you do not need to manually set many of the message's meta fields (i.e. `message_id`, `target`, `dialogue_reference`, `counterparty`). Those can be set for you  automatically because the message is in the context of a dialogue. This significantly reduces error in creating new messages. All you need to specify for a new message in this context, are the performative and contents, and the target message if you are replying to one.
- If the agent wants to create a new dialogue with a new message, `dialogues.create(...)` must be used instead of `dialogues.update(...)`. `dialogues.update(...)` used to be a "magical" method that does too many things. This change makes it so `dialogues.update(...)` is only for "updating the model of dialogues with an incoming message from the opponent" now.
- If the agent wants to reply to an existing message with a new message, `dialogue.reply(...)` is used. 

Since base `dialogue` and `dialogues` classes in the current design are expected to manipulate dialogue objects, they should be made aware of the concrete classes of the objects they are expected to work with. These are: a concrete `Message` subclass for base `Dialogue`, and concrete `Message` and `Dialogue` subclasses for base `Dialogues`. Hence the object properties `_message_class` and `_dialogue_class` in base `Dialogue` and `Dialogues`.

Then for the Fipa instance, the concrete `FipaMessage` and `FipaDialogue` are fed into `FipaDialogue` and `FipaDialogues`. 

If `FipaDialogue` is further subclassed, as in the generic buyer skill, then that class is fed to the corresponding dialogues class.

This means the `create_dialogue` method, whose purpose was merely to construct a concrete instance of the dialogue we want to manipulate, could be removed from base `Dialogues` and `FipaDialogues` classes and any classes derived from `FipaDialogues` (e.g. in skills).

## Types of changes

## Checklist


aea/helpers/dialogue/base.py
```diff
@@ -222,6 +222,7 @@ def __str__(self):
     def __init__(
         self,
         dialogue_label: DialogueLabel,
+        message_class,
```
DavidMinarsch, line 225: Missing type hint; should be "Type[Message]".

aea/helpers/dialogue/base.py
```diff
@@ -400,6 +404,24 @@ def update(self, message: Message) -> bool:
...
+            performative=performative,
+            **kwargs,
+        )
+        reply.counterparty = self.dialogue_label.dialogue_opponent_addr
+        result = self.update(reply)
+
+        if result:
+            return reply
+        else:
+            return None
```
DavidMinarsch, line 423: I think we should throw an error; because the dialogue not being constructed means we have specified something incorrectly in the kwargs; which should be thrown rather than handled.

packages/fetchai/skills/generic_buyer/dialogues.py
```diff
@@ -161,7 +161,9 @@ def __init__(self, **kwargs) -> None:
         :return: None
         """
         Model.__init__(self, **kwargs)
-        BaseFipaDialogues.__init__(self, self.context.agent_address)
+        BaseFipaDialogues.__init__(
+            self, self.context.agent_address, dialogue_class=FipaDialogue
+        )
```
5A11, line 166: Specifying the dialogue object to be manipulated by `Dialogues` as the subclassed `FipaDialogue` above.

aea/helpers/dialogue/base.py
```diff
@@ -614,7 +636,11 @@ class Dialogues(ABC):
     """The dialogues class keeps track of all dialogues for an agent."""
 
     def __init__(
-        self, agent_address: Address, end_states: FrozenSet[Dialogue.EndState]
+        self,
+        agent_address: Address,
+        end_states: FrozenSet[Dialogue.EndState],
+        message_class,
```
DavidMinarsch, line 642: type hints; see above

aea/helpers/dialogue/base.py
```diff
@@ -656,6 +688,31 @@ def new_self_initiated_dialogue_reference(self) -> Tuple[str, str]:
         """
         return str(self._dialogue_nonce + 1), ""
 
+    def create(self, counterparty, performative, **kwargs,) -> Tuple[Message, Dialogue]:
+        initial_message = self._message_class(
+            dialogue_reference=("", ""),
```
DavidMinarsch, line 693: wrong; we have a constructor method for the dialogue reference which needs to be used here.

aea/helpers/dialogue/base.py
```diff
@@ -656,6 +688,31 @@ def new_self_initiated_dialogue_reference(self) -> Tuple[str, str]:
...
+            message_id=1,
+            target=0,
```
DavidMinarsch, line 695: should be expressed in constants (we already have it i think)

aea/helpers/dialogue/base.py
```diff
@@ -656,6 +688,31 @@ def new_self_initiated_dialogue_reference(self) -> Tuple[str, str]:
         """
         return str(self._dialogue_nonce + 1), ""
 
+    def create(self, counterparty, performative, **kwargs,) -> Tuple[Message, Dialogue]:
```
DavidMinarsch, line 691: type hints missing

aea/helpers/dialogue/base.py
```diff
@@ -656,6 +688,31 @@ def new_self_initiated_dialogue_reference(self) -> Tuple[str, str]:
...
+            target=0,
+            performative=performative,
+            **kwargs,
+        )
+        initial_message.counterparty = counterparty
+
+        dialogue = self._create_self_initiated(
+            dialogue_opponent_addr=counterparty,
+            role=self.role_from_first_message(initial_message),
+        )
```
DavidMinarsch, line 704: Why don't we provide role as an argument as well?

DavidMinarsch: See above; we create dialogue label before we pass it to this...

aea/helpers/dialogue/base.py
```diff
@@ -656,6 +688,31 @@ def new_self_initiated_dialogue_reference(self) -> Tuple[str, str]:
...
+        initial_message.set(
+            "dialogue_reference", dialogue.dialogue_label.dialogue_reference
+        )
```
DavidMinarsch, line 707: remove; see above

aea/helpers/dialogue/base.py
```diff
@@ -656,6 +688,31 @@ def new_self_initiated_dialogue_reference(self) -> Tuple[str, str]:
...
+
+        successfully_updated = dialogue.update(initial_message)
+        assert (
+            successfully_updated
+        ), "Cannot create the a dialogue with the specified performative and contents."
```
DavidMinarsch, line 712: This is a problem in code; so it should throw rather than being handled.

aea/helpers/dialogue/base.py
5A11: ok I wasn't aware how type hints for a class type is. Done

aea/helpers/dialogue/base.py
5A11: Agree.

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: Agreed

aea/helpers/dialogue/base.py
5A11: When a message is being created in the context of a dialogue, its `dialogue_reference` is set to the dialogue's `dialogue_reference` (e.g. in `dialogue.reply(...)`).

But for an initial message here, the dialogue hasn't been created yet, and the dialogue cannot be created before the message because for it we need the message to set the role. That's why I create the initial message first, with some arbitrary dialogue_reference`, then I create an empty dialogue off of this message. Then I update the message's `dialogue_reference` to be the dialogue's, then add the message to the dialogue and return.

DavidMinarsch: That's incorrect. Look at skills usage; we always create the label before we create the dialogue...https://github.com/fetchai/agents-aea/blob/master/packages/fetchai/skills/generic_buyer/handlers.py#L423

aea/helpers/dialogue/base.py
```diff
@@ -679,18 +736,6 @@ def update(self, message: Message) -> Optional[Dialogue]:
                 dialogue_reference=dialogue_reference,
                 role=self.role_from_first_message(message),
             )  # type: Optional[Dialogue]
-        elif (  # new dialogue by self
```
DavidMinarsch, line 682: Cannot be removed; we want to be compatible and also see my other comment above

aea/helpers/dialogue/base.py
```diff
@@ -829,24 +879,16 @@ def _create_opponent_initiated(
...
+            message_class=self._message_class,
+            agent_address=self.agent_address,
+            role=role,
+        )
         self.dialogues.update({dialogue_label: dialogue})
 
         return dialogue
 
-    @abstractmethod
-    def create_dialogue(
```
DavidMinarsch, line 838: Let's keep the method (so we keep the interface definition) until 0.6 but remove the decorator. You can also add a note that it is deprecated...

packages/fetchai/protocols/fipa/dialogues.py
```diff
@@ -161,20 +167,6 @@ def __init__(self, agent_address: Address) -> None:
             self,
             agent_address=agent_address,
             end_states=cast(FrozenSet[Dialogue.EndState], self.END_STATES),
+            dialogue_class=dialogue_class,
+            message_class=message_class,
         )
-
-    def create_dialogue(
```
DavidMinarsch, line 166: Here it can be removed indeed

aea/helpers/dialogue/base.py
5A11: Well, after changing the errors I realised its going to change the interface (the return values of methods). So need to hold this off for the next release if we want to be backwards compatible for this release. I make a list of all breaking changes for the next release to not forget anything.

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: See above. Refactoring errors to be done for the next release as it would break the interface.

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: Done

packages/fetchai/protocols/fipa/dialogues.py
5A11: Yep

aea/helpers/dialogue/base.py
DavidMinarsch: reply doesn;t exist in 0.5.0 so you can do it here: raise if reply is None

aea/helpers/dialogue/base.py
DavidMinarsch: create is a new method so you can raise here and ensure it only returns on success

aea/helpers/dialogue/base.py
5A11: Done

PR #1481, opened 2020-07-07
[AEA-645] [AEA-687] Improve libp2p tests stability - Fully async libp2p connection

lrahmani:
## Proposed changes

- Increase timeouts for AEATestCase tests and address lookup in DHTPeer
- Remove `flaky` marker from all libp2p tests
- Extract all (DHTPeer and DHTClient) timeouts as constants
- In case of address lookup failure, retry
- Fully async `go build`: run in a separate thread, to avoid blocking the agent loop
- Add libp2p node logging in case of libp2p python tests failures


aea/test_tools/test_cases.py
```diff
@@ -660,6 +660,9 @@ def missing_from_output(
             cls.terminate_agents(process)
         if missing_strings != []:
             print("Non-empty missing strings, stderr:\n" + cls.stderr[process.pid])
+            print("=====================")
```
DavidMinarsch, line 663: We use logger throughout; please change.

Issue #1286, opened 2020-05-27
Probably blocking operation.


solarw:
https://github.com/fetchai/agents-aea/blob/4fcd5c55ae03373d790321a00305b6c98d6518c9/packages/fetchai/connections/oef/connection.py#L727

looks like blocking operation that can freeze multiplexer loop

underlying function is:
```
 def connect(self, timeout=2) -> bool:
        self.state = "connecting"
        self.conn = Connection(self.core, logger=self.logger)
        self.conn.connect(url=self.url, success=self.my_on_connect_ok, failure=self.my_on_connect_fail, public_key=self.public_key)

        starttime = time.perf_counter();

        while self.state == "connecting":
            delta = time.perf_counter() - starttime;
            if delta > timeout:
                self.my_on_connect_timeout(self.conn, self.url)
            time.sleep(0.1)
        return self.state == "connected"
```

PR #1486, opened 2020-07-09
[AEA-686] OEF connection full test coverage and resolve blocking logic

solarw:
## Proposed changes

asyncio friendly oef connection and full test coverage

## Fixes

https://github.com/fetchai/agents-aea/issues/1286

## Types of changes

## Checklist


packages/fetchai/connections/oef/connection.py
```diff
@@ -82,6 +44,9 @@
 )
 from packages.fetchai.protocols.oef_search.message import OefSearchMessage
 
+from .object_translator import OEFObjectTranslator
```
DavidMinarsch, line 47: please provide an absolute rather than relative path

packages/fetchai/connections/oef/connection.py
```diff
@@ -427,12 +171,9 @@ def on_message(
...
         # We are not using the 'msg_id', 'dialogue_id' and 'origin' parameters because 'content' contains a
         # serialized instance of 'Envelope', hence it already contains this information.
-        assert self.in_queue is not None
-        assert self.loop is not None
+        self._check_loop_and_queue()
         envelope = Envelope.decode(content)
-        asyncio.run_coroutine_threadsafe(
-            self.in_queue.put(envelope), self.loop
-        ).result()
+        asyncio.run_coroutine_threadsafe(self.in_queue.put(envelope), self.loop)
```
DavidMinarsch, line 176: do we not need to call result?

packages/fetchai/connections/oef/connection.py
```diff
@@ -534,16 +271,17 @@ def on_search_result(self, search_id: int, agents: List[Address]) -> None:
...
-            return
+            logger.warning(
+                "Could not find dialogue for search_id={}".format(search_id)
+            )  # pragma: nocover
+            return  # pragma: nocover
         last_msg = oef_search_dialogue.last_incoming_message
         if last_msg is None:
-            logger.warning("Could not find last message.")
-            return
+            logger.warning("Could not find last message.")  # pragma: nocover
```
DavidMinarsch, line 283: if you place `pragma: no cover` after the if it works better

packages/fetchai/connections/oef/connection.py
```diff
@@ -573,20 +309,21 @@ def on_oef_error(
...
-            return
+            logger.warning(
+                "Could not find dialogue for answer_id={}".format(answer_id)
+            )  # pragma: nocover
+            return  # pragma: nocover
         last_msg = oef_search_dialogue.last_incoming_message
         if last_msg is None:
-            logger.warning("Could not find last message.")
-            return
+            logger.warning("Could not find last message.")  # pragma: nocover
```
DavidMinarsch, line 325: see above

packages/fetchai/connections/oef/connection.py
```diff
@@ -682,8 +415,8 @@ def send_oef_message(self, envelope: Envelope) -> None:
         if oef_search_dialogue is None:
             logger.warning(
                 "Could not create dialogue for message={}".format(oef_message)
-            )
-            return
+            )  # pragma: nocover
```
DavidMinarsch, line 418: see above

packages/fetchai/connections/oef/connection.py
solarw: the result will block and this result even not returned from the function

packages/fetchai/connections/oef/connection.py
solarw: yes, but the main idea do not hide all if block from coverage, in case i'd like to add some other code into this if, I can forget to remove  nocover directive and new code will be skipped from coverage as well, so decided to use nocover only for specific line

Issue #1118, opened 2020-04-19
Add 'agent name' to AEA logger messages by default 


marcofavorito:
**Is your feature request related to a problem? Please describe.**
It would be useful to have the 'agent_name' by default in the logger messages, e.g. in the ones generated by a skill. It would be useful especially in the case of `aea launch` commands, in order to distinguish between agent logs.

**Describe the solution you'd like**
Prepend the agent name in all the log message produced by an AEA.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

Issue #1050, opened 2020-04-03
All loggers of components should have `{agent_name}` as prefix


marcofavorito:
**Is your feature request related to a problem? Please describe.**
Currently, we cannot distinguish between log messages of different agents because the loggers used by the components don't contain the agent name. This is related to #607, #1033 .
 
**Describe the solution you'd like**
Add the agent name to the logger name of any component. Probably there will be the need to set it dynamically, and not hard-coded at the top of the module, as it is done in some package.

**Describe alternatives you've considered**

**Additional context**
For the module name: 
https://docs.python.org/3/library/logging.html#logrecord-attributes

However, it does not handle nested modules in a package correctly. So the best solution is to build the dotted path of the logger name dynamically.

PR #1491, opened 2020-07-13
AEA-660 Set logger for each component and prepend agent_name to log messages

### LINKED ISSUE #1327: Agent level logging context
current loggers do not refers to an agent, so they can not print the agent's name.
create root logger bound to an agent to print agent names in every logline.

create root logger bound to agent and initiate loggers from it, probably use agent context.
need to find a nice solution: especially in the respect we can run multiple agents in the same process, so we can not modify logger behavior globally.

requires better overview

marcofavorito:
## Proposed changes

 Set logger for each component and prepend agent_name to log messages

## Fixes

Fixes #1327, #1118 , #1050

## Types of changes

## Checklist


aea/aea_builder.py
```diff
@@ -879,11 +880,12 @@ def build(self, connection_ids: Optional[Collection[PublicId]] = None,) -> AEA:
...
         identity = self._build_identity_from_wallet(wallet)
-        self._load_and_add_components(ComponentType.PROTOCOL, resources)
-        self._load_and_add_components(ComponentType.CONTRACT, resources)
+        self._load_and_add_components(ComponentType.PROTOCOL, resources, identity)
+        self._load_and_add_components(ComponentType.CONTRACT, resources, identity)
         self._load_and_add_components(
             ComponentType.CONNECTION,
             resources,
+            identity,
             identity=identity,
```
DavidMinarsch, line 889: identity twice?

aea/helpers/logging.py
```diff
@@ -0,0 +1,66 @@
...
+
+        :param agent_name: the agent name.
+        """
+        super().__init__(logger, dict(agent_name=agent_name))
+
+    def process(
+        self, msg: Any, kwargs: MutableMapping[str, Any]
+    ) -> Tuple[Any, MutableMapping[str, Any]]:
+        """Prepend the agent name to every log message."""
+        return "[%s] %s" % (self.extra["agent_name"], msg), kwargs
```
DavidMinarsch, line 40: either f string or .format; we need some consistency with string interpolation

packages/fetchai/connections/gym/connection.py
```diff
@@ -57,6 +57,7 @@ def __init__(self, address: Address, gym_env: gym.Env):
         self._threaded_pool: ThreadPoolExecutor = ThreadPoolExecutor(
             self.THREAD_POOL_SIZE
         )
+        self.logger: Union[logging.Logger, logging.LoggerAdapter] = logger
```
DavidMinarsch, line 60: where is logger initiated?

packages/fetchai/connections/http_client/connection.py
```diff
@@ -85,9 +85,12 @@ def __init__(
         )  # type: Optional[asyncio.AbstractEventLoop]  # pragma: no cover
         self.excluded_protocols = excluded_protocols
         self.is_stopped = True
-        logger.info("Initialised the HTTP client channel")
         self._tasks: Set[Task] = set()
 
+        self.logger = logger
+        # TODO logger at this point is the module-level one, not with the agent name.
```
DavidMinarsch, line 91: suggests we should pass logger during construction?

packages/fetchai/connections/oef/connection.py
```diff
@@ -139,6 +139,8 @@ def __init__(
 
         self._threaded_pool = ThreadPoolExecutor(self.THREAD_POOL_SIZE)
 
+        self.aea_logger = logger
```
DavidMinarsch, line 142: can we use `logger` or does this clash with oef?

packages/fetchai/connections/oef/connection.py
```diff
@@ -563,12 +568,14 @@ async def _connection_check(self) -> None:
             if not self.channel.get_state() == "connected":  # pragma: no cover
                 self.connection_status.is_connected = False
                 self.connection_status.is_connecting = True
-                logger.warning(
+                self.logger.warning(
```
DavidMinarsch, line 571: why now `logger` and not `aea_logger`?

packages/fetchai/connections/p2p_client/connection.py
```diff
@@ -63,7 +63,8 @@ def __init__(
         self.thread = Thread(target=self.receiving_loop)
         self.lock = threading.Lock()
         self.stopped = True
-        logger.info("Initialised the peer to peer channel")
+        self.logger = logger
```
DavidMinarsch, line 66: Why don't we delegate this to base?

aea/aea_builder.py
marcofavorito: this is because one is a keyword argument to be forwarded to the component constructor; the first one instead is for retrieve the agent name. I could replace the first with the agent name.

packages/fetchai/connections/gym/connection.py
marcofavorito: right!

aea/helpers/logging.py
marcofavorito: right

packages/fetchai/connections/gym/connection.py
marcofavorito: the logger on the right is the top-level module logger.
It is overwritten by `GymConnection.connect` to replace it with the wrapped logger (i.e. with the agent name prepended).
We can't to that at init time because we create the loggers after the component is instantiated. We could have done the other way, but would have been too intrusive (we would have to modify the init of many components)

packages/fetchai/connections/http_client/connection.py
marcofavorito: here we are in the channel; that means we should pass the logger during construction also for the connection.

packages/fetchai/connections/oef/connection.py
marcofavorito: it does clash

packages/fetchai/connections/oef/connection.py
marcofavorito: because we are in `OEFConnection` (whose `logger` is inherited from `Component(WithLogger)`) and it does not clash with anything unlike not `OEFChannel(OEFAgent)`.

packages/fetchai/connections/p2p_client/connection.py
marcofavorito: because the `logger` on the right is the module logger, which is accessible only in the inheriting class.

packages/fetchai/connections/http_client/connection.py
DavidMinarsch: @marcofavorito I don't think we should add TODO here in package;

packages/fetchai/connections/p2p_stub/connection.py
```diff
@@ -29,9 +28,7 @@
 from aea.identity.base import Identity
 from aea.mail.base import Envelope
 
-logger = logging.getLogger(__name__)
```
DavidMinarsch, line 32: Wasn't used; so removed it @marcofavorito

aea/aea_builder.py
DavidMinarsch: Ugly, but ok..

marcofavorito: fixed that

packages/fetchai/connections/http_client/connection.py
marcofavorito: removed, thanks

PR #1540, opened 2020-07-25
update tox.ini

### LINKED ISSUE #1239: Clean up tox testenvs
Some test environments might be out-to-date:
- Currently, we treat Python 3.8 builds in a different way than the previous Python versions. That was due because we introduced Python 3.8 as soon as it has been released, and some dependencies were not up-to-date.
- many libraries for linting might be updated to the latest version.
- for some test environments (especially linting), we don't need to install the package. Use `skipdist=True` where possible.
- in `workflow.yml` we still install `pipenv`, although it's not used anymore in `tox`.

Address the above.

n/a

**Additional context**
n/a

marcofavorito:
## Proposed changes

- Skip the distribution for some testenvs when possible to save time (e.g. 'flake8', 'mypy' etc.)
- simplify test envs for main Python tests.

## Fixes

Fixes #1239 

## Types of changes

## Checklist

## Further comments
n/a

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -135,6 +135,7 @@ skipsdist = True
 skip_install = True
 deps = mypy==0.761
        aiohttp==3.6.2
+       packaging==20.4
 commands = mypy aea benchmark examples packages scripts tests
 
 [testenv:pylint]
--- a/tox.ini
+++ b/tox.ini
@@ -64,7 +64,6 @@ commands = black aea benchmark examples packages scripts tests --check --verbose
 
 [testenv:copyright_check]
 skipsdist = True
-usedevelop = False
 skip_install = True
 deps =
 commands = {toxinidir}/scripts/check_copyright_notice.py --directory {toxinidir}
@@ -89,7 +88,6 @@ commands = {toxinidir}/scripts/check_package_dependencies.py
 
 [testenv:docs]
 skipsdist = True
-usedevelop = False
 skip_install = True
 description = Build the documentation.
 deps = markdown==3.2.1
@@ -102,7 +100,6 @@ commands = pip3 install git+https://github.com/pugong/mkdocs-mermaid-plugin.git#
 
 [testenv:docs-serve]
 skipsdist = True
-usedevelop = False
 skip_install = True
 description = Run a development server for working on documentation.
 deps = markdown==3.2.1
@@ -117,7 +114,6 @@ commands = pip3 install git+https://github.com/pugong/mkdocs-mermaid-plugin.git#
 
 [testenv:flake8]
 skipsdist = True
-usedevelop = False
 skip_install = True
 deps = flake8==3.7.9
        flake8-bugbear==20.1.4
@@ -128,7 +124,7 @@ commands = flake8 aea benchmark examples packages scripts tests
 
 [testenv:liccheck]
 skipsdist = True
-usedevelop = False
+usedevelop = True
 skip_install = False
 deps = liccheck==0.4.3
 commands = {toxinidir}/scripts/freeze_dependencies.py -o {envtmpdir}/requirements.txt
@@ -136,7 +132,6 @@ commands = {toxinidir}/scripts/freeze_dependencies.py -o {envtmpdir}/requirement
 
 [testenv:mypy]
 skipsdist = True
-usedevelop = False
 skip_install = True
 deps = mypy==0.761
        aiohttp==3.6.2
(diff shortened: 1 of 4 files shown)


tox.ini
```diff
@@ -33,73 +30,24 @@ deps =
     pytest-rerunfailures==9.0
 
 commands =
-    pip install git+https://github.com/pytoolz/cytoolz.git#egg=cytoolz==0.10.1.dev0
+    ; ignore exit code of tensorflow because it does not work on Python 3.8
```
DavidMinarsch, line 33: It might be worth checking if we can upgrade the ml example to 2.+ of tensorflow without any work

tox.ini
```diff
@@ -1,46 +1,14 @@
...
-envlist = bandit-main, bandit-tests, black, black-check, copyright_check, docs, flake8, liccheck, mypy, py3.8, py3.7, py3.6
-skipsdist = False
-ignore_basepython_conflict = True
+envlist = bandit, black, black-check, copyright_check, docs, flake8, liccheck, mypy, py{3.6,3.7,3.8}
 
 [testenv]
+basepython = python3
+skipsdist = False
+usedevelop = False
+skip_install = False
```
marcofavorito, line 8: by default, we:
- don't skip the creation of the source distribution (the output of `python setup.py sdist`)
- don't use `develop` installation
- don't skip install.

They were the default settings; but migth be good to state them explicitly for self-documentation. Thoughts? O/w if we think it's too verbose we can get rid of them.

tox.ini
```diff
@@ -1,46 +1,14 @@
...
 whitelist_externals = /bin/sh
-
-[testenv:py3.8]
-basepython = python3.8
-passenv = *
-deps =
-    Cython
```
marcofavorito, line 13: this has been excluded because not needed anymore (it was needed for `cytoolz`)

tox.ini
```diff
@@ -1,46 +1,14 @@
...
-    black==19.10b0
-    mistune==2.0.0a4
-    aiohttp==3.6.2
-    SQLAlchemy==1.3.16
-    pynacl==1.3.0
-    pexpect==4.8.0
-    pytest-rerunfailures==9.0
-
-commands =
-    pip install git+https://github.com/pytoolz/cytoolz.git#egg=cytoolz==0.10.1.dev0
```
marcofavorito, line 36: We are at `0.10.1` now: https://pypi.org/project/cytoolz/

tox.ini
marcofavorito: We could use `develop` in the main testenv to save some time, but better to test also the packaging.

tox.ini
```diff
@@ -64,72 +32,65 @@ deps =
...
-commands =
-    pip install .[all]
-    pip install -i https://test.pypi.org/simple/ fetch-p2p-api==0.0.2
-    pytest -rfE --doctest-modules aea packages/fetchai/protocols packages/fetchai/connections tests/ --cov-report=html --cov-report=xml --cov-report=term --cov=aea --cov=packages/fetchai/protocols --cov=packages/fetchai/connections {posargs}
+[testenv:py3.8]
+basepython = python3.8
 
 [testenv:bandit]
+skipsdist = True
+skip_install = True
```
marcofavorito, line 49: here we skip both the source distribution and the installation.

tox.ini
```diff
@@ -64,72 +32,65 @@ deps =
...
 deps = bandit==1.6.2
 commands = bandit -s B101 -r aea benchmark examples packages scripts tests
 
 [testenv:black]
+skipsdist = True
+skip_install = True
```
marcofavorito, line 55: see above re. `bandit`

tox.ini
```diff
@@ -64,72 +32,65 @@ deps =
...
 deps = black==19.10b0
 commands = black aea benchmark examples packages scripts tests
 
 [testenv:black-check]
+skipsdist = True
+skip_install = True
```
marcofavorito, line 61: see above

tox.ini
```diff
@@ -64,72 +32,65 @@ deps =
...
+skipsdist = True
+usedevelop = True
+deps = ipfshttpclient
 commands = {toxinidir}/scripts/generate_ipfs_hashes.py --check {posargs}
 
 [testenv:package_version_checks]
-deps = python-dotenv
+skipsdist = True
+usedevelop = True
+deps =
```
marcofavorito, line 81: the only dependency is the `aea` package, installed with source installation via `usedevelop = True` and `skip_install = False` (by default)

marcofavorito:
notably, after the latest changes, we got rid of `pip install .[all]` in the `commands:` list of the test envs.

tox.ini
DavidMinarsch: I think we can document in the tox.ini as a comment. We should not use develop. I prefer to catch packaging issues early.

PR #1549, opened 2020-07-27
AEA-773 custom Contract request dispatching

### LINKED ISSUE #1539: Add test coverage and correct abstract methods for contracts
The current contracts ABC is not properly reflecting the requirements of the ledger connection. It also does not have enough tests.

Add tests and correct abstract method signatures.

marcofavorito:
## Proposed changes

This PR adds a new optional way to dispatch Contract API requests coming from the `LedgerConnection` (more precisely, from the `ContractApiRequestDispatcher`).

It provides the developer to specify four methods of its `Contract` subclass:
- `get_deploy_transaction`
- `get_raw_transaction`
- `get_raw_message`
- `get_state`

If implemented, the `ContractApiRequestDispatcher` will call them instead of calling the `message.callable` attribute of the `Contract` instance.

## Fixes

Fixes #1539

## Types of changes

## Checklist

## Further comments

n/a


aea/contracts/base.py
```diff
@@ -121,3 +121,62 @@ def from_config(cls, configuration: ContractConfig) -> "Contract":
         #     contract_interface = json.load(interface_file)
 
         return contract_class(configuration)
+
+    # TODO if 'ContractApiMessage' was a default package
+    #   we could import it and remove the 'type ignore' tag.
+    #   the same applies to the below methods.
```
marcofavorito, line 127: the idea here is that methods like `get_deploy_transaction` let the developer to manually handle contract API requests (in this case `GET_DEPLOY_TRANSACTION` requests). That would mean read the `callable` field to dispatch to the right method, or just processing the request (hence ignoring the `callable` field). The same applies to the other methods below.

packages/fetchai/connections/ledger/contract_dispatcher.py
```diff
@@ -130,6 +132,30 @@ def get_error_message(
         dialogue.update(response)
         return response
 
+    def _handle_request(
```
marcofavorito, line 135: this is the method that does all the work; it is parametrized to save some line of code.

packages/hashes.csv
```diff
@@ -21,19 +21,19 @@ fetchai/agents/weather_station,QmaRVcpYHcyUR6nA1Y5J7zvaYanPr3jTqVtkCjUB4r9axp
 fetchai/connections/gym,QmXpTer28dVvxeXqsXzaBqX551QToh9w5KJC2oXcStpKJG
```
marcofavorito, line 21: for some reason many hashes were out-to-date, the PR doesn't change many of those packages.

aea/contracts/base.py
marcofavorito: An alternative would be to have the exact parameters of a "get deploy transaction callable", that is `(api, **kwargs)`, but in that case, the developer cannot do dispatching because he has not access to the `message`.

packages/fetchai/connections/ledger/contract_dispatcher.py
```diff
@@ -237,19 +248,80 @@ def get_raw_message(
...
+            "get_raw_message",
+            "get_raw_transaction",
+            "get_deploy_transaction",
+        ]
+        method: Callable[[LedgerApi, ContractApiMessage], bytes] = getattr(
+            contract, message.performative.value
+        )
+        data = method(api, message)
+        return data
+    except NotImplementedError:
```
DavidMinarsch, line 280: If the developer does remove the stub method we need to catch the exception too here.

packages/fetchai/connections/ledger/contract_dispatcher.py
```diff
@@ -237,19 +248,80 @@ def get_raw_message(
...
+):
+    """
+    Validate a Contract callable, given the performative.
+
+    In particular:
+    - if the performative is either 'get_state' or 'get_raw_transaction', the signature
+      must accept ledger api as first argument and contract address as second argument,
+      plus keyword arguments.
+    - if the performative is either 'get_deploy_transaction' or 'get_raw_message', the signature
+      must accept ledger api as first argument, plus keyword arguments.
```
DavidMinarsch, line 300: Note; you only check quantity of args; not that they are actually including api and address

aea/contracts/base.py
```diff
@@ -121,3 +121,62 @@ def from_config(cls, configuration: ContractConfig) -> "Contract":
...
+    def get_deploy_transaction(
+        self, api: LedgerApi, message: "ContractApiMessage"  # type: ignore  # noqa
```
DavidMinarsch, line 129: missing kwargs

aea/contracts/base.py
```diff
@@ -121,3 +121,62 @@ def from_config(cls, configuration: ContractConfig) -> "Contract":
...
+        Implement this method in the sub class if you want
+        to handle the contract requests manually.
+
+        :param api: the ledger apis.
+        :param message: the contract API request.
+        :return: the bytes representing the state.
+        """
+        raise NotImplementedError
+
+    def get_raw_transaction(
```
DavidMinarsch, line 143: see above

aea/contracts/base.py
```diff
@@ -121,3 +121,62 @@ def from_config(cls, configuration: ContractConfig) -> "Contract":
...
+        Implement this method in the sub class if you want
+        to handle the contract requests manually.
+
+        :param api: the ledger apis.
+        :param message: the contract API request.
+        :return: the bytes representing the state.
+        """
+        raise NotImplementedError
+
+    def get_raw_message(self, api: LedgerApi, message: "ContractApiMessage") -> bytes:  # type: ignore   # noqa
```
DavidMinarsch, line 158: see above

aea/contracts/base.py
```diff
@@ -121,3 +121,62 @@ def from_config(cls, configuration: ContractConfig) -> "Contract":
...
+        Implement this method in the sub class if you want
+        to handle the contract requests manually.
+
+        :param api: the ledger apis.
+        :param message: the contract API request.
+        :return: the bytes representing the state.
+        """
+        raise NotImplementedError
+
+    def get_state(self, api: LedgerApi, message: "ContractApiMessage") -> bytes:  # type: ignore   # noqa
```
DavidMinarsch, line 171: see above

aea/contracts/base.py
```diff
@@ -121,3 +121,62 @@ def from_config(cls, configuration: ContractConfig) -> "Contract":
         #     contract_interface = json.load(interface_file)
 
         return contract_class(configuration)
+
+    # TODO if 'ContractApiMessage' was a default package
```
DavidMinarsch, line 125: Ok; makes sense to have it as a default package in that case

aea/contracts/base.py
DavidMinarsch, line 171: `ledger_api` not `api`

packages/fetchai/connections/ledger/contract_dispatcher.py
marcofavorito: the names can change; a check on the name would be useless.

aea/contracts/base.py
marcofavorito: ok

aea/contracts/base.py
DavidMinarsch: None of the methods should be instance methods; there is no meaningful instance. We give the illusion that we keep instances alive when we don't. So either class or static method please.

aea/contracts/base.py
marcofavorito: I think it should be included by default, like the default protocol and the error skill.

packages/hashes.csv
DavidMinarsch: concerning

DavidMinarsch: are you on the right branch?

aea/contracts/base.py
marcofavorito: the reason is, the ledger connection is the only reasonable way the framework helps the agent to perform its economic activity, which is the essence of an AEA.

packages/fetchai/contracts/erc1155/contract.py
```diff
@@ -67,8 +67,9 @@ def _generate_id(index: int, token_type: int) -> int:
         token_id = (token_type << 128) + index
         return token_id
 
+    # TODO rename this method to avoid conflicts with the base class method.
     @classmethod
-    def get_deploy_transaction(
+    def get_deploy_transaction(  # type: ignore
```
DavidMinarsch, line 72: this should not require changing

aea/contracts/base.py
DavidMinarsch: Actually; we should not expose the message here at all. That conflates concepts. So this is mute.

packages/fetchai/connections/ledger/contract_dispatcher.py
marcofavorito: The most popular or easy way to remove an attribute are:

aea/contracts/base.py
DavidMinarsch: I am in favour of the approach we already implemented for the existing contract: `ledger_api, **kwargs`. The developer should not have access to the message here. All the message content is available so the developer is not constraint at all.

aea/contracts/base.py
marcofavorito: Ok

aea/contracts/base.py
marcofavorito: Ok

aea/contracts/base.py
marcofavorito: Ok

aea/contracts/base.py
marcofavorito: ... and `contract_address` for some performatives.

packages/hashes.csv
marcofavorito: I remind you we don't check hashes anymore.

packages/fetchai/connections/ledger/contract_dispatcher.py
```diff
@@ -130,6 +132,30 @@ def get_error_message(
         dialogue.update(response)
         return response
 
+    def _handle_request(
+        self,
+        api: LedgerApi,
+        message: ContractApiMessage,
+        dialogue: ContractApiDialogue,
+        response_builder: Callable[[bytes], ContractApiMessage],
+    ) -> ContractApiMessage:
```
DavidMinarsch, line 141: Missing docstring

packages/fetchai/connections/ledger/contract_dispatcher.py
```diff
@@ -144,22 +170,17 @@ def get_state(
         :param dialogue: the contract API dialogue
         :return: None
         """
-        contract = self.contract_registry.make(message.contract_id)
-        method_to_call = getattr(contract, message.callable)
-        try:
-            data = method_to_call(api, message.contract_address, **message.kwargs.body)
-            response = ContractApiMessage(
+
+        def build_response(data: bytes) -> ContractApiMessage:
```
DavidMinarsch, line 174: Missing docstring

packages/fetchai/connections/ledger/contract_dispatcher.py
```diff
@@ -237,19 +248,97 @@ def get_raw_message(
         :param dialogue: the contract API dialogue
         :return: None
         """
-        contract = self.contract_registry.make(message.contract_id)
-        method_to_call = getattr(contract, message.callable)
-        try:
-            rm = method_to_call(api, message.contract_address, **message.kwargs.body)
-            response = ContractApiMessage(
+
+        def build_response(rm: bytes) -> ContractApiMessage:
```
DavidMinarsch, line 252: missing docstring

packages/fetchai/connections/ledger/contract_dispatcher.py
```diff
@@ -206,22 +222,17 @@ def get_raw_transaction(
         :param dialogue: the contract API dialogue
         :return: None
         """
-        contract = self.contract_registry.make(message.contract_id)
-        method_to_call = getattr(contract, message.callable)
-        try:
-            tx = method_to_call(api, message.contract_address, **message.kwargs.body)
-            response = ContractApiMessage(
+
+        def build_response(tx: bytes) -> ContractApiMessage:
```
DavidMinarsch, line 226: missing docstring

packages/fetchai/connections/ledger/contract_dispatcher.py
```diff
@@ -175,22 +196,17 @@ def get_deploy_transaction(
         :param dialogue: the contract API dialogue
         :return: None
         """
-        contract = self.contract_registry.make(message.contract_id)
-        method_to_call = getattr(contract, message.callable)
-        try:
-            tx = method_to_call(api, **message.kwargs.body)
-            response = ContractApiMessage(
+
+        def build_response(tx: bytes) -> ContractApiMessage:
```
DavidMinarsch, line 200: missing docstring

packages/fetchai/connections/ledger/contract_dispatcher.py
```diff
@@ -237,19 +248,97 @@ def get_raw_message(
...
-        except Exception as e:  # pylint: disable=broad-except  # pragma: nocover
-            response = self.get_error_message(e, api, message, dialogue)
-        return response
+
+        return self._handle_request(api, message, dialogue, build_response)
+
+
+def _call_stub(
+    api: LedgerApi, message: ContractApiMessage, contract: Contract
+) -> Optional[bytes]:
```
DavidMinarsch, line 266: missing docstring

packages/fetchai/connections/ledger/contract_dispatcher.py
DavidMinarsch: we know what the first one/two positional arguments should be though...

packages/hashes.csv
DavidMinarsch: we do check hashes

DavidMinarsch: its concerning that you have locally differrent outcomes than me

packages/fetchai/connections/ledger/contract_dispatcher.py
marcofavorito: it's not a public method.

packages/fetchai/connections/ledger/contract_dispatcher.py
marcofavorito: it's not a public method.

packages/fetchai/connections/ledger/contract_dispatcher.py
marcofavorito: it's not a public method.

packages/fetchai/connections/ledger/contract_dispatcher.py
marcofavorito: it's not a public method.

packages/hashes.csv
marcofavorito: not for all packages:
https://github.com/fetchai/agents-aea/blob/develop/scripts/generate_ipfs_hashes.py#L445-L451

packages/fetchai/connections/ledger/contract_dispatcher.py
marcofavorito: I don't understand your comment; could you elaborate?

PR #1557, opened 2020-07-29
Added generate CosmWasm transactions support

MissingNO57:
## Proposed changes

Implementation of CosmWasm contract deployment, InitMsg, HandleMsg, QueryMsg, and CosmWasm compatible tx broadcast function 

After deploy contract, each deployed code gets it's own code_id
To initialize contract you need to know code_id
Each initialized contract has it's contract_address - to run HandleMsg or QueryMsg contract_address must be known

Deploy message example:
```
contract_filename = "../erc1155/contract.wasm"

# Deploy contract
tx = api.get_wasm_deploy_transaction(contract_owner_crypto.address, contract_filename, gas=900000)
signed_tx = contract_owner_crypto.sign_transaction(tx)
receipt = api.send_signed_cosmwasm_transaction(signed_tx)
print (receipt)
```

InitMsg example:
```
# Init deployed contract
tx = api.get_wasm_init_transaction(contract_owner_crypto.address, 50000, code_id, {}, 200000, "stake", "erc1155")
signed_tx = contract_owner_crypto.sign_transaction(tx)
receipt = api.send_signed_cosmwasm_transaction(signed_tx)
print (receipt)
```

HandleMsg example
```
# Create single
msg = {"create_single": {"item_owner": item_owner_address, "id": "1234", "path": "SOME_URI"}}
tx = api.get_wasm_handle_transaction(contract_owner_crypto.address, contract_address, msg)
signed_tx = contract_owner_crypto.sign_transaction(tx)
receipt = api.send_signed_cosmwasm_transaction(signed_tx)
print (receipt)
```

QueryMsg example
```
msg = {
    "balance":
        {
            "address": str(item_owner_address),
            "id": "1234"
        }
}
res = api.execute_wasm_query(contract_address,msg)
print(res)
```

## Types of chang

## Checklist


aea/crypto/cosmos.py
```diff
@@ -447,6 +621,25 @@ def _try_get_account_number_and_sequence(
             )
         return result
 
+    def send_signed_cosmwasm_transaction(
```
DavidMinarsch, line 624: This should be called from `send_signed_transaction`. In `send_signed_transaction` you need to identify whether to call this or continue with old approach based on signed message.

MissingNO57: it might be a bit tricky because right now I am running wasmcli in block mode which means that I get receipt/timeout in one function call. If I get it right the original approach is to non-blocking broadcast, wait, get_receipt rather than blocking broadcast which gives you a receipt when it's done.

PR #1560, opened 2020-07-30
[AEA-767] better runtime state support on agent

### LINKED ISSUE #1323: Get rid of agent's liveness object
liveness object represents a limited set of agent states: started/stopped but actually there is a bigger amount of states that can be represented by the runtime
agent.state should be overviews and updated corresponding to runtime states.

tests and documentation should be updated accordingly 

use runtime state as an agent state source: add several helpers methods like is_running, is_stopped is_connected for easy check of agent states.

probably stay liveness, but it will duplicate states and less flexible

solarw:
## Proposed changes

expose runtime.state to agent. runtime state small refactor. fixes. coverage fixes

## Fixes

#1323

## Types of changes

## Checklist

### DIFF
--- a/aea/helpers/async_utils.py
+++ b/aea/helpers/async_utils.py
@@ -74,7 +74,7 @@ class AsyncState:
         """
         self._state = initial_state
         self._watchers: Set[Future] = set()
-        self._callbacks: List[Callable] = []
+        self._callbacks: List[Callable[[Any], None]] = []
         self._states_enum = states_enum
 
     def set(self, state: Any) -> None:
@@ -90,7 +90,7 @@ class AsyncState:
         self._state_changed(state)
         self._state = state
 
-    def add_callback(self, callback_fn: Callable) -> None:
+    def add_callback(self, callback_fn: Callable[[Any], None]) -> None:
         """
         Add callback to track state changes.
 
--- a/aea/helpers/async_utils.py
+++ b/aea/helpers/async_utils.py
@@ -31,6 +31,7 @@ from typing import (
     Any,
     Awaitable,
     Callable,
+    Container,
     List,
     Optional,
     Sequence,
@@ -63,30 +64,41 @@ def ensure_list(value: Any) -> List:
 class AsyncState:
     """Awaitable state."""
 
-    def __init__(self, initial_state: Any = None):
+    def __init__(
+        self, initial_state: Any = None, states_enum: Optional[Container[Any]] = None
+    ):
         """Init async state.
 
         :param initial_state: state to set on start.
+        :param states_enum: container of valid states if not provided state not checked on set.
         """
         self._state = initial_state
         self._watchers: Set[Future] = set()
-
-    @property
-    def state(self) -> Any:
-        """Return current state."""
-        return self.get()
-
-    @state.setter
-    def state(self, state: Any) -> None:
-        """Set state."""
-        self.set(state)
+        self._callbacks: List[Callable] = []
+        self._states_enum = states_enum
 
     def set(self, state: Any) -> None:
         """Set state."""
+        if self._states_enum is not None and state not in self._states_enum:
+            raise ValueError(
+                f"Unsupported state: {state}. Valid states are {self._states_enum}"
+            )
+
         if self._state == state:  # pragma: no cover
             return
+
         self._state_changed(state)
         self._state = state
+
+    def add_callback(self, callback_fn: Callable) -> None:
+        """
+        Add callback to track state changes.
+
+        :param callback_fn: callable object to be called on state changed.
+
+        :return: None
+        """
+        self._callbacks.append(callback_fn)
 
     def get(self) -> Any:
         """Get state."""
@@ -94,6 +106,12 @@ class AsyncState:
 
     def _state_changed(self, state: Any) -> None:
         """Fulfill watchers for state."""
+        for callback_fn in self._callbacks:
+            try:
+                callback_fn(state)
+            except Exception:  # pylint: disable=broad-except
+                logger.exception(f"Exception on calling {callback_fn}")
+
         for watcher in list(self._watchers):
             if state not in watcher._states:  # type: ignore # pylint: disable=protected-access  # pragma: nocover
                 continue
--- a/aea/helpers/exec_timeout.py
+++ b/aea/helpers/exec_timeout.py
@@ -203,7 +203,9 @@ class ExecTimeoutThreadGuard(BaseExecTimeout):
 
             cls._loop = asyncio.new_event_loop()
             cls._stopped_future = Future(loop=cls._loop)
-            cls._supervisor_thread = threading.Thread(target=cls._supervisor_event_loop)
+            cls._supervisor_thread = threading.Thread(
+                target=cls._supervisor_event_loop, daemon=True
+            )
             cls._supervisor_thread.start()
 
     @classmethod
@@ -227,6 +229,7 @@ class ExecTimeoutThreadGuard(BaseExecTimeout):
                 if cls._supervisor_thread and cls._supervisor_thread.is_alive():
                     cls._supervisor_thread.join()
                 cls._supervisor_thread = None
+                cls._start_count = 0
 
     @classmethod
     def _supervisor_event_loop(cls) -> None:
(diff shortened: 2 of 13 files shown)


solarw:
https://github.com/fetchai/agents-aea/blob/feature/agent-statuses/aea/agent.py#L181

I think we can remove it but should check is it still useful and do we need extend runtime states to get full compatibility with agent_state property

aea/helpers/async_utils.py
```diff
@@ -63,37 +64,54 @@ def ensure_list(value: Any) -> List:
 class AsyncState:
     """Awaitable state."""
 
-    def __init__(self, initial_state: Any = None):
+    def __init__(
+        self, initial_state: Any = None, states_enum: Optional[Container[Any]] = None
```
DavidMinarsch, line 68: state should be an enum or class imo. not Any

aea/helpers/async_utils.py
```diff
@@ -63,37 +64,54 @@ def ensure_list(value: Any) -> List:
...
-        self.set(state)
+        self._callbacks: List[Callable] = []
+        self._states_enum = states_enum
 
     def set(self, state: Any) -> None:
         """Set state."""
+        if self._states_enum is not None and state not in self._states_enum:
+            raise ValueError(
+                f"Unsupported state: {state}. Valid states are {self._states_enum}"
+            )
```
DavidMinarsch, line 85: see above; state should not be any

aea/helpers/async_utils.py
```diff
@@ -63,37 +64,54 @@ def ensure_list(value: Any) -> List:
...
+
+        :return: None
+        """
+        self._callbacks.append(callback_fn)
+
     def get(self) -> Any:
         """Get state."""
         return self._state
 
     def _state_changed(self, state: Any) -> None:
```
DavidMinarsch, line 107: see above

aea/helpers/async_utils.py
```diff
@@ -63,37 +64,54 @@ def ensure_list(value: Any) -> List:
...
-    @property
-    def state(self) -> Any:
-        """Return current state."""
-        return self.get()
-
-    @state.setter
-    def state(self, state: Any) -> None:
-        """Set state."""
-        self.set(state)
+        self._callbacks: List[Callable] = []
```
DavidMinarsch, line 77: Callables can be further typed as they are expected to consume single arg `state`

aea/helpers/exec_timeout.py
```diff
@@ -227,6 +229,7 @@ def stop(cls, force: bool = False) -> None:
                 if cls._supervisor_thread and cls._supervisor_thread.is_alive():
                     cls._supervisor_thread.join()
                 cls._supervisor_thread = None
+                cls._start_count = 0
```
DavidMinarsch, line 232: what's that doing?

aea/runtime.py
```diff
@@ -100,12 +115,12 @@ def _stop(self) -> None:  # pragma: nocover
...
     def is_running(self) -> bool:
         """Get running state of the runtime."""
-        return self._state.get() == RuntimeStates.started
+        return self._state.get() == RuntimeStates.running
 
     @property
     def is_stopped(self) -> bool:
         """Get stopped state of the runtime."""
-        return self._state.get() in [RuntimeStates.stopped, RuntimeStates.initial]
+        return self._state.get() in [RuntimeStates.stopped]
```
DavidMinarsch, line 123: should we also have properties `is_starting` and `is_stopping`?

aea/runtime.py
```diff
@@ -156,28 +180,24 @@ def _start(self) -> None:
...
             self._stopping_task = None
 
     async def run_runtime(self) -> None:
         """Run agent and starts multiplexer."""
+        self._state.set(RuntimeStates.starting)
         try:
-            self._state.set(RuntimeStates.starting)
-            self._agent.setup_multiplexer()
             await self._start_multiplexer()
-            self._agent.start_setup()
```
DavidMinarsch, line 180: @solarw and @marcofavorito ; I would have thought we should have `starting` where we call setup on multiplexer and agent, then when finished setting up, call `start` on both of them.

tests/test_agent.py
```diff
@@ -100,25 +102,20 @@ def test_run_agent():
                 error_msg="Agent loop not started!'",
             )
             wait_for_condition(
-                lambda: agent.agent_state == AgentState.RUNNING,
+                lambda: agent.state == RuntimeStates.running,
                 timeout=5,
                 error_msg="Agent state must be 'running'",
             )
+            wait_for_condition(
+                lambda: agent.agent_state == AgentState.RUNNING,
```
DavidMinarsch, line 110: why do we have both `agent_state` and `state`?

tests/test_aea.py
```diff
@@ -112,6 +115,27 @@ def test_start_stop():
...
+    agent = builder.build()
+
+    with run_in_thread(agent.start, timeout=20):
+        wait_for_condition(
+            lambda: agent._main_loop and agent._main_loop.is_running, timeout=10
+        )
+        t = Thread(target=agent.start)
+        t.start()
+        sleep(1)
+        assert not t.is_alive()
```
DavidMinarsch, line 135: should it not be `assert is_alive()`?

aea/helpers/async_utils.py
marcofavorito: I agree. In other modules, I used to annotate arguments like this as:
```
arg: Union[SomeEnum, str]
```
and then, in the body of the function:
```
SomeEnum(arg)
```
which works both for `SomeEnum` and `str`, but in the latter case, it validates whether the string is a legal value.

aea/helpers/async_utils.py
marcofavorito: yes, and also by return value, if it is relevant in this context.

aea/helpers/async_utils.py
```diff
@@ -63,37 +64,54 @@ def ensure_list(value: Any) -> List:
...
+                f"Unsupported state: {state}. Valid states are {self._states_enum}"
+            )
+
         if self._state == state:  # pragma: no cover
             return
+
         self._state_changed(state)
         self._state = state
 
+    def add_callback(self, callback_fn: Callable) -> None:
```
marcofavorito, line 93: The same applies here re `Callable` annotation

aea/helpers/async_utils.py
solarw: it's a general-purpose class, the state can be any value, enum is optional and it can be not an enum but any container with any values inside.
this is why Any annotation is used. i see no reason to limit it by types, it does not know anything about what kind of states are set.

aea/helpers/exec_timeout.py
solarw: it's a counter of number it was started, kind of usage references, and here is set to 0 that means it's not in use. this class can be shared across several agents but should have only one thread started

aea/runtime.py
```diff
@@ -192,12 +212,14 @@ async def _multiplexer_disconnect(self) -> None:
...
         """Call multiplexer connect asynchronous way."""
+        self._agent.setup_multiplexer()
         await AsyncMultiplexer.connect(self._agent.multiplexer)
 
     async def _start_agent_loop(self) -> None:
         """Start agent main loop asynchronous way."""
         logger.debug("[{}]: Runtime started".format(self._agent.name))
-        self._state.set(RuntimeStates.started)
+        self._agent.start_setup()
+        self._state.set(RuntimeStates.running)
```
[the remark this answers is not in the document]
solarw: actually it's a loop and it does not return control back until finished.
so i can not set state in run_runtime

aea/helpers/async_utils.py
solarw: done

tests/test_aea.py
solarw: it's correct, agent should not start the second time and thread should exit instead of keep running

aea/runtime.py
solarw: we can add it if required.

aea/helpers/async_utils.py
solarw: done

tests/test_agent.py
solarw: it was my first comment in this pr about what to do with this agent_state.

now it removed.

aea/runtime.py
marcofavorito: isn't this the case already? Line 198 sets the state to `starting`

Issue #922, opened 2020-03-12
Improve `ConnectionStatus` to include many other cases


marcofavorito:
**Is your feature request related to a problem? Please describe.**

Currently, a connection can be in status connected or not connected.
However, there are other cases, e.g.:
1. not connected
2. trying to connect
3. attempt to connect failed
4. attempt to connect timed out
5. connection established
6. connection lost

**Describe the solution you'd like**

Trying to find a good trade-off between richness of the enumeration and easiness of usage.

**Describe alternatives you've considered**

**Additional context**

DavidMinarsch:
After discussion we settled on: “disconnected”, “connecting”, “connected”, “disconnecting” states

PR #1573, opened 2020-08-03
[AEA-766] Add ConnectionState to connections instead of ConnectionStatus

solarw:
## Proposed changes

use connection state object to set state: connected, disconnected, connecting, disconnecting

## Fixes

https://github.com/fetchai/agents-aea/issues/922

## Types of changes

## Checklist


packages/fetchai/connections/http_server/connection.py
```diff
@@ -525,20 +525,25 @@ async def connect(self) -> None:
...
-            self.channel.logger = self.logger
-            await self.channel.connect(loop=self.loop)
-            self.connection_status.is_connected = not self.channel.is_stopped
+        if self.is_connected:
+            return
+        self._state.set(ConnectionStates.connecting)
+        self.channel.logger = self.logger
+        await self.channel.connect(loop=self.loop)
+        if not self.channel.is_stopped:
+            self._state.set(ConnectionStates.connected)
```
marcofavorito, line 534: if `self.channel.is_stopped`, the connection will remain in `connecting` state. Is it correct? Maybe we should set it to disconnected, or raise `ConnectionError`.

marcofavorito: It's also a matter of making it clearer at interface level what's the expected behaviour in such cases. Worth an issue imo.

packages/fetchai/connections/p2p_client/connection.py
```diff
@@ -174,22 +174,27 @@ async def connect(self) -> None:
...
-            self.channel.logger = self.logger
-            self.connection_status.is_connected = True
-            self.channel.in_queue = asyncio.Queue()
-            self.channel.loop = self.loop
-            self.channel.connect()
+        if self.is_connected:
+            return
+        self._state.set(ConnectionStates.connecting)
+        self.channel.logger = self.logger
+        self.channel.logger = self.logger
```
marcofavorito, line 181: remove line duplicate

packages/fetchai/connections/http_server/connection.py
solarw: nice point, thanks. it was previous implementation can looks this was incorrect solution

updated. now it will be disconnected

packages/fetchai/connections/p2p_client/connection.py
solarw: done

aea/connections/base.py
```diff
@@ -45,6 +46,15 @@
 logger = logging.getLogger(__name__)
 
 
+class ConnectionStates(Enum):
```
DavidMinarsch, line 49: Ok, why do we have ConnectionStates and ConnectionStatus now? Seems like ConnectionStatus needs to be removed..

packages/fetchai/connections/ledger/ledger_dispatcher.py
```diff
@@ -194,7 +194,7 @@ def get_transaction_receipt(
         while (
             not is_settled
             and attempts < self.MAX_ATTEMPTS
-            and self.connection_status.is_connected
+            and self.connection_state.get() == ConnectionStates.connected
```
DavidMinarsch, line 197: I think you have a property for this now; is_connected

packages/fetchai/connections/ledger/ledger_dispatcher.py
```diff
@@ -207,7 +207,7 @@ def get_transaction_receipt(
         while (
             transaction is None
             and attempts < self.MAX_ATTEMPTS
-            and self.connection_status.is_connected
+            and self.connection_state.get() == ConnectionStates.connected
```
DavidMinarsch, line 210: see above

solarw: I pass state not connection, there is no is connected property

packages/fetchai/connections/ledger/ledger_dispatcher.py
solarw: I pass state not connection, there is no is connected property

aea/connections/base.py
solarw: it used by  multiplexer

solarw: it used by  multiplexer

PR #1581, opened 2020-08-03
update 'generate_ipfs_hashes' script

marcofavorito:
## Proposed changes

The update was needed because of the recent upgrade of the package `ipfshttpclient`. With the new version, the flag that turns on the recursive computation of hashes of a directory was set to `False` by default, which wasn't the desired outcome.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments
n/a


tests/data/hashes.csv
```diff
@@ -1,4 +1,4 @@
-dummy_author/agents/dummy_aea,QmR4SuPKbe4qjgF5DVwVwHPm5eFqCx83sdavs4QEXgF1mR
+dummy_author/agents/dummy_aea,QmVjFRYLEwTF9zK2g4sDTvb3Djo2We1QFiNUHT6EJedgsx
```
marcofavorito, line 1: this is likely due to symlinks. `follow_symlinks=True` didn't solve the difference.

tox.ini
```diff
@@ -74,7 +74,7 @@ commands = {toxinidir}/scripts/check_copyright_notice.py --directory {toxinidir}
 [testenv:hash_check]
 skipsdist = True
 usedevelop = True
-deps = ipfshttpclient
+deps = ipfshttpclient==0.6.0.post1
```
marcofavorito, line 77: better to stick with one version to avoid surprises.

DavidMinarsch: Needs to be properly bumped in pipenv too

tests/data/hashes.csv
DavidMinarsch: Concerning. We need to understand why before we merge this.

DavidMinarsch:
Closed in favour of #1588

PR #1597, opened 2020-08-05
Release v0.5.3

DavidMinarsch:
## Release summary

Version number: v0.5.3

## Release details

This release:
- Adds support for re-starting agent after stopping it
- Adds full test coverage for protocols generator
- Adds support for dynamically adding handlers
- Improves p2p connection startup reliability
- Addresses p2p connection race condition with long running processes
- Adds connection states in connections
- Applies consistent logger usage throughout
- Adds key rotation and randomised locations for integration tests
- Adds request delays in SOEF connection to avoid request limits
- Exposes runtime states on agent and removes agent liveness object
- Adds readme files in protocols and connections
- Improves edge case handling in dialogue models
- Adds support for cosmwasm message signing
- Adds test coverage for test tools
- Adds dialogues models in all connections where required
- Transitions erc1155 skills and simple search to SOEF and p2p
- Adds full test coverage for skills modules
- Multiple docs updates
- Multiple additional tests and test stability fixes

## Checklist

- [x] I am making a pull request against the `master` branch (left side), from `develop`
- [x] Lint and unit tests pass locally
- [x] I have checked the fingerprint hashes are correct by running (`scripts/generate_ipfs_hashes.py`)
- [x] I have regenerated the latest API docs
- [ ] I built the documentation and updated it with the latest changes
- [x] I have added an item in `HISTORY.md` for this release
- [x] I bumped the version number in the `aea/__version__.py` file.
- [x] I bumped the version number in the `docs/version.md` file
- [x] I bumped the version number in every Docker image of the repo and published it. Also, I built and published them with tag `latest`  
      (check the READMEs of [`aea-develop`](../develop-image/README.md#publish) 
      and [`aea-deploy`](../deploy-image/README.md#publish))
- [ ] I have pushed the latest packages to the registry.

## Further comments

-


.github/workflows/workflow.yml
```diff
@@ -82,6 +86,8 @@ jobs:
           tar xvfz go-ipfs.tar.gz
           sudo mv go-ipfs/ipfs /usr/local/bin/ipfs
           ipfs init
+      - name: Pipenv lock
+        run: pipenv lock
```
solarw, line 90: think pipenv lock should be done locally and lock file added to repo. it's quite slow operation and should be done only on pipenv file changed. also, lock file guarantees that test environment will be the same. matching all packages versions

DavidMinarsch: Since these are development dependencies only we removed the Lockfile some time ago. It did only create problems and slowed down development. This step in CI is ok as most other jobs take longer. We should automate all critical tasks rather than do them locally.

PR #1620, opened 2020-08-07
[AEA-644] [AEA-666] libp2p public dht add comprehensive tests

lrahmani:
## Proposed changes

Add connectivity and communication tests for public dht nodes services.
This includes, for both relay and delegate services:
- connectivity tests: connect a local connection to every public node
- communication tests: envelope delivery between two local connection using the same public node and using two different public nodes
- aea cli connectivity tests


tests/test_packages/test_connections/test_p2p_libp2p/test_public_dht.py
```diff
@@ -60,19 +64,121 @@ def setup_class(cls):
...
+        for maddr in PUBLIC_DHT_MADDRS:
             connection = _make_libp2p_connection(
                 DEFAULT_PORT + 1, relay=False, entry_peers=[maddr]
             )
             multiplexer = Multiplexer([connection])
             self.log_files.append(connection.node.log_file)
             multiplexer.connect()
 
-            assert connection.is_connected is True
+            assert (
```
DavidMinarsch, line 76: If your tests go wrong (assertions don't pass) the multiplexer will not disconnect. Either have individual test classes and do disconnect in teardown so it always happens or wrap in try finally blocks.

lrahmani: Indeed

PR #1629, opened 2020-08-10
AEA-807 Set ipfshttpclient version to 0.4.12 in setup.py

marcofavorito:
## Proposed changes

Set `ipfshttpclient` version to 0.4.12 in setup.py

## Fixes

Fixes #1629 

## Types of changes

## Checklist

## Further comments
n/a

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,7 +30,7 @@ def get_all_extras() -> Dict:
 
     fetch_ledger_deps = ["fetchai-ledger-api==1.1.0"]
 
-    ethereum_ledger_deps = ["web3==5.2.2", "eth-account==0.4.0"]
+    ethereum_ledger_deps = ["web3==5.2.2", "ipfshttpclient==0.4.12", "eth-account==0.4.0"]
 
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -30,7 +30,7 @@ def get_all_extras() -> Dict:
 
     fetch_ledger_deps = ["fetchai-ledger-api==1.1.0"]
 
-    ethereum_ledger_deps = ["web3==5.2.2", "eth-account==0.4.0"]
+    ethereum_ledger_deps = ["web3==5.2.2", "ipfshttpclient==0.4.12", "eth-account==0.4.0"]
```
DavidMinarsch, line 33: It's not an ethereum ledger dependency; it's a general dependency...

marcofavorito: It's a sub-dependency of `web3==5.2.2`. In fact, `ipfshttpclient` was not used from the framework; only in `scripts/generate_ipfs_hashes.py`.

DavidMinarsch: Ah ok; I see now; makes sense!

Issue #1643, opened 2020-08-11
Dialogue reference as bytes instead of strings in *.proto files


marcofavorito:
**Is your feature request related to a problem? Please describe.**
Currently, `dialogue_starter_reference` and `dialogue_responder_reference` in any protocol are of Protobuf type `string`. 

**Describe the solution you'd like**
Use bytes as it is a more compact type.

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

PR #1644, opened 2020-08-11
AEA-817 Make the nonce for dialogue reference randomly generated

marcofavorito:
## Proposed changes

Make the nonce for dialogue reference randomly generated.

We use `secrets`, a Python Standard Library module designed to be cryptographically secure, since it relies on source of randomness provided by the operating system.

https://docs.python.org/3/library/secrets.html

I think that it's enough for modern operating systems, and for our purposes (relatively low scale atm).

Useful resources:
- https://www.2uo.de/myths-about-urandom/
- https://security.stackexchange.com/a/3939/146388
- https://man7.org/linux/man-pages/man2/getrandom.2.html

## Fixes

n/a

## Types of changes

## Checklist

### DIFF
--- a/aea/helpers/dialogue/base.py
+++ b/aea/helpers/dialogue/base.py
@@ -31,6 +31,7 @@ from abc import ABC, abstractmethod
 from enum import Enum
 from typing import Callable, Dict, FrozenSet, List, Optional, Set, Tuple, Type, cast
 
+from aea.helpers.base import is_hexadecimal
 from aea.mail.base import Address
 from aea.protocols.base import Message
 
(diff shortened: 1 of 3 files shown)


aea/helpers/dialogue/base.py
```diff
@@ -1116,11 +1118,11 @@ def role_from_first_message(message: Message) -> Dialogue.Role:
         """
         pass  # pragma: no cover
 
-    def _next_dialogue_nonce(self) -> int:
+    @staticmethod
+    def _generate_dialogue_nonce() -> str:
```
marcofavorito, line 1122: renamed because this is a more proper name.

aea/helpers/dialogue/base.py
```diff
@@ -1037,7 +1039,7 @@ def _create_opponent_initiated(
         )
         new_dialogue_reference = (
             dialogue_reference[0],
-            str(self._next_dialogue_nonce()),
+            str(self._generate_dialogue_nonce()),
```
DavidMinarsch, line 1042: remove str

aea/helpers/dialogue/base.py
```diff
@@ -34,6 +35,9 @@
 from aea.protocols.base import Message
 
 
+NONCE_BYTES_NB = 32
```
marcofavorito, line 38: Why 32?

- Here it is considered sufficient to use 256 bits -> 32 bytes: https://docs.python.org/3/library/secrets.html#how-many-bytes-should-tokens-use
- too many bytes would make it too expensive in terms of network resources (although unless #1643 is fixed, the message field for the nonce will actually take at least twice of the bytes, because the nonce is in hexadecimal format.)
- According to the birthday bound, to have a probability of collision of 50% with 256 bits it is required to have ~ `2^(n/2)`, i.e. `2^128`, which is a massive number. 
Put in other terms: the following is the needed number of messages to have a collision probability of `1e-9` with `256` bits:

```python
def n_approx(H, p):
    return math.sqrt(2 * H * math.log(1 / (1 - p)))

n_approx(2**256, 1e-9)
# 1.5217890705357334e+34
```

marcofavorito: Another good argument for `32` bytes and not more is the following, stated here: https://man7.org/linux/man-pages/man2/getrandom.2.html

> If the urandom source has been initialized, reads of up to 256 bytes will always return as many bytes as requested and will not be interrupted by signals.  No such guarantees apply for larger buffer sizes.  For example, if the call is interrupted by a signal handler, it may return a partially filled buffer, or fail with the error EINTR.

aea/helpers/dialogue/base.py
```diff
@@ -52,10 +57,59 @@ def __init__(
 
         :return: None
         """
+        # TODO make it an assert
+        if not self.is_dialogue_reference(dialogue_reference):
+            pass
```
marcofavorito, line 62: TODO before the merge!

DavidMinarsch:
Closed in favour of the less invasive #1660. All credit goes to this PR.

Issue #1261, opened 2020-05-22
Dialogue To-Dos


5A11:
### To-Dos:

**Current issues/ugly designs**

- [x] The `create_dialogue` method in Dialogues is ugly, and in general, the 3 layered subclass dependencies (base, instance such as Fipa, derived dialogue) is not the cleanest. Make dialogue creation explicit. [v0.6!]  (Stage 3)
- [x] The current implementation means: the role of an agent in a dialogue is always determined via the first move (whether the move is by the agent or its counterparty) via the `role_from_first_message()` method (NOTES) [v0.5] > create(cfp_msg, role, counterparty) replaces update()
- [x] Two places can update the dialogue reference on receiving second message, `_update_self_initiated_dialogue_label_on_second_message()` in `Dialogues` and `Dialogue`. [v0.5.1]
- [x] After studying how dialogues are used in the skills, I propose adding the following 3 methods that would further simplify their usage. The idea is, one should not worry about `message_id`, `dialogue_reference`, `counterparty` and sometimes `target`, if one is creating messages in the context of a dialogue. Those fields can be automatically set, which should result in reduced accidental errors in setting those fields. (USAGE)
In `Dialogue`:
```
reply(self, message_target, performative, **kwargs) -> Dialogue
```
- [x] Apply usage to all skills:  (Stage 2)
*************************
dialogues.update(incoming_message) —> Optional[Dialogue]
dialogues.create(counterparty, performative, **kwargs) —> Tuple[Message, Dialogue]
dialogue.reply(target_message, performative, **kwargs) —> Message
*************************
- [x] Modify update to not have create & reply. (Stage 1)
- [x] handle incoming (update > handle & log); raise outgoing (reply > raise, create > raise) (Stage 1)
- [x] assert the replied target message exists. (Stage 1)
- [x] apply couterparty/is_incoming deprecation throughout all code (Stage 1)

**Docs**
- [ ] Docs update

**Test**
- [x] Tests for the dialogue

**Refactoring**
- [x] Consolidate `_incoming` and `_outgoing` into one.  (Stage 1)
- [x] `sender` and `receiver` fields in messages is cleaner than `is_incoming`, `counterparty`. Both can be automatically assigned so developer does not have to worry about it.

### Done

[...]

Issue #1542, opened 2020-07-25
Update protocol generator and skills to incorporate new dialogues api


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**
..

PR #1645, opened 2020-08-11
AEA-724, AEA-855 Refactoring Dialogue Module

5A11:
This PR implements the following:
 - Removes deprecated methods from dialogue module
 - Replaces `counterparty` and `is_incoming` from `Message` with `to` and `sender`
 - Ensures dialogue tests pass with the above changes applied

Fixes:
#1261
#1542

## Types of changes

## Checklist


aea/protocols/base.py
```diff
@@ -222,7 +187,7 @@ def __eq__(self, other):
         return (
             isinstance(other, Message)
             and self.body == other.body
-            and self._counterparty == other._counterparty
+            and self._sender == other._sender
```
DavidMinarsch, line 190: Missing comparison on other properties

aea/helpers/dialogue/base.py
```diff
@@ -484,16 +462,42 @@ def is_belonging_to_dialogue(self, message: Message) -> bool:
...
+                        message.dialogue_reference[0],
+                        Dialogue.OPPONENT_STARTER_REFERENCE,
+                    ),
+                    message.sender,
+                    self.agent_address,
+                )
+            else:
+                raise SyntaxError(
+                    "neither 'to' nor 'sender' match the agent's address in this dialogue."
+                )
```
DavidMinarsch, line 486: refactor

aea/helpers/dialogue/base.py
```diff
@@ -484,16 +462,42 @@ def is_belonging_to_dialogue(self, message: Message) -> bool:
...
+                    message.dialogue_reference, message.to, message.to
+                )
+            elif message.to == self.agent_address:
+                other_initiated_dialogue_label = DialogueLabel(
+                    message.dialogue_reference, message.sender, message.sender
+                )
+            else:
+                raise SyntaxError(
+                    "neither 'to' nor 'sender' match the agent's address in this dialogue."
+                )
```
DavidMinarsch, line 500: refactor

aea/helpers/dialogue/base.py
```diff
@@ -736,20 +740,13 @@ def __init__(
...
+        assert issubclass(dialogue_class, Dialogue)
         self._dialogue_class = dialogue_class
 
-        if role_from_first_message is not None:
-            self._role_from_first_message = role_from_first_message
-        else:
-            self._role_from_first_message = (
-                self.role_from_first_message
-            )  # pragma: no cover
+        self._role_from_first_message = role_from_first_message
```
DavidMinarsch, line 749: add check on role_from_first_message

aea/helpers/dialogue/base.py
```diff
@@ -834,12 +831,12 @@ def update(self, message: Message) -> Optional[Dialogue]:
...
         dialogue_reference = message.dialogue_reference
 
-        if not message.has_counterparty:
+        if not message.has_to:
+            raise ValueError("The message's 'to' field is not set {}".format(message))
+        if not message.has_sender:
             raise ValueError(
-                "The message counterparty field is not set {}".format(message)
+                "The message's 'sender' field is not set {}".format(message)
             )
```
DavidMinarsch, line 839: only raise when the message is being received

aea/helpers/dialogue/base.py
```diff
@@ -938,15 +939,24 @@ def get_dialogue(self, message: Message) -> Optional[Dialogue]:
...
+        elif message.to == self.agent_address:
+            self_initiated_dialogue_label = DialogueLabel(
+                message.dialogue_reference, message.sender, self.agent_address
+            )
+            other_initiated_dialogue_label = DialogueLabel(
+                message.dialogue_reference, message.sender, message.sender
+            )
+        else:
+            raise SyntaxError(
+                "neither 'to' nor 'sender' match the agent's address in this dialogue."
```
DavidMinarsch, line 958: simplify

aea/helpers/dialogue/base.py
```diff
@@ -1076,46 +1086,15 @@ def _create(
...
-
-    @staticmethod
-    def role_from_first_message(message: Message) -> Dialogue.Role:
-        """
-        Infer the role of the agent from an incoming or outgoing first message.
-
-        :param message: an incoming/outgoing first message
-        :return: the agent's role
-        """
-        pass  # pragma: no cover
```
DavidMinarsch, line 1117: ensure dynamically provided method has this signature

aea/protocols/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: I did implement some checks using `inspect.signature` but that might be a bit of an overkill IMO, wouldn't the type hint (`Callable...`) be sufficient?

aea/helpers/dialogue/base.py
5A11: See above.

aea/helpers/dialogue/base.py
```diff
@@ -738,20 +757,17 @@ def __init__(
...
-        if role_from_first_message is not None:
-            self._role_from_first_message = role_from_first_message
-        else:
-            self._role_from_first_message = (
-                self.role_from_first_message
-            )  # pragma: no cover
+        sig = signature(role_from_first_message)
+        assert len(sig.parameters.keys()) == 1
+        assert list(sig.parameters.values())[0].annotation == Message
+        assert sig.return_annotation == Dialogue.Role
```
DavidMinarsch, line 769: Please provide helpful error messages for failure mode; same above in in lines 760, 763

aea/helpers/dialogue/base.py
DavidMinarsch: No; because type hints are not used at runtime.

aea/helpers/dialogue/base.py
```diff
@@ -773,13 +789,19 @@ def dialogue_stats(self) -> DialogueStats:
         """
         return self._dialogue_stats
 
+    def is_message_by_self(self, message: Message) -> bool:
+        return message.sender == self.agent_address
+
+    def is_message_by_other(self, message: Message) -> bool:
+        return not self.is_message_by_self(message)
```
DavidMinarsch, line 796: missing docstrings

aea/helpers/dialogue/base.py
```diff
@@ -830,106 +849,88 @@ def update(self, message: Message) -> Optional[Dialogue]:
...
-        :param message: a new message
+        :param message: a new incoming message
         :return: the new or existing dialogue the message is intended for, or None in case of any errors.
         """
-        dialogue_reference = message.dialogue_reference
+        assert self.is_message_by_other(
+            message
+        ), "Invalid 'update' usage. Update must only be used with a message by another agent."
+        assert message.has_to, "The message's 'to' field is not set {}".format(message)
+        # the message's sender field is ensured to have been set via the is_message_by_other call above
```
DavidMinarsch, line 859: You can remove this check; it will automatically raise if the to field is accessed and not present; this check adds nothing here

aea/helpers/dialogue/base.py
```diff
@@ -830,106 +849,88 @@ def update(self, message: Message) -> Optional[Dialogue]:
...
-            # we can allow a dialogue to have incomplete reference
-            # as multiple messages can be sent before one is received with complete reference
-            dialogue = self.get_dialogue(message)
-        else:
-            self._update_self_initiated_dialogue_label_on_message_with_complete_reference(
-                message
-            )
+        else:  # non-initial message for existing dialogue
+            self._complete_dialogue_reference(message)
             dialogue = self.get_dialogue(message)
```
DavidMinarsch, line 883: keep incomplete label logic!

aea/helpers/dialogue/base.py
```diff
@@ -830,106 +849,88 @@ def update(self, message: Message) -> Optional[Dialogue]:
...
-    ) -> None:
+    def _complete_dialogue_reference(self, message: Message) -> None:
         """
         Update a self initiated dialogue label with a complete dialogue reference from counterparty's first message.
 
         :param message: A message in the dialogue (the first by the counterparty with a complete reference)
         :return: None
         """
-        dialogue_reference = message.dialogue_reference
+        complete_dialogue_reference = message.dialogue_reference
```
DavidMinarsch, line 905: currently wrong var name; the `complete_dialogue_reference` can currently be complete or incomplete; once the change above (line 868 onwards) is reverted it will be by default complete

aea/helpers/dialogue/base.py
```diff
@@ -939,14 +940,15 @@ def get_dialogue(self, message: Message) -> Optional[Dialogue]:
...
+            opponent = message.to
+        else:
+            opponent = message.sender
         self_initiated_dialogue_label = DialogueLabel(
-            dialogue_reference, counterparty, self.agent_address
+            message.dialogue_reference, opponent, self.agent_address
         )
         other_initiated_dialogue_label = DialogueLabel(
-            dialogue_reference, counterparty, counterparty
+            message.dialogue_reference, opponent, message.to
```
DavidMinarsch, line 951:
```suggestion
            message.dialogue_reference, opponent, opponent
```

aea/helpers/dialogue/base.py
```diff
@@ -412,21 +362,25 @@ def last_message(self) -> Optional[Message]:
...
+        assert message_id <= len(
+            list_of_all_messages
+        ), "This dialogue does not have a message with id {}. Last message id is {}.".format(
+            message_id, len(list_of_all_messages)
+        )
 
-        return result
+        for message in list_of_all_messages:
+            if message.message_id == message_id:
+                return message
```
DavidMinarsch, line 383: do you really want this to throw if the message_id does not make sense; why? isn't it better to return None?

aea/helpers/dialogue/base.py
```diff
@@ -437,65 +391,67 @@ def is_empty(self) -> bool:
...
-            return False
-
-        is_extendable = self.is_valid_next_message(message)
-        if is_extendable:
-            if message.is_incoming:
-                self._incoming_messages.extend([message])
-            else:
-                self._outgoing_messages.extend([message])
-        return is_extendable
+            raise InvalidDialogueMessage("message does not belong to this dialogue.")
```
DavidMinarsch, line 415: Provide some more info like message id and dialogue label so we know what context the error occurs in

aea/helpers/dialogue/base.py
```diff
@@ -437,65 +391,67 @@ def is_empty(self) -> bool:
...
-        :return: None
-        """
-        counterparty = None  # type: Optional[str]
-        try:
-            counterparty = message.counterparty
-        except AssertionError:
-            # assume message belongs to dialogue
-            message.counterparty = self.dialogue_label.dialogue_opponent_addr
+        if not is_valid_result:
+            raise InvalidDialogueMessage(is_valid_message)
```
DavidMinarsch, line 420: Better error message needed

aea/helpers/dialogue/base.py
```diff
@@ -543,13 +496,24 @@ def is_valid_next_message(self, message: Message) -> bool:
...
+                msg_additional_validation,
+            ) = self._additional_validation(message)
+            if not result_additional_validation:
+                return False, msg_additional_validation
+            else:
+                result_is_valid, msg_is_valid = self.is_valid(message)
+                if not result_is_valid:
+                    return False, msg_is_valid
+                else:
+                    return True, "Message is valid with respect to this dialogue."
```
DavidMinarsch, line 514: Compare left and right; what is more readable? This can be done better!

aea/helpers/dialogue/base.py
```diff
@@ -602,13 +616,17 @@ def _additional_rules(self, message: Message) -> bool:
...
-        return result
+            if message.target != last_target + 1:
+                return (
+                    False,
+                    "Invalid target. Expected {}. Found {}.".format(
+                        last_target + 1, message.target
+                    ),
+                )
+
+        return True, "The message passes additional validation."
```
DavidMinarsch, line 629: This is the opposite of a guard clause; why not do
```
if self.last_message is None:
     return True, "The message passes additional validation."

...
```

DavidMinarsch: This is much easier to read and avoids one level of nesting!

aea/helpers/dialogue/base.py
```diff
@@ -543,13 +496,24 @@ def is_valid_next_message(self, message: Message) -> bool:
...
+                return False, msg_additional_validation
+            else:
+                result_is_valid, msg_is_valid = self.is_valid(message)
+                if not result_is_valid:
+                    return False, msg_is_valid
+                else:
+                    return True, "Message is valid with respect to this dialogue."
 
-    def _basic_rules(self, message: Message) -> bool:
+    def _basic_validation(self, message: Message) -> Tuple[bool, str]:
```
DavidMinarsch, line 516: Compare left and right; the code has blown up and become very hard to read. Firstly why? secondly, it needs to be done better.

aea/helpers/dialogue/base.py
DavidMinarsch: One thing' you don't need the else as the method will already have returned otherwise

aea/helpers/dialogue/base.py
```diff
@@ -437,65 +391,67 @@ def is_empty(self) -> bool:
...
+                ),
+                opponent,
                 self.agent_address,
             )
             result = self_initiated_dialogue_label in self.dialogue_labels
         else:
+            if self.is_message_by_self(message):
+                opponent = message.to
+            else:
+                opponent = message.sender
```
DavidMinarsch, line 452: one liner possible...also since you have a need to get the opponent in multiple places; why not expose it as a property on the message?

aea/protocols/base.py
```diff
@@ -105,49 +103,15 @@ def has_to(self) -> bool:
...
-
-    @property
-    def counterparty(self) -> Address:
-        """
-        Get the counterparty of the message in Address form.
-
-        :return the address
-        """
-        assert self._counterparty is not None, "Counterparty must not be None."
-        return self._counterparty
```
DavidMinarsch, line 130: It's worth keeping this or relabelling as `opponent` and keeping as you keep using it above

aea/helpers/dialogue/base.py
```diff
@@ -167,72 +172,7 @@ class Dialogue(ABC):
 
     STARTING_MESSAGE_ID = 1
     STARTING_TARGET = 0
-    OPPONENT_STARTER_REFERENCE = ""
-
-    class Rules:
```
DavidMinarsch, line 172: Rules class to be reinstated

aea/helpers/dialogue/base.py
5A11: You are supposed to use `get_message` with a correct `message_id`. If it sometimes returns None and sometimes Message, one must always check that it does not return None in every correct use.

I added two other methods `has_message_id` and `has_method` to check whether a message exists in the dialogue.

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: Not a fair comparison as previously you didn't know what's wrong with the message in case of an error, now there is detailed information as to which part failed.

Made some improvements (removed elif, elif, ...)

aea/helpers/dialogue/base.py
5A11: There would then be two `else` clauses returning `True` + string info, one the opposite of `if self.last_message is not None:` the other the opposite of `if message.target != last_target + 1`

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: Discussed. It is not reached here given update is now only for incoming messages. 

The scenario where multiple messages are sent before one is received now happens with `Dialogue.reply`, for which there is an associated test now.

aea/helpers/dialogue/base.py
5A11: See above

aea/helpers/dialogue/base.py
```diff
@@ -939,14 +940,15 @@ def get_dialogue(self, message: Message) -> Optional[Dialogue]:
         :param message: a message
         :return: the dialogue, or None in case such a dialogue does not exist
         """
-        dialogue_reference = message.dialogue_reference
-        counterparty = message.counterparty
-
+        if self.is_message_by_self(message):
+            opponent = message.to
+        else:
+            opponent = message.sender
```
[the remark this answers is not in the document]
5A11: Done

aea/helpers/dialogue/base.py
5A11: Done as above

aea/protocols/base.py
5A11: See above. Cannot do that in message. Message does not have a notion of `this` agent, dialogue does. 

Added the method in the dialogue for reuse.

aea/helpers/dialogue/base.py
```diff
@@ -180,11 +189,9 @@ def __init__(
         ) -> None:
             """
             Initialize a dialogue.
-
```
DavidMinarsch, line 183: We use the pattern to separate first line and params throughout. It is not good practice to change formatting of code you are not touching. Just leave it as it is and focus on your own contributions.

aea/helpers/dialogue/base.py
```diff
@@ -412,90 +395,145 @@ def last_message(self) -> Optional[Message]:
...
+    @property
+    def is_empty(self) -> bool:
+        """
+        Check whether the dialogue is empty.
+
+        :return: True if empty, False otherwise
+        """
+        return len(self._outgoing_messages) == 0 and len(self._incoming_messages) == 0
+
+    def counterparty_from_message(self, message: Message) -> Address:
```
DavidMinarsch, line 407: This should clearly be a property of the `message` not the `dialogue`!

aea/helpers/dialogue/base.py
```diff
@@ -412,90 +395,145 @@ def last_message(self) -> Optional[Message]:
...
+        if self.is_message_by_self(message):
+            counterparty = message.to
+        else:
+            counterparty = message.sender
```
DavidMinarsch, line 411: One liner please where possible:
```suggestion
        counterparty = message.to if self.is_message_by_self(message) else  message.sender
```

DavidMinarsch: Missing docstrings

aea/helpers/dialogue/base.py
DavidMinarsch: I see; in this case I was wrong and we should keep it indeed.

aea/helpers/dialogue/base.py
DavidMinarsch: Actually, on consideration, this api is still available so it needs to support the case where it is used directly without reply. We don't enforce usage of `reply`, we just make it available after all.

aea/helpers/dialogue/base.py
DavidMinarsch: Ok

aea/helpers/dialogue/base.py
```diff
@@ -412,90 +395,145 @@ def last_message(self) -> Optional[Message]:
...
+    def get_message(self, message_id: int) -> Message:
         """
         Get the message whose id is 'message_id'.
 
-        :param message_id_to_find: the id of the message
+        :param message_id: the id of the message
         :return: the message if it exists, None otherwise
         """
-        result = None  # type: Optional[Message]
+        assert not self.is_empty, "This dialogue is empty"
```
DavidMinarsch, line 440: Consistency in the error messages would be nice; i.e. always end with a dot for instance

aea/helpers/dialogue/base.py
```diff
@@ -412,90 +395,145 @@ def last_message(self) -> Optional[Message]:
...
+            self.last_message.message_id, message_id
+        )
+
         list_of_all_messages = self._outgoing_messages + self._incoming_messages
         for message in list_of_all_messages:
-            if message.message_id == message_id_to_find:
-                result = message
-                break
+            if message.message_id == message_id:
+                return message
```
DavidMinarsch, line 450: You could have another method which wraps this section like so:
````
def try_get_method(self, message_id: int) -> Optional[Message]:
    """
    ...
    """
    result = None  # type: Optional[Message]
    list_of_all_messages = self._outgoing_messages + self._incoming_messages
    for message in list_of_all_messages:
          if message.message_id == message_id_to_find:
                 result = message
                 break
      return  result
```
(i.e. the old `get_message`) and then you use this in the new `get_message`; that way you can then add a line in the new `get_message`:
```
message = try_get_message(message_id)
assert message is not None, "Message not present"
return message
```

aea/helpers/dialogue/base.py
```diff
@@ -412,90 +395,145 @@ def last_message(self) -> Optional[Message]:
...
+        Check whether a message with the supplied message id exists in this dialogue.
 
-    def update(self, message: Message) -> bool:
+        :param message_id: the message id
+        :return: True if message with that id exists in this dialogue, False otherwise
+        """
+        if self.is_empty:
+            return False
+
+        return 1 <= message_id <= self.last_message.message_id
```
DavidMinarsch, line 478: replace 1 with the constant `STARTING_MESSAGE_ID`

aea/helpers/dialogue/base.py
```diff
@@ -412,90 +395,145 @@ def last_message(self) -> Optional[Message]:
...
+                    message.message_id,
+                    message.dialogue_reference,
+                    self.dialogue_label.dialogue_reference,
+                )
+            )
 
-    def ensure_counterparty(self, message: Message) -> None:
-        """
-        Ensure the counterparty is set (set if not) correctly.
+        is_valid_result, is_valid_message = self.is_valid_next_message(message)
```
DavidMinarsch, line 500: The second variable should not start with `is_` as it is a string; maybe call it `validation_msg`

aea/helpers/dialogue/base.py
```diff
@@ -412,90 +395,145 @@ def last_message(self) -> Optional[Message]:
...
-            counterparty = message.counterparty
-        except AssertionError:
-            # assume message belongs to dialogue
-            message.counterparty = self.dialogue_label.dialogue_opponent_addr
+        if not is_valid_result:
+            raise InvalidDialogueMessage(
+                "Message {} is invalid with respect to this dialogue. Error: ".format(
+                    message.message_id
+                )
+                + is_valid_message
```
DavidMinarsch, line 507: don't use `+`, you already use format, so add another `{}` and pass it to format

aea/helpers/dialogue/base.py
```diff
@@ -412,90 +395,145 @@ def last_message(self) -> Optional[Message]:
...
+                self.counterparty_from_message(message),
                 self.agent_address,
             )
             result = self_initiated_dialogue_label in self.dialogue_labels
         else:
             other_initiated_dialogue_label = DialogueLabel(
-                message.dialogue_reference, message.counterparty, message.counterparty
+                message.dialogue_reference,
+                self.counterparty_from_message(message),
+                self.counterparty_from_message(message),
```
DavidMinarsch, line 536: I'd assign to helper var and do a single call to `self.counterparty_from_message(message)`

aea/helpers/dialogue/base.py
```diff
@@ -412,90 +395,145 @@ def last_message(self) -> Optional[Message]:
...
+            return False
+
+        return 1 <= message_id <= self.last_message.message_id
+
+    def update(self, message: Message) -> None:
         """
         Extend the list of incoming/outgoing messages with 'message', if 'message' belongs to dialogue and is valid.
 
         :param message: a message to be added
         :return: True if message successfully added, false otherwise
```
DavidMinarsch, line 485: Update docstring return type and mention it can raise `:raises: ...`

aea/helpers/dialogue/base.py
```diff
@@ -543,13 +580,24 @@ def is_valid_next_message(self, message: Message) -> bool:
         :param message: the message to be validated
         :return: True if yes, False otherwise.
         """
-        return (
-            self._basic_rules(message)
-            and self._additional_rules(message)
-            and self.is_valid(message)
-        )
+        result_basic_validation, msg_basic_validation = self._basic_validation(message)
```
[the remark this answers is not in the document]
DavidMinarsch: Note; all the methods with return a tuple now can no longer start with `is`, `has` or `can`... you could raise early so you don't have to pass all these messages around...

aea/helpers/dialogue/base.py
```diff
@@ -567,30 +615,80 @@ def _basic_rules(self, message: Message) -> bool:
...
-        else:
+        if self.last_message is None:  # Initial message
+            if not dialogue_reference[0] == self.dialogue_label.dialogue_reference[0]:
+                return (
+                    False,
+                    "Invalid dialogue_reference[0]. Expected {}. Found {}.".format(
+                        self.dialogue_label.dialogue_reference[0], dialogue_reference[0]
+                    ),
+                )
+            if not message_id == Dialogue.STARTING_MESSAGE_ID:
```
DavidMinarsch, line 626:
```suggestion
            if message_id != Dialogue.STARTING_MESSAGE_ID:
```

DavidMinarsch: same below

aea/helpers/dialogue/base.py
```diff
@@ -567,30 +615,80 @@ def _basic_rules(self, message: Message) -> bool:
...
 
-        if self.last_message is None:
-            result = (
-                dialogue_reference[0] == self.dialogue_label.dialogue_reference[0]
-                and message_id == Dialogue.STARTING_MESSAGE_ID
-                and target == Dialogue.STARTING_TARGET
-                and performative in self.rules.initial_performatives
-            )
-        else:
+        if self.last_message is None:  # Initial message
```
DavidMinarsch, line 618: There are a lot of lines here in one method; suggest breaking it up and having some more helper methods (maybe content of each block

aea/helpers/dialogue/base.py
DavidMinarsch: yes; just have a few guard clauses, easier to read:
```
if ...
 return
if ...
  return
...
return
```

aea/helpers/dialogue/base.py
```diff
@@ -773,13 +895,39 @@ def dialogue_stats(self) -> DialogueStats:
...
+        :param message: the message
+        :return: True if message is by the counterparty agent in this dialogue, False otherwise
+        """
+        return not self.is_message_by_self(message)
+
+    def counterparty_from_message(self, message: Message) -> Address:
+        if self.is_message_by_self(message):
+            counterparty = message.to
+        else:
+            counterparty = message.sender
```
DavidMinarsch, line 920:
```suggestion
        counterparty = message.to if self.is_message_by_self(message) else message.sender
```

docs/api/helpers/dialogue/base.md
```diff
@@ -521,11 +521,11 @@ Extend the list of incoming/outgoing messages with 'message', if 'message' belon
 
 True if message successfully added, false otherwise
 
-<a name="aea.helpers.dialogue.base.Dialogue.ensure_counterparty"></a>
```
DavidMinarsch, line 524: don't bother changing api docs; it's auto generated

aea/helpers/dialogue/base.py
5A11: There is no change. Not sure why GH shows a removed line.

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: They were meant to be read as the bool result returned by `is_valid` and the message returned by `is_valid`.

Anyway changed.

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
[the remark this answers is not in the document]
5A11: That means changing `is_valid` abstract method which has implication on generator and other dialogue packages. Is this desired?

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: This would only add further messiness to this already bloated module, where we support multiple ways of achieving the same thing, and leads to repetition of code, obscure errors, ...

5A11: If we are making breaking changes, as its result developers are to update their code, then we should create the clean state we are after. Otherwise, when is a good time to make this separation of responsibilities between the three methods (reply, create, update) if not when we are doing a major restructuring?

aea/helpers/dialogue/base.py
5A11: I added this, but I think this adds unnecessary repetition to the module and confusion in usage (which method to use) while adding very minor benefits.

aea/helpers/dialogue/base.py
```diff
@@ -510,10 +565,10 @@ def reply(self, target_message: Message, performative, **kwargs) -> Message:
```
DavidMinarsch, line 565: I think the `target_message` should be optional argument and by default (i.e. if not provided ink wargs) equal to the last message. That will simplify usage, as in most cases you reply to last message in dialogue

aea/helpers/dialogue/base.py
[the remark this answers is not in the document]
DavidMinarsch: Well, ideally we would avoid it. But consistency is important. For now we could leave it and monitor the situation for `is_valid`. We should change it for the other internal methods.

aea/helpers/dialogue/base.py
```diff
@@ -939,14 +1132,15 @@ def get_dialogue(self, message: Message) -> Optional[Dialogue]:
...
-            dialogue_reference, counterparty, self.agent_address
+            message.dialogue_reference,
+            self.counterparty_from_message(message),
+            self.agent_address,
         )
         other_initiated_dialogue_label = DialogueLabel(
-            dialogue_reference, counterparty, counterparty
+            message.dialogue_reference,
+            self.counterparty_from_message(message),
+            message.to,
```
DavidMinarsch, line 1143:
```suggestion
            message.dialogue_reference,
            self.counterparty_from_message(message),
            self.counterparty_from_message(message),
```

aea/helpers/dialogue/base.py
DavidMinarsch: There is no repetition; it's two methods offering different behaviours on accessing the message from its id.

aea/helpers/dialogue/base.py
DavidMinarsch: We have a convention and we are sticking to it.

aea/helpers/dialogue/base.py
5A11: Good idea.

Done

docs/api/helpers/dialogue/base.md
5A11: It was an automatically changed due to PyCharm refactoring. I'll revert back.

aea/helpers/dialogue/base.py
5A11: Done

aea/protocols/base.py
```diff
@@ -88,8 +86,8 @@ def sender(self) -> Address:
 
         :return the address
         """
-        assert self._sender is not None, "Sender must not be None."
-        return self._sender
+        assert self.has_sender, "Message's 'Sender' field must be set."
+        return self._sender  # type: ignore
```
DavidMinarsch, line 90: Role back the changes to avoid using type ignore...

aea/protocols/base.py
```diff
@@ -105,49 +103,15 @@ def has_to(self) -> bool:
     @property
     def to(self) -> Address:
         """Get address of receiver."""
-        assert self._to is not None, "To must not be None."
-        return self._to
+        assert self.has_to, "Message's 'To' field must be set."
```
DavidMinarsch, line 106: See above ...

aea/protocols/base.py
5A11: Done

aea/protocols/base.py
5A11: Done

tests/test_helpers/test_dialogue/test_base.py
```diff
@@ -81,17 +82,9 @@ def __init__(
...
-                cast(FrozenSet[Message.Performative], self.TERMINAL_PERFORMATIVES),
-                cast(
-                    Dict[Message.Performative, FrozenSet[Message.Performative]],
-                    self.VALID_REPLIES,
-                ),
-            ),
         )
 
-    def is_valid(self, message: Message) -> bool:
+    def is_valid(self, message: Message) -> Tuple[bool, str]:
```
DavidMinarsch, line 87: Can we actually rename this (as discussed above) and make it a default implementation in base class. Then if the user wants to; they can always override it? Saves us having to have this unnecessary code here.

tests/test_helpers/test_dialogue/test_base.py
```diff
@@ -280,102 +257,331 @@ def test_dialogue_properties(self):
...
-    def test_update_negative_not_is_extendible(self):
-        """Negative test for the 'update' method: dialogue is not extendable with the input message."""
-        invalid_message_id = DefaultMessage(
+    def test_is_message_by_self(self):
+        """Test the 'is_message_by_self' method."""
+        message_1 = DefaultMessage(
             dialogue_reference=(str(1), ""),
-            message_id=0,
+            message_id=1,
             target=0,
```
DavidMinarsch, line 297: Throughout you can drop message id and target when they are equal to the default values

tests/test_helpers/test_dialogue/test_base.py
```diff
@@ -280,102 +257,331 @@ def test_dialogue_properties(self):
...
+
+        message_2 = DefaultMessage(
+            dialogue_reference=(str(1), ""),
+            message_id=1,
+            target=0,
+            performative=DefaultMessage.Performative.BYTES,
+            content=b"Hello",
+        )
+        message_2.sender = self.opponent_address
+        message_2.to = self.agent_address
```
DavidMinarsch, line 337: The same two messages are used throughout; why not construct them once in the setup and then assert different properties/behaviours

tests/test_helpers/test_dialogue/test_base.py
```diff
@@ -280,102 +257,331 @@ def test_dialogue_properties(self):
...
+            target=0,
+            performative=DefaultMessage.Performative.BYTES,
+            content=b"Hello",
+        )
+        message_1.sender = self.agent_address
+        message_1.to = self.opponent_address
+
+        self.dialogue.update(message_1)
+
+        assert self.dialogue.try_get_message(2) is None
```
DavidMinarsch, line 358: Magic values like the 2 here make tests hard to read for others. Declare a var like `unused_message_id = 2` and the next person will get it.

tests/test_helpers/test_dialogue/test_base.py
```diff
@@ -280,102 +257,331 @@ def test_dialogue_properties(self):
...
+            target=1,
+            performative=DefaultMessage.Performative.BYTES,
+            content=b"Hello II",
+        )
+        message_2.sender = self.agent_address
+        message_2.to = self.opponent_address
+
+        self.dialogue.update(message_2)
+
+        assert self.dialogue.try_get_message(2) == message_2
```
DavidMinarsch, line 372: see above

DavidMinarsch: here 2 is meant to match the message_id of the message; so declare a var and ensure the two indeed match! Next person rewriting the code will otherwise easily miss this.

tests/test_helpers/test_dialogue/test_base.py
```diff
@@ -280,102 +257,331 @@ def test_dialogue_properties(self):
...
+            performative=DefaultMessage.Performative.BYTES,
+            content=b"Hello",
+        )
+        message_1.sender = self.agent_address
+        message_1.to = self.opponent_address
+
+        self.dialogue.update(message_1)
+
+        with pytest.raises(AssertionError) as cm:
+            self.dialogue.get_message(2)
```
DavidMinarsch, line 393: see above

tests/test_helpers/test_dialogue/test_base.py
```diff
@@ -280,102 +257,331 @@ def test_dialogue_properties(self):
...
+            target=1,
+            performative=DefaultMessage.Performative.BYTES,
+            content=b"Hello II",
+        )
+        message_2.sender = self.agent_address
+        message_2.to = self.opponent_address
+
+        self.dialogue.update(message_2)
+
+        assert self.dialogue.get_message(2) == message_2
```
DavidMinarsch, line 408: see above

packages/fetchai/skills/erc1155_client/handlers.py
```diff
@@ -555,10 +559,11 @@ def _handle_signed_message(
             performative=FipaMessage.Performative.ACCEPT_W_INFORM,
             info={"tx_signature": signing_msg.signed_message.body},
         )
-        inform_msg.counterparty = last_fipa_msg.counterparty
+        inform_msg.sender = last_fipa_msg.to
```
DavidMinarsch, line 562: note; currently update sets the sender, so this is not strictly needed

packages/fetchai/skills/erc1155_client/behaviours.py
```diff
@@ -57,7 +57,9 @@ def setup(self) -> None:
             ledger_id=strategy.ledger_id,
             address=cast(str, self.context.agent_addresses.get(strategy.ledger_id)),
         )
-        ledger_api_msg.counterparty = LEDGER_API_ADDRESS
+        # ledger_api_msg.sender = ???
```
DavidMinarsch, line 60: see below; update takes care of sender

packages/fetchai/skills/echo/handlers.py
```diff
@@ -125,7 +126,8 @@ def _handle_bytes(self, message: DefaultMessage, dialogue: DefaultDialogue):
             target=message.message_id,
             content=message.content,
         )
-        reply.counterparty = message.sender
+        reply.sender = message.sender
```
DavidMinarsch, line 129: message.to

packages/fetchai/skills/echo/handlers.py
```diff
@@ -90,7 +90,8 @@ def _handle_unidentified_dialogue(self, message: DefaultMessage) -> None:
             error_msg="Invalid dialogue.",
             error_data={"default_message": message.encode()},
         )
-        reply.counterparty = message.sender
+        reply.sender = message.sender
```
DavidMinarsch, line 93: reply.sender = message.to

packages/fetchai/connections/webhook/connection.py
```diff
@@ -173,11 +173,13 @@ async def to_envelope(self, request: web.Request) -> Envelope:
             bodyy=payload_bytes if payload_bytes is not None else b"",
             dialogue_reference=self._dialogues.new_self_initiated_dialogue_reference(),
         )
-        http_message.counterparty = self.agent_address
+        # ToDo
+        # http_message.sender = ???
```
DavidMinarsch, line 177: update sets sender....

tests/test_helpers/test_dialogue/test_base.py
5A11: I'll rename it to `custom_validation`

tests/test_helpers/test_dialogue/test_base.py
5A11: Done

tests/test_helpers/test_dialogue/test_base.py
5A11: Done

tests/test_decision_maker/test_default.py
```diff
@@ -350,13 +350,12 @@ def test_handle_tx_signing_fetchai(self):
...
         )
-        signing_msg.counterparty = "decision_maker"
+        signing_msg.to = "decision_maker"
         signing_dialogue = signing_dialogues.update(signing_msg)
         assert signing_dialogue is not None
         self.decision_maker.message_in_queue.put_nowait(signing_msg)
         signing_msg_response = self.decision_maker.message_out_queue.get(timeout=2)
-        signing_msg_response.counterparty = signing_msg.counterparty
-        signing_msg_response.is_incoming = True
+        signing_msg_response.to = signing_msg.sender
```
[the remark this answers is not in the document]
5A11: Done

tests/test_decision_maker/test_default.py
```diff
@@ -386,13 +385,12 @@ def test_handle_tx_signing_ethereum(self):
...
```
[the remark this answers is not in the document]
5A11: Done

tests/test_decision_maker/test_default.py
```diff
@@ -425,13 +423,12 @@ def test_handle_tx_signing_unknown(self):
...
```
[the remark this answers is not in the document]
5A11: Done

tests/test_decision_maker/test_default.py
```diff
@@ -461,13 +458,12 @@ def test_handle_message_signing_fetchai(self):
...
```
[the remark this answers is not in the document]
5A11: Done

tests/test_decision_maker/test_default.py
```diff
@@ -497,13 +493,12 @@ def test_handle_message_signing_ethereum(self):
...
```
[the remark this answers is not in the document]
5A11: Done

tests/test_decision_maker/test_default.py
```diff
@@ -533,13 +528,12 @@ def test_handle_message_signing_ethereum_deprecated(self):
...
```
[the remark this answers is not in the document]
5A11: Done

tests/test_decision_maker/test_default.py
```diff
@@ -570,13 +564,12 @@ def test_handle_message_signing_unknown_and_two_dialogues(self):
...
```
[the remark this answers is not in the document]
5A11: Done

packages/fetchai/skills/tac_control/dialogues.py
```diff
@@ -143,36 +124,45 @@ def __init__(self, **kwargs) -> None:
...
+        BaseTacDialogues.__init__(
+            self,
+            agent_address=self.context.agent_address,
+            role_from_first_message=role_from_first_message,
         )
-        self._dialogue_by_address[dialogue_label.dialogue_opponent_addr] = dialogue
-        return dialogue
+        self._dialogue_by_address = {}  # type: Dict[str, Dialogue]
+
+    # ToDo replace the dialogue_by_address functionality if required with a method that dynamically
```
5A11, line 146: Note here. The deprecated `create` method updated an auxiliary `dict` as well as creating a dialogue, which is the only place this method does something more than creating a dialogue. 

Moving to the new approach means this functionality is lost. 

If `dialogue_by_address`'s functionality cannot be achieved using standard fields and methods I suggest creating a method that upon accessing it dynamically goes through all dialogues and constructs the auxiliary dict. 

This is used 3 times in the behaviour and handler classes.

packages/fetchai/skills/ml_train/dialogues.py
```diff
@@ -234,32 +208,22 @@ def __init__(self, **kwargs) -> None:
...
-    ) -> OefSearchDialogue:
-        """
-        Create an instance of oef_search dialogue.
-
-        :param dialogue_label: the identifier of the dialogue
-        :param role: the role of the agent this dialogue is maintained for
+            :param message: an incoming/outgoing first message
+            :param receiver_address: the address of the receiving agent
+            :return: The role of the agent
+            """
```
DavidMinarsch, line 220: missing role

packages/fetchai/skills/tac_control/dialogues.py
```diff
@@ -143,36 +124,45 @@ def __init__(self, **kwargs) -> None:
         :return: None
         """
         Model.__init__(self, **kwargs)
-        BaseTacDialogues.__init__(self, self.context.agent_address)
-        self._dialogue_by_address = {}  # type: Dict[str, Dialogue]
-
-    @property
-    def dialogue_by_address(self) -> Dict[str, Dialogue]:
```
DavidMinarsch, line 150: Don't remove properties...

packages/fetchai/skills/tac_control/dialogues.py
DavidMinarsch: This functionality needs to be added again. I don't see why we cannot simply store it as before..

packages/fetchai/skills/ml_train/dialogues.py
5A11: Done

packages/fetchai/skills/tac_control/dialogues.py
5A11: Functionality restored

aea/helpers/dialogue/base.py
```diff
@@ -972,6 +973,20 @@ def dialogue_stats(self) -> DialogueStats:
...
+        """
+        Get the dialogue by address.
+
+        :param counterparty: the counterparty
+        :return: The dialogue is one exists with the counterparty, None otherwise.
+        """
+        if counterparty in self._dialogue_by_address:
+            return self._dialogue_by_address[counterparty]
+        else:
+            return None
```
DavidMinarsch, line 988:
```suggestion
        return self._dialogue_by_address.get(counterparty, None)
```

aea/helpers/dialogue/base.py
```diff
@@ -412,185 +403,361 @@ def last_message(self) -> Optional[Message]:
...
-        if is_extendable:
-            if message.is_incoming:
-                self._incoming_messages.extend([message])
-            else:
-                self._outgoing_messages.extend([message])
-        return is_extendable
+        return self.STARTING_MESSAGE_ID <= message_id <= self.last_message.message_id  # type: ignore
 
-    def ensure_counterparty(self, message: Message) -> None:
+    def update(self, message: Message) -> None:
```
DavidMinarsch, line 500: We should make this method private (i.e. relabel to `_update`). In the new API concept, the `update` functionality is either called from `reply` or from the `Dialogues`' `create` and `reply`, we should avoid it being called directly. By making this private, we can also maintain the old method `update` with its old signature (returning a bool) (which wraps `_update` in a try, except block). This makes upgrading easier and we can furnish the old `update` with a `DeprecationWarning`. This change should be done after we have converted all skills to new API (so not as part of this PR).

aea/helpers/dialogue/base.py
```diff
@@ -412,185 +403,361 @@ def last_message(self) -> Optional[Message]:
...
         :param target_message: the message to reply to.
         :param performative: the performative of the reply message.
         :param kwargs: the content of the reply message.
 
         :return: the reply message if it was successfully added as a reply, None otherwise.
         """
-        assert (
-            self._message_class is not None
-        ), "No 'message_class' argument was provided to this class on construction."
         assert self.last_message is not None, "Cannot reply in an empty dialogue!"
```
DavidMinarsch, line 577:
```suggestion
        last_message = self.last_message 
        assert last_message  is not None, "Cannot reply in an empty dialogue!"
```

aea/helpers/dialogue/base.py
```diff
@@ -412,185 +403,361 @@ def last_message(self) -> Optional[Message]:
...
 
+        if target_message is None:
+            target_message = self.last_message
```
DavidMinarsch, line 580:
```suggestion
            target_message = last_message
```

aea/helpers/dialogue/base.py
```diff
@@ -412,185 +403,361 @@ def last_message(self) -> Optional[Message]:
...
+            assert self.has_message(
+                target_message  # type: ignore
+            ), "The target message does not exist in this dialogue."
+
         reply = self._message_class(
             dialogue_reference=self.dialogue_label.dialogue_reference,
-            message_id=self.last_message.message_id + 1,
-            target=target_message.message_id,
+            message_id=self.last_message.message_id + 1,  # type: ignore
+            target=target_message.message_id,  # type: ignore
```
DavidMinarsch, line 589:
```suggestion
            message_id=last_message.message_id + 1,
            target=target_message.message_id,
```

aea/helpers/dialogue/base.py
DavidMinarsch: By doing an assignment we can avoid multiple calls to `self.last_message` and satisfy mypy (see comments below)

aea/helpers/dialogue/base.py
DavidMinarsch: Old update can then be removed fully with `0.7.0`

aea/helpers/dialogue/base.py
5A11: I think this is not necessary as after I changed the assertion from `is_empty` to checking explicitly that `last_message is not None`, I forgot to remove `type: ignore` form the lines. 

Though for performance reasons it is cleaner to make the call once and then reuse `last_message` across the method so I'm applying your suggestion.

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: Done

aea/helpers/dialogue/base.py
5A11: Done

aea/protocols/base.py
```diff
@@ -297,36 +265,6 @@ def decode(obj: bytes) -> Message:
         return msg
 
 
-class JSONSerializer(Serializer):
```
DavidMinarsch, line 300: We cannot maintain this Serializer as it does not preserve all structure on the objects passed in `body`. It is also not used, so can be safely removed.

aea/protocols/base.py
```diff
@@ -217,8 +180,13 @@ def __eq__(self, other):
         """Compare with another object."""
         return (
             isinstance(other, Message)
+            and self._sender == other._sender
+            and self._to == other._to
+            # and self.dialogue_reference == other.dialogue_reference  # noqa: E800
+            # and self.message_id == other.message_id  # noqa: E800
+            # and self.target == other.target  # noqa: E800
+            # and self.performative == other.performative  # noqa: E800
```
DavidMinarsch, line 188: Currently, this is not possible. We first need to implement the `message` format in protobuf and add the default fields so we can properly serialise.

tests/test_protocols/test_base.py
```diff
@@ -97,25 +131,6 @@ def test_default_protobuf_serialization(self):
         actual_msg = self.message
         assert expected_msg == actual_msg
 
-    def test_default_json_serialization(self):
```
DavidMinarsch, line 100: see above why removed

aea/helpers/dialogue/base.py
5A11: I agree with privatising update (`update` to `_update`) to discourage use.

I don't quite like postponing the cleanups to another version and introduce interim functions. Developers cannot avoid changing their code if they want to upgrade to 0.6.0 (which is the only distinction between major/minor versions which we placed ourselves, i.e. backwards compatibility). 

I don't see why them not upgrading one part of their code in 0.6.0 (`Dialogue.update`) and doing it in 0.7.0 is helpful. On the other side, we just have to come back to this part of the code again in a few weeks and make the move completely.

aea/protocols/base.py
5A11: Agree

aea/mail/base.py
```diff
@@ -404,6 +406,21 @@ def connection_id(self) -> Optional[PublicId]:
             connection_id = self.context.connection_id
         return connection_id
 
+    @staticmethod
+    def _check_consistency(message: Message, to: str, sender: str) -> Message:
```
DavidMinarsch, line 410: Envelopes need to match their messages and vice versa.

tests/test_packages/test_connections/test_gym/test_gym.py
```diff
@@ -178,8 +190,6 @@ async def test_send_close_negative(self):
             dialogue_reference=self.dialogues.new_self_initiated_dialogue_reference(),
         )
         msg.to = self.gym_address
-        dialogue = self.dialogues.update(msg)
```
DavidMinarsch, line 194: Something got lost here...

5A11: Here you are trying to 1. create a wrong message, 2. add it in a dialogue, 3. send it to the connection, and 4. check the connection raises an error. 

With the new error system for dialogue, it is not possible to do 2. So the incorrect message is just created and sent to the connection.

5A11: Here you are trying to 1. create a wrong message, 2. add it in a dialogue, 3. send it to the connection, and 4. check the connection raises an error.

With the new error system for dialogue, it is not possible to do 2. So the incorrect message is just created and sent to the connection.

DavidMinarsch: @5A11 sounds good. but then this should be commented here in code; the next person will not immediately see this.

tests/test_packages/test_protocols/test_fipa.py
```diff
@@ -48,12 +48,9 @@ def test_fipa_cfp_serialization():
...
         query=query,
     )
-    msg.counterparty = "receiver"
+    msg.to = "receiver"
     envelope = Envelope(
-        to="receiver",
-        sender="sender",
-        protocol_id=FipaMessage.protocol_id,
-        message=msg,
+        to=msg.to, sender="sender", protocol_id=FipaMessage.protocol_id, message=msg,
```
DavidMinarsch, line 53: msg.sender must be set too!

tests/test_packages/test_connections/test_local/test_search_services.py
```diff
@@ -540,19 +533,22 @@ def setup_class(cls):
...
+        # search_services_request = OefSearchMessage(
+        #     performative=OefSearchMessage.Performative.SEARCH_SERVICES,
+        #     dialogue_reference=self.dialogues1.new_self_initiated_dialogue_reference(),
+        #     query=query,
+        # )
+        # search_services_request.to = str(OEFLocalConnection.connection_id)
+        # sending_dialogue = cast(
+        #     Optional[OefSearchDialogue], self.dialogues1.update(search_services_request)
+        # )
+        # assert sending_dialogue is not None
```
DavidMinarsch, line 546: remove comments

tests/test_packages/test_connections/test_http_server/test_http_server.py
```diff
@@ -127,19 +157,28 @@ async def test_get_200(self):
...
+        #     version=incoming_message.version,
+        #     headers=incoming_message.headers,
+        #     message_id=incoming_message.message_id + 1,
+        #     target=incoming_message.message_id,
+        #     status_code=200,
+        #     status_text="Success",
+        #     bodyy=b"Response body",
+        # )
+        # message.to = incoming_message.sender
+        # assert dialogue.update(message)
```
DavidMinarsch, line 172: remove comment

packages/fetchai/connections/http_server/connection.py
```diff
@@ -188,6 +188,15 @@ def to_envelope_and_set_id(
...
         context = EnvelopeContext(connection_id=connection_id, uri=uri)
+        # http_message, dialogue = dialogues.create(
+        #     counterparty=agent_address,
+        #     performative=HttpMessage.Performative.REQUEST,
+        #     method=self.method,
+        #     url=url,
+        #     headers=self.parameters.header,
+        #     bodyy=self.body if self.body is not None else b"",
+        #     version="",
+        # )
```
DavidMinarsch: add it

tests/test_packages/test_connections/test_gym/test_gym.py
5A11: Done

tests/test_packages/test_connections/test_local/test_search_services.py
5A11: Done

tests/test_packages/test_connections/test_http_server/test_http_server.py
5A11: Done

packages/fetchai/connections/http_server/connection.py
5A11: Done

DavidMinarsch (approved):
LGTM

Issue #1426, opened 2020-06-25
loop property for Connection abstract class


solarw:
**Is your feature request related to a problem? Please describe.**
the main problem Connection constructed outside multiplexer, and multiplexer can create own loop,
so referring loop inside connection can cause some problems with using entities bound to different loops,

**Describe the solution you'd like**
create a loop property to provide access to the current running loop for connection's code, but only inside that loop.
so get current loop and loop has to be running, or raise an exception

PR #1646, opened 2020-08-12
[AEA-814] Connection loop refactoring

solarw:
## Proposed changes

move the connection loop under the property. remove all direct usage of `_loop` property.
some connections refactoring

## Fixes
https://github.com/fetchai/agents-aea/issues/1426

## Types of changes

## Checklist


solarw:
Initial problem was:
Connections are created outside of multiplexer, and every asyncio object created in `Connection.__init__` is bound to the default event loop, same time multiplexer(in threaded mode) creates own loop, and all objects created in connection init are useless.

Assume that all asyncio loop bound objects have to be constructed in `Connection.connect` called within the multiplexer event loop.
Assume that all public api of connection are called within one event loop (the multiplexer one).

There are two possible solutions:
1. always use current event loop so  `self.loop == asyncio.get_event_loop()`. I see no cases that connection will be managed directly from another event loop without an explicit selection of the target loop.
2. set `connection.loop` on `connection.connect` and every method called in connections will use loop set on connect. this will allow us to call some methods outside of the initial event loop, but not sure it actually required.


some checks showed me that all tests are passed with the first solution. so not sure we need to set loop on connection and can use current one with connection.loop property.

only one thing developer has to remember about: is to create objects within connect coroutine.


also in python 3.10 loop parameter will be removed from some asyncio objects, so loop will be current one all the time.

aea/connections/base.py
```diff
@@ -100,22 +101,25 @@ def __init__(
...
+        assert asyncio.get_event_loop().is_running()
+        return asyncio.get_event_loop()
+
+    def _ensure_connected(self) -> None:  # pragma: nocover
+        """Raise exception if connection is not connected."""
+        if not self.is_connected:
+            raise ConnectionError("Connection is not connected! Connect first!")
+
+    @contextmanager
+    def _connect_context(self) -> Generator:
```
DavidMinarsch, line 115: why not make them public?

aea/connections/base.py
```diff
@@ -100,22 +101,25 @@ def __init__(
...
+        """Set state connecting, disconnecteing, dicsconnected during connect method."""
+        with self._state.transit(
+            initial=ConnectionStates.connecting,
+            success=ConnectionStates.connected,
+            fail=ConnectionStates.disconnected,
+        ):
+            yield
```
DavidMinarsch, line 122: 👍

aea/connections/base.py
```diff
@@ -86,7 +88,6 @@ def __init__(
         assert (
             super().public_id == self.connection_id
         ), "Connection ids in configuration and class not matching."
-        self._loop: Optional[AbstractEventLoop] = None
```
DavidMinarsch, line 89: @marcofavorito is it safe to remove this? I think we are loosing an ability to configure the object; so I don't see why this is beneficial.

aea/connections/base.py
solarw: it's only for internal use. especially _connect_context, cause it changes internal state.

aea/connections/base.py
solarw: keep it simple. i made checks that the current test base works well without this feature.
also, this loop was set via property access outside the connection, that's tricky a bit and may confuse. in some code, loop was set directly, in some not. anyway, all the time loop was set to the current one.


for the moment it looks redundant without objective cases.

aea/connections/base.py
```diff
@@ -100,22 +101,25 @@ def __init__(
...
-        Set the event loop.
-
-        :param loop: the event loop.
-        :return: None
-        """
-        assert (
-            self._loop is None or not self._loop.is_running()
-        ), "Cannot set the loop while it is running."
-        self._loop = loop
+        assert asyncio.get_event_loop().is_running()
```
marcofavorito, line 106: can I ask you why the loop must be running? In some implementations, I might just want to get the loop.
The check could be done either manually or by having another property like `ensure_running_loop` that does the check.

aea/connections/base.py
marcofavorito: An argument in favour of public methods is that among framework users there will be some that will develop their connection, and those methods could be designed with that in mind. Anyway, I'm happy to keep them private at the moment.

packages/fetchai/connections/http_server/connection.py
```diff
@@ -533,6 +533,7 @@ async def connect(self) -> None:
         """
         if self.is_connected:
             return
+
         self._state.set(ConnectionStates.connecting)
         self.channel.logger = self.logger
         await self.channel.connect(loop=self.loop)
```
marcofavorito, line 539: Regarding the loop parameter; I guess this and other connection channels could be changed accordingly (just use `asyncio` loop getter)? Not crucial though; it could be done in another moment.

aea/connections/base.py
marcofavorito: ... maybe `loop` and `running_loop`, respectively?

Or even, don't have loop properties at all. Connection developers can just use `asyncio.get_event_loop` and `asyncio.get_running_loop` respectively. Do you see benefits, or is that too extreme?

aea/connections/base.py
marcofavorito: I agree with the change. It simplifies the APIs and almost surely `asyncio.get_event_loop()` will return the right one, either if the connection run by the multiplexer or as a stand-alone connection.

In fact, modern Python versions deprecated `loop` parameters in many `asyncio` APIs, and I believe the same arguments for removal apply in this case too.

Good readings:
https://stackoverflow.com/a/60315290/7150172
https://vorpus.org/blog/some-thoughts-on-asynchronous-api-design-in-a-post-asyncawait-world/#getting-in-touch-with-your-event-loop

aea/helpers/async_utils.py
```diff
@@ -61,6 +64,9 @@ def ensure_list(value: Any) -> List:
     return [value]
 
 
+not_set = object()
```
marcofavorito, line 67: Could you explain to me why this can't be just `None`? I might miss something.

aea/connections/base.py
solarw: probably this check is redundant and kind of legacy from the previous implementation

packages/fetchai/connections/http_server/connection.py
solarw: supposed that channel is internal implementation and used by the connection, channel have to share the same loop as connection uses, so would stay explicit loop set for channel. and if we will change the loop parameter of connection it will be easier to change just loop property of connection instead of changing loop property of connection and channel

aea/helpers/async_utils.py
solarw: cause None is the valid value for the state, and when we want check for value is not defined bu can be set to None we should use something like NonSet value

aea/connections/base.py
solarw: also this should help to avoid loop usage outside of asyncio coroutines.

so 
```
class Connection:
   def __init__():
        self.arg = something(self.loop)
```

will fail if __init__ called outside of asyncio coroutine

Issue #798, opened 2020-02-19
Option to set all configs from `aea-config.yaml`


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
Currently, configs need to be set at each level.

**Describe the solution you'd like**
Allow for an override option where configs set on top level (`aea-config.yaml`) override all the configs in the packages. This way, an `aea-config.yaml` fully specifies an agent.

DavidMinarsch:
Ideally something like a multi-page yaml doc:

```
aea_version: ">=0.2.0"
agent_name: myagent
author: fetchai
version: 0.1.0
license: Apache-2.0
fingerprint: {}
connections:
- fetchai/oef:0.1.0
default_connection: fetchai/oef:0.1.0
default_ledger: fetchai
private_key_paths: {}
protocols:
- fetchai/oef_search:0.1.0
- fetchai/default:0.1.0
- fetchai/tac:0.1.0
- fetchai/fipa:0.1.0
skills:
- fetchai/echo:0.1.0
description: "An example of agent configuration file for testing purposes."
logging_config:
  disable_existing_loggers: false
  version: 1
registry_path: ../../packages
ledger_apis:
   fetchai:
     args:
       address: example.fetch-ai.com
       port: 8080
   ethereum:
     args:
       address: example.ethereum.com
       chain_id: 8080
---
name: dummy
author: dummy_author
version: 0.1.0
description: a dummy_skill for testing purposes.
license: Apache-2.0
aea_version: '>=0.2.0, <0.3.0'
fingerprint:
  __init__.py: Qmd3mY5TSBA632qYRixRVELXkmBWMZtvnxJyQC1oHDTuEm
  behaviours.py: QmWKg1GfJpuJSoCkEKW1zUskkNo4Rsoan1AD2cXpe2E93C
  dummy.py: QmeV6FBPAkmQC49gATSmU1aq8S1SKPv7cm2zSH8cnuqoLT
  handlers.py: QmVe1VGT7TBdjQ5xG8X9djY77f8hebHeUeX3eDrQ6FR3yt
  tasks.py: Qmegg4QsYSqSZN3q2zuRiBAToQ2LEiWrAPtUo7rCMrxjGJ
fingerprint_ignore_patterns: []
contracts: []
protocols:
- fetchai/default:0.1.0
behaviours:
  dummy:
    args:
      behaviour_arg_1: 1
      behaviour_arg_2: '2'
    class_name: DummyBehaviour
handlers:
[...]

PR #1667, opened 2020-08-16
Add 'type' field in json schema and update configurations

marcofavorito:
## Proposed changes

at the moment the field is optional for backward compatibility with old registry packages.

## Fixes

Preparatory PR for #798 

## Types of changes

## Checklist

## Further comments

The `type` field should be made required at some point. Either just before release `0.6.0`, or the absence of it deprecated and made mandatory in `0.7.0`.


aea/configurations/base.py
```diff
@@ -636,6 +636,7 @@ class PackageConfiguration(Configuration, ABC):
     """
 
     default_configuration_filename: str
+    package_type: PackageType
```
marcofavorito, line 639: put it as class attribute so it can be retrieved in the loader just from the Configuration class.

aea/configurations/loader.py
```diff
@@ -181,6 +181,27 @@ def from_configuration_type(
...
+        :param json_data:
+        :return: None
+        :raises ValidationError: if the file doesn't comply with the JSON schema.
+              | ValueError: if other consistency checks fail.
+        """
+        # this might raise ValidationError.
+        self.validator.validate(instance=json_data)
+
+        expected_type = self.configuration_class.package_type
+        # TODO 'type' is optional for backward compatibility
```
marcofavorito, line 197: keep in mind this before the release

marcofavorito: after the latest versions of all the packages will be updated, the `"type" in json_data` should be removed.

aea/configurations/schemas/connection-config_schema.json
```diff
@@ -23,6 +23,9 @@
     "version": {
       "$ref": "definitions.json#/definitions/package_version"
     },
+    "type": {
```
marcofavorito, line 26: in the `"required"` list, `"type"` is temporarily missing (see above).

aea/configurations/schemas/aea-config_schema.json
```diff
@@ -7,7 +7,6 @@
     "aea_version",
     "agent_name",
     "author",
-    "description",
```
DavidMinarsch, line 10: Why are you removing the description?

aea/protocols/signing/protocol.yaml
```diff
@@ -1,6 +1,7 @@
 name: signing
 author: fetchai
 version: 0.2.0
+type: protocol
```
DavidMinarsch, line 4: The protocol generator needs to be updated to account for this extra field!

aea/configurations/schemas/aea-config_schema.json
marcofavorito: I don't see why that should be a mandatory field

DavidMinarsch: It is a mandatory field on registry; add it back please.

aea/protocols/generator/base.py
```diff
@@ -1900,6 +1900,9 @@ def _protocol_yaml_str(self) -> str:
         protocol_yaml_str = "name: {}\n".format(self.protocol_specification.name)
         protocol_yaml_str += "author: {}\n".format(self.protocol_specification.author)
         protocol_yaml_str += "version: {}\n".format(self.protocol_specification.version)
+        protocol_yaml_str += "type: {}\n".format(
```
solarw, line 1903: why do yaml  generated as string not using yaml dump? to preserve order of the keys?

PR #1668, opened 2020-08-16
Make 'setup' and 'teardown' of SkillComponent non-abstract method

marcofavorito:
## Proposed changes

Make 'setup' and 'teardown' of SkillComponent non-abstract method.

This will make the programmatic usage a bit less verbose in case `setup` and `teardown` don't need to be specified. It mostly changes the thing for `Handler` class, since `Model` and `SimpleBehaviour` (but not `Behaviour) already overwrote the methods.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments
Not very important PR.

### DIFF
--- a/aea/skills/base.py
+++ b/aea/skills/base.py
@@ -279,7 +279,6 @@ class SkillComponent(ABC):
         """Get the config of the skill component."""
         return self.configuration.args
 
-    @abstractmethod
     def setup(self) -> None:
         """
         Implement the setup.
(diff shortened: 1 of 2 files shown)


DavidMinarsch:
Whilst in general I agree with the benefit this simplification brings, I would like to see this change happen when we build the second layer handler APIs. The cost of adding the abstract methods back in is higher than keeping them a little longer in my opinion.

PR #1674, opened 2020-08-17
[AEA-506] Handle multipage AEA configuration 

### LINKED ISSUE #798: Option to set all configs from `aea-config.yaml`

marcofavorito:
## Proposed changes

(merge after #1667 )
Handle multipage AEA configurations (see #798)

## Fixes

Fix #798

## Types of changes

## Checklist

## Further comments
n/a


aea/configurations/base.py
```diff
@@ -1310,6 +1311,9 @@ def __init__(
         )  # type: Dict[PublicId, PublicId]
         self.loop_mode = loop_mode
         self.runtime_mode = runtime_mode
```
DavidMinarsch, line 1313: what about from/to json methods?

tests/data/aea-config.example_multipage.yaml
```diff
@@ -0,0 +1,74 @@
...
+   ethereum: eth_private_key.txt
+connection_private_key_paths:
+   cosmos: cosmos_private_key.txt
+   ethereum: eth_private_key.txt
+registry_path: ../../packages
+---
+name: dummy
+author: dummy_author
+version: 0.1.0
+type: skill
```
DavidMinarsch, line 35: Note; this should by definition have to be also listed as a dependency above under skills

aea/configurations/base.py
marcofavorito: will fix that.

tests/data/aea-config.example_multipage.yaml
marcofavorito: right!

aea/configurations/schemas/aea-config_schema.json
```diff
@@ -59,15 +59,6 @@
         }
       }
     },
-    "ledger_apis": {
```
DavidMinarsch, line 62: we can also drop `path_to_contract_interface` in contract json now as it was only kept for backwards comp

tests/data/aea-config.example_multipage.yaml
```diff
@@ -0,0 +1,74 @@
...
+      handler_arg_2: '2'
+    class_name: DummyInternalHandler
+models:
+  dummy:
+    args:
+      model_arg_1: 1
+      model_arg_2: '2'
+    class_name: DummyModel
+skills: []
+dependencies: {}
```
DavidMinarsch, line 73: I don't think we should include all the default fields here; we should only include the configurable parts...most of the above is fixed!

aea/configurations/schemas/connection-config_schema.json
```diff
@@ -10,7 +10,6 @@
     "license",
     "aea_version",
     "class_name",
-    "config",
```
marcofavorito, line 13: I don't think `config` field in connection configuration should be required.

aea/helpers/base.py
```diff
@@ -45,15 +45,11 @@
...
 
 
-def yaml_load(stream: TextIO) -> Dict[str, str]:
-    """
-    Load a yaml from a file pointer in an ordered way.
-
-    :param stream: the file pointer
-    :return: the yaml
-    """
+def _ordered_loading(fun: Callable):
```
marcofavorito, line 48: this is just a refactoring of previous helper functions, to reuse the code for `yaml_(load|dump)_all` helpers

scripts/generate_ipfs_hashes.py
```diff
@@ -131,7 +131,12 @@ def sort_configuration_file(config: PackageConfiguration):
     # load config file to get ignore patterns, dump again immediately to impose ordering
     assert config.directory is not None
     configuration_filepath = config.directory / config.default_configuration_filename
-    yaml_dump(config.ordered_json, configuration_filepath.open("w"))
+    if config.package_type == PackageType.AGENT:
```
marcofavorito, line 134: since we are not using the configuration loader, we have to manually remove the `component_configurations` from `agent_config.json`.
This is because `agent_config.json` is not actually the same JSON object of the YAML file. In the future, we should have that `agent_config.json` is a JSON Object List, but that requires a refactoring of the interface.

aea/configurations/base.py
```diff
@@ -922,7 +923,7 @@ def __init__(
         )
         self.dependencies = dependencies if dependencies is not None else {}
         self.description = description
-        self.config = config
+        self.config = config if len(config) > 0 else {}
```
DavidMinarsch, line 926: when can it be length zero?

aea/configurations/loader.py
```diff
@@ -202,6 +204,198 @@ def _validate(self, json_data: Dict) -> None:
...
+        """
+        configurable_fields: Set[str] = set()
+        if component_id.package_type == PackageType.PROTOCOL:
+            configurable_fields = set()
+        elif component_id.package_type == PackageType.CONNECTION:
+            configurable_fields = {"config"}
+        elif component_id.package_type == PackageType.CONTRACT:
+            configurable_fields = set()
+        elif component_id.package_type == PackageType.SKILL:
+            configurable_fields = {"handlers", "behaviours", "models"}
```
DavidMinarsch, line 390: should be a constant on top of file

aea/configurations/loader.py
```diff
@@ -202,6 +204,198 @@ def _validate(self, json_data: Dict) -> None:
...
+        # common to every package
+        temporary_config["name"] = component_id.name
+        temporary_config["author"] = component_id.author
+        temporary_config["version"] = component_id.version
+        temporary_config["license"] = "some_license"
+        temporary_config["aea_version"] = "0.1.0"
+        if component_id.component_type == ComponentType.PROTOCOL:
+            pass  # no other required field
+        elif component_id.component_type == ComponentType.CONNECTION:
+            temporary_config["class_name"] = "SomeClassName"
```
DavidMinarsch, line 354: what's going on here? also, declare constant, no magic var

aea/configurations/loader.py
```diff
@@ -202,6 +204,198 @@ def _validate(self, json_data: Dict) -> None:
...
+        :raises ValueError: if the configuration is not valid.
+        """
+        # we need to populate the required fields to validate the configurations.
+        temporary_config = deepcopy(configuration)
+        # common to every package
+        temporary_config["name"] = component_id.name
+        temporary_config["author"] = component_id.author
+        temporary_config["version"] = component_id.version
+        temporary_config["license"] = "some_license"
+        temporary_config["aea_version"] = "0.1.0"
```
DavidMinarsch, line 350: constants

aea/configurations/loader.py
```diff
@@ -202,6 +204,198 @@ def _validate(self, json_data: Dict) -> None:
...
+        Split component id and configuration.
+
+        :param i: the position of the component configuration in the agent config file..
+        :param component_configuration_json: the JSON object to process.
+        :return: the component id and the configuration object.
+        :raises ValueError: if the component id cannot be extracted.
+        """
+        result = deepcopy(component_configuration_json)
+        # author, name, version, type are mandatory fields
+        missing_fields = {"author", "name", "version", "type"}.difference(
```
DavidMinarsch, line 312: declare constant

aea/configurations/schemas/connection-config_schema.json
DavidMinarsch: It's used everywhere atm and seems little cost to keep at is? Are we sure making it optional has no unintended consequences?

scripts/generate_ipfs_hashes.py
DavidMinarsch: would it not be easier to just use configuration loader here?

tests/data/aea-config.example_multipage.yaml
DavidMinarsch: I think this should include also a connection spec; so we test overriding that too

tests/test_configurations/test_base.py
```diff
@@ -206,7 +206,10 @@ class TestAgentConfig:
     def test_from_json_and_to_json(self, agent_path):
         """Test the 'from_json' method and 'to_json' work correctly."""
         f = open(agent_path)
-        original_json = yaml.safe_load(f)
+        original_jsons = list(yaml.safe_load_all(f))
+        components = original_jsons[1:]
+        original_json = original_jsons[0]
+        original_json["component_configurations"] = components
```
DavidMinarsch, line 212: This does not look like a good solution; we should not require changes here..

DavidMinarsch: what am I missing?

aea/configurations/base.py
marcofavorito: when the keyword arguments are zero. See above the constructor `**config`

scripts/generate_ipfs_hashes.py
marcofavorito: it is too invasive for the purposes of the script

tests/test_configurations/test_base.py
marcofavorito: `safe_load_all` returns a list and it is the only reasonable way to load a multi-paged YAML file with the `YAML` APIs.

marcofavorito: Also, the `AgentConfig.json` doesn't return a list, but a dictionary. 
Please read this comment: https://github.com/fetchai/agents-aea/pull/1674/files#r472072161

aea/configurations/loader.py
marcofavorito: doesn't make sense to have them as constants. They are just dummy values to feed the component configuration validator.

aea/configurations/loader.py
marcofavorito: see above, please.

aea/configurations/loader.py
marcofavorito: see above

aea/configurations/loader.py
marcofavorito: maybe a custom class for "AgentComponentConfiguration" is better, but at this stage, I don't think there's much value in adding that constant...

aea/aea_builder.py
```diff
@@ -1350,6 +1356,11 @@ def _load_and_add_components(
                     )
             else:
                 configuration = deepcopy(configuration)
+                configuration.update(
+                    self._custom_component_configurations.get(
+                        configuration.component_id, {}
+                    )
+                )
```
marcofavorito, line 1363: we do the replacement only for components added from a directory, so initialization from CLI works.

aea/cli/scaffold.py
```diff
@@ -146,18 +146,19 @@ def scaffold_item(ctx: Context, item_type: str, item_name: str) -> None:
...
             ctx.cwd, item_type_plural, item_name, default_config_filename
         )
-        config = loader.load(config_filepath.open())
+        with config_filepath.open() as fp:
+            config = loader.load(fp)
         config.name = item_name
         config.author = author_name
-        loader.dump(config, open(config_filepath, "w"))
+        with config_filepath.open("w") as fp:
+            loader.dump(config, fp)
```
marcofavorito, line 161: This solved the test issues on Windows. Without context managers, the mocked exception left the file opened, that made the `@clean_after` decorator to fail when attempting to delete the resource folder.

DavidMinarsch (approved):
LGTM

aea/configurations/loader.py
solarw: probably create a set of dummy configs for every component that should pass validation and update those configs with variables from agent config and check will.loader? also requires tests for that configs validation

solarw (approved):
LGTM

please check one comment

aea/configurations/loader.py
marcofavorito: @solarw thanks that sounds like a good idea

PR #1683, opened 2020-08-19
Updated Token Faucet Implementation

ejfitzgerald:
## Proposed changes

Update internal faucet logic to match new faucet implementation on `agent-land` (and subsequent) networks

:construction: **Do not merge yet - faucet rollout in progress** :construction:

### DIFF
--- a/tests/test_crypto/test_cosmos.py
+++ b/tests/test_crypto/test_cosmos.py
@@ -26,11 +26,11 @@ from unittest.mock import MagicMock, call
 import pytest
 
 from aea.crypto.cosmos import (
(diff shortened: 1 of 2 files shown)


tests/test_crypto/test_cosmos.py
```diff
@@ -183,6 +204,82 @@ def get_wealth(address: str):
     return balance
 
 
+@pytest.mark.ledger
```
DavidMinarsch, line 207: You can remove the `ledger` mark as this does not actually call the test net

DavidMinarsch (commented):
LGTM, some parts not covered according to `coverage_checks`, please add additional tests or pragma no cover as you see fit.

PR #1684, opened 2020-08-19
[AEA-834] Add MultiAddr class: computes multiaddress from host, port, and Crypto

lrahmani:
## Proposed changes

Add  MultiAddr class that allows PeerID computation from Private/Public key.
Inspired from https://github.com/libp2p/py-libp2p


aea/helpers/multiaddr/base.py
```diff
@@ -0,0 +1,96 @@
...
+
+import base58
+
+import multihash
+
+from binascii import unhexlify
+
+from aea.crypto.base import Crypto
+
+from .crypto_pb2 import KeyType, PublicKey
```
DavidMinarsch, line 28: Please use the absolute path (`aea. ...`)

aea/helpers/multiaddr/base.py
```diff
@@ -0,0 +1,96 @@
...
+    # Pad odd-length hex strings.
+    return hexed if not len(hexed) & 1 else "0" + hexed
+
+
+def _hex_to_bytes(hexed):
+    return _pad_scalar(unhexlify(_pad_hex(hexed)))
+
+
+class MultiAddr:
+    def __init__(self, host: str, port: int, key: Crypto):
```
DavidMinarsch, line 75: We should pass the public key rather than the whole Crypto object. That makes the MultiAddr class usable in settings where all you have is the public key (and port and host) of some other agent, say.

aea/helpers/multiaddr/base.py
```diff
@@ -0,0 +1,96 @@
...
+        )
+        key_serialized = key_protobuf.SerializeToString()
+        algo = multihash.Func.sha2_256
+        if ENABLE_INLINING and len(key_serialized) <= MAX_INLINE_KEY_LENGTH:
+            algo = IDENTITY_MULTIHASH_CODE
+        key_mh = multihash.digest(key_serialized, algo)
+        return base58.b58encode(key_mh.encode()).decode()
+
+    def to_string(self) -> str:
+        return "/dns/" + self._host + "/tcp/" + str(self._port) + "/p2p/" + self._peerid
```
DavidMinarsch, line 93: f string or `format` approach are preferred

aea/helpers/multiaddr/crypto.proto
```diff
@@ -0,0 +1,20 @@
...
+
+message PublicKey {
+  required KeyType key_type = 1;
+  required bytes data = 2;
+}
+
+message PrivateKey {
+  required KeyType key_type = 1;
+  required bytes data = 2;
+}
```
DavidMinarsch, line 20: I don't see that being used; I'd remove

DavidMinarsch: Unless you want to have a generic protobuf rep of crypto module; then the proto files should be moved to `aea.crypto`

tests/test_helpers/test_multiaddr.py
```diff
@@ -0,0 +1,63 @@
...
+    key = make_crypto(DEFAULT_LEDGER)
+    maddr1 = MultiAddr(HOST, PORT, key)
+
+    tmpdir = tempfile.mkdtemp()
+    key_file = tmpdir + "/key"
+    with open(key_file, "wb+") as k:
+        key.dump(k)
+
+    key2 = make_crypto(DEFAULT_LEDGER, private_key_path=key_file)
+    maddr2 = MultiAddr(HOST, PORT, key2)
```
DavidMinarsch, line 45: See above; rather than passing the whole key pair/crypto object; just pass the `key2.public_key` so we make the `MultiAddr` more usable.

DavidMinarsch (commented):
Looks good so far, I provided some comments. Re `Address static analyzer and linter with regards to protobuf objects`, please update `setup.cfg` so the files are ignored. Agreed that docstrings are missing, so I did not comment on that.

aea/helpers/multiaddr/base.py
marcofavorito:
```
f"/dns/{self._host}/tcp/{self._port}/p2p/{self._peerid}"
```

aea/helpers/multiaddr/base.py
```diff
@@ -0,0 +1,96 @@
...
+        IDENTITY_MULTIHASH_CODE, "identity", hash_new=lambda: IdentityHash()
+    )
+
+
+def _pad_scalar(scalar):
+    return (ZERO * (KEY_SIZE - len(scalar))) + scalar
+
+
+def _pad_hex(hexed):
+    # Pad odd-length hex strings.
```
marcofavorito, line 66: (minor) here you could use docstrings instead of a comment.

aea/helpers/multiaddr/base.py
marcofavorito: I think you can get rid of `to_string` and move the implementation in `__str__`

aea/helpers/multiaddr/base.py
lrahmani: Right. I made Crypto object as part of the API to ensure users will pass the correct format of the public key.
I'll change it to a str object, with a TODO comment to add correct format checks

aea/helpers/multiaddr/base.py
lrahmani: > I think you can get rid of `to_string` and move the implementation in `__str__`

I prefer to have an independent separate public method for it, in addition to string representation. Will rename it to `format`

aea/helpers/multiaddr/crypto.proto
lrahmani: This proto file is the reference one used by libp2p. I kept it as is to avoid having any impact on the serialization algorithm.

Issue #1578, opened 2020-08-03
Option to set skill id in envelope context


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
Not possible atm

**Describe the solution you'd like**
Envelope context to include skill_id too. This is needed for routing from multiplexer to skills.

Issue #1514, opened 2020-07-20
Upgrade internal messaging system


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
Currently, routing from skills to connections is well-defined and expressive. Addressing agent vs its components is not well-defined. Routing from connections to skills is too simplistic.

**Describe the solution you'd like**
Add full routing capabilities from connections to skills. Allow for components (skills & connections only) to be addressed in envelopes. Handle all message manipulation (set `is_incoming` & `counterparty`).

Issue #1409, opened 2020-06-23
Improve envelope routing to skills


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
Currently, envelopes are routed to all skills with relevant handlers.

**Describe the solution you'd like**
Like with outgoing envelopes; we should allow for routing based on two additional concerns:
- default routing table
- envelope context

PR #1685, opened 2020-08-20
Feature/routing

DavidMinarsch:
## Proposed changes

This PR completes the routing functionality provided by envelope context. So far, the routing to skills required a uri to be set, now optionally a skill id can be set explicitly. This functionality is helpful to target a specific skill with a given Envelope, rather than all skills with a matching handler for the given protocol id.

This PR does not fully resolve the routing issue. There are multiple unresolved inconsistencies which need to be addressed in future PRs:
- currently, routing uses information contained in envelope context; but dialogues are maintained using the `sender` and `counterpart` fields.
- the envelope is not an object used in the skills; it is inconsistent that the developer uses the envelope context there.

## Fixes

- #1578
- #1409 (the idea of a routing table from connections to skills was rejected; this role is already taken by the protocol ids specified in handlers)
- partially #1514

## Types of changes

## Checklist

## Further comments

-


aea/protocols/default/README.md
```diff
@@ -1,15 +1,5 @@
 # Default Protocol
 
-**Name:** default
```
DavidMinarsch, line 3: This is unnecessary duplication and just lead's to outdated versions

packages/fetchai/connections/oef/connection.py
```diff
@@ -421,6 +460,7 @@ def send_oef_message(self, envelope: Envelope) -> None:
                 "Could not create dialogue for message={}".format(oef_message)
             )  # pragma: nocover
             return  # pragma: nocover
+        oef_search_dialogue.envelope_context = envelope.context
```
DavidMinarsch, line 463: We use the same context on response. If a skill_id is set in the context it gets applied by AEA when routing the envelope.

packages/fetchai/connections/soef/connection.py
```diff
@@ -388,6 +425,7 @@ async def process_envelope(self, envelope: Envelope) -> None:
             raise ValueError(
                 "Could not create dialogue for message={}".format(oef_message)
             )
+        oef_search_dialogue.envelope_context = envelope.context
```
DavidMinarsch, line 428: see oef; same idea

packages/fetchai/protocols/tac/dialogues.py
```diff
@@ -57,7 +57,13 @@ class TacDialogue(Dialogue):
                 TacMessage.Performative.CANCELLED,
             }
         ),
-        TacMessage.Performative.TAC_ERROR: frozenset(),
+        TacMessage.Performative.TAC_ERROR: frozenset(
```
DavidMinarsch, line 60: This was not correct previously

packages/fetchai/skills/tac_negotiation/behaviours.py
```diff
@@ -106,7 +107,10 @@ def _register_agent(self) -> None:
         oef_search_msg.counterparty = self.context.search_service_address
         oef_search_dialogue = oef_search_dialogues.update(oef_search_msg)
         assert oef_search_dialogue is not None, "OefSearchDialogue not created."
-        self.context.outbox.put_message(message=oef_search_msg)
+        envelope_context = EnvelopeContext(skill_id=self.context.skill_id)
```
DavidMinarsch, line 110: By setting a skill_id in envelope context this skill is the only one receiving the response; without this other skills (e.g. `tac_participation`) would also receive the same envelope.

Issue #1676, opened 2020-08-17
Replace all occurrences of `assert` in core and packages


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
Python's `assert` is a debugging tool (https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement) and we use it wrongly.

**Describe the solution you'd like**
Replace all occurrences of `assert condition, exception_text` with `enforce(condition, exception_text)` where `enforce` is a framework method and defined as follows:
```
def enforce(condition, exception_text) -> None:
    """
    Evaluate a condition and raise an exception with the provided text if it is not satisfied.

    :param condition: the condition
    :param exception_text: the exception to be raised
    """
    if not condition:
        ValueError(exception_text)
```
(Optionally, have a custom exception class.)

PR #1686, opened 2020-08-20
AEA-843 assert replacement

DavidMinarsch:
## Proposed changes

Initial attempt to address #1676 

## Fixes

- #1676 

## Types of changes

## Checklist

## Further comments

-


aea/aea_builder.py
```diff
@@ -765,10 +765,11 @@ def _build_identity_from_wallet(self, wallet: Wallet) -> Identity:
         :param wallet: the wallet
         :return: the identity
         """
-        assert self._name is not None, "You must set the name of the agent."
+        if self._name is None:
```
DavidMinarsch, line 768: Note; in cases where we check for None type, we cannot use `enforce`. Mypy does not recognise it implements the check.

aea/cli/install.py
```diff
@@ -92,7 +92,7 @@ def _install_dependency(dependency_name: str, dependency: Dependency):
         if return_code == 1:
             # try a second time
             return_code = _run_install_subprocess(command)
-        assert return_code == 0, "Return code != 0."
+        enforce(return_code == 0, "Return code != 0.")
```
DavidMinarsch, line 95: In cases where we try to enforce a condition other than for mypy it does fulfil its purpose.

aea/cli/interact.py
```diff
@@ -107,18 +108,23 @@ def _process_envelopes(
...
-        else:
-            click.echo("Received no new envelope!")
+        _check_for_incoming_envelope(inbox)
     else:
         outbox.put(envelope)
         click.echo(_construct_message("sending", envelope))
 
 
-def _construct_message(action_name, envelope):
+def _check_for_incoming_envelope(inbox: InBox):
```
DavidMinarsch, line 117: small refactor

aea/cli/interact.py
```diff
@@ -166,7 +172,7 @@ def _try_construct_envelope(
         )
         msg.counterparty = agent_name
         msg.sender = sender
-        assert dialogues.update(msg) is not None
+        enforce(dialogues.update(msg) is not None, "Dialogue was not created.")
```
DavidMinarsch, line 175: This is one of those cases where the replacement of assert is actually crucial. The line would not be evaluated on the LHS if debug flag is not set! Our enforce will always be evaluated!

aea/exceptions.py
```diff
@@ -26,3 +26,18 @@ class AEAException(Exception):
...
+
+def enforce(is_valid_condition: bool, exception_text: str) -> None:
+    """
+    Evaluate a condition and raise an exception with the provided text if it is not satisfied.
+
+    :param is_valid_condition: the valid condition
+    :param exception_text: the exception to be raised
+    """
+    if not is_valid_condition:
+        AEAEnforceError(exception_text)
```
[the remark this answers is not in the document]
5A11: I suggest parameterising the exception so it would be possible to raise any exception depending on the situation. Could have a default value, say ValueError, to make it easier/faster to use.

tests/test_helpers/test_async_utils.py
```diff
@@ -172,8 +172,8 @@ async def test_ensure_loop():
     loop = asyncio.new_event_loop()
     assert ensure_loop(loop) == loop
 
-    loop = asyncio.get_event_loop()
-    assert ensure_loop(loop) != loop
+    # loop = asyncio.get_event_loop()  # TODO: don't understand! # noqa: E800
```
DavidMinarsch, line 175: @solarw this worked before but no longer...why?

aea/helpers/async_utils.py
```diff
@@ -267,9 +267,11 @@ def ensure_loop(loop: AbstractEventLoop = None) -> AbstractEventLoop:
     """
     try:
         loop = loop or asyncio.new_event_loop()
-        assert not loop.is_closed()
-        assert not loop.is_running()
-    except (RuntimeError, AssertionError):
+        if loop.is_closed():
+            ValueError("Event loop closed.")  # pragma: nocover
```
solarw, line 271: `raise ValueError`

DavidMinarsch: nice; which explains the other issue.

aea/skills/base.py
```diff
@@ -84,12 +84,13 @@ def logger(self) -> Logger:
 
     @logger.setter
     def logger(self, logger_: Logger) -> None:
-        assert self._logger is None, "Logger already set."
```
DavidMinarsch, line 87: Note; this raises an error when enabled now ....

Issue #1657, opened 2020-08-13
CLI functionality to generate multiaddress


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
Currently, multi address is generated on go side, need it on python side too

**Describe the solution you'd like**
CLI functionality to generate multiaddress

marcofavorito:
Reporting here the comment from: https://github.com/fetchai/agents-aea/pull/1695#discussion_r476244094

> As discussed, there are two ways this can be used, if called without connection_id, we simply load the private key from either private_key_paths or connection_private_key_paths (if flag provided), based on provided ledger_id and then compute the peer id using the multi address class.
> If the public id is provided, we first check if the package is present, then load the config from the connection specified, then use the config to retrieve host and port to construct the multi-address fully. You probably need another optional arg there to specify which field in config contains the url.
> We need this for the p2p_libp2p packages, so use them for orientation.

PR #1695, opened 2020-08-25
AEA-848 Add get-multiaddress command

marcofavorito:
## Proposed changes

Add get-multiaddress command, as described in detail in #1657 .

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a


aea/cli/get_multiaddress.py
```diff
@@ -0,0 +1,63 @@
...
+@click.command()
+@click.argument(
+    "ledger_id",
+    metavar="TYPE",
+    type=click.Choice(list(crypto_registry.supported_ids)),
+    required=True,
+)
+@click.option(
+    "--connection", type=PublicIdParameter(), required=False, default=None,
+)
```
DavidMinarsch, line 42: A further optional command needs to be provided: `connection_id` which must be a public id type.

aea/cli/get_multiaddress.py
```diff
@@ -0,0 +1,63 @@
...
+@click.pass_context
+@check_aea_project
+def get_multiaddress(click_context, ledger_id, connection: Optional[PublicId]):
+    """Get the address associated with the private key."""
+    address = _try_get_multiaddress(click_context, ledger_id, connection)
+    click.echo(address)
```
DavidMinarsch, line 48: As discussed, there are two ways this can be used, if called without `connection_id`, we simply load the private key from either `private_key_paths` or `connection_private_key_paths` (if flag provided), based on provided `ledger_id` and then compute the peer id using the multi address class.

If the public id is provided, we first check if the package is present, then load the `config` from the connection specified, then use the config to retrieve `host` and `port` to construct the multi-address fully. You probably need another optional arg there to specify which field in `config` contains the url.

We need this for the `p2p_libp2p` packages, so use them for orientation.

tests/test_cli/test_misc.py
```diff
@@ -58,34 +58,35 @@ def test_flag_help():
...
+  list              List the installed resources.
+  login             Login to Registry account.
+  logout            Logout from Registry account.
+  publish           Publish Agent to Registry.
+  push              Push item to Registry or save it in local packages.
+  register          Register a new Registry account.
+  remove            Remove a resource from the agent.
+  run               Run the agent.
+  scaffold          Scaffold a resource for the agent.
+  search            Search for components in the registry.
```
marcofavorito, line 90: I had to add a space to each line because `get-multiaddress` was too long

scripts/whitelist.py
```diff
@@ -41,6 +41,7 @@
 _.convert  # unused method (aea/cli/utils/click_utils.py:104)
 _.convert  # unused method (aea/cli/utils/click_utils.py:129)
 _.formatter  # unused attribute (aea/cli/utils/loggers.py:92)
+is_item_present_unified  # unused function (aea/cli/utils/package_utils:394)
```
marcofavorito, line 44: I added the method just for completeness with the analogous `get_package_path_unified`, but I didn't use it. That's why it is here; we could remove it anyway.

aea/helpers/multiaddr/base.py
```diff
@@ -102,15 +102,21 @@ def __init__(self, host: str, port: int, public_key: str):
...
 
         self._public_key = public_key
-        self._peerid = self._compute_peerid()
+        self._peerid = self.compute_peerid(self._public_key)
 
-    def _compute_peerid(self) -> str:
-        """
-        Compute base58 representation of libp2p PeerID from Bitcoin EC encoded Secp256k1 public key
+    @staticmethod
+    def compute_peerid(public_key: str) -> str:
```
marcofavorito, line 108: a slight refactoring to make `compute_peerid` usable from other parts of the code (in particular, I used it in the `get_multiaddress` implementation).

DavidMinarsch:
3 issues:
- [ ] When host and port are not present in the configs it shows the full stack trace; it should only show the clean error message (use click error handling):
Screenshot 2020-08-28 at 08 35 39

- [ ] this command is almost exclusively used for the `p2p_libp2p` connection. Obviously, we don't want to tie it to it; but we could make it more useful by looking for both `host` and `port` pair as well as `public_uri` in the config.

- [ ] -h -p options are not used to construct the full multi address
Screenshot 2020-08-28 at 08 37 46

aea/cli/get_multiaddress.py
```diff
@@ -0,0 +1,172 @@
...
+    except Exception as e:
+        raise ClickException(str(e))
+
+
+def _try_get_connection_multiaddress(
+    click_context,
+    crypto: Crypto,
+    connection_id: PublicId,
+    host_field: str,
+    port_field: str,
```
DavidMinarsch, line 135: Provided host and port not used correctly

marcofavorito: @DavidMinarsch why not?

marcofavorito:
@DavidMinarsch 
> this command is almost exclusively used for the p2p_libp2p connection. Obviously, we don't want to tie it to it; but we could make it more useful by looking for both host and port pair as well as public_uri in the config

What about having another option `-u`, `--uri` that has higher priority than `-h` and `-p` and points to the configuration field to retrieve the URI?

DavidMinarsch:
I like the `uri` option.

DavidMinarsch (approved):
LGTM

Issue #1438, opened 2020-06-30
Contract configs need to be shared with skills and connections


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
Currently, the registry registers the contract configs as well as the class.

**Describe the solution you'd like**
The configs should not be registered as they might differ for different agents in the same process. Instead; contract configs should be made available to the connections and skills so the contracts can be `made` in place with the right config.

DavidMinarsch:
Currently, this seems no longer relevant as the contracts do not have any configuration.

PR #1696, opened 2020-08-25
Move contract loading into Contract.from_config

### LINKED ISSUE #1459: Contract loading to registry revision
Currently, contracts are loaded into registry when `build` is called on the `AEABuilder`.

Instead, load a contract during call to `from_config` method of `Contract` class. This means the method has one more side effect but it already has one. It also means we avoid code duplication and make it less likely that adding to registry is forgotten.

marcofavorito:
## Proposed changes

Move contract loading into Contract.from_config

## Fixes

Fixes #1459 

## Types of changes

## Checklist

## Further comments

n/a


aea/contracts/base.py
```diff
@@ -191,3 +192,25 @@ def get_state(
         :return: the tx
         """
         raise NotImplementedError
+
+
+def _try_to_register_contract(configuration: ContractConfig):
```
DavidMinarsch, line 197: Looks like this needs to be moved elsewhere to avoid circular imports, e.g. `aea.contracts.helpers`

marcofavorito: merged #1697 in this branch. Better to merge the other PR before this one.

aea/contracts/base.py
```diff
@@ -191,3 +198,24 @@ def get_state(
...
+        logger.warning(
+            f"Skipping registration of contract {configuration.public_id} since already registered."
+        )
+        return
+    logger.debug(f"Registering contract {configuration.public_id}")  # pragma: nocover
+    try:  # pragma: nocover
+        contract_registry.register(
+            id_=str(configuration.public_id),
+            entry_point=f"{configuration.prefix_import_path}.contract:{configuration.class_name}",
+            class_kwargs={"contract_interface": configuration.contract_interfaces},
```
marcofavorito, line 215: This parameter will make the registry to populate the contract class with the attributes specified in the dictionary.
However, this cannot be overwritten by `registry.make`, so the class attributes will be set in the first registration and never changed (in the same process). This might lead to unexpected behaviours of the framework, and related to issue #1438

aea/helpers/base.py
```diff
@@ -169,44 +167,6 @@ def locate(path: str) -> Any:
     return object_
 
 
-def load_aea_package(configuration: ComponentConfiguration) -> None:
```
marcofavorito, line 172: See #1697

aea/components/base.py
```diff
@@ -98,3 +101,41 @@ def directory(self, path: Path) -> None:
         """Set the directory. Raise error if already set."""
         assert self._directory is None, "Directory already set."
         self._directory = path
+
+
+def load_aea_package(configuration: ComponentConfiguration) -> None:
```
marcofavorito, line 106: See #1697

aea/contracts/base.py
DavidMinarsch: That's a feature rather than a bug from my perspective

DavidMinarsch (commented):
Coverage checks show some errors

DavidMinarsch (approved):
LGTM

Issue #1689, opened 2020-08-22
Finish implementation and usage of multiplexer `ConnectionStatus` 


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
Currently, `ConnectionStatus` is disjoint from the status of all the individual connections.

**Describe the solution you'd like**
ConnectionStatus should be implemented fully so it reflects the global connection status:
`is_connecting` -> if at least one connection `is_connecting` and the others are `connected` or `is_connecting`
`connected` -> if and only if all connections are `connected`
`disconnected` -> if and only if all connections are `disconnected`
`disconnecting` -> if at least one connection is `disconnecting` and others are in arbitrary states

The ConnectionStatus should then be used in agent runner to ensure agent is properly torn down if key board interrupt happens during `is_connecting`

PR #1707, opened 2020-08-27
AEA-851 MultiplexerStatus

solarw:
## Proposed changes

Improvement of the multiplexer status object

## Fixes

https://github.com/fetchai/agents-aea/issues/1689

## Types of changes

## Checklist

### DIFF
--- a/aea/multiplexer.py
+++ b/aea/multiplexer.py
@@ -28,7 +28,7 @@ from aea.configurations.base import PublicId
 from aea.connections.base import Connection, ConnectionStates
 from aea.exceptions import enforce
 from aea.helpers.async_friendly_queue import AsyncFriendlyQueue
-from aea.helpers.async_utils import ThreadedAsyncRunner, cancel_and_wait
+from aea.helpers.async_utils import AsyncState, ThreadedAsyncRunner, cancel_and_wait
 from aea.helpers.logging import WithLogger
 from aea.mail.base import (
     AEAConnectionError,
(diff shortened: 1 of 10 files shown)


aea/multiplexer.py
```diff
@@ -201,30 +218,36 @@ async def connect(self) -> None:
...
                 return
             try:
+                self.connection_status.set(ConnectionStates.connecting)
                 await self._connect_all()
-                enforce(self.is_connected, "At least one connection failed to connect!")
-                self._connection_status.is_connected = True
+
+                if all(c.is_connected for c in self._connections):
+                    self.connection_status.set(ConnectionStates.connected)
+                elif not all(
```
DavidMinarsch, line 226: why not simply `else`? If not all connections are connected then it means `connect` of at least one connection has been implemented wrongly or failed so we should raise?

aea/multiplexer.py
```diff
@@ -201,30 +218,36 @@ async def connect(self) -> None:
...
         async with self._lock:
-            if not self.connection_status.is_connected:
+            if self.connection_status.is_disconnected:
                 self.logger.debug("Multiplexer already disconnected.")
                 await asyncio.wait_for(self._stop(), timeout=60)
                 return
             try:
+                self.connection_status.set(ConnectionStates.disconnecting)
                 await asyncio.wait_for(self._disconnect_all(), timeout=60)
                 await asyncio.wait_for(self._stop(), timeout=60)
```
DavidMinarsch, line 250: at the end we should check:
```
if all(c.is_disconnected for c in self._connections):
       self.connection_status.set(ConnectionStates.disconnected)
else:
     raise ...
```

aea/multiplexer.py
```diff
@@ -253,6 +276,10 @@ async def _stop(self) -> None:
             if c.state in (ConnectionStates.connecting, ConnectionStates.connected)
         ]:
             await connection.disconnect()
+
```
DavidMinarsch, line 279: I am confused how this related to `disconnect` above. Should we not be calling `disconnect` from here and implement the disconnection in `disconnect`?

solarw: updated.

aea/multiplexer.py
solarw: updated, but should discuss

aea/multiplexer.py
solarw: updated, but should discuss

aea/multiplexer.py
```diff
@@ -40,17 +40,34 @@
...
     def __init__(self):
         """Initialize the connection status."""
-        self.is_connected = False  # type: bool
-        self.is_connecting = False  # type: bool
+        super().__init__(
+            initial_state=ConnectionStates.disconnected, states_enum=ConnectionStates
+        )
+
+    @property
+    def is_connected(self) -> bool:  # pragma: nocover
```
DavidMinarsch, line 53: I think it is worth testing these properties...

DavidMinarsch (approved):
LGTM

PR #1714, opened 2020-08-30
remove callback ids from signing protocol

### LINKED ISSUE #1261: Dialogue To-Dos

DavidMinarsch:
## Proposed changes

This PR removes an inconsistency which means decision maker is tied to signing protocol.

## Fixes

#1261, 
#1514 

## Types of changes

## Checklist

## Further comments

-


aea/registries/filter.py
```diff
@@ -110,11 +109,8 @@ def handle_internal_message(self, internal_message: Optional[Message]) -> None:
         """Handlle internal message."""
         if internal_message is None:
             logger.warning("Got 'None' while processing internal messages.")
-        elif isinstance(internal_message, SigningMessage):
-            internal_message = cast(SigningMessage, internal_message)
-            self._handle_signing_message(internal_message)
-        else:
-            logger.warning("Cannot handle a {} message.".format(type(internal_message)))
+            return
+        self._handle_internal_message(internal_message)
```
DavidMinarsch, line 113: any message can now be processed here

aea/registries/filter.py
```diff
@@ -148,22 +144,22 @@ def _handle_new_handlers(self) -> None:
...
-        for skill_id in skill_callback_ids:
-            handler = self.resources.handler_registry.fetch_by_protocol_and_skill(
-                signing_message.protocol_id, skill_id,
+    def _handle_internal_message(self, message: Message):
+        """Handle message from the Decision Maker."""
+        try:
+            skill_id = PublicId.from_str(message.to)
+        except ValueError:
+            logger.warning("Invalid public id={}".format(message.to))
+            return
```
DavidMinarsch, line 153: Will eventually be done in message object directly

aea/registries/filter.py
```diff
@@ -148,22 +144,22 @@ def _handle_new_handlers(self) -> None:
...
+        handler = self.resources.handler_registry.fetch_by_protocol_and_skill(
+            message.protocol_id, skill_id,
+        )
+        if handler is not None:
+            logger.debug(
+                "Calling handler {} of skill {}".format(type(handler), skill_id)
+            )
+            handler.handle(cast(Message, message))
```
DavidMinarsch, line 161: remove unnecessary cast

PR #1724, opened 2020-09-02
Some performance check scripts

solarw:
## Proposed changes

some performance simple check scripts

## Types of changes

## Checklist


benchmark/checks/__init__.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# ------------------------------------------------------------------------------
+#
+#   Copyright 2018-2019 Fetch.AI Limited
```
DavidMinarsch, line 4: New files should be 2018-2020

benchmark/checks/check_mem_usage.py
```diff
@@ -0,0 +1,90 @@
...
+    agent.resources.add_skill(make_skill(agent, handlers={"test": TestHandler}))
+    t = Thread(target=agent.start, daemon=True)
+    t.start()
+    wait_for_condition(lambda: agent.is_running, timeout=5)
+
+    connection.enable()
+    time.sleep(duration)
+    connection.disable()
+    process = psutil.Process(os.getpid())
+    mem_usage = 1.0 * process.memory_info().rss / 1024 ** 2
```
DavidMinarsch, line 78: wrap in helper so reusable and separated concerns

benchmark/checks/check_mem_usage.py
```diff
@@ -0,0 +1,90 @@
...
+    agent.stop()
+    t.join(5)
+    rate = connection._count_in / duration
```
DavidMinarsch, line 81: wrap (see above)

benchmark/checks/check_mem_usage.py
```diff
@@ -0,0 +1,90 @@
...
+        """Handle incoming message."""
+        self.context.outbox.put(make_envelope(message.to, message.sender))
+
+
+@click.command()
+@click.option("--duration", default=3, help="Run time in seconds.")
+@click.option(
+    "--runtime_mode", default="async", help="Runtime mode: async or threaded."
+)
+def main(duration, runtime_mode):
```
DavidMinarsch, line 63: We need to run main multiple times to be able to measure the variance in the measurements as well. I'd start with 100 times.

benchmark/checks/check_reactive.py
```diff
@@ -0,0 +1,114 @@
...
+        """Handle incoming message."""
+        self.context.outbox.put(make_envelope(message.to, message.sender))
+
+
+@click.command()
+@click.option("--duration", default=3, help="Run time in seconds.")
+@click.option(
+    "--runtime_mode", default="async", help="Runtime mode: async or threaded."
+)
+def main(duration, runtime_mode):
```
DavidMinarsch, line 84: See above re running multiple times

benchmark/checks/check_reactive.py
```diff
@@ -0,0 +1,114 @@
...
+    """Test memory usage."""
+    click.echo(f"Start test for {duration} seconds in runtime mode: {runtime_mode}")
+    agent = make_agent(runtime_mode)
+    connection = TestConnection.make()
+    agent.resources.add_connection(connection)
+    agent.resources.add_skill(make_skill(agent, handlers={"test": TestHandler}))
```
DavidMinarsch, line 90: why not slightly alter `make_agent` so we can pass it the handler and connection class? So:
```
agent = make_agent(runtime_mode, TestConnection, TestHandler)
```

benchmark/checks/check_reactive.py
```diff
@@ -0,0 +1,114 @@
...
+    time.sleep(duration)
+    connection.disable()
+    agent.stop()
+    t.join(5)
+
+    latency = mean(
+        map(lambda x: x[1] - x[0], zip(connection._sends, connection._recvs,))
+    )
+    total_amount = len(connection._recvs)
+    rate = total_amount / duration
```
DavidMinarsch, line 105: extract and isolate in helpers

benchmark/checks/cosmos_private_key.txt
```diff
@@ -0,0 +1 @@
+81b0352f99a08a754b56e529dda965c4ce974edb6db7e90035e01ed193e1b7bc
```
DavidMinarsch, line 1: Add to gitignore

benchmark/checks/ethereum_private_key.txt
```diff
@@ -0,0 +1 @@
+0x6F611408F7EF304947621C51A4B7D84A13A2B9786E9F984DA790A096E8260C64
```
DavidMinarsch, line 1: See above

benchmark/checks/fetchai_private_key.txt
```diff
@@ -0,0 +1 @@
+6d56fd47e98465824aa85dfe620ad3dbf092b772abc6c6a182e458b5c56ad13b
```
DavidMinarsch, line 1: See above

benchmark/checks/utils.py
```diff
@@ -0,0 +1,186 @@
...
+        self._state.set(ConnectionStates.disconnected)
+
+    async def send(self, envelope: "Envelope") -> None:
+        """Handle incoming envelope."""
+        self._count_in += 1
+
+    async def receive(self, *args, **kwargs) -> Optional["Envelope"]:
+        """Generate an envelope."""
+        while not self._enabled:
+            await asyncio.sleep(0.0001)
```
DavidMinarsch, line 130: declare value as constant, no magic values.

Also what is the "right" value here?

benchmark/checks/check_reactive.py
solarw: It's just a math and currently used one time right here.

benchmark/checks/check_reactive.py
solarw: it's not the goal to create universal tool. also name parameter is required for handlers.
LocalConnection requires extra option: local_node, so it's not possible to pass just connection class.

benchmark/checks/utils.py
solarw: there is no right value, it's done only for specific testing purposes

benchmark/checks/ethereum_private_key.txt
solarw: removed

benchmark/checks/fetchai_private_key.txt
solarw: removed

solarw: removed

benchmark/checks/__init__.py
solarw: updated

benchmark/checks/utils.py
```diff
@@ -0,0 +1,222 @@
...
+
+    :return: list of tuples of results
+    """
+    results = []
+    for _ in range(num_runs):
+        p = Pool(1)
+        results.append(p.apply(fn, tuple(args)))
+        p.terminate()
+        del p
+    mean_values = map(mean, zip(*(map(lambda x: x[1], i) for i in results)))
```
DavidMinarsch, line 222: Variance and standard deviation should also be reported

DavidMinarsch: Please add rounding to ensure fixed decimal places

benchmark/checks/check_mem_usage.py
```diff
@@ -0,0 +1,101 @@
...
+    def teardown(self) -> None:
+        """Noop teardown."""
+
+    def handle(self, message: Message) -> None:
+        """Handle incoming message."""
+        self.context.outbox.put(make_envelope(message.to, message.sender))
+
+
+def run(duration, runtime_mode):
+    """Check memory usage."""
```
DavidMinarsch, line 59: For the memory usage; can we add the reactive and proactive here? So this would also run the behaviour at the same time.

benchmark/checks/check_reactive.py
```diff
@@ -0,0 +1,125 @@
...
+        ("rate: {} envelopes/second", rate),
+    ]
+
+
+@click.command()
+@click.option("--duration", default=1, help="Run time in seconds.")
+@click.option(
+    "--runtime_mode", default="async", help="Runtime mode: async or threaded."
+)
+@click.option("--number_of_runs", default=10, help="How many times run teste.")
```
DavidMinarsch, line 113: Can we add an option to run the connection synced versus not synced?

marcofavorito:
Looks good, thank you. As a "nice to have":
- show the progress of an experiment (to answer questions like "at which run are we?"). Either simple logging/printing (with timestamp would be great), or something like https://github.com/tqdm/tqdm
- allow for saving the output to files. Both JSON and CSV have their advantages/disadvantages; I think CSV is good because lets you append data easily, while you have to load the entire JSON before modifying it; on the other hand, JSON is more flexible and could let you to store configuration information in a specific object.

Of course, both features must not be invasive in terms of measurements of performances.

benchmark/run_benchmark.sh
```diff
@@ -0,0 +1,71 @@
...
+REPO=https://github.com/fetchai/agents-aea.git
+BRANCH=feature/benchmarks
+TMP_DIR=$(mktemp -d -t bench-XXXXXXXXXX)
+git clone --branch $BRANCH $REPO $TMP_DIR
+
+CURDIR=`pwd`
+cd $TMP_DIR
+pip install --upgrade aea
+pipenv lock --requirements > requirements.txt
+pip install -r requirements.txt
```
DavidMinarsch, line 11: ? why

solarw: cause it uses `psutil = "==5.7.0"` that is part of dev requirements
so it's better to install aea and all development requirements

DavidMinarsch (approved):
LGTM

Issue #1738, opened 2020-09-11
Local node shows connection errors on sending or receiving


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
When sending or receiving envelopes, the local node occasionally shows disconnection errors.
```
[agent119] Error in the sending loop: Connection is not connected! Connect first!
Exception in callback Future.set_result(True)
handle: <Handle Future.set_result(True)>
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/asyncio/events.py", line 81, in _run
    self._context.run(self._callback, *self._args)
asyncio.exceptions.InvalidStateError: invalid state
```

**Describe the solution you'd like**
This can be investigated by running `run_benchmark.sh`.

We need to identify the issue and fix it to reliably run benchmarks.

solarw:
it's a multiplexer's issue.  disconnect order. it disconnects connections first while send loop tries to use connection already disconnected.

PR #1744, opened 2020-09-15
[AEA-903] multiplexer cancel recv send loops before disconnecting connections

solarw:
## Proposed changes

multiplexer cancel recv send loops before disconnecting connections

## Fixes

https://github.com/fetchai/agents-aea/issues/1738

fixes issue when multiplexer tries to handle messages in queues but connections alraedy disconnected.
now it stops message processing loops and disconnect connections

## Types of changes

## Checklist

### DIFF
--- a/aea/multiplexer.py
+++ b/aea/multiplexer.py
@@ -272,8 +272,8 @@ class AsyncMultiplexer(WithLogger):
                 self.logger.exception("Exception on disconnect:")
                 raise AEAConnectionError("Failed to disconnect the multiplexer.")
 
-    async def _stop_recv_send_loops(self) -> None:
-        """Stop recv and send loops."""
+    async def _stop_receive_send_loops(self) -> None:
+        """Stop receive and send loops."""
         self.logger.debug("Stopping recv loop...")
         if self._recv_loop_task:
             self._recv_loop_task.cancel()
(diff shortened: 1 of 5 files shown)


aea/multiplexer.py
```diff
@@ -263,17 +263,36 @@ async def disconnect(self) -> None:
...
             try:
                 self.connection_status.set(ConnectionStates.disconnecting)
-                await asyncio.wait_for(self._disconnect_all(), timeout=60)
                 await asyncio.wait_for(self._stop(), timeout=60)
                 self.logger.debug("Multiplexer disconnected.")
             except (CancelledError, Exception):
                 self.logger.exception("Exception on disconnect:")
                 raise AEAConnectionError("Failed to disconnect the multiplexer.")
 
+    async def _stop_recv_send_loops(self) -> None:
```
DavidMinarsch, line 275: Please don't abbreviate `recv` ..it is inconsistent with the code base (spelling things out!) and confusing. Relabel `receive`.

DavidMinarsch (approved):
LGTM, except some naming issues

DavidMinarsch:
Please also add a test that ensures the cancel send receive happens before disconnect so we don't reintroduce the problem when refactoring

aea/multiplexer.py
solarw: done

solarw:
> Please also add a test that ensures the cancel send receive happens before disconnect so we don't reintroduce the problem when refactoring

done

DavidMinarsch (approved):
LGTM

aea/multiplexer.py
```diff
@@ -263,41 +263,58 @@ async def disconnect(self) -> None:
         async with self._lock:
             if self.connection_status.is_disconnected:
                 self.logger.debug("Multiplexer already disconnected.")
-                await asyncio.wait_for(self._stop(), timeout=60)
```
lrahmani, line 266: @solarw why was this needed in the first place?
Just making the observation in case this was here for a reason (to handle an edge case for e.g.)

I understand that in the previous version disconnecting all connections and then cancelling background coroutines  *in this order* was problematic. However they were still both executed after a call of `disconnect()`.

lrahmani (commented):
LGTM, just one question

aea/multiplexer.py
solarw: @lrahmani i think there is no reason to call stop second time if already disconnected, this produced me an error in tests after the refactoring. now all the tests are passing ok, so not sure it's required anymore.

PR #1747, opened 2020-09-15
Update package version script 2

marcofavorito:
## Proposed changes

Apply minor fixes:
- avoid useless removal of newline
- make items ambiguous only by public id prefix.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments
n/a


scripts/update_package_versions.py
```diff
@@ -292,7 +298,9 @@ def _get_ambiguous_public_ids(
             operator.itemgetter(0),
             filter(
                 lambda x: x[1] > 1,
-                Counter(id_.public_id for id_ in all_package_ids_to_update).items(),
+                Counter(
+                    _extract_prefix(id_.public_id) for id_ in all_package_ids_to_update
```
marcofavorito, line 302: now two public ids are ambiguous if they have the same prefix. Otherwise, it might lead to wrong updates (e.g. public ids updated twice if the conflicting ids have consecutive semantic versions (e.g. `fetchai/ml_data_provider:0.10.0` and `fetchai/ml_data_provider:0.11.0`).

marcofavorito: alternative: detect this corner case, and process the public ids from the greatest version to the smallest.

scripts/update_package_versions.py
```diff
@@ -403,7 +415,7 @@ def _ask_to_user_and_replace_if_allowed(content, old_string, new_string, type_)
         content = content.replace(old_string, new_string)
         return content
 
-    lines = content.splitlines()
+    lines = content.splitlines(keepends=True)
```
marcofavorito, line 418: this will solve the disappearing `\n` at the end of the files

scripts/update_package_versions.py
DavidMinarsch: No; we cannot do that because processing order MUST be determined by the package type; not the package name!

DavidMinarsch (approved):
LGTM

PR #1750, opened 2020-09-16
[DO NOT MERGE] load_aea_package do not replace modules if exist

solarw:
## Proposed changes

fix for existing module replacement in load_aea_package.

## Fixes

fix for existing module replacement in load_aea_package.
cause it replaces already imported module

## Types of changes

## Checklist


DavidMinarsch:
@marcofavorito says:

it means only the former module is preserved

this might lead to inconsistencies if the prefixes for some reasons are the same in general but there are submodules not contained in the former package

anyway, as I often said, the long-term solution is to patch at runtime the import system properly 

Hence I am closing this.

PR #1751, opened 2020-09-16
AEA-904 soef generic command. integration tests added

solarw:
## Proposed changes

]soef generic command. integration tests added

## Types of changes

## Checklist


packages/fetchai/connections/soef/connection.py
```diff
@@ -506,6 +514,47 @@ async def _ping_handler(self, service_description: Description) -> None:
...
+        self._check_data_model(service_description, ModelNames.generic_command)
+        command = service_description.values.get("command", None)
+        params = service_description.values.get("parameters", {})
+        if params:
+            params = urllib.parse.parse_qs(params)
+
+        content = await self._generic_oef_command(command, params)
+
+        message = oef_search_dialogue.reply(
+            performative=OefSearchMessage.Performative.SUCCESS,
```
DavidMinarsch, line 543: Success response should be provided for register and unregister too!

packages/fetchai/connections/soef/connection.py
```diff
@@ -506,6 +514,47 @@ async def _ping_handler(self, service_description: Description) -> None:
...
+            target_message=oef_message,
+            service_description=Description(
```
DavidMinarsch, line 545: This needs to be `agent_info=AgentInfo(...)`

DavidMinarsch: But I don't understand why the info is being sent back to the skill...

DavidMinarsch: Shouldn't you return the info returned from the soef call?

DavidMinarsch (approved):
LGTM

Issue #1727, opened 2020-09-04
Revise maximum length of string types.


marcofavorito:
**Is your feature request related to a problem? Please describe.**
Currently, we don't impose maximum length for relevant strings like author name, package name, version string. This may cause subtle corner cases.

**Describe the solution you'd like**

Better validation of data in Python code, e.g. configuration, Protobuf messages etc. To elaborate.

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

PR #1761, opened 2020-09-21
CLI reset_password command implemented AEA-887

panasevychol:
## Proposed changes

CLI reset_password command implemented AEA-887

## Types of changes

## Checklist

### DIFF
--- a/aea/cli/reset_password.py
+++ b/aea/cli/reset_password.py
@@ -16,32 +16,28 @@
 #   limitations under the License.
 #
 # ------------------------------------------------------------------------------
-"""Registry utils used for CLI login command."""
 
-from aea.cli.registry.utils import request_api
+"""Implementation of the 'aea reset_password' subcommand."""
 
+import click
 
-def registry_login(username: str, password: str) -> str:
-    """
-    Login into Registry account.
-
-    :param username: str username.
-    :param password: str password.
-
-    :return: str token
-    """
-    resp = request_api(
-        "POST", "/rest-auth/login/", data={"username": username, "password": password}
-    )
-    return resp["key"]
+from aea.cli.registry.login import registry_reset_password
 
 
-def registry_reset_password(email: str) -> None:
+@click.command(name="reset_password", help="Reset password of Registry account.")
+@click.argument("email", type=str, required=True)
+def reset_password(email):
+    """Command to request Registry to reset password."""
+    _do_password_reset(email)
+
+
+def _do_password_reset(email: str) -> None:
     """
     Request Registry to reset password.
 
-    :param email: user email.
+    :param email: str email.
 
-    :return: None.
+    :return:
     """
-    request_api("POST", "/rest-auth/password/reset/", data={"email": email})
+    registry_reset_password(email)
+    click.echo("A letter with password reset link was sent to {}".format(email))
--- a/aea/cli/reset_password.py
+++ b/aea/cli/reset_password.py
@@ -40,4 +40,4 @@ def _do_password_reset(email: str) -> None:
     :return:
     """
     registry_reset_password(email)
-    click.echo("A letter with password reset link was sent to {}".format(email))
+    click.echo("An email with a password reset link was sent to {}".format(email))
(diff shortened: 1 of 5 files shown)


aea/cli/reset_password.py
```diff
@@ -0,0 +1,43 @@
...
+def _do_password_reset(email: str) -> None:
+    """
+    Request Registry to reset password.
+
+    :param email: str email.
+
+    :return:
+    """
+    registry_reset_password(email)
+    click.echo("A letter with password reset link was sent to {}".format(email))
```
DavidMinarsch, line 43: change to: `An email with a password reset link was sent to {}`

DavidMinarsch (approved):
LGTM, minor modification required before merging

tests/test_cli/test_misc.py
```diff
@@ -85,6 +85,7 @@ def test_flag_help():
   push              Push item to Registry or save it in local packages.
   register          Register a new Registry account.
   remove            Remove a resource from the agent.
+  reset_password    Reset password of Registry account.
```
marcofavorito, line 88: @DavidMinarsch @panasevychol in the long term, I propose having a subcommand `auth` with

aea/cli/reset_password.py
```diff
@@ -0,0 +1,43 @@
...
+"""Implementation of the 'aea reset_password' subcommand."""
+
+import click
+
+from aea.cli.registry.login import registry_reset_password
+
+
+@click.command(name="reset_password", help="Reset password of Registry account.")
+@click.argument("email", type=str, required=True)
+def reset_password(email):
```
marcofavorito, line 29: why the email? And why not the `username`? Why not supporting both?

aea/cli/reset_password.py
```diff
@@ -0,0 +1,43 @@
...
+
+def _do_password_reset(email: str) -> None:
+    """
+    Request Registry to reset password.
+
+    :param email: str email.
+
+    :return:
+    """
+    registry_reset_password(email)
```
marcofavorito, line 42: The following command:
```
aea reset_password this-is-a-wrong-email   
```

Gave me:
```
Error: {'email': ['Enter a valid email address.']}
```

That means the request was sent, with a wrong email.
In my opinion, the validation of the content of `email` should be done locally, and not delegated to the remote APIs. 

- First of all, because the command would be more responsive. 
- Also, because we give the right feedback to the user; the command failed because the input was wrong, not for connectivity issues. 
- Although not a problem at the current state of the situation, we reduce some avoidable workload to the server.
- I also think it's a better approach on the security side of things. We allow requests with arbitrary content as arguments, which may break both the client and the server unexpectedly (of course we won't solve all the security risks on the server-side, but at least the official clients behave correctly). This is part of a broader set of issues, some of them included into #1727.

Anyway, I don't think it's a priority, but I suggest to keep track of it in that case.

marcofavorito:
```
aea reset_password marco.favorito@gmail.com
```
gave me
```
Error: Registry internal server error.
```

Could you investigate that?

aea/cli/reset_password.py
panasevychol: Simply because:
- it's what auth package requires from us
- you receive a letter on your email for password recovery. So it's not required to remember a username but if you do remember a username how would you know which email will receive the letter.
- it's like a standard, email is used everywhere for that.

panasevychol:
> ```
> aea reset_password marco.favorito@gmail.com
> ```
> 
> gave me
> 
> ```
> Error: Registry internal server error.
> ```
> 
> Could you investigate that?

It's very simple - it's not deployed to prod yet :) Gonna be there soon. If you want to test it you can try it on Staging, it works there.

aea/cli/reset_password.py
panasevychol: @marcofavorito can I ask you for assistance with that? You've implemented a PublicIdParameter that validates Public IDs, can you do the same for email? In separate ticket/pr

tests/test_cli/test_misc.py
panasevychol: I agree. We can have a ticket for that. If we have plans to add more auth related commands it would be nice to organize that way.

aea/cli/reset_password.py
panasevychol: Done.

aea/cli/reset_password.py
marcofavorito: Yes, sure

marcofavorito (approved):
LGTM

PR #1762, opened 2020-09-21
Update version bumper script

marcofavorito:
## Proposed changes

- shrink the set of ambiguous public ids. The idea is to
  overcome the "multiple-update", due to versions with consecutive minor numbers,
  by sorting the package id in descending order on the version number. May lead to other unforeseen issues.
- update the yamls independently of whether the package id is ambiguous or not,
  since in that case we always disambiguate thanks to the context;
- update protocol's READMEs, and in particular the protocol specifications,
  separately, by disambiguating from the path (the parent of the parent must be 'protocols').

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a


scripts/update_package_versions.py
```diff
@@ -260,66 +289,79 @@ def get_public_ids_to_update() -> Set[PackageId]:
             deployed_public_id = public_id_in_registry(type_, name)
             difference = minor_version_difference(current_public_id, deployed_public_id)
             # check if the public ids of the local package and the package in the registry are already the same.
+            package_info = f"Package `{name}` of type `{type_}`"
```
marcofavorito, line 292: this is just for aesthetic reasons.

scripts/update_package_versions.py
```diff
@@ -260,66 +289,79 @@ def get_public_ids_to_update() -> Set[PackageId]:
...
-    """Extract (author, package_name) from public id."""
-    return public_id.author, public_id.name
-
-
-def _get_ambiguous_public_ids(
-    all_package_ids_to_update: Set[PackageId],
-) -> Set[PublicId]:
+def _get_ambiguous_public_ids() -> Set[PublicId]:
     """Get the public ids that are the public ids of more than one package id."""
+    all_package_ids = get_all_package_ids()
```
marcofavorito, line 308: this is important... ambiguous public ids must be computed from all the packages in the repo, not just the one being updated.

DavidMinarsch (approved):
LGTM

PR #1771, opened 2020-09-24
AEA-913 Skill Test Tool

5A11:
DRAFT

This PR adds a new test tool for skills.

## Types of changes

## Checklist


tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
```diff
@@ -0,0 +1,163 @@
...
+    path_to_skill = Path(ROOT_DIR, "packages", "fetchai", "skills", "generic_buyer")
+
+    def test_search_behaviour_setup_a(self):
+        """Test the setup method of the search behaviour."""
+        outbox = self.skill.skill_context.outbox
+        search_behaviour = cast(
+            GenericSearchBehaviour, self.skill.skill_context.behaviours.search
+        )
+
+        assert outbox.empty()
```
DavidMinarsch, line 49: In some sense this check is unnecessary as the test tool guarantees this

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
```diff
@@ -0,0 +1,163 @@
...
+
+    def test_search_behaviour_setup_b(self):
+        """Test the setup method of the search behaviour."""
+        outbox = self.skill.skill_context.outbox
+        strategy = cast(GenericStrategy, self.skill.skill_context.strategy)
+        strategy._is_ledger_tx = False
+        search_behaviour = cast(
+            GenericSearchBehaviour, self.skill.skill_context.behaviours.search
+        )
+        assert outbox.empty()
```
DavidMinarsch, line 73: see above

aea/test_tools/test_skill.py
```diff
@@ -0,0 +1,85 @@
...
+        envelope = self._multiplexer.out_queue.get_nowait()
+        return envelope.message
+
+    @classmethod
+    def setup(cls) -> None:
+        """Set up the skill test case."""
+        identity = Identity("test_agent_name", "test_agent_address")
+
+        cls._multiplexer = AsyncMultiplexer()
+        cls._multiplexer._out_queue = asyncio.Queue()
```
DavidMinarsch, line 69: Multiplexer can be removed here if we instantiate the out_queue in the outbox..can be trivially done but should be a separate PR (i.e. on connect call multiplexer calls outbox to set out_queue)

aea/test_tools/test_skill.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def get_message_from_outbox(self) -> Optional[Message]:
+        """Get message from outbox."""
+        if self._outbox.empty():
+            return None
+        envelope = self._multiplexer.out_queue.get_nowait()
+        return envelope.message
+
+    @classmethod
+    def setup(cls) -> None:
```
5A11, line 64: As we talked about before, a test might change the state of the loaded skill and mess up the execution of other tests (e.g. `test_search_behaviour_setup_b` below changes `strategy._is_ledger_tx`).

The code block here in `setup` does not get called before every test function does it? 
If no, perhaps this code block can be moved into a autouse pytest fixture?

marcofavorito: > The code block here in setup does not get called before every test function does it?
Correct. It gets called only once for all the tests in the test class.

>  If no, perhaps this code block can be moved into a autouse pytest fixture?
I think the burden to avoid side-effects, or to restore the original state, can be left to the developer. It is more in line with `pytest` test classes behaviour.

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
marcofavorito: It depends on what we decide (see the discussion above)
- If we set up the skill at the beginning of each test, then yes, that's true.
- If not, then that line works as a "sanity check". But I'd say it's not best practice, since if tests would be well written there shouldn't be the need for it.

aea/test_tools/test_skill.py
DavidMinarsch: My bad, I was under the impression it would be reset on each method call. Of course it needs to be a clean skill state for each test, that's the whole point..will fix. I disagree that the burden should be on the developer. The skill has too many side-effects. It will render the whole tool largely pointless...

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
DavidMinarsch: See below; I am in favour of clean skill reset for each test method.

aea/test_tools/test_skill.py
marcofavorito: My bad; `setup` seems to be called for each method; although I couldn't find confirmation in the relevant documentation:
https://docs.pytest.org/en/2.8.7/xunit_setup.html#class-level-setup-teardown

aea/test_tools/test_skill.py
DavidMinarsch, line 69:
```suggestion
        cls._multiplexer._out_queue = asyncio.Queue()  # pylint: disable=protected-access
```

aea/test_tools/test_skill.py
```diff
@@ -0,0 +1,85 @@
...
+        """Get the skill."""
+        try:
+            value = self._skill
+        except AttributeError:
+            raise ValueError("Ensure skill is set during setup_class.")
+        return value
+
+    def get_quantity_in_outbox(self) -> int:
+        """Get the quantity of envelopes in the outbox."""
+        return self._outbox._multiplexer.out_queue.qsize()
```
DavidMinarsch, line 54:
```suggestion
        return self._multiplexer.out_queue.qsize()
```

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
```diff
@@ -0,0 +1,389 @@
...
+        assert type(actual_message) == DefaultMessage
+        assert actual_message.performative == expected_performative
+        assert actual_message.to == expected_counterparty
+        assert actual_message.error_code == expected_error_code
+        assert actual_message.error_msg == expected_error_msg
+        assert actual_message.error_data == expected_error_data
+
+    def test_fipa_handler_handle_propose(self):
+        """Test the _handle_propose method of the fipa handler."""
+        # ToDo need to mock affordable and acceptable values
```
5A11, line 183: This is an example that shows mocking is needed. Need to be able to mock `is_acceptable_proposal` and `is_affordable_proposal`, otherwise it would be too much manual configuration to ensure those values are what you want, also making the test not independent of the skill's configuration yaml.

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
```diff
@@ -0,0 +1,389 @@
...
+            end_state_numbers,
+        ) in fipa_dialogues.dialogue_stats.self_initiated.items():
+            if end_state == FipaDialogue.EndState.DECLINED_ACCEPT:
+                assert end_state_numbers == 1
+            else:
+                assert end_state_numbers == 0
+
+    def test_fipa_handler_handle_match_accept(self):
+        """Test the _handle_match_accept method of the fipa handler."""
+        # ToDo does not work; fails on dialogues.update(). Investigate why.
```
5A11, line 317: This test does not work (failing on dialogues.update() line 363).

5A11: It also shows that for some unit tests, a lot of setup is needed to get some part of a skill code working. For instance to get `handle_match_accept` working, you need to pass it a dialogue containing all other messages prior to the `match_accept`  (cfp, propose, accept).

5A11: This highlights the issue that for complex skills, to unit test a part of the code that depends on a large state, that whole state has to be replicated to get the code working. 

Maybe we can create something to make that easy? For this test, an easy way of populating a dialogue object, and given heavy reliance on dialogues, this would be useful across tests.

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
```diff
@@ -0,0 +1,381 @@
...
+        res_bool, res_str = self.message_has_attributes(
+            actual_message=self.get_message_from_outbox(),
+            message_type=LedgerApiMessage,
+            performative=LedgerApiMessage.Performative.GET_BALANCE,
+            to=str(LEDGER_PUBLIC_ID),
+            sender=self.skill.skill_context.agent_address,
+            ledger_id=FETCHAI,
+            address=self.skill.skill_context.agent_address,
+        )
+        assert res_bool is True, res_str
```
DavidMinarsch, line 73:
```suggestion
        assert res_bool, res_str
```

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
```diff
@@ -0,0 +1,381 @@
...
+
+        # before
+        assert self.get_quantity_in_outbox() == 0
+
+        # operation
+        search_behaviour.setup()
+
+        # after
+        assert self.get_quantity_in_outbox() == 1
+        res_bool, res_str = self.message_has_attributes(
```
DavidMinarsch, line 64:
```suggestion
        has_attributes, error_str = self.message_has_attributes(
```

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
```diff
@@ -0,0 +1,381 @@
...
+
+    def test_search_behaviour_setup_is_ledger_tx(self):
+        """Test the setup method of the search behaviour where is_ledger_tx is True."""
+        # setup
+        search_behaviour = cast(
+            GenericSearchBehaviour, self.skill.skill_context.behaviours.search
+        )
+
+        # before
+        assert self.get_quantity_in_outbox() == 0
```
DavidMinarsch, line 57: drop, unnecessary

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
```diff
@@ -0,0 +1,381 @@
...
+        """Test the act method of the search behaviour where is_searching is True."""
+        # setup
+        search_behaviour = cast(
+            GenericSearchBehaviour, self.skill.skill_context.behaviours.search
+        )
+        strategy = cast(GenericStrategy, self.skill.skill_context.strategy)
+        strategy._is_searching = True
+
+        # before
+        assert self.get_quantity_in_outbox() == 0
```
DavidMinarsch, line 103: see above

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
```diff
@@ -0,0 +1,381 @@
...
+
+        res_bool, res_str = self.message_has_attributes(
+            actual_message=self.get_message_from_outbox(),
+            message_type=OefSearchMessage,
+            performative=OefSearchMessage.Performative.SEARCH_SERVICES,
+            to=self.skill.skill_context.search_service_address,
+            sender=self.skill.skill_context.agent_address,
+            query=self.skill.skill_context.strategy.get_location_and_service_query(),
+        )
+        assert res_bool is True, res_str
```
DavidMinarsch, line 119: see above

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
```diff
@@ -0,0 +1,381 @@
...
+        # setup
+        strategy = cast(GenericStrategy, self.skill.skill_context.strategy)
+        strategy._is_searching = False
+        search_behaviour = cast(
+            GenericSearchBehaviour, self.skill.skill_context.behaviours.search
+        )
+
+        # before
+        assert self.get_quantity_in_outbox() == 0
+
```
DavidMinarsch, line 132: see above

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
```diff
@@ -0,0 +1,381 @@
...
+        # setup
+        fipa_handler = cast(GenericFipaHandler, self.skill.skill_context.handlers.fipa)
+        incoming_message = self.build_incoming_message(
+            message_type=FipaMessage,
+            dialogue_reference=("1", "0"),
+            performative=FipaMessage.Performative.ACCEPT,
+        )
+
+        # before
+        assert self.get_quantity_in_outbox() == 0
```
DavidMinarsch, line 156: see above

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
```diff
@@ -0,0 +1,381 @@
...
+
+        # operation
+        search_behaviour.setup()
+
+        # after
+        assert self.get_quantity_in_outbox() == 1
```
DavidMinarsch, line 63:
```suggestion
        assert self.get_quantity_in_outbox() == 1, "No message in outbox."
```

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
```diff
@@ -0,0 +1,381 @@
...
+            message_type=FipaMessage,
+            dialogue_reference=("1", "0"),
+            message_id=2,
+            target=1,
+            performative=FipaMessage.Performative.PROPOSE,
+            proposal=proposal,
+        )
+
+        # before
+        assert self.get_quantity_in_outbox() == 0
```
DavidMinarsch, line 200: see above

aea/test_tools/test_skill.py
```diff
@@ -0,0 +1,171 @@
...
+    def get_message_from_outbox(self) -> Optional[Message]:
+        """Get message from outbox."""
+        if self._outbox.empty():
+            return None
+        envelope = self._multiplexer.out_queue.get_nowait()
+        return envelope.message
+
+    @staticmethod
+    def message_has_attributes(
+        actual_message: Message, message_type: Type[Message], **kwargs,
```
DavidMinarsch, line 67: See if you can infer the type from the performative provided. Then all that's needed is the performative as a required argument, reducing one more line of code

aea/test_tools/test_skill.py
```diff
@@ -0,0 +1,171 @@
...
+        :param message_id: the message_id
+        :param target: the target
+        :param performative: the performative
+        :param to: the 'to' address
+        :param sender: the 'sender' address
+        :param kwargs: other attributes
+
+        :return: the created incoming message
+        """
+        d = dict()
```
DavidMinarsch, line 130: single value variables are discouraged, use `body = {}` for instance

aea/test_tools/test_skill.py
```diff
@@ -0,0 +1,171 @@
...
+                    )
+
+        return True, "The message has the provided expected attributes."
+
+    def build_incoming_message(
+        self,
+        message_type: Type[Message],
+        dialogue_reference: Optional[Tuple[str, str]] = None,
+        message_id: Optional[int] = None,
+        target: Optional[int] = None,
```
DavidMinarsch, line 108: why not give them defaults? 1, 0 respectively?

aea/test_tools/test_skill.py
```diff
@@ -0,0 +1,171 @@
...
+        performative: Optional[Message.Performative] = None,
```
DavidMinarsch, line 109: performative seems to be a required rather than optional argument

aea/test_tools/test_skill.py
```diff
@@ -0,0 +1,171 @@
...
+                            expected_value,
+                        ),
+                    )
+
+        return True, "The message has the provided expected attributes."
+
+    def build_incoming_message(
+        self,
+        message_type: Type[Message],
+        dialogue_reference: Optional[Tuple[str, str]] = None,
```
DavidMinarsch, line 106: An incoming message means that the dialogue is held somewhere outside the skill. So you can simply generate the dialogue reference inside the method rather than have it passed as an argument.

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
```diff
@@ -0,0 +1,381 @@
...
+
+        # operation
+        fipa_handler._handle_propose(
+            cast(FipaMessage, incoming_message),
+            FipaDialogue(
+                DialogueLabel(("1", "1"), "counterparty", "me"),
+                "me",
+                FipaDialogue.Role.BUYER,
+            ),
+        )
```
DavidMinarsch, line 210: If you feed the right arguments (i.e. the correctly defined incoming message) to the `handle` method then the skill will take care of this. Don't repeat work the skill already does. So instead, create a propose like message and feed it to the `handle` method. To ensure the skill's dialogue is at the correct stage, we can either mock the skill's dialogue or manually "prep" it with another helper method.

DavidMinarsch (changes requested):
Comment on current state, looks good so far

aea/test_tools/test_skill.py
5A11: If we change those default values then we should change these too. 
With None, whatever those defaults values are will be used.

aea/test_tools/test_skill.py
5A11: I thought in the most generic case, one might just want to check the message is of a certain type (of course if you want to assert the performative, you can still do it).

aea/test_tools/test_skill.py
5A11: Sometimes you want to create an incoming message with a specific incorrect dialogue reference (e.g. testing `_handle_unidentified_dialogue()`). This accommodates this kind of test.

If the dialogue_reference isn't passed, the default ("", "") will be used.

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
5A11: Yep, I am working on the quick dialogue builder.

aea/test_tools/test_skill.py
DavidMinarsch: ok, but then don't fill them in in your tests...unnecessary loc

aea/test_tools/test_skill.py
DavidMinarsch: In the current case I always have to provide the message_type as it is not optional..see your tests, two lines are needed where we could do with one

aea/test_tools/test_skill.py
DavidMinarsch: The default dialogue reference will not be accepted for addition in the receiving dialogue as it is an invalid reference.

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
5A11: Done

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
5A11: Done

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
5A11: These line (and the similar ones below) is not strictly needed, but it makes it explicit for the reader what this function is testing, i.e. there should be 0 outgoing message before the test, and 1 after.

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
5A11: Done

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
5A11: Done

aea/test_tools/test_skill.py
5A11: Done

aea/test_tools/test_skill.py
5A11: okI change it so if no dialogue_reference is provided, a new correct reference is used.

aea/test_tools/test_skill.py
5A11: Sorry my above reply was for your other comment on the `message_has_attributes`.

I make performative non-optional.

aea/test_tools/test_skill.py
5A11: I couldn't immediately find a solution for this. I posted a question online, if I get an answer I'll apply it.

aea/protocols/dialogue/base.py
```diff
@@ -611,6 +611,51 @@ def reply(
 
         return reply
 
+    def reply_with_target(
```
5A11, line 614: An overload of reply which instead of specifying the target message, you just specify the target message_id.

aea/test_tools/test_skill.py
```diff
@@ -127,23 +133,110 @@ def build_incoming_message(
...
+            new_message = (
+                (False, *message) if last_incoming is True else (True, *message)
+            )
+        if type(new_message[1]) != int:  # pylint: disable=unidiomatic-typecheck
+            new_message_list = list(new_message)
+            new_message_list.insert(1, message_id - 1)
+            new_message = tuple(new_message_list)
+        return new_message
+
+    def build_dialogue(
```
5A11, line 169: quick dialogue builder

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
```diff
@@ -311,53 +311,21 @@ def test_fipa_handler_handle_match_accept(self):
...
-            proposal=Description(
-                {
-                    "ledger_id": "some_ledger_id",
-                    "price": "some_price",
-                    "currency_id": "some_currency_id",
-                    "service_id": "some_service_id",
-                    "quantity": "some_quantity",
-                    "tx_nonce": "some_tx_nonce",
-                }
+        fipa_dialogue = self.build_dialogue(
```
5A11, line 316: Example of a quick dialogue

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
```diff
@@ -311,53 +311,21 @@ def test_fipa_handler_handle_match_accept(self):
         fipa_handler = cast(GenericFipaHandler, self.skill.skill_context.handlers.fipa)
         strategy = cast(GenericStrategy, self.skill.skill_context.strategy)
         strategy._is_ledger_tx = False
-        counterparty_address = "counterparty"
 
         fipa_dialogues = cast(FipaDialogues, self.skill.skill_context.fipa_dialogues)
```
5A11, line 315: Example of a quick dialogue >

aea/protocols/dialogue/base.py
DavidMinarsch: DRY! don't repeat all the code please...

Also, on review, why not simply add another optional argument to `reply`? This is not a neat implementation at all at present.

aea/test_tools/test_skill.py
```diff
@@ -0,0 +1,264 @@
...
+        :return: the created incoming message
+        """
+        message_attributes = dict()
+        message_attributes["dialogue_reference"] = Dialogues._generate_dialogue_nonce()  # pylint: disable=protected-access
+        for arg_name in inspect.getfullargspec(self.build_incoming_message)[0][1:]:
+            if (
+                arg_name not in {"message_type", "to", "sender"}
+                and locals()[arg_name] is not None
+            ):
+                message_attributes[arg_name] = locals()[arg_name]
```
DavidMinarsch, line 143: I don't like the usage of locals...can lead to hard to debug errors if the method is extended. Also, pylint above shows you're not using the keyword arguments..just use them properly and avoid using locals.

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
```diff
@@ -0,0 +1,349 @@
...
+        fipa_dialogue = self.build_dialogue(
+            dialogues=fipa_dialogues,
+            messages=(
+                (FipaMessage.Performative.CFP, {"query": "some_query"}),
+                (FipaMessage.Performative.PROPOSE, {"proposal": "some_proposal"}),
+                (FipaMessage.Performative.ACCEPT, {}),
+                (
+                    FipaMessage.Performative.MATCH_ACCEPT_W_INFORM,
+                    {"info": {"address": "some_term_sender_address"}},
+                ),
```
DavidMinarsch, line 325: Why four messages; why not just provide the last? Then you might have to use the private methods of the dialogue models but that's much preferable than unnecessarily bloating the test here...

aea/test_tools/test_skill.py
```diff
@@ -0,0 +1,264 @@
...
+        dialogues: Dialogues,
+        messages: Tuple[Tuple[Any, ...]],
+        counterparty: Address = "counterparty",
+    ) -> Dialogue:
+        """
+        Quickly create a dialogue.
+
+        The 'messages' argument is a tuple of "compact messages".
+        Each compact message is represented as a tuple (inc, target, performative, contents), where
+         - 'inc' is boolean, representing whether the message is incoming (True), or outgoing (False).
```
DavidMinarsch, line 180: I don't think creating these "compact messages" is a good idea. It is yet another concept which has to be understood and maintained. The developer might as well just provide the message object at this point...

DavidMinarsch (changes requested):

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
```diff
@@ -0,0 +1,349 @@
...
+        )
+        assert has_attributes, error_str
+
+    def test_fipa_handler_handle_propose(self):
+        """Test the _handle_propose method of the fipa handler."""
+        # ToDo need to mock affordable and acceptable values
+        # setup
+        fipa_handler = cast(GenericFipaHandler, self.skill.skill_context.handlers.fipa)
+        fipa_dialogues = cast(FipaDialogues, self.skill.skill_context.fipa_dialogues)
+        fipa_dialogue = self.build_dialogue(
```
DavidMinarsch, line 181: How about
```
self.prepare_dialogue(fipa_dialogues, expected_next_performative=FipaMessage.Performative.PROPOSE)
```

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
```diff
@@ -0,0 +1,349 @@
...
+        )
+        incoming_message = fipa_dialogue.last_message
+
+        # before
+        assert self.get_quantity_in_outbox() == 0
+
+        # operation
+        fipa_handler._handle_propose(
+            cast(FipaMessage, incoming_message), cast(FipaDialogue, fipa_dialogue)
+        )
```
DavidMinarsch, line 210:
```
fipa_handler.handle(incoming_message)
```

aea/protocols/dialogue/base.py
5A11: Done

aea/test_tools/test_skill.py
5A11: Done

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
5A11: Done

aea/test_tools/test_skill.py
5A11: There is no concept. It is just a list of performatives and contents, which is the bare minimum of providing messages for a dialogue (with the option to specify targets and incoming/outgoing if the messages do not have the "standard" pattern of each targeting the one before and alternating between incoming/outgoing).

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
5A11: Changed all handlers to the main public handler

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
5A11: Done

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
5A11: Done

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
5A11: Done

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
5A11: Done

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
5A11: Done

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
5A11: Done

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
5A11: Not possible to just add the last message in a dialogue. Dialogue's structure does not allow an arbitrary population of messages (e.g. ACCEPT as a first message).

aea/test_tools/test_skill.py
5A11: I suggest we introduce it in `aea.protocols.dialogue.base` since it is effectively the representation of a message that is useful only in a dialogue.

aea/protocols/dialogue/base.py
```diff
@@ -569,15 +580,15 @@ def _is_belonging_to_dialogue(self, message: Message) -> bool:
     def reply(
         self,
         performative: Message.Performative,
-        target_message: Optional[Message] = None,
+        target: Optional[Union[int, Message]] = None,
```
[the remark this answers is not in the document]
5A11: Done

aea/test_tools/test_skill.py
```diff
@@ -0,0 +1,283 @@
...
+            "performative": performative,
+            "dialogue_reference": dialogue_reference,
+            "message_id": message_id,
+            "target": target,
+        }
+
+        for arg_name, arg_value in args_dict.items():
+            if arg_value is not None:
+                message_attributes[arg_name] = arg_value
+        message_attributes.update(kwargs)
```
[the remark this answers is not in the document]
5A11: Done

aea/test_tools/test_skill.py
```diff
@@ -0,0 +1,283 @@
...
+                    message_type=dialogues._message_class,  # pylint: disable=protected-access
+                    dialogue_reference=dialogue_reference,
+                    message_id=message_id,
+                    target=target,
+                    performative=performative,
+                    to=to,
+                    sender=sender,
+                    **contents,
+                )
+                dialogue = dialogues.update(message)
```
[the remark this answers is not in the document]
5A11: This does not add the last incoming message which the user wants to test the handler with, but adds those in the chain of messages the user wants to put in the dialogue which are supposed to have been from the opponent.

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
```diff
@@ -0,0 +1,302 @@
...
+                "service_id": "some_service_id",
+                "quantity": "some_quantity",
+                "tx_nonce": "some_tx_nonce",
+            }
+        )
+        incoming_message = self.prepare_dialogue(
+            dialogues=fipa_dialogues,
+            messages=((FipaMessage.Performative.CFP, {"query": "some_query"}),),
+            incoming_message=(FipaMessage.Performative.PROPOSE, {"proposal": proposal}),
+        )
```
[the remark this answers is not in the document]
5A11: I did it slightly differently. Instead of the above, you first prepare the dialogue, then build the incoming_message based on it:

```
dialogue = self.prepare_dialogue(dialogues, messages=(...))
incoming_message = self.build_incoming_message(dialogue, ...)
```

aea/protocols/dialogue/base.py
```diff
@@ -30,13 +30,65 @@
...
+            raise AEAEnforceError("target is None.")
+        return self._target
+
+    @property
+    def performative(self) -> Message.Performative:
+        """Get message's performative."""
+        return self._performative
+
+    @property
+    def contents(self) -> Dict[str, Any]:
```
DavidMinarsch, line 87: At a minimum this class also needs one class method `from_tuple` and one instance method `to_tuple`

aea/test_tools/test_skill.py
```diff
@@ -0,0 +1,345 @@
...
+                if attribute != expected_value:
+                    return (
+                        False,
+                        "The '{}' fields do not match. Actual '{}': {}. Expected '{}': {}".format(  # pylint: disable=duplicate-string-formatting-argument
+                            attribute_name,
+                            attribute_name,
+                            attribute,
+                            attribute_name,
+                            expected_value,
+                        ),
```
DavidMinarsch, line 102: use f-string to save space and avoid pylint warning

aea/test_tools/test_skill.py
```diff
@@ -0,0 +1,345 @@
...
+        incoming_message.sender = sender
+        incoming_message.to = (
+            self.skill.skill_context.agent_address if to is None else to
+        )
+
+        return incoming_message
+
+    def build_incoming_message_for_dialogue(
+        self,
+        dialogue: Dialogue,
```
DavidMinarsch, line 163: Let's discuss...

aea/protocols/dialogue/base.py
DavidMinarsch: but the other question is whether we actually need to define a class or whether it is enough to define a type using typing or something like a namedtuple (https://docs.python.org/3/library/collections.html#collections.namedtuple)

aea/test_tools/test_skill.py
5A11: Done

aea/protocols/dialogue/base.py
5A11: NamedTuple seems to be a good fit here.

Makefile
```diff
@@ -78,7 +78,7 @@ common_checks: security misc_checks lint static docs
 
 .PHONY: test
 test:
-	pytest -rfE --doctest-modules aea packages/fetchai/protocols packages/fetchai/connections tests/ --cov-report=html --cov-report=xml --cov-report=term-missing --cov-report=term --cov=aea --cov=packages/fetchai/protocols --cov=packages/fetchai/connections  --cov-config=.coveragerc
+	pytest -rfE --doctest-modules aea packages/fetchai/protocols packages/fetchai/connections packages/fetchai/skills/generic_* tests/ --cov-report=html --cov-report=xml --cov-report=term-missing --cov-report=term --cov=aea --cov=packages/fetchai/protocols --cov=packages/fetchai/connections  --cov=packages/fetchai/skills/generic_* --cov-config=.coveragerc
```
DavidMinarsch, line 81: Make paths explicit:
```suggestion
	pytest -rfE --doctest-modules aea packages/fetchai/protocols packages/fetchai/connections packages/fetchai/skills/generic_buyer tests/ --cov-report=html --cov-report=xml --cov-report=term-missing --cov-report=term --cov=aea --cov=packages/fetchai/protocols --cov=packages/fetchai/connections  --cov=packages/fetchai/skills/generic_buyer --cov-config=.coveragerc
```

aea/test_tools/test_skill.py
```diff
@@ -0,0 +1,345 @@
...
+        Evaluates whether a message's attributes match the expected attributes provided.
+
+        :param actual_message: the actual message
+        :param message_type: the expected message type
+        :param kwargs: other expected message attributes
+
+        :return: boolean result of the evaluation and accompanied message
+        """
+        if (
+            type(actual_message)  # pylint: disable=unidiomatic-typecheck
```
DavidMinarsch, line 80: https://vald-phoenix.github.io/pylint-errors/plerr/errors/basic/C0123

aea/test_tools/test_skill.py
DavidMinarsch: I disagree, different representations of the message should be close together. But let's see when it is applied throughout...

aea/test_tools/test_skill.py
```diff
@@ -0,0 +1,345 @@
...
+        if type(new_message[1]) != int:  # pylint: disable=unidiomatic-typecheck
+            new_message_list = list(new_message)
+            new_message_list.insert(1, message_id - 1)
+            new_message = tuple(new_message_list)
+        return new_message
+
+    def prepare_dialogue(
+        self,
+        dialogues: Dialogues,
+        messages: Tuple[Tuple[Any, ...]],
```
DavidMinarsch, line 220:
```suggestion
        messages: Tuple[Tuple[Any, ...]],
```
needs to be replaced with the new type/namedtuple typing

aea/test_tools/test_skill.py
```diff
@@ -0,0 +1,345 @@
...
+            to=self.skill.skill_context.agent_address,
+            sender=dialogue.dialogue_label.dialogue_opponent_addr,
+            **kwargs,
+        )
+
+        return incoming_message
+
+    @staticmethod
+    def _complete_compact_message(
+        message: Tuple[Any, ...], last_incoming: Optional[bool], message_id: int
```
DavidMinarsch, line 204: change `message_id: Optional[int] = None`. As by default, message_id = target + 1

aea/test_tools/test_skill.py
```diff
@@ -0,0 +1,345 @@
...
+            if dialogue is None:
+                raise AEAEnforceError(
+                    "Cannot update the dialogue with message number {}".format(
+                        message_id
+                    )
+                )
+        else:  # messages from self
+            _, dialogue = dialogues.create(
+                counterparty=counterparty, performative=performative, **contents
+            )
```
DavidMinarsch, line 276: messy...missing abstractions on either message or compact_message or some helper method

aea/protocols/dialogue/base.py
5A11: The only thing is we cannot do type-checking on the elements.

aea/test_tools/test_skill.py
```diff
@@ -0,0 +1,345 @@
...
+                incomplete_message,
+                last_incoming=last_incoming,
+                message_id=messages.index(incomplete_message) + 1,
+            )
+            last_incoming = complete_compact_message[0]
+            is_incoming = complete_compact_message[0]
+            message_id = messages.index(incomplete_message) + 1
+            target = complete_compact_message[1]
+            performative = complete_compact_message[2]
+            contents = complete_compact_message[3]
```
DavidMinarsch, line 289: what's the point of `_complete_compact_message` if you then need another 6 lines to unpack it?

aea/test_tools/test_skill.py
```diff
@@ -0,0 +1,345 @@
...
+        Quickly create a dialogue.
+
+        The 'messages' argument is a tuple of "compact messages".
+        Each compact message is represented as a tuple (inc, target, performative, contents), where
+         - 'inc' is boolean, representing whether the message is incoming (True), or outgoing (False).
+            - 'inc' is optional, if not provided: for the first message is assumed False (outgoing),
+            for any other message, it is the opposite of the one preceding it.
+         - 'target' is integer, representing the target of the message.
+            - 'target' is optional, if not provided: for the first message is assumed 0,
+            for any other message, it is the index of the message before it in the tuple of messages + 1.
```
DavidMinarsch, line 233: All these rules around compact message sequences need to be validated somewhere and this logic needs to be implemented elsewhere to keep concerns separated.

aea/test_tools/test_skill.py
```diff
@@ -0,0 +1,345 @@
...
+            contents = complete_compact_message[3]
+
+            if is_incoming:  # messages from the opponent
+                dialogue_reference = (
+                    dialogue.dialogue_label.dialogue_reference[0],
+                    Dialogues._generate_dialogue_nonce()  # pylint: disable=protected-access
+                    if dialogue.dialogue_label.dialogue_reference[1]
+                    == Dialogue.UNASSIGNED_DIALOGUE_REFERENCE
+                    else dialogue.dialogue_label.dialogue_reference[1],
+                )
```
DavidMinarsch, line 298: probably extract in a helper method as this is quite hard to read/understand and would benefit from a named method

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
```diff
@@ -0,0 +1,312 @@
...
+    """Test behaviours of generic buyer."""
+
+    path_to_skill = Path(ROOT_DIR, "packages", "fetchai", "skills", "generic_buyer")
+
+    def test_search_behaviour_setup_is_ledger_tx(self):
+        """Test the setup method of the search behaviour where is_ledger_tx is True."""
+        # setup
+        search_behaviour = cast(
+            GenericSearchBehaviour, self.skill.skill_context.behaviours.search
+        )
```
DavidMinarsch, line 53: since each method does something with the search_behaviour, you can extract this in `setup` method

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
```diff
@@ -0,0 +1,312 @@
...
+
+    def test_fipa_handler_handle_unidentified_dialogue(self):
+        """Test the _handle_unidentified_dialogue method of the fipa handler."""
+        # setup
+        fipa_handler = cast(GenericFipaHandler, self.skill.skill_context.handlers.fipa)
+        incoming_message = self.build_incoming_message(
+            message_type=FipaMessage,
+            dialogue_reference=("", ""),
+            performative=FipaMessage.Performative.ACCEPT,
+        )
```
DavidMinarsch, line 143:
```suggestion
        invalid_dialogue_ref = ("", ""),
        incoming_message = self.build_incoming_message(
            message_type=FipaMessage,
            dialogue_reference=invalid_dialogue_ref,
            performative=FipaMessage.Performative.ACCEPT,
        )
```
makes it easier for the next developer to understand

aea/test_tools/test_skill.py
```diff
@@ -0,0 +1,283 @@
...
+            new_message_list.insert(1, message_id - 1)
+            new_message = tuple(new_message_list)
+        return new_message
+
+    def prepare_dialogue(
+        self,
+        dialogues: Dialogues,
+        messages: Tuple[Tuple[Any, ...]],
+        incoming_message: Tuple[Any, ...],
+        counterparty: Address = "counterparty",
```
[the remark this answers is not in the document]
5A11: Done

aea/test_tools/test_skill.py
5A11: Done

aea/test_tools/test_skill.py
5A11: This is the `message_id` of this message from which the target is derived if not specified.

aea/test_tools/test_skill.py
5A11: This method gives a semantics to the unspecified `target` and `is_incoming`. 
The former becomes `last_message_id +1` and the latter becomes the opposite of the preceding message.

5A11: Maybe a bad name. I can also make it so unpacking is easier

5A11: I refactored them, so hopefully it's more clear.

aea/test_tools/test_skill.py
DavidMinarsch: optional args for message_id, target

aea/test_tools/test_skill.py
```diff
@@ -0,0 +1,345 @@
...
+        self,
+        dialogue: Dialogue,
+        performative: Message.Performative,
+        target: Optional[int] = None,
+        **kwargs,
+    ) -> Message:
+        """
+        Quickly create an incoming message with the provided attributes for a dialogue.
+
+        For any attribute not provided, the corresponding default value in message is used.
```
DavidMinarsch, line 171: Method to be used with care. The dialogue provided is part of the object which is being tested (the skill). By providing the dialogue and no message_id, target, and dialogue_ref, the test is by designed insured to pass for these three....

aea/test_tools/test_skill.py
5A11: Done

aea/test_tools/test_skill.py
5A11: Done

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
5A11: Done (also for handler)

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
5A11: Done

aea/test_tools/test_skill.py
5A11: Done

aea/test_tools/test_skill.py
5A11: Done

aea/protocols/dialogue/base.py
```diff
@@ -28,15 +28,25 @@
...
-from aea.exceptions import enforce
+from aea.exceptions import AEAEnforceError, enforce
 from aea.protocols.base import Message
 
 
+DialogueMessage = namedtuple(
+    "DialogueMessage",
+    ["performative", "contents", "is_incoming", "target"],
+    rename=False,
+    defaults=[None, None],
```
DavidMinarsch, line 45: should default of `is_incoming` not be a boolean? (edit: I see this is not possible based on usage below)

DavidMinarsch: what about default for contents being `{}`/`dict`?

aea/test_tools/test_skill.py
```diff
@@ -0,0 +1,368 @@
...
+            if dialogue_reference is not None
+            else dialogue.dialogue_label.dialogue_reference
+        )
+        message_id = (
+            message_id
+            if message_id is not None
+            else dialogue.last_message.message_id + 1
+        )
+        target = target if target is not None else dialogue.last_message.message_id
+        to = to if to is not None else self.skill.skill_context.agent_address
```
DavidMinarsch, line 211: although correct, we should be consistent and use the dialogue here:
```suggestion
        to = to if to is not None else dialogue.self_address
```

aea/test_tools/test_skill.py
```diff
@@ -0,0 +1,368 @@
...
+        :param kwargs: other attributes
+
+        :return: the created incoming message
+        """
+        message_attributes = dict()  # type: Dict[str, Any]
+
+        default_dialogue_reference = (
+            Dialogues._generate_dialogue_nonce(),  # pylint: disable=protected-access
+            "",
+        )
```
DavidMinarsch, line 133: Would be nice if we could turn the `new_self_initiated_dialogue_reference` into a class method in the base `Dialogues` class and then call here `Dialogues.new_self_initiated_dialogue_reference()`. That should be possible without causing anything to break.

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
```diff
@@ -0,0 +1,329 @@
...
+            sender=self.skill.skill_context.agent_address,
+            ledger_id=FETCHAI,
+            address=self.skill.skill_context.agent_address,
+        )
+        assert has_attributes, error_str
+
+    def test_search_behaviour_setup_not_is_ledger_tx(self):
+        """Test the setup method of the search behaviour where is_ledger_tx is False."""
+        # setup
+        strategy = cast(GenericStrategy, self.skill.skill_context.strategy)
```
DavidMinarsch, line 78: could also be set in setup, no real cost if we don't use it..and saves some loc

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
```diff
@@ -0,0 +1,329 @@
...
+                "tx_nonce": "some_tx_nonce",
+            }
+        )
+        fipa_dialogue = self.prepare_dialogue(
+            dialogues=self.fipa_dialogues,
+            messages=(
+                DialogueMessage(FipaMessage.Performative.CFP, {"query": "some_query"}),
+            ),
+        )
+        incoming_message = self.build_incoming_message_for_dialogue(
```
DavidMinarsch, line 188: maybe we should name the methods `prepare_skill_dialogue` and `build_incoming_message_for_skill_dialogue`? I'm also ok with how it is.

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
```diff
@@ -0,0 +1,329 @@
...
+
+        fipa_dialogue = self.prepare_dialogue(
+            dialogues=self.fipa_dialogues,
+            messages=(
+                DialogueMessage(FipaMessage.Performative.CFP, {"query": "some_query"}),
+                DialogueMessage(
+                    FipaMessage.Performative.PROPOSE, {"proposal": "some_proposal"}
+                ),
+                DialogueMessage(FipaMessage.Performative.ACCEPT, {}),
+            ),
```
DavidMinarsch, line 307: This list could be defined once in setup; then just reused again and again by accessing slices of it.

aea/test_tools/test_skill.py
```diff
@@ -0,0 +1,368 @@
...
+            for any other message, it is the opposite of the one preceding it.
+            - if 'target' is not provided: for the first message it is assumed 0,
+            for any other message, it is the index of the message before it in the tuple of messages + 1.
+
+        :param dialogues: a dialogues class
+        :param counterparty: the message_id
+        :param messages: the dialogue_reference
+
+        :return: the created incoming message
+        """
```
DavidMinarsch, line 276: First we need to check the messages list is non-zero length.

aea/test_tools/test_skill.py
```diff
@@ -0,0 +1,368 @@
...
+            if dialogue is None:
+                raise AEAEnforceError(
+                    "Cannot update the dialogue with message number {}".format(
+                        message_id
+                    )
+                )
+        else:  # messages from self
+            _, dialogue = dialogues.create(
+                counterparty=counterparty, performative=performative, **contents
+            )
```
DavidMinarsch, line 307: Can be extracted in appropriate private method; then we can use it below again in the for loop. Diff between create and reply can be easily controlled with flag.

aea/test_tools/test_skill.py
```diff
@@ -0,0 +1,368 @@
...
+        is_incoming = default_is_incoming if message[2] is None else message[2]
+
+        default_target = message_id - 1
+        target = default_target if message[3] is None else message[3]
+        return is_incoming, target
+
+    @staticmethod
+    def _non_initial_incoming_message_dialogue_reference(
+        dialogue: Dialogue,
+    ) -> Tuple[str, str]:
```
DavidMinarsch, line 245: missing docstrings.

DavidMinarsch (commented):
Looking good, please also make sure that all linters pass so the tests can run.

aea/protocols/dialogue/base.py
DavidMinarsch: https://stackoverflow.com/questions/56780005/how-to-validate-namedtuple-values ...but brings overhead as well.

aea/test_tools/test_skill.py
5A11: Agree

aea/protocols/dialogue/base.py
5A11: Done

aea/test_tools/test_skill.py
5A11: Done

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
5A11: Yes

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
5A11: Agree. Done

tests/test_packages/test_skills/test_generic_buyer/test_behaviours.py
5A11: Done

aea/test_tools/test_skill.py
5A11: Done

aea/test_tools/test_skill.py
5A11: I took the extraction part into its own private method.

As I said, the treatment of the first message and the rest could be combined into one for loop, but the way `dialogue_reference` is defined then causes a flake8/mypy error. I did not manage to come up with a way to address that. 

If you want, I can submit that code and you can see if you can resolve the flake8/mypy issue?

5A11:
> Looking good, please also make sure that all linters pass so the tests can run.

The test for the proposal method, i.e. `test_fipa_handler_handle_propose` needs a mock which I did not manage to implement. This part needs help from you/others.

aea/test_tools/test_skill.py
DavidMinarsch: I don't think anything needs to be updated as a result of this change. The old usage `self.new_self_initiated_dialogue_reference()` is automatically still supported...

5A11: Done

aea/test_tools/test_skill.py
```diff
@@ -0,0 +1,402 @@
...
+            - if 'target' is not provided: for the first message it is assumed 0,
+            for any other message, it is the index of the message before it in the tuple of messages + 1.
+
+        :param dialogues: a dialogues class
+        :param counterparty: the message_id
+        :param messages: the dialogue_reference
+
+        :return: the created incoming message
+        """
+        if not len(messages) >= 1:
```
DavidMinarsch, line 309: More idiomatic:
```suggestion
        if len(messages) == 0:
```

aea/test_tools/test_skill.py
```diff
@@ -0,0 +1,402 @@
...
+
+    def _extract_message_fields(
+        self,
+        message: DialogueMessage,
+        messages: Tuple[DialogueMessage, ...],
+        last_is_incoming: bool,
+    ):
+        performative = message[0]
+        contents = message[1]
+        message_id = messages.index(message) + 1
```
DavidMinarsch, line 281: erm just provide the index in the args? Decouples the logic a bit.
```suggestion
        message_id = index + 1
```

aea/test_tools/test_skill.py
```diff
@@ -0,0 +1,402 @@
...
+            if dialogue.dialogue_label.dialogue_reference[1]
+            == Dialogue.UNASSIGNED_DIALOGUE_REFERENCE
+            else dialogue.dialogue_label.dialogue_reference[1],
+        )
+        return dialogue_reference
+
+    def _extract_message_fields(
+        self,
+        message: DialogueMessage,
+        messages: Tuple[DialogueMessage, ...],
```
DavidMinarsch, line 276: I don't think the list of messages should be passed here, just provide the index (see below). That makes it easier to read and less expensive.

aea/test_tools/test_skill.py
```diff
@@ -0,0 +1,402 @@
...
+        if not len(messages) >= 1:
+            raise AEAEnforceError("the list of messages must be positive.")
+
+        (
+            performative,
+            contents,
+            message_id,
+            is_incoming,
+            target,
+        ) = self._extract_message_fields(messages[0], messages, True)
```
DavidMinarsch, line 318:
```suggestion
        ) = self._extract_message_fields(messages[0], index=0, last_is_incoming=True)
```

DavidMinarsch: (why is last_is_incoming=True by default, should be None?)

aea/test_tools/test_skill.py
```diff
@@ -0,0 +1,402 @@
...
+
+        if is_incoming:  # messages from the opponent
```
DavidMinarsch, line 320:
```suggestion
        if is_incoming:  # first message from the opponent
```

aea/test_tools/test_skill.py
```diff
@@ -0,0 +1,402 @@
...
+                **contents,
+            )
+            dialogue = cast(Dialogue, dialogues.update(message))
+            if dialogue is None:
+                raise AEAEnforceError(
+                    "Cannot update the dialogue with message number {}".format(
+                        message_id
+                    )
+                )
+        else:  # messages from self
```
DavidMinarsch, line 339:
```suggestion
        else:  # first message from self
```

aea/test_tools/test_skill.py
```diff
@@ -0,0 +1,402 @@
...
+            _, dialogue = dialogues.create(
+                counterparty=counterparty, performative=performative, **contents
+            )
+
+        for incomplete_message in messages[1:]:  # type: ignore
```
DavidMinarsch, line 344:
```suggestion
        for idx, dialogue_message in enumerate(messages[1:]):  # type: ignore
```
why is there a type ignore?

aea/test_tools/test_skill.py
```diff
@@ -0,0 +1,402 @@
...
+            (
+                performative,
+                contents,
+                message_id,
+                is_incoming,
+                target,
+            ) = self._extract_message_fields(incomplete_message, messages, is_incoming)
```
DavidMinarsch, line 351:
```suggestion
            ) = self._extract_message_fields(dialogue_message, idx, is_incoming)
```

aea/test_tools/test_skill.py
5A11: The default `is_incoming` value, if not specified, is the opposite of the `is_incoming` of the last message, and for the first message it is outgoing. 

For the first message, the `last_is_incoming` above is set to `True` to make its `is_incoming` (if unspecified) False.

aea/protocols/dialogue/base.py
```diff
@@ -27,16 +27,36 @@
...
+        rename=False,
+        module="aea.protocols.dialogues.base",
+    )
+    DialogueMessage.__new__.__defaults__ = (dict(), None, None)
+else:
+    DialogueMessage = namedtuple(
+        "DialogueMessage",
+        ["performative", "contents", "is_incoming", "target"],
+        rename=False,
+        defaults=[dict(), None, None],
```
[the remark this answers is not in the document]
5A11: Done

aea/protocols/dialogue/base.py
```diff
@@ -27,16 +27,36 @@
...
 
 
+if sys.version_info < (3, 7):
+    DialogueMessage = namedtuple(
+        "DialogueMessage",
+        ["performative", "contents", "is_incoming", "target"],
+        rename=False,
+        module="aea.protocols.dialogues.base",
+    )
+    DialogueMessage.__new__.__defaults__ = (dict(), None, None)
```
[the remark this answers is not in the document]
5A11: No I don't have 3.6. Found this solution in an stackoverflow answer.

aea/test_tools/test_skill.py
```diff
@@ -0,0 +1,402 @@
...
+    def build_incoming_message_for_skill_dialogue(
+        self,
+        dialogue: Dialogue,
+        performative: Message.Performative,
+        message_type: Optional[Type[Message]] = None,
+        dialogue_reference: Optional[Tuple[str, str]] = None,
+        message_id: Optional[int] = None,
+        target: Optional[int] = None,
+        to: Optional[Address] = None,
+        sender: Address = None,
```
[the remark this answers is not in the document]
5A11: Done

aea/test_tools/test_skill.py
```diff
@@ -0,0 +1,402 @@
...
+        if dialogue is None:
+            raise AEAEnforceError("dialogue cannot be None.")
+
+        if dialogue.last_message is None:
+            raise AEAEnforceError("dialogue cannot be empty.")
+
+        message_type = (
+            message_type
+            if message_type is not None
+            else dialogue._message_class  # pylint: disable=protected-access
```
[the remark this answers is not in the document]
5A11: Done

aea/test_tools/test_skill.py
5A11: Done

aea/test_tools/test_skill.py
5A11: Agree. Done.

aea/test_tools/test_skill.py
5A11: Agree. Done.

aea/test_tools/test_skill.py
5A11: Done

aea/test_tools/test_skill.py
5A11: Done

aea/test_tools/test_skill.py
5A11: Done

aea/test_tools/test_skill.py
5A11: Done

aea/protocols/dialogue/base.py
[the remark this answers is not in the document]
5A11: https://stackoverflow.com/questions/11351032/named-tuple-and-default-values-for-optional-keyword-arguments

aea/test_tools/test_skill.py
```diff
@@ -0,0 +1,419 @@
...
+        return self._multiplexer.out_queue.qsize()
+
+    def get_message_from_outbox(self) -> Optional[Message]:
+        """Get message from outbox."""
+        if self._outbox.empty():
+            return None
+        envelope = self._multiplexer.out_queue.get_nowait()
+        return envelope.message
+
+    def get_quantity_in_decision_maker_outbox(self) -> int:
```
DavidMinarsch, line 69: this is the inbox of the decision maker..
```suggestion
    def get_quantity_in_decision_maker_inbox(self) -> int:
```

aea/test_tools/test_skill.py
```diff
@@ -0,0 +1,419 @@
...
+        """Get the quantity of envelopes in the decision maker outbox."""
+        return self._skill.skill_context.decision_maker_message_queue.qsize()
+
+    def get_message_from_decision_maker_outbox(self) -> Optional[Message]:
+        """Get message from decision maker outbox."""
+        if self._skill.skill_context.decision_maker_message_queue.empty():
+            return None
+        envelope = self._skill.skill_context.decision_maker_message_queue.get_nowait()
+        return envelope.message
```
DavidMinarsch, line 78: there are no envelopes in this queue, just messages

tests/test_protocols/test_dialogue/test_base.py
```diff
@@ -123,6 +125,31 @@ def role_from_first_message(  # pylint: disable=unused-argument
...
+    dialogue_message = DialogueMessage(DefaultMessage.Performative.BYTES)
+    assert isinstance(dialogue_message.performative, Message.Performative)
+
+    assert dialogue_message.performative == DefaultMessage.Performative.BYTES
+    assert dialogue_message.contents == {}
+    assert dialogue_message.is_incoming is None
+    assert dialogue_message.target is None
+
+
+def test_dialogue_message_python_3_6():
```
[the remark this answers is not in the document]
5A11: Done

aea/test_tools/test_skill.py
5A11: Done

aea/test_tools/test_skill.py
5A11: Done

tests/test_packages/test_skills/test_generic_buyer/test_dialogues.py
```diff
@@ -0,0 +1,219 @@
...
+        _, dialogue = self.default_dialogues.create(
+            counterparty=COUNTERPARTY_NAME,
+            performative=DefaultMessage.Performative.BYTES,
+            content=b"some_content",
+        )
+        assert dialogue.role == DefaultDialogue.Role.AGENT
+        assert dialogue.self_address == self.skill.skill_context.agent_address
+
+    def test_fipa_dialogue(self):
+        """Test the DefaultDialogues class."""
```
DavidMinarsch, line 88: wrong docstring

DavidMinarsch (approved):
LGTM

PR #1772, opened 2020-09-25
[AEA-670 AEA-752] Libp2p node add monitoring

lrahmani:
## Proposed changes

Add prometheus monitoring to libp2p node
- to enable it pass monitoring uri as configuration to the connection
- when prometheus monitoring is disabled, the node will print metrics to file

Metrics are:
- ACN primitive operations: dht store and dht lookup latency
- ACN main service: Register and Route latency, number of ongoing route requests, total number of route request, total number of successful route requests
- ACN delegate service: number of concurrent clients, total number of clients
- ACN relay service: total number of clients


packages/fetchai/connections/p2p_libp2p/utils/utils.go
```diff
@@ -50,6 +50,10 @@ import (
 	"libp2p_node/aea"
 )
 
+const (
```
DavidMinarsch, line 53: Why this (particular) bound?

packages/fetchai/connections/p2p_libp2p/dht/dhtpeer/dhtpeer.go
```diff
@@ -762,13 +891,21 @@ func (dhtPeer *DHTPeer) handleAeaNotifStream(stream network.Stream) {
...
+		duration := timer.GetTimer(start)
+		opLatencyRegister.Observe(float64(duration.Microseconds()))
 	}
 	dhtPeer.addressAnnounced = true
 }
 
 func (dhtPeer *DHTPeer) handleAeaRegisterStream(stream network.Stream) {
 	lerror, _, linfo, _ := dhtPeer.getLoggers()
 
+	nbrClients, _ := dhtPeer.monitor.GetCounter("service_relay_clients_count_all")
```
DavidMinarsch, line 903: This strings could be defined as constants elsewhere so we aid reuse and avoid magic values

packages/fetchai/connections/p2p_libp2p/connection.py
```diff
@@ -548,6 +559,10 @@ def __init__(self, **kwargs):
         if libp2p_delegate_uri is not None:
             delegate_uri = Uri(libp2p_delegate_uri)
 
+        monitoring_uri = None
```
DavidMinarsch, line 562: monitoring_uri: Optional[Uri] = None

packages/fetchai/connections/p2p_libp2p/utils/utils.go
lrahmani: it is arbitrary considering the regular messages that agents currently exchange

DavidMinarsch: We should keep track of this choice in docs

PR #1779, opened 2020-09-30
Overwrite agent configurations in Agent Manager

marcofavorito:
## Proposed changes

Overwrite agent configurations in Agent Manager

## Fixes

n/a

## Types of changes

## Checklist

## Further comments
n/a

### DIFF
--- a/aea/configurations/base.py
+++ b/aea/configurations/base.py
@@ -733,6 +733,11 @@ class ComponentId(PackageId):
             self.public_id.author, self.component_type.to_plural(), self.public_id.name
         )
 
+    @property
+    def json(self) -> Dict:
+        """Get the JSON representation."""
+        return dict(**self.public_id.json, type=str(self.component_type))
+
 
 ProtocolId = PublicId
 ContractId = PublicId
--- a/aea/configurations/base.py
+++ b/aea/configurations/base.py
@@ -27,7 +27,7 @@ import pprint
 import re
 from abc import ABC, abstractmethod
 from collections import OrderedDict
-from copy import deepcopy
+from copy import copy, deepcopy
 from enum import Enum
 from pathlib import Path
 from typing import (
@@ -833,6 +833,14 @@ class PackageConfiguration(Configuration, ABC):
         """Get the package dependencies."""
         return set()
 
+    def update(self, data: Dict) -> None:
+        """
+        Update configuration with other data.
+
+        :param data: the data to replace.
+        :return: None
+        """
+
 
 class ComponentConfiguration(PackageConfiguration, ABC):
     """Class to represent an agent component configuration."""
@@ -915,14 +923,6 @@ class ComponentConfiguration(PackageConfiguration, ABC):
         :raises ValueError if the version of the aea framework falls within a specifier.
         """
         _check_aea_version(self)
-
-    def update(self, data: Dict) -> None:
-        """
-        Update configuration with other data.
-
-        :param data: the data to replace.
-        :return: None
-        """
 
 
 class ConnectionConfig(ComponentConfiguration):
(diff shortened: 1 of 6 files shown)


aea/configurations/base.py
```diff
@@ -1537,7 +1537,18 @@ def default_ledger(self, ledger_id: str):
 
     def component_configurations_json(self) -> List[OrderedDict]:
         """Get the component configurations in JSON format."""
-        return list(map(OrderedDict, self.component_configurations.values()))
+        result: List[OrderedDict] = []
```
marcofavorito, line 1540: the difference with the previous implementation is that earlier we included `author`, `name` and `version` in the values of the dictionary `self.component_configurations`; however, to avoid inconsistencies, those information about the package can be extracted from the key.

aea/configurations/base.py
```diff
@@ -1647,10 +1658,10 @@ def from_json(cls, obj: Dict):
         component_configurations = {}
         for config in obj.get("component_configurations", []):
             tmp = deepcopy(config)
-            name = tmp["name"]
-            author = tmp["author"]
-            version = tmp["version"]
-            type_ = tmp["type"]
+            name = tmp.pop("name")
```
marcofavorito, line 1661: see above; `tmp` at the end of the iteration will contain only the configurable fields.

aea/configurations/base.py
```diff
@@ -1665,6 +1676,30 @@ def from_json(cls, obj: Dict):
...
+        result: Dict[ComponentId, Dict] = copy(self.component_configurations)
+        for component_id, obj in new_component_configurations.items():
+            if component_id not in result:
+                result[component_id] = obj
+            else:
+                recursive_update(result[component_id], obj)
+        self.component_configurations = result
+
+        # update other fields
+        # TODO - currently not supported.
```
marcofavorito, line 1701: how to deal with those?

aea/manager.py
```diff
@@ -551,20 +545,35 @@ def _make_config(
...
+        agent_configuration_file_path: Path = AEABuilder.get_configuration_file_path(
+            project_path
         )
-        agent_config = self._update_dict(json_data[0], agent_overrides)
-
-        components_configs = {PublicId.from_json(obj): obj for obj in json_data[1:]}
-
+        loader = ConfigLoaders.from_package_type(PackageType.AGENT)
+        with agent_configuration_file_path.open() as fp:
+            agent_config: AgentConfig = loader.load(fp)
```
marcofavorito, line 553: we use `AgentConfig` instead of bare JSONs, the idea is that hopefully will make the code more robust, since we call the `agent_config.update` method to override the configurations, and there we can delegate the consistency checks

aea/manager.py
```diff
@@ -551,20 +545,35 @@ def _make_config(
...
-        return [agent_config] + list(components_configs.values())
+            component_id = ComponentId(obj.pop("type"), PublicId(author, name, version))
+            components_configs[component_id] = obj
+        agent_update_dictionary["component_configurations"] = components_configs
+        # do the override (and valiation)
+        agent_config.update(agent_update_dictionary)
+
+        # return the multi-paged JSON object.
+        json_data = agent_config.ordered_json
+        result: List[Dict] = [json_data] + json_data.pop("component_configurations")
```
marcofavorito, line 575: this was done in other places as well.

Basically, `Configuration.json` returns a dictionary; however, for `AgentConfig` the interface is not ideal. So what we do is to create a dictionary with `component_configurations` as the list of following overrides; then the burden it's up to the client of the class.

solarw (approved):
lgtm

DavidMinarsch (approved):
LGTM

PR #1780, opened 2020-09-30
Let to update all configurable fields of the main part of an AEA configuration.

marcofavorito:
## Proposed changes

Let to update all configurable fields of the main part of an `AgentConfig` configuration via `AgentConfig.update`.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a


aea/configurations/base.py
```diff
@@ -1703,7 +1705,14 @@ def update(self, data: Dict) -> None:
         self.component_configurations = result
 
         # update other fields
-        # currently not supported.
+        for item_id, value in data.get("private_key_paths", {}).items():
+            self.private_key_paths.update(item_id, value)
```
marcofavorito, line 1709: is it correct we only update existing keys?

aea/configurations/base.py
```diff
@@ -1703,7 +1705,14 @@ def update(self, data: Dict) -> None:
...
+
+        for item_id, value in data.get("connection_private_key_paths", {}).items():
+            self.connection_private_key_paths.update(item_id, value)
+
+        self.logging_config = data.get("logging_config", self.logging_config)
```
marcofavorito, line 1714: here we don't apply granular set... I think it's OK

aea/configurations/base.py
DavidMinarsch: No, at the moment we need to be able to add and override, otherwise it is not very useful

DavidMinarsch (changes requested):

DavidMinarsch (approved):
LGTM

aea/configurations/base.py
marcofavorito: OK, the behaviour of `CRUDCollection.update` now creates an item if it the key doesn't exists yet.

solarw (approved):
LGTM

PR #1791, opened 2020-10-06
[DO NOT MERGE] CI Send remaining funds back to faucet

steevi:
This PR sends the unused funds requested by the faucet, back to the faucet account.

I'm not sure if there is anywhere in similar code needs to be added to the cli tests.
Waiting for your comments.
Thanks

### DIFF
--- a/tests/test_crypto/test_fetchai.py
+++ b/tests/test_crypto/test_fetchai.py
@@ -647,10 +647,11 @@ def test_execute_shell_command(mock_api_call):
 
 
 def send_remaining_funds_back_to_faucet(account: FetchAICrypto):
-    """Sends remainind funds back to faucet"""
+    """Sends remaining funds back to faucet"""
     faucet_address = "fetch193vvag846gz3pt3q0mdjuxn0s5jrt39fsjrays"
     fetchai_api = FetchAIApi(**FETCHAI_TESTNET_CONFIG)
     balance = fetchai_api.get_balance(account.address)
+    assert balance is not None, "balance is of type None"
 
     tx_fee = 1000
     amount = balance - tx_fee
--- a/tests/test_crypto/test_fetchai.py
+++ b/tests/test_crypto/test_fetchai.py
@@ -654,9 +654,7 @@ def send_remaining_funds_back_to_faucet(account: FetchAICrypto):
 
     tx_fee = 1000
     amount = balance - tx_fee
-    nonce = FetchAIApi.generate_tx_nonce(
-        seller=account.address, client=faucet_address
-    )
+    nonce = FetchAIApi.generate_tx_nonce(seller=account.address, client=faucet_address)
     transfer_transaction = fetchai_api.get_transfer_transaction(
         sender_address=account.address,
         destination_address=faucet_address,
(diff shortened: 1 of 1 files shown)


tests/test_crypto/test_fetchai.py
```diff
@@ -639,3 +644,62 @@ def test_execute_shell_command(mock_api_call):
...
     res = cosmos_api._execute_shell_command(["test", "command"])
     assert res == {"SOME": "RESULT"}
+
+
+def send_remaining_funds_back_to_faucet(account: FetchAICrypto):
+    """Sends remaining funds back to faucet"""
+    faucet_address = "fetch193vvag846gz3pt3q0mdjuxn0s5jrt39fsjrays"
+    fetchai_api = FetchAIApi(**FETCHAI_TESTNET_CONFIG)
+    balance = fetchai_api.get_balance(account.address)
+    assert balance is not None, "balance is of type None"
```
DavidMinarsch, line 654: If the balance is None it should simply return imo

tests/test_crypto/test_fetchai.py
```diff
@@ -639,3 +644,62 @@ def test_execute_shell_command(mock_api_call):
...
+
+    tx_fee = 1000
```
DavidMinarsch, line 656: Is this the minimum required tx fee? Please define a constant above and use here (e.g. `tx_fee = MINIMUM_TX_FEE`).

tests/test_crypto/test_fetchai.py
```diff
@@ -639,3 +644,62 @@ def test_execute_shell_command(mock_api_call):
...
+        and isinstance(signed_transaction["tx"]["signatures"], list)
+    ), "Incorrect signed_transaction constructed."
+
+    transaction_digest = fetchai_api.send_signed_transaction(signed_transaction)
+    assert transaction_digest is not None, "Failed to submit transfer transaction!"
+
+    not_settled = True
+    elapsed_time = 0
+    while not_settled and elapsed_time < 20:
+        elapsed_time += 1
```
DavidMinarsch, line 683: I don't think we should actually wait...this is a cleanup, so let's send and then be done with it. If it goes through, good, if not, well no-one's going to do anything about it so let's just wrap up.

tests/test_crypto/test_fetchai.py
```diff
@@ -639,3 +644,62 @@ def test_execute_shell_command(mock_api_call):
...
+
+    tx = fetchai_api.get_transaction(transaction_digest)
+    is_valid = fetchai_api.is_transaction_valid(
+        tx, faucet_address, account.address, "", amount
+    )
+    assert is_valid, "Failed to settle tx correctly!"
+    assert tx == transaction_receipt, "Should be same!"
+
+
+def check_balance_0(account: FetchAICrypto):
```
DavidMinarsch, line 701: Remove?

tests/test_crypto/test_fetchai.py
```diff
@@ -639,3 +644,62 @@ def test_execute_shell_command(mock_api_call):
 
     res = cosmos_api._execute_shell_command(["test", "command"])
     assert res == {"SOME": "RESULT"}
+
+
+def send_remaining_funds_back_to_faucet(account: FetchAICrypto):
```
DavidMinarsch, line 649: Ideally this is implemented as a decorator so we can decorate test functions with it.

DavidMinarsch (changes requested):
Thanks for this @steevi . I'd make a few changes, please see my comments.

PR #1796, opened 2020-10-09
[AEA-950] Add libp2p bench

lrahmani:
## Proposed changes

Add benchmark files used in the paper


packages/fetchai/connections/p2p_libp2p/dht/dhtpeer/dhtpeer.go
```diff
@@ -831,7 +837,16 @@ func (dhtPeer *DHTPeer) registerAgentAddress(addr string) error {
 	linfo().Str("op", "register").
 		Str("addr", addr).
 		Msgf("Announcing address to the dht with cid key %s", addressCID.String())
+	//var elapsed time.Duration
+	//start := time.Now()
 	err = dhtPeer.dht.Provide(ctx, addressCID, true)
+	//elapsed = time.Since(start)
```
panasevychol, line 843: Do we need this commented code here and above? If yes maybe it worth to have a comment that explains why it's here.

packages/fetchai/connections/p2p_libp2p/dht/dhtpeer/benchmarks_test.go
```diff
@@ -0,0 +1,378 @@
...
+
+/* **********
+* How to run
+* ***********
+
+	$ go test -p 1 -count 20 libp2p_node/dht/dhtpeer/ -run=XXX  -bench .  -benchtime=20x
+
+*/
+
+var keysFilePath = "/path/to/file/benchmark_keys.txt"
```
solarw, line 43: this path a bit confusing, probably I miss something?

solarw (approved):
LGTM

packages/fetchai/connections/p2p_libp2p/dht/dhtpeer/benchmarks_test.go
```diff
@@ -0,0 +1,378 @@
...
+var addrsFilePath = "/path/to/file/benchmark_addrs.txt"
+var tcpUri = "localhost:12345"
+
+/* **********************************
+ * baseline TCP connection benchmark
+ * ********************************** */
+
+// helpers
+
+func acceptAndEcho(server net.Listener) {
```
panasevychol, line 53: Just wondering with the level of nesting in this function. Wasn't it possible to move from level 3 to a separate func?

panasevychol (approved):
I didn't dig much but brief impression is good. Just a few comments rather then changes requests.
And pls fix the conflicts with a target branch.

packages/fetchai/connections/p2p_libp2p/dht/dhtpeer/benchmarks_test.go
lrahmani: you are right, considering that it is intended to be used only once I didn't want to add an extra symbol for it

packages/fetchai/connections/p2p_libp2p/dht/dhtpeer/benchmarks_test.go
lrahmani: addressed

packages/fetchai/connections/p2p_libp2p/dht/dhtpeer/dhtpeer.go
lrahmani: addressed

DavidMinarsch (approved):
LGTM

Issue #1702, opened 2020-08-27
Refactor configurations to avoid dependency cycles and address TODOs


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
In loader:
```
# TODO: instead of this, create custom loader and use it
#       by wrapping yaml.safe_load to use it
_config_loader()
```

In loader
```python
    def _validate(self, json_data: Dict) -> None:
        """
        Validate a configuration file.

        :param json_data: the JSON object of the configuration file to validate.
        :return: None
        :raises ValidationError: if the file doesn't comply with the JSON schema.
              | ValueError: if other consistency checks fail.
        """
        # this might raise ValidationError.
        self.validate(json_data)

        expected_type = self.configuration_class.package_type
        # TODO 'type' is optional for backward compatibility
        if expected_type != PackageType.AGENT and "type" in json_data:
            actual_type = PackageType(json_data["type"])
            if expected_type != actual_type:
                raise ValueError(
                    f"The field type is not correct: expected {expected_type}, found {actual_type}."
                )
```

```
is_satisfiable: this is not the full story. 
```

- multiple in base

**Describe the solution you'd like**
-

DavidMinarsch:
```
# TODO implement a proper class to represent this type.
Dependency = dict
```
- [ ] 
![Screenshot 2020-08-28 at 11 38 24](https://user-images.githubusercontent.com/19314578/91552121-2e694700-e923-11ea-9348-6b32124ad18f.png)
![Screenshot 2020-08-28 at 11 38 34](https://user-images.githubusercontent.com/19314578/91552127-30330a80-e923-11ea-845a-b9e1d03b5561.png)

marcofavorito:
- [ ] refactor YAML loading 
- TODO

PR #1802, opened 2020-10-12
Refactor YAML loader/dumper

marcofavorito:
## Proposed changes

- Move yaml related utils to their own module
- introduce `_AEAYamlLoader` and `_AEAYamlDumper` classes, instead of decorators and automatically called functions at import time.

## Fixes

Partially addresses #1702

## Types of changes

## Checklist

## Further comments

n/a


aea/helpers/yaml_utils.py
```diff
@@ -0,0 +1,153 @@
+# -*- coding: utf-8 -*-
```
marcofavorito, line 1: this module contains pieces of code both from aea.helpers.base and aea.configurations.loader, nothing really new.

aea/helpers/yaml_utils.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+def yaml_load(stream: TextIO) -> Dict[str, Any]:
+    """
+    Load a yaml from a file pointer in an ordered way.
+
+    :param stream: file pointer to the input file.
+    :return: the dictionary object with the YAML file content.
+    """
+    return yaml.load(stream, Loader=_AEAYamlLoader)  # nosec
```
marcofavorito, line 121: this is changed from `aea.helpers.base`, but you can't see the differences. Also each of the function below has changed analogously.

aea/helpers/yaml_utils.py
```diff
@@ -0,0 +1,153 @@
...
+        super().__init__(*args, **kwargs)
+        _AEAYamlLoader.add_constructor(
+            yaml.resolver.BaseResolver.DEFAULT_MAPPING_TAG, self._construct_mapping
+        )
+        _AEAYamlLoader.add_constructor(self.envvar_key, self._envvar_constructor)
+        self._add_implicit_resolver_if_not_present_already()
+
+    def _add_implicit_resolver_if_not_present_already(self) -> None:
+        """Add implicit resolver for environment variables, if not present already."""
+        if self.envvar_key in dict(self.yaml_implicit_resolvers.get(None, [])):
```
marcofavorito, line 59: this is to prevent the `add_implicit_resolver` method adds the resolver twice.

aea/helpers/yaml_utils.py
```diff
@@ -0,0 +1,153 @@
...
+
+    @staticmethod
+    def _construct_mapping(loader, node):
+        """Construct a YAML mapping with OrderedDict."""
+        object_pairs_hook = OrderedDict
+        loader.flatten_mapping(node)
+        return object_pairs_hook(loader.construct_pairs(node))
+
+    @staticmethod
+    def _envvar_constructor(_loader, node):  # pragma: no cover
```
DavidMinarsch, line 72: nice!

aea/helpers/yaml_utils.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def _add_implicit_resolver_if_not_present_already(self) -> None:
+        """Add implicit resolver for environment variables, if not present already."""
+        if self.envvar_key in dict(self.yaml_implicit_resolvers.get(None, [])):
+            _AEAYamlLoader.add_implicit_resolver(
+                self.envvar_key, self.envvar_matcher, None
+            )
+
+    @staticmethod
+    def _construct_mapping(loader, node):
```
DavidMinarsch, line 65: typing?

aea/helpers/yaml_utils.py
```diff
@@ -0,0 +1,157 @@
...
+        match = _AEAYamlLoader.envvar_matcher.match(node_value)
+        match = cast(Match[str], match)
+        env_var = match.group()[2:-1]
+
+        # check for defaults
+        var_name, default_value = env_var.split(":")
+        var_name = var_name.strip()
+        default_value = default_value.strip()
+        var_value = os.getenv(var_name, default_value)
+        return var_value + node_value[match.end() :]
```
marcofavorito, line 87: beware: the following will be applied to _any_ YAML field. That means also non-configurable ones because currently validation is done _after_ custom YAML loading. This should be reported as issue. It could be easily solved by first loading normally and then loading it with replacement. Or, just doing the variable replacement by hand only for configurable fields...

marcofavorito: see https://github.com/fetchai/agents-aea/blob/master/aea/configurations/loader.py#L262 and https://github.com/fetchai/agents-aea/blob/master/aea/configurations/loader.py#L327

marcofavorito: I suspect this is linked to the broader issue of having both descriptive and configurable fields in the same file. If we had one "manifest" file and a "configuration" field we would only apply this approach to the latter file.

DavidMinarsch (approved):
LGTM

PR #1826, opened 2020-10-15
[AEA-860] libp2p print human readable config

### LINKED ISSUE #1530: Proper logging and exit in p2p startup
Currently, when the p2p connection is started with non-matching configs it is not clear from the logs what's wrong. We also do not gracefully terminate the agent.

Provide multiple configuration scenarios (a: genesis peer with client endpoint; b: genesis peer with no client endpoint; c: non-genesis peer with client end-pont; d: non-genesis peer with no client endpoint; e: behind-nats-peer with client endpoint; f: behind-nats peer with no client endpoint) and properly log which one is selected during startup.

If inconsistent config is provided log this clearly and exit gracefully where necessary.

lrahmani:
## Proposed changes

- Print human readable configuration description of peer following a successful start
- Check that peers public uri ip address matches entry peers ip addresses classes (loopbak-private-public)

Fixes #1530


packages/fetchai/connections/p2p_libp2p/connection.py
```diff
@@ -536,6 +570,15 @@ def __init__(self, **kwargs):
                     "Local Uri must be set when Public Uri is provided. "
                     "Hint: they are the same for local host/network deployment"
                 )
+            else:
+                # check if node's public host and entry peers hosts are either
+                #  both private or both public
+                if not _ip_all_private_or_all_public(
```
DavidMinarsch, line 576: avoid nesting with `elif`?

packages/fetchai/connections/p2p_libp2p/connection.py
```diff
@@ -402,7 +437,8 @@ def get_libp2p_node_multiaddrs(self) -> Sequence[MultiAddr]:
             if found:
                 elem = line.strip()
                 if elem != LIST_END:
-                    multiaddrs.append(MultiAddr(elem))
+                    if len(elem) != 0:
```
DavidMinarsch, line 440: avoid nesting; use `and`?

DavidMinarsch (approved):
LGTM, but significant coverage gaps are present. Please fill (especially for helpers)

packages/fetchai/connections/p2p_libp2p/connection.py
lrahmani: removed the nesting

PR #1829, opened 2020-10-16
AEA-948 fix bug on import order computation in the 'AEABuilder'

marcofavorito:
## Proposed changes

The bug has been introduced by #1822. The problem was due to wrong computation of the input to provide to the topological order algorithm.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a

### DIFF
--- a/aea/aea_builder.py
+++ b/aea/aea_builder.py
@@ -1282,7 +1282,8 @@ class AEABuilder(WithLogger):  # pylint: disable=too-many-public-methods
                 ),
             )
 
-            dependency_to_supported_dependencies[skill_id] = set()
+            if skill_id not in dependency_to_supported_dependencies:
+                dependency_to_supported_dependencies[skill_id] = set()
             for dependency in configuration.skills:
                 dependency_to_supported_dependencies[
                     ComponentId(ComponentType.SKILL, dependency)


aea/aea_builder.py
```diff
@@ -1282,7 +1282,8 @@ def _find_import_order(
                 ),
             )
 
-            dependency_to_supported_dependencies[skill_id] = set()
+            if skill_id not in dependency_to_supported_dependencies:
+                dependency_to_supported_dependencies[skill_id] = set()
```
marcofavorito, line 1286: @DavidMinarsch for the specific issue you showed me:

Say we got `skill_ids = [carpark_detection, generic_seller, ...]`

First iteration: we process `carpark_detection`. and we find that `generic_seller` supports `carpark_detection` (or in other words, `carpark_detection` depends on `generic_seller`)

Second iteration: we process `generic_seller` and... well, we used to reset the "supported" skills. Hence, at the end of the for loop, `carpark_detection` doesn't depend on anything, hence the problem.

marcofavorito: the reason why sometimes worked is because `skill_ids` is a set, and the iterator order is not always the same across different runs.

DavidMinarsch: Nice explanation. Thanks a lot and I think this solves this issue indeed. Let's see the run complete  :)

DavidMinarsch (approved):
LGTM

PR #1837, opened 2020-10-17
AEA-982 rename body in base message

### LINKED ISSUE #1817: Change `message` field `body` to `_body`
Currently, the `body` property on message means no protocol can have a `body` field. This is an issue for the http protocol where we need to name the field `bodyy` as a result.

Make private `__body` and public `_body` property. Then change http protocol.

DavidMinarsch:
## Proposed changes

This PR renames the `body` property to `_body` to avoid `body` being a reserved property for protocols.

## Fixes

#1817 

## Types of changes

## Checklist

## Further comments

-


aea/protocols/base.py
```diff
@@ -63,8 +63,8 @@ def __init__(self, body: Optional[Dict] = None, **kwargs):
         """
         self._to = None  # type: Optional[Address]
         self._sender = None  # type: Optional[Address]
-        self._body = copy(body) if body else {}  # type: Dict[str, Any]
-        self._body.update(kwargs)
+        self.__body = copy(_body) if _body else {}  # type: Dict[str, Any]
+        self.__body.update(kwargs)
```
marcofavorito, line 67: the leading double underscore has a special meaning in Python. I think it is OK to store `_body`

marcofavorito: https://dbader.org/blog/meaning-of-underscores-in-python

DavidMinarsch: That's what I'm after (or at least don't mind the properties it provides)

Issue #1378, opened 2020-06-17
Refactor aea cli launch command implementation


solarw:
**Is your feature request related to a problem? Please describe.**
Launch command implementation has two pretty similar functions for threads and subprocesses exclude a couple of workarounds (for windows and python 3.6/3.7)

**Describe the solution you'd like**
Those functions can be refactored into one.

**Describe alternatives you've considered**
refactor is the best solution, the current approach is quite good only in case of test coverage, cause test launch command involves AEARunner and AEALauncher, after refactoring it's supposed only AEAlauncher will be used.

**Additional context**

PR #1840, opened 2020-10-19
[AEA-964] cli.launch refactoring. windows fixes

solarw:
## Proposed changes

cli launch command refactoring and fixes for windows

## Fixes

https://github.com/fetchai/agents-aea/issues/1378

## Types of changes

## Checklist


aea/cli/launch.py
```diff
@@ -24,15 +24,11 @@
...
 
-from aea.aea import AEA
-from aea.aea_builder import AEABuilder
 from aea.cli.utils.click_utils import AgentDirectory
 from aea.cli.utils.context import Context
 from aea.cli.utils.loggers import logger
-from aea.helpers.base import cd
 from aea.helpers.multiple_executor import ExecutorExceptionPolicies
 from aea.launcher import AEALauncher
-from aea.runner import AEARunner
```
DavidMinarsch, line 35: So we no longer need AEARunner?

DavidMinarsch (approved):
LGTM

aea/cli/install.py
```diff
@@ -65,6 +66,7 @@ def do_install(ctx: Context, requirement: Optional[str] = None) -> None:
             logger.debug("Installing all the dependencies...")
             dependencies = ctx.get_dependencies()
             for name, d in dependencies.items():
+                click.echo("Installing {}...".format(pprint.pformat(name)))
```
marcofavorito, line 69:
```suggestion
                click.echo(f"Installing {pprint.pformat(name)}...")
```

aea/cli/launch.py
```diff
@@ -100,29 +81,5 @@ def _launch_subprocesses(click_context: click.Context, agents: List[Path]) -> in
...
-    try:
-        runner.start(threaded=True)
-        runner.join_thread()  # for some reason on windows and python 3.7/3.7 keyboard interuption exception gets lost so run in threaded mode to catch keyboard interruped
-    except KeyboardInterrupt:
-        logger.info("Keyboard interrupt detected.")
-    finally:
-        runner.stop()
-    return runner.num_failed
+    logger.debug(f"Exit cli. code: {launcher.num_failed}")
+    sys.exit(launcher.num_failed)
```
marcofavorito, line 85: I think this should be upadted. I think this is not a good practice. 

Say there is a failure in our system such that all the agents fail for the same reason (e.g. no networking).We return a different error code depending on the number of agents. Why not just `1`? This could potentially overlap with other error codes, e.g. `2` in UNIX has a special meaning, misuse of command line syntax.

aea/cli/launch.py
marcofavorito: It is used in benchmark `benchmark/checks/check_multiagent.py`

aea/cli/launch.py
DavidMinarsch: Yes let's return `1` indeed. I agree.

aea/cli/launch.py
solarw: i suppose it can be replaced with manager

aea/cli/launch.py
solarw: done

aea/cli/install.py
solarw: done

PR #1843, opened 2020-10-19
[AEA-1002] proper handling upgrade and remove dependencies for non vendor packages

solarw:
## Proposed changes

proper handling upgrade and remove dependencies for non vendor packages

## Types of changes

## Checklist

### DIFF
--- a/aea/cli/remove.py
+++ b/aea/cli/remove.py
@@ -27,6 +27,7 @@ from typing import Dict, Generator, Optional, Set, Tuple, cast
 
 import click
 
+from aea.aea_builder import AEABuilder
 from aea.cli.utils.click_utils import PublicIdParameter
 from aea.cli.utils.config import load_item_config
 from aea.cli.utils.context import Context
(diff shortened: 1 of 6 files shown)


aea/cli/remove.py
```diff
@@ -373,36 +386,9 @@ def agent_name(self) -> str:
...
-                    )
-                )
-            if (
-                item_folder.exists()
-                and not self.agent_config.author == self.item_id.author
-            ):  # pragma: no cover
-                raise click.ClickException(
-                    "{} {} author is different from {} agent author. "
-                    "Please fix the author field.".format(
-                        self.item_name, self.item_type, self.agent_name
```
DavidMinarsch, line 399: This check should be maintained. It could be moved under `get_component_directory`?

DavidMinarsch (approved):
LGTM

aea/cli/remove.py
solarw: did in `get item config` but it's better to do an overall refactoring, cause component config loading and checks are done in various code places, so the codebase can be unified.

marcofavorito (approved):
LGTM

PR #1862, opened 2020-10-21
Add block query to ledger_api protocol

jrriehl:
## Proposed changes

It would be useful to be able to query a block, for example to get the block height or random beacon value, both for the fetchai and ethereum ledgers. The block query of beacon-land is currently performed by a skill, but this might be better suited as a message using the ledger_api protocol. The DRB oracle also needs to get the block height of the target chain, which is currently ethereum-based.

## Types of changes

## Checklist


DavidMinarsch:
Please open an issue with this proposal

PR #1871, opened 2020-10-26
Fix/benchmark messages memory usage

solarw:
## Proposed changes

Perform various duration tests to  check memory usage across the time.

## Types of changes

## Checklist


benchmark/checks/check_mem_usage.py
```diff
@@ -57,6 +57,10 @@ def handle(self, message: Message) -> None:
 
 def run(duration, runtime_mode):
     """Check memory usage."""
+    # pylint: disable=import-outside-toplevel
+    # import causer tricky import in decision_maker
```
DavidMinarsch, line 61:
```suggestion
    # import manually due to some lazy imports in decision_maker
```

benchmark/checks/check_multiagent.py
```diff
@@ -105,6 +105,10 @@ def handle(self, message: Message) -> None:
 
 def run(duration, runtime_mode, runner_mode, start_messages, num_of_agents):
     """Test multiagent message exchange."""
+    # pylint: disable=import-outside-toplevel
+    # import causer tricky import in decision_maker
```
DavidMinarsch, line 109: see above

benchmark/checks/check_reactive.py
```diff
@@ -84,6 +84,10 @@ def handle(self, message: Message) -> None:
 
 def run(duration, runtime_mode, connection_mode):
     """Test memory usage."""
+    # pylint: disable=import-outside-toplevel
+    # import causer tricky import in decision_maker
```
DavidMinarsch, line 88: see above

benchmark/checks/check_proactive.py
```diff
@@ -61,6 +61,10 @@ def act(self):
 
 def run(duration, runtime_mode):
     """Test act message generate performance."""
+    # pylint: disable=import-outside-toplevel
+    # import causer tricky import in decision_maker
```
DavidMinarsch, line 65: see above

benchmark/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+ENV PYTHONPATH "$PYTHONPATH:/usr/lib/python3.8/site-packages"
+RUN apk add --no-cache make git bash wget
+RUN apk add --no-cache gcc musl-dev python3-dev libffi-dev openssl-dev
+RUN apk add --update --no-cache py3-numpy py3-scipy py3-pillow
+RUN apk add --update --no-cache gfortran freetype-dev libpng-dev openblas-dev g++ py3-numpy-dev
+RUN apk add --no-cache go
+
+RUN pip install --upgrade pip pipenv
+RUN pip install protobuf colorlog graphviz
```
DavidMinarsch, line 11: I don't think we need colorlog graphviz any more. It was needed for the old oef. Please check

benchmark/checks/check_dialogues_memory_usage.py
```diff
@@ -0,0 +1,128 @@
...
+    @property
+    def random_string(self) -> str:
+        """Get random string on every access."""
+        return uuid.uuid4().hex
+
+    def process_message(self) -> None:
+        """Process a message with dialogues."""
+        message = self.create()
+        dialogue = self.update(message)
+        self.reply(dialogue, message)
```
DavidMinarsch, line 65: This is wrong. The dialogues model has a `create` method. use this, see the skills for examples!

DavidMinarsch: e.g. generic buyer skill, behaviours.py line 81

benchmark/checks/run_benchmark_messages_mem.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+DATE=`LC_ALL=C date +"%d.%m.%Y_%H:%M"`
+echo "Performance report for $DATE"
+echo -e "-----------------------------\n"
+
+AGENTS=10
+NUM_RUNS=10
```
DavidMinarsch, line 7: Number runs should be higher to get more reliable measures. I'd say 100

benchmark/checks/check_dialogues_memory_usage.py
solarw: every message generated is sent from a new address.
so the message has to be generated manually,  existing dialogues can not be used, cause the same dialogue can not be used for creating and reply (checked!)

benchmark/checks/run_benchmark_messages_mem.sh
solarw: will update a bit later

benchmark/checks/check_multiagent.py
solarw: done

benchmark/checks/check_mem_usage.py
solarw: done

benchmark/checks/check_proactive.py
solarw: done

benchmark/checks/check_reactive.py
solarw: done

benchmark/Dockerfile
solarw: fixed

benchmark/checks/check_dialogues_memory_usage.py
DavidMinarsch: Yes you'd need to create two dialogues. That's better though as it keeps it consistent. The message should be created outside the handler (e.g. in a behaviour) and just then processed by the handler. Your custom logic here does not reflect what we have in the framework.

solarw: realistic case implemented in `check_multiagent_http_dialogues.py`, a limited amount of agents are interacting, there is one agent interacting with like an unlimited amount of agent, so every message generated from the new address

DavidMinarsch (approved):
LGTM

PR #1882, opened 2020-10-30
AEA-990 Feature/optimize message generator

solarw:
## Proposed changes

Better message class generation for less memory usage

## Types of changes

## Checklist


aea/protocols/base.py
```diff
@@ -56,23 +56,42 @@ def __str__(self):
...
         :param kwargs: any additional value to add to the body. It will overwrite the body values.
         """
+        self._slots = self._SlotsCls()
+
         self._to: Optional[Address] = None
         self._sender: Optional[Address] = None
-        self.__body: Dict[str, Any] = copy(_body) if _body else {}
-        self.__body.update(kwargs)
+
+        self._update_slots_from_dict(copy(_body) if _body else {})
```
DavidMinarsch, line 82: Why copy?

aea/protocols/base.py
```diff
@@ -56,23 +56,42 @@ def __str__(self):
             """Get the string representation."""
             return str(self.value)
 
+    class _SlotsCls:  # pylint: disable=too-few-public-methods
+        __slots__: Tuple[str, ...] = (
+            "performative",
+            "dialogue_reference",
+            "message_id",
+            "target",
+            "content",
```
DavidMinarsch, line 65: Not sure why we allow for the field `content` here. We don't want anyone to use the Message class directly. It should probably be made ABC...thoughts @marcofavorito ?

DavidMinarsch (commented):
LGTM, some minor points and discussion points

aea/protocols/base.py
solarw: @DavidMinarsch 
it was used in some test cases, anyway it's redefined for every Message class

aea/protocols/base.py
solarw: it was the previous implementation, so decided to stay as is, probably to be sure data passed will not be changed inside the message with side effects

aea/protocols/base.py
DavidMinarsch: So the test cases are wrong imo. The Message class is not to be used as it does not reside in a complete protocol.

marcofavorito: > Not sure why we allow for the field `content` here. We don't want anyone to use the Message class directly. It should probably be made ABC...thoughts @marcofavorito ?

I agree

marcofavorito (approved):
LGTM

aea/protocols/base.py
DavidMinarsch: @solarw please fix

solarw: @DavidMinarsch @marcofavorito 
I found that the class being tested is not used anywhere instead of this test, so it moved to test to cover base serializer.

aea/protocols/base.py
```diff
@@ -125,17 +138,20 @@ def _body(self) -> Dict:
...
     def _body(self, body: Dict) -> None:
         """
-        Set the body of hte message.
+        Set the body of the message.
 
         :param body: the body.
         :return: None
         """
-        self.__body = body
+        self._slots = self._SlotsCls()  # new instsance to clean up all data
```
DavidMinarsch, line 153:
```suggestion
        self._slots = self._SlotsCls()  # new instance to clean up all data
```

aea/protocols/base.py
```diff
@@ -266,90 +290,6 @@ class Serializer(Encoder, Decoder, ABC):
     """The implementations of this class defines a serialization layer for a protocol."""
 
 
-class ProtobufSerializer(Serializer):
```
DavidMinarsch, line 269: I assume this is not used anywhere now and therefore removed?

DavidMinarsch: @marcofavorito  safe to remove?

marcofavorito: well since we deleted the possibility to serialized `Message` object directly then I'd say yes.

marcofavorito (approved):
LGTM

DavidMinarsch (approved):
LGTM, great work!

PR #1924, opened 2020-11-11
AEA-969 Storage 

solarw:
## Proposed changes

generic storage

## Types of changes

## Checklist


aea/helpers/storage.py
```diff
@@ -0,0 +1,432 @@
+# -*- coding: utf-8 -*-
+# ------------------------------------------------------------------------------
+#
+#   Copyright 2018-2020 Fetch.AI Limited
+#
```
DavidMinarsch, line 5: Suggest splitting: all apart from sql lite goes into `aea/helpers/storage/base.py` and then specific adaptors go into `aea/helpers/storage/sqlite.py` etc

aea/protocols/base.py
```diff
@@ -81,6 +81,22 @@ def __init__(self, _body: Optional[Dict] = None, **kwargs):
         except Exception as e:  # pylint: disable=broad-except
             _default_logger.error(e)
 
+    def to_json(self) -> dict:
```
DavidMinarsch, line 84: just use existing encode and decode logic and create a generic JsonSerializer please

aea/aea.py
```diff
@@ -406,6 +407,7 @@ def log_exception(e, fn):
             self.logger.exception(f"<{e}> raised during `{fn}`")
 
         if self._skills_exception_policy == ExceptionPolicyEnum.propagate:
+            log_exception(exception, function)
```
DavidMinarsch, line 410: Please merge latest develop so we can see the true diff

aea/helpers/storage/backends/base.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def _check_collection_name(self, collection_name: str) -> None:
+        """
+        Check collection name is valid.
+
+        raises ValueError if bad collection name provided.
+        """
+        if not self.VALID_COL_NAME.match(collection_name):
+            raise ValueError(
+                f"Invalid collection name: {collection_name}, should contains ony a-z and _"
```
DavidMinarsch, line 43:
```suggestion
                f"Invalid collection name: {collection_name}, should contain only alpha-numeric characters and _"
```

aea/helpers/storage/backends/base.py
```diff
@@ -0,0 +1,108 @@
...
+        Remove object from the collection.
+
+        :param collection_name: str.
+        :param object_id: str object id
+
+        :return: None
+        """
+
+    @abstractmethod
+    async def find(self, collection_name: str, field: str, equals: Any) -> List[Dict]:
```
DavidMinarsch, line 99: can it really be type Any?

aea/helpers/storage/backends/sqlite.py
```diff
@@ -0,0 +1,162 @@
...
+        :param collection_name: str.
+        :param object_id: str object id
+
+        :return: dict if object exists in collection otherwise None
+        """
+        self._check_collection_name(collection_name)
+        sql = f"""SELECT object_body FROM {collection_name} WHERE object_id = ? LIMIT 1;"""
+        result = await self._executute_sql(sql, [object_id])
+        if result:
+            return json.loads(result[0][0])
```
DavidMinarsch, line 130: can this raise key error?

aea/runtime.py
```diff
@@ -81,10 +83,16 @@ def __init__(
         self._multiplexer: AsyncMultiplexer = self._get_multiplexer_instance()
         self._task_manager = TaskManager()
         self._decision_maker: Optional[DecisionMaker] = None
+        self._storage: Optional[Storage] = None
```
DavidMinarsch, line 86: Storage backend should be configurable via the AEA config, so please add as optional field in aea-config and pass down

aea/runtime.py
```diff
@@ -270,14 +278,26 @@ async def stop_runtime(self) -> None:
...
         """Run agent and starts multiplexer."""
         self._state.set(RuntimeStates.starting)
-        await asyncio.gather(self._start_multiplexer(), self._start_agent_loop())
+        await asyncio.gather(
+            self._start_multiplexer(), self._start_agent_loop(), self._start_storage()
+        )
+
+    async def _start_storage(self) -> None:
+        """Start storage component."""
+        self._storage = Storage("sqlite://:memory:", threaded=True)
```
DavidMinarsch, line 298: No magic values; needs to be configurable

tests/test_act_storage.py
```diff
@@ -0,0 +1,85 @@
...
+        self.counter = 0
+
+    def act(self) -> None:
+        """Make an action."""
+        if self.context.storage and self.context.storage.is_connected:
+            col = self.context.storage.get_sync_collection(self.COL_NAME)
+            col.put(self.OBJ_ID, self.OBJ_BODY)
+        self.counter += 1
+
+
```
DavidMinarsch, line 50: Let's have the same test in a handler

DavidMinarsch (commented):
LGTM, some modifications requested.

aea/helpers/storage/backends/sqlite.py
solarw: `if result` does the check required.

aea/helpers/storage/backends/base.py
solarw: fixed

aea/protocols/base.py
solarw: `from_json` logic can not be ported to generic serializer decode, cause it requires a class of a Message type to restore a message.

tests/test_act_storage.py
solarw: done

aea/runtime.py
solarw: done

aea/runtime.py
solarw: done

aea/helpers/storage/backends/binaries/README.txt
```diff
@@ -0,0 +1 @@
+json1.dll - is sqlite extension for windows python<3.9
```
DavidMinarsch, line 1: you need to make sure `.dll` is included in the MANIFEST.in so this gets distributed properly

aea/helpers/storage/backends/sqlite.py
```diff
@@ -0,0 +1,200 @@
...
+        Create collection if not exits.
+
+        :param collection_name: str.
+        :return: None
+        """
+        self._check_collection_name(collection_name)
+        sql = f"""CREATE TABLE IF NOT EXISTS {collection_name} (
+            object_id TEXT PRIMARY KEY,
+            object_body JSON1 NOT NULL)
+        """  # nosec
```
DavidMinarsch, line 122: use the appropriate way to avoid nosec

aea/helpers/storage/backends/sqlite.py
```diff
@@ -0,0 +1,200 @@
...
+
+        :param collection_name: str.
+        :param object_id: str object id
+        :param object_body: python dict, json compatible.
+        :return: None
+        """
+        self._check_collection_name(collection_name)
+        sql = f"""INSERT OR REPLACE INTO {collection_name} (object_id, object_body)
+            VALUES (?, ?);
+        """  # nosec
```
DavidMinarsch, line 139: see above

aea/helpers/storage/backends/sqlite.py
```diff
@@ -0,0 +1,200 @@
...
+        """
+        Get object from the collection.
+
+        :param collection_name: str.
+        :param object_id: str object id
+
+        :return: dict if object exists in collection otherwise None
+        """
+        self._check_collection_name(collection_name)
+        sql = f"""SELECT object_body FROM {collection_name} WHERE object_id = ? LIMIT 1;"""  # nosec
```
DavidMinarsch, line 152: see above

aea/helpers/storage/backends/sqlite.py
```diff
@@ -0,0 +1,200 @@
...
+        """
+        Remove object from the collection.
+
+        :param collection_name: str.
+        :param object_id: str object id
+
+        :return: None
+        """
+        self._check_collection_name(collection_name)
+        sql = f"""DELETE FROM {collection_name} WHERE object_id = ?;"""  # nosec
```
DavidMinarsch, line 168: see above

aea/helpers/storage/backends/sqlite.py
```diff
@@ -0,0 +1,200 @@
...
+        Get objects from the collection by filtering by field value.
+
+        :param collection_name: str.
+        :param field: field name to search: example "parent.field"
+        :param equals: value field should be equal to
+
+        :return: None
+        """
+        self._check_collection_name(collection_name)
+        sql = f"""SELECT object_body FROM {collection_name} WHERE json_extract(object_body, ?) = ?;"""  # nosec
```
DavidMinarsch, line 184: see above

aea/helpers/storage/backends/sqlite.py
```diff
@@ -0,0 +1,200 @@
...
+
+    async def list(self, collection_name: str) -> List[Tuple[str, JSON_TYPES]]:
+        """
+        List all objects with keys from the collection.
+
+        :param collection_name: str.
+        :return: Tuple of objects keys, bodies.
+        """
+        self._check_collection_name(collection_name)
+        sql = f"""SELECT object_id, object_body FROM {collection_name};"""  # nosec
```
DavidMinarsch, line 199: see above

aea/protocols/base.py
```diff
@@ -83,6 +84,26 @@ def __init__(self, _body: Optional[Dict] = None, **kwargs):
...
+    @classmethod
+    def from_json(cls, data: dict) -> "Message":
+        """Construct message instance from json data."""
+        instance = cls.decode(b64decode(data["body"]))
+        sender = data["sender"]
+        if sender:
+            instance.sender = sender
+        to = data["to"]
+        if to:
+            instance.to = to
```
DavidMinarsch, line 104: let's use .get instead

aea/protocols/dialogue/base.py
```diff
@@ -338,6 +339,52 @@ def __init__(
...
+
+        :param message_class: type of message used with this dialogue
+        :param data: dict with data exported with Dialogue.to_json() method
+
+        :return: Dialogue instance
+        """
+        obj = cls(
+            dialogue_label=DialogueLabel.from_json(data["dialogue_label"]),
+            message_class=message_class,
+            self_address=Address(data["self_address"]),
```
DavidMinarsch, line 377: This is the agent's address; we don't need to save it in storage

setup.py
```diff
@@ -141,4 +141,5 @@ def parse_readme():
         "Bug Reports": "https://github.com/fetchai/agents-aea/issues",
         "Source": "https://github.com/fetchai/agents-aea",
     },
+    packages_data={"aea": ["helpers/storage/backends/binaries/*.*"]},
```
DavidMinarsch, line 144: include in MANIFEST.in in instead

aea/protocols/dialogue/base.py
```diff
@@ -948,6 +995,220 @@ def add_dialogue_endstate(
             self._other_initiated[end_state] += 1
 
 
+def find_caller_object(object_type: Type):
```
DavidMinarsch, line 998: How reliable is this approach?

aea/helpers/storage/backends/sqlite.py
DavidMinarsch: https://security.openstack.org/guidelines/dg_parameterize-database-queries.html

aea/protocols/base.py
DavidMinarsch: Or at least catch key error and print a more useful exception, like data corrupted in db or something like this

aea/protocols/dialogue/base.py
```diff
@@ -338,6 +339,52 @@ def __init__(
         )
         self._message_class = message_class
 
+    def __eq__(self, other) -> bool:
+        """Compare two dialogues."""
+        return (
+            type(self) == type(other)  # pylint: disable=unidiomatic-typecheck
```
DavidMinarsch, line 345: why not use isinstance?

aea/protocols/dialogue/base.py
```diff
@@ -338,6 +339,52 @@ def __init__(
...
+            and self.dialogue_label == other.dialogue_label
+            and self.message_class == other.message_class
+            and self._incoming_messages == other._incoming_messages
+            and self._outgoing_messages == other._outgoing_messages
+            and self.role == other.role
```
DavidMinarsch, line 350: self_address should also be compared

aea/protocols/dialogue/base.py
```diff
@@ -338,6 +339,52 @@ def __init__(
...
+        )
+
+    def json(self) -> dict:
+        """Get json representation of the dialogue."""
+        data = {
+            "dialogue_label": self._dialogue_label.json,
+            "self_address": self.self_address,
+            "role": self._role.value,
+            "incoming_messages": [i.json() for i in self._incoming_messages],
+            "outgoing_messages": [i.json() for i in self._outgoing_messages],
```
DavidMinarsch, line 360: what about message class?

aea/protocols/dialogue/base.py
```diff
@@ -948,6 +995,220 @@ def add_dialogue_endstate(
...
+        """
+        Add dialogue to storage.
+
+        :param dialogue: dialogue to add.
+        :return: None
+        """
+        self._dialogues_by_dialogue_label[dialogue.dialogue_label] = dialogue
+        self._dialogue_by_address[
+            dialogue.dialogue_label.dialogue_opponent_addr
+        ].append(dialogue)
```
DavidMinarsch, line 1042: Would be cheaper to change this to `_dialogue_label_by_address` and then use the `_dialogues_by_dialogue_label` to retrieve the dialogues! Means, we only store dialogues once

aea/protocols/dialogue/base.py
```diff
@@ -948,6 +995,220 @@ def add_dialogue_endstate(
...
+        """
+        Get dialogue stored by it's label.
+
+        :param dialogue_label: label of the dialogue
+        :return: dialogue if presents or None
+        """
+        return self.dialogues.get(dialogue_label, None)
+
+    @property
+    def dialogues(self) -> Dict[DialogueLabel, Dialogue]:
```
DavidMinarsch, line 1067: Can we name this please `dialogues_by_dialogue_label` to keep it consistent

aea/protocols/dialogue/base.py
```diff
@@ -948,6 +995,220 @@ def add_dialogue_endstate(
...
+
+    def _get_collection_name(self) -> Optional[str]:
+        """Generate collection name based on the dialogues class name and skill component."""
+        if not self._skill_component:  # pragma: nocover
+            return None
+        return "_".join(
+            [
+                self._skill_component.skill_id.author,
+                self._skill_component.skill_id.name,
+                self._skill_component.name,
```
DavidMinarsch, line 1134: also skill component type should feature here to be sure?

aea/protocols/dialogue/base.py
```diff
@@ -966,11 +1227,8 @@ def __init__(
...
         """
-        self._dialogues_by_dialogue_label = {}  # type: Dict[DialogueLabel, Dialogue]
-        self._dialogue_by_address = {}  # type: Dict[Address, List[Dialogue]]
-        self._incomplete_to_complete_dialogue_labels = (
-            {}
-        )  # type: Dict[DialogueLabel, DialogueLabel]
+
+        self._dialogues_storage = PersistDialoguesStorage(self)
         self._self_address = self_address
         self._dialogue_stats = DialogueStats(end_states)
```
DavidMinarsch, line 1233: Let's discuss

aea/runtime.py
```diff
@@ -81,10 +83,22 @@ def __init__(
...
         self._decision_maker: Optional[DecisionMaker] = None
+        self._storage: Optional[Storage] = self._get_storage(agent)
 
         self._loop_mode = loop_mode or self.DEFAULT_RUN_LOOP
         self.main_loop: BaseAgentLoop = self._get_main_loop_instance(self._loop_mode)
 
+    @staticmethod
+    def _get_storage(agent) -> Optional[Storage]:
+        if agent.storage_uri:
+            return Storage(agent.storage_uri, threaded=True)
```
DavidMinarsch, line 94: should the mode be configurable?

aea/skills/base.py
```diff
@@ -323,6 +330,9 @@ def teardown(self) -> None:
 
         :return: None
         """
+        super_obj = super()
+        if hasattr(super_obj, "teardown"):
+            super_obj.teardown()  # type: ignore  # pylint: disable=no-member
```
DavidMinarsch, line 335: why are we adding this here in the abstract methods?

aea/protocols/base.py
solarw: updated with base64 and message.encode

aea/helpers/storage/backends/sqlite.py
solarw: @DavidMinarsch 
not possible in my case, table names do not support substitution as a parameter.
check collection name performed before every query.

aea/protocols/dialogue/base.py
```diff
@@ -1020,11 +1278,6 @@ def __init__(
         )
         self._role_from_first_message = role_from_first_message
 
-    @property
-    def dialogues(self) -> Dict[DialogueLabel, Dialogue]:
-        """Get dictionary of dialogues in which the agent engages."""
-        return self._dialogues_by_dialogue_label
```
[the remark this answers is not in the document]
solarw: it's not in use in any module, not sure should be public, can return it back if needed.

aea/protocols/dialogue/base.py
solarw: the type has to be right the same, isinstsance(anydialogue, Basedialogue) == True, but that's not correct.

aea/protocols/dialogue/base.py
solarw: nice point. fixed

aea/protocols/dialogue/base.py
solarw: done

aea/protocols/dialogue/base.py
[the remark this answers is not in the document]
solarw: also, I don't want to provide r/w access to this data, cause it can be modified directly with no proper control

aea/protocols/dialogue/base.py
solarw: added. nice point

aea/runtime.py
solarw: no, only threaded. sync calls from skills do not allow to share event loop.

aea/skills/base.py
solarw: I'm not sure it's not the most elegant way with this hasattr/super, but every place it was added it's required.

The main problem is multiple inheritances, so it can be resolved in various ways. depending on base classes order.

aea/protocols/dialogue/base.py
solarw: I don't like this solution, but it works for cases I tested.

Probably can try to find out something with a global variable like a context. probably will be a better solution.
only one point is to create  a thread local variable, but it's provided by python.

aea/protocols/base.py
solarw: done

aea/protocols/dialogue/base.py
solarw: not possible to serialize this, module and class path can be provided, but it's not as easy just to load it in every case.
in this case, the message class will be restored by the dialogues.

aea/protocols/dialogue/base.py
solarw: actually, it does not copy a dialogue object, it's just a reference. it does not dumped, and can be restored easily from dialogues.

also, this structure used in a couple of cases only, probably tac? so logic overview can be helpful.

setup.py
solarw: done

aea/protocols/dialogue/base.py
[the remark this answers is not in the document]
DavidMinarsch: Yes, I am ok with removal. Just a question

aea/runtime.py
DavidMinarsch: please add a comment in code so the next developer knows why it must be so.

aea/protocols/dialogue/base.py
DavidMinarsch: This functionality is essential even if currently it is just used in tac.

aea/runtime.py
solarw: done

DavidMinarsch (approved):
LGTM, a great feature to have :)

PR #1930, opened 2020-11-12
Add state/get-state to ledger_api protocol (AEA-1016)

jrriehl:
## Proposed changes

Added `get_state` and `state` performatives to ledger_api protocol, to allow for arbitrary methods to be called on the underlying ledger_api.

## Fixes

## Types of changes

## Checklist


aea/crypto/ethereum.py
```diff
@@ -294,6 +294,15 @@ def _try_get_balance(self, address: Address) -> Optional[int]:
...
         return self._api.eth.getBalance(address)  # pylint: disable=no-member
 
+    def get_block(self, block_id: Union[int, str] = "latest") -> Optional[Any]:
+        """Get the block header and metadata."""
+        return self._try_get_block(block_id)
+
+    @try_decorator("Unable to retrieve block: {}", logger_method="warning")
+    def _try_get_block(self, block_id: Union[int, str] = "latest") -> Optional[Any]:
+        """Get the block header and metadata."""
+        return self._api.eth.getBlock(block_id)  # pylint: disable=no-member
```
DavidMinarsch, line 304:
```suggestion
    def _try_get_block(self, block_id: Union[int, str] = "latest") -> Optional[Any]:
        """Get the block header and metadata."""
        args =  [block_id]
        kwargs = {}
        callable_name = getBlock
        callable = getattr(self._api.eth, callable_name)
        result = callable(*args, **kwargs)
        return result
```

DavidMinarsch: Of course, we would actually not hard code the args and kwargs ...see above

packages/fetchai/connections/ledger/ledger_dispatcher.py
```diff
@@ -135,6 +136,35 @@ def get_balance(
...
+        """
+        Send the request 'get_state'.
+
+        :param api: the API object.
+        :param message: the Ledger API message
+        :return: None
+        """
+        callable = getattr(api, message.callable)
+        args = tuple(message.kwargs.body.values())
+        data = callable(*args)
```
DavidMinarsch, line 151: If the ledger api has generic `get_state` than here we can be explicit and move the generics down the the ledger api (see above)

packages/fetchai/protocols/ledger_api/README.md
```diff
@@ -51,10 +62,12 @@ ct:TransactionReceipt: |
...
 ---
-initiation: [get_balance, get_raw_transaction, send_signed_transaction, get_transaction_receipt]
+initiation: [get_balance, get_state, get_raw_transaction, send_signed_transaction, get_transaction_receipt]
 reply:
   get_balance: [balance, error]
   balance: []
+  get_state: [state, error]
+  state: []
   get_raw_transaction: [raw_transaction, error]
   raw_transaction: [send_signed_transaction]
```
DavidMinarsch, line 72: termination is missing `state`

aea/crypto/base.py
```diff
@@ -248,6 +248,17 @@ def get_balance(self, address: Address) -> Optional[int]:
         :return: the balance.
         """
 
+    @abstractmethod
+    def get_block(self, block_id: Union[int,str]) -> Optional[Any]:
```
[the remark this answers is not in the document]
jrriehl: Absolutely, thanks for the clarification. The updated implementation is generalized as above.

aea/crypto/cosmos.py
```diff
@@ -397,6 +397,28 @@ def _try_get_balance(self, address: Address) -> Optional[int]:
...
+        return block
+
+    @try_decorator(
+        "Encountered exception when trying get block: {}",
+        logger_method=_default_logger.warning,
+    )
+    def _try_get_block(self, block_id: Union[int,str]) -> Optional[Any]:
+        """Try to get the block header and metadata."""
+        block = None  # type: Optional[int]
+        url = self.network_address + f"/block"
```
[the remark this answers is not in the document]
jrriehl: Done, it is now `_try_get_state`.

aea/crypto/base.py
```diff
@@ -248,6 +248,18 @@ def get_balance(self, address: Address) -> Optional[int]:
         :return: the balance.
         """
 
+    @abstractmethod
+    def get_state(self, callable_name: str, *args) -> Optional[Any]:
```
DavidMinarsch, line 252: for generality, probably add `kwargs`?

aea/crypto/base.py
```diff
@@ -248,6 +248,18 @@ def get_balance(self, address: Address) -> Optional[int]:
...
+        """
+        Call a specified function on the underlying ledger API.
+
+        This usually takes the form of a web request to be waited synchronously.
+
+        :param callable_name: the name of the API function to be called.
+        :param kwargs: the keyword arguments for the API function
```
DavidMinarsch, line 259: currently there are only args

aea/crypto/cosmos.py
```diff
@@ -397,6 +397,25 @@ def _try_get_balance(self, address: Address) -> Optional[int]:
...
+        return response
+
+    @try_decorator(
+        "Encountered exception when trying get state: {}",
+        logger_method=_default_logger.warning,
+    )
+    def _try_get_state(self, callable_name: str, *args) -> Optional[Any]:
+        """Try to call a function on the ledger API."""
+        result = None  # type: Optional[Any]
+        query = "/".join(args)
```
DavidMinarsch, line 412: are we sure this is correct and its now a parameter list or similar?

packages/fetchai/protocols/ledger_api/README.md
```diff
@@ -33,6 +33,13 @@ speech_acts:
     transaction_digest: ct:TransactionDigest
   transaction_receipt:
     transaction_receipt: ct:TransactionReceipt
+  get_state:
+    ledger_id: pt:str
+    callable: pt:str
+    args: ct:Args
```
DavidMinarsch, line 39: Args is a list of strings, so why not: `pt:list[pt:str]`?

DavidMinarsch: But we do probably also want `Kwargs` (see above)

packages/fetchai/protocols/ledger_api/custom_types.py
```diff
@@ -31,3 +36,63 @@
 Terms = BaseTerms
 TransactionDigest = BaseTransactionDigest
 TransactionReceipt = BaseTransactionReceipt
+State = BaseState
+
+
+class Args:
```
DavidMinarsch, line 42: Args must be string in python, so don't think we need this

tests/test_packages/test_protocols/test_ledger_api.py
```diff
@@ -78,6 +78,47 @@ def test_get_balance_serialization():
...
+    assert expected_envelope.sender == actual_envelope.sender
+    assert expected_envelope.protocol_id == actual_envelope.protocol_id
+    assert expected_envelope.message != actual_envelope.message
+
+    actual_msg = LedgerApiMessage.serializer.decode(actual_envelope.message)
+    actual_msg.to = actual_envelope.to
+    actual_msg.sender = actual_envelope.sender
+    expected_msg = msg
+    print(msg)
+    print(actual_msg)
```
DavidMinarsch, line 118: remove please

DavidMinarsch (commented):
LGTM, only thing to change imo is the args/kwargs

aea/crypto/cosmos.py
jrriehl: This is based on the cosmos REST API specification here: https://cosmos.network/rpc/v0.38.3, which takes a path as defined above. I took the convention of defining the root of the path (`txs`, `blocks`, etc.) as the `callable_name` and the rest of the path as arguments, partly to keep some consistency with the ethereum ledger interface, but it could also be handled as a simple list of arguments.

packages/fetchai/protocols/contract_api/custom_types.py
```diff
@@ -31,64 +28,4 @@
 RawMessage = BaseRawMessage
 RawTransaction = BaseRawTransaction
 State = BaseState
-
```
jrriehl, line 34: Moved `Kwargs` to `aea.helpers.transaction.base` because it is now imported by multiple packages.

aea/crypto/cosmos.py
DavidMinarsch: It'd probably move that description into line 401 so we have it near the code for future devs.

aea/crypto/ethereum.py
```diff
@@ -294,6 +294,22 @@ def _try_get_balance(self, address: Address) -> Optional[int]:
...
+        return response
+
+    @try_decorator("Unable to get state: {}", logger_method="warning")
+    def _try_get_state(  # pylint: disable=unused-argument
+        self, callable_name: str, *args, **kwargs
+    ) -> Optional[Any]:
+        """Try to call a function on the ledger API."""
+
+        function = getattr(self._api.eth, callable_name)
+        response = function(*args)
```
DavidMinarsch, line 309:
```suggestion
        response = function(*args, **kwargs)
```

packages/fetchai/protocols/contract_api/custom_types.py
DavidMinarsch: In general this would be the right approach. But for this particular class, I prefer it if we have replication. The reason: "kwargs" class should really not be a class and just a dict with arbitrary primitive types inside. The problem: protocol generator does not yet support it. If we include it in the protocols, it's easier to upgrade in the future as we just change the protocol rather than an AEA class. So please just replicate across custom_types of the different protocols.

aea/helpers/transaction/base.py
```diff
@@ -402,6 +402,67 @@ def __str__(self):
         return "State: ledger_id={}, body={}".format(self.ledger_id, self.body)
 
 
+class Kwargs:
```
DavidMinarsch, line 405: See comment below

tests/test_packages/test_connections/test_ledger/test_ledger_api.py
```diff
@@ -142,6 +143,59 @@ async def test_get_balance(
...
+    await asyncio.sleep(0.01)
+    response = await ledger_apis_connection.receive()
+
+    assert response is not None
+    assert type(response.message) == LedgerApiMessage
+    response_msg = cast(LedgerApiMessage, response.message)
+    response_dialogue = ledger_api_dialogues.update(response_msg)
+    assert response_dialogue == ledger_api_dialogue
+
+    print(response_msg)
```
DavidMinarsch, line 189: use logger not print

5A11 (approved):
LGTM

Issue #1935, opened 2020-11-13
MultiAgentManager: implement save-load functionality


panasevychol:
**Is your feature request related to a problem? Please describe.**
Currently MAM save-load functionality is missing. That is required for AEA GUI.

**Describe the solution you'd like**
Save-load on start-stop.

PR #1936, opened 2020-11-13
Save-load functionality of MultiAgentManager implemented incl. tests AEA-1061

panasevychol:
## Proposed changes

Save-load functionality of MultiAgentManager implemented incl. tests. #1935 

## Types of changes

## Checklist


aea/configurations/project.py
```diff
@@ -60,10 +60,12 @@ def load(
         ctx.set_config("skip_consistency_check", skip_consistency_check)
         path = os.path.join(working_dir, public_id.author, public_id.name)
         target_dir = os.path.join(public_id.author, public_id.name)
-        if is_local:
-            fetch_agent_locally(ctx, public_id, target_dir=target_dir)
-        else:
-            fetch_agent(ctx, public_id, target_dir=target_dir)
+
+        if not os.path.exists(target_dir):
```
DavidMinarsch, line 64: Better to explicitly enable this skip via a default kwarg. Otherwise no error is thrown when we want to.

aea/manager.py
```diff
@@ -37,6 +38,42 @@
...
+        }
+
+
+class MAMState:
+    """MultiAgentManager state."""
+
+    def __init__(self, projects: List[PublicId], agents: List[MAMAgentState]):
+        """Initialize MultiAgentManager state."""
+        self.projects = projects
+        self.agents = agents
```
DavidMinarsch, line 66:
```suggestion
        self.agent_states = agent_states
```
similarly elsewhere

aea/manager.py
```diff
@@ -607,3 +664,53 @@ def _create_private_key(self, name, ledger) -> str:
...
+
+        :return: None
+        :raises: ValueError if failed to load state.
+        """
+        if not os.path.exists(self._save_path):
+            return
+
+        save_json = {}
+        with open(self._save_path) as f:
+            save_json = json.load(f)
```
DavidMinarsch, line 682: can throw an error..handle it

aea/manager.py
```diff
@@ -607,3 +664,53 @@ def _create_private_key(self, name, ledger) -> str:
...
+        except ValueError as e:
+            raise ValueError(f"Failed to load state. {e}")
+
+    def _save_state(self) -> None:
+        """
+        Save MultiAgentManager state.
+
+        :return: None.
+        """
+        agents_state = [
```
DavidMinarsch, line 706:
```suggestion
        agent_states = [
```

aea/manager.py
```diff
@@ -607,3 +664,53 @@ def _create_private_key(self, name, ledger) -> str:
...
+            MAMAgentState(
+                public_id=alias.project.public_id,
+                agent_name=alias.agent_name,
+                config=alias.config,
+            )
```
DavidMinarsch, line 711: Why don't we just use Alias class and make it possible to save it to json?

aea/manager.py
```diff
@@ -524,23 +579,25 @@ def _build_agent_alias(
...
+        config: Optional[List[dict]] = None,
     ) -> AgentAlias:
         """Create agent alias for project, with given name and overrided values."""
-        json_config = self._make_config(
-            project.path, agent_overrides, component_overrides
-        )
+        if not config:
+            config = self._make_config(
+                project.path, agent_overrides, component_overrides
+            )
```
DavidMinarsch, line 588: what if config and overrides are misaligned?

aea/manager.py
```diff
@@ -211,25 +251,31 @@ def add_error_callback(
...
         self._is_running = True
         self._thread = Thread(target=self._run_thread, daemon=True)
         self._thread.start()
         self._started_event.wait(self.DEFAULT_TIMEOUT_FOR_BLOCKING_OPERATIONS)
         return self
 
-    def stop_manager(self) -> "MultiAgentManager":
+    def stop_manager(
+        self, cleanup: bool = True, save: bool = False
+    ) -> "MultiAgentManager":
```
DavidMinarsch, line 270: booleans should be named with `is_`,`has_` or `can_`

aea/manager.py
panasevychol: Changed to `agent_aliases`.

aea/manager.py
panasevychol: Which kind of error? If file doesn't exist it creates one, if exists - overwrites one.

aea/manager.py
panasevychol: Changed to `agent_aliases`.

aea/manager.py
panasevychol: The full config is passed only on state recreation. We override if no full config passed here.

aea/manager.py
panasevychol: There're a lot of bool params here named `local` or `cleanup`. IMO `save` option sounds consistent on this case.

aea/manager.py
DavidMinarsch: The problem is, in `add_agent`, a public method, I can now pass inconsistent data. This needs to be avoided. There should be a check that either `agent_overrides` and `component_overrides` are passed (one optionally none) or `config`.

aea/manager.py
```diff
@@ -337,6 +376,7 @@ def add_agent(
             agent_name=agent_name,
             agent_overrides=agent_overrides,
             component_overrides=component_overrides,
+            config=config,
```
DavidMinarsch, line 379: Not ok, checks need to be added so config or overrides can be passed, not both

aea/manager.py
```diff
@@ -309,6 +346,7 @@ def add_agent(
         agent_name: Optional[str] = None,
         agent_overrides: Optional[dict] = None,
         component_overrides: Optional[List[dict]] = None,
+        config: Optional[List[dict]] = None,
```
DavidMinarsch, line 349: see below

aea/manager.py
```diff
@@ -37,6 +38,23 @@
 from aea.crypto.helpers import create_private_key
 
 
+class MAMState:
```
DavidMinarsch, line 41: I think a function or helper method to convert to dict is enough, we don't need this object instantiated as far as I can tell.

aea/configurations/project.py
DavidMinarsch: So this is not addressed yet.

panasevychol: My bad, now fixed.

aea/manager.py
panasevychol: Agree. Done.

aea/manager.py
panasevychol: Done.

aea/manager.py
panasevychol: Fine. Done.

DavidMinarsch (approved):
LGTM

aea/configurations/project.py
```diff
@@ -61,10 +62,13 @@ def load(
         ctx.set_config("skip_consistency_check", skip_consistency_check)
         path = os.path.join(working_dir, public_id.author, public_id.name)
         target_dir = os.path.join(public_id.author, public_id.name)
-        if is_local:
-            fetch_agent_locally(ctx, public_id, target_dir=target_dir)
-        else:
-            fetch_agent(ctx, public_id, target_dir=target_dir)
+
```
solarw, line 65: `path` has to be checked with os.path.exists

solarw (approved):
LGTM

aea/configurations/project.py
panasevychol: It's checked in `fetch_agent` and if we trying to fetch new project (not to restore) that will fail with an error which is an expected behavior.

panasevychol: If it's absent for some reason but `is_restore == True` we proceed to fetch it.

Issue #1923, opened 2020-11-10
Address multiple development flow issues


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
Multiple issues have been collected:
- [x] setting up working dir for development with an IDE or similar
- [x] interaction between agent and local registry not clearly described
- [x] cli command for automated dev setup?

**Describe the solution you'd like**
Address above issues.

DavidMinarsch:
- actions not currently supported:
  - [ ] vendor package push to local packages
  - [ ] fetch/add directly to packages
  - [ ] upgrade packages
- package import paths not found by IDE
- update contrib details/main readme

PR #1942, opened 2020-11-14
add symlink during scaffolding

DavidMinarsch:
## Proposed changes

This PR proposes adding a symlink from `vendor` to `non-vendor` packages during scaffolding. This, together with a symlink with `touch packages && ln -s vendor packages` in AEA dir will mean all import paths are detected.

Still to do:
- make sure local push is adjusted to still work

## Fixes

- #1923 (partially)

## Types of changes

## Checklist

## Further comments

-


aea/helpers/sym_link.py
```diff
@@ -0,0 +1,94 @@
...
+#
+# ------------------------------------------------------------------------------
+"""Sym link implementation for Linux, MacOS, and Windows."""
+
+import contextlib
+import os
+from functools import reduce
+from pathlib import Path
+
+
```
DavidMinarsch, line 27: straight copy from scripts

marcofavorito: OK

pbukva:
@DavidMinarsch  I know that this is being used only for CI at the moment, so just comment: Simlinks are highly dependent on the File System and OS. The File System being the biggest potential hurdle here (since all relevant modern OSes with support symlinks).Issue w

aea/cli/fetch.py
```diff
@@ -142,7 +150,7 @@ def _fetch_agent_deps(ctx: Context) -> None:
     :return: None
     :raises: ClickException re-raises if occurs in add_item call.
     """
-    for item_type in ("protocol", "contract", "connection", "skill"):
+    for item_type in (SKILL, PROTOCOL, CONNECTION, CONTRACT):
```
DavidMinarsch, line 153: fix

DavidMinarsch:
> @DavidMinarsch I know that this is being used only for CI at the moment, so just comment: Simlinks are highly dependent on the File System and OS. The File System being the biggest potential hurdle here (since all modern and relevant OSes support symlinks at least for directories).
> Issue can pop specially on external disk drives which are usually formatted using FS with with NO symlink support (e.g. exFAT), and network drives (e.g. nfs, smb, etc ..., where FS server can be configured that way that creation symlinks via network api might be disabled, or not support at all). Also, Windows OS has `Create symbolic link` policy which needs to be enabled for groups/users in order to make creation of symlinks working ...

It's used for development related operations on CLI only. So these problems are unlikely to present themselves in this form.

Issue #1846, opened 2020-10-20
If standalone agent is used connections get added twice to multiplexer


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
gym ex: run, connections added twice in multiplexer!

**Describe the solution you'd like**
Fix and refactor connections adding

PR #1970, opened 2020-11-19
fix for multiplexer setup connections replace

solarw:
## Proposed changes

connection added twice in gym ex. caused cause unproper cleanup on multiplexer.setup

## Fixes

https://github.com/fetchai/agents-aea/issues/1846

## Types of changes

## Checklist

### DIFF
--- a/aea/multiplexer.py
+++ b/aea/multiplexer.py
@@ -612,7 +612,11 @@ class AsyncMultiplexer(Runnable, WithLogger):
         :return: None.
         """
         self.default_routing = default_routing or {}
+
+        # replace connections
         self._connections = []
+        self._id_to_connection = {}
+
         for c in connections:
             self.add_connection(c, c.public_id == default_connection)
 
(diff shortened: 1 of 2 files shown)


aea/multiplexer.py
```diff
@@ -612,7 +612,11 @@ def setup(
         :return: None.
         """
         self.default_routing = default_routing or {}
+
+        # replace connections
         self._connections = []
+        self._id_to_connection = {}
```
DavidMinarsch, line 618: How is this fixing it? Can you give any context @solarw ?

solarw: setup method replaces all connections added before,  `self._id_to_connection` is just mapping for `self._connections`
it's populated with add_connection, but not reset on setup.

i think `setup` method quite tricky, would be nice to get rid out of it.

DavidMinarsch: Thanks!

DavidMinarsch (approved):
LGTM

PR #1988, opened 2020-11-23
Add basic oracle skills and oracle contract

jrriehl:
## Proposed changes

The idea is to provide some basic skills that can be used to build oracle agents, including fetching data from relevant APIs and deploying and updating values in smart contracts. This is an initial version which will be expanded to cover more use cases and to integrate with oracle agent networks.

## Types of changes

## Checklist

## Further comments

Documentation in progress.


packages/fetchai/skills/coin_price/behaviours.py
```diff
@@ -0,0 +1,109 @@
...
+        coin_id = model.coin_id
+        currency = model.currency
+
+        self.context.logger.info(
+            f"Fetching price of {coin_id} in {currency} from CoinPrice"
+        )
+
+        query = f"simple/price?ids={coin_id}&vs_currencies={currency}"
+
+        self.send_http_request_message("GET", url + query)
```
DavidMinarsch, line 101:
```suggestion
        model = cast(CoinPriceModel, self.context.coin_price_model)

        self.context.logger.info(
            f"Fetching price of {model.coin_id} in {model.currency} from CoinPrice"
        )

        url = f"{model.url}simple/price?ids={model.coin_id}&vs_currencies={model.currency}"

        self.send_http_request_message("GET", url)
```

packages/fetchai/skills/coin_price/handlers.py
```diff
@@ -0,0 +1,192 @@
...
+
+from aea.mail.base import EnvelopeContext
+from aea.protocols.base import Message
+from aea.skills.base import Handler
+
+from packages.fetchai.protocols.http.message import HttpMessage
+from packages.fetchai.skills.coin_price.dialogues import HttpDialogue, HttpDialogues
+
+
+# Skill can be run with or without http_server
```
DavidMinarsch, line 33: I'd prefer this was explicitly configured on the skill config and then the configuration was cross-checked with the values here to avoid surprises.

packages/fetchai/skills/coin_price/handlers.py
```diff
@@ -0,0 +1,192 @@
...
+                    self.context.logger.info("failed to get price: no price listed")
+                else:
+                    oracle_data = {
+                        "value": value,
+                        "decimals": model.decimals,
+                    }
+                    self.context.shared_state["oracle_data"] = oracle_data
+                    self.context.logger.info(
+                        f"{model.coin_id} price = {price} {model.currency}"
+                    )
```
DavidMinarsch, line 103: extract into private helper method `_handle_response`

packages/fetchai/skills/coin_price/handlers.py
```diff
@@ -0,0 +1,192 @@
...
+        :return: None
+        """
+        http_response = http_dialogue.reply(
+            performative=HttpMessage.Performative.RESPONSE,
+            target_message=http_msg,
+            version=http_msg.version,
+            status_code=200,
+            status_text="Success",
+            headers=http_msg.headers,
+            body=json.dumps(self.context.shared_state["oracle_data"]).encode("utf-8"),
```
DavidMinarsch, line 148: what if it's not there, you'll get a key error here. use `get` instead

packages/fetchai/skills/coin_price/handlers.py
```diff
@@ -0,0 +1,192 @@
...
+            "received invalid message={}, unidentified dialogue.".format(msg)
+        )
+
+    def teardown(self) -> None:
+        """
+        Teardown the handler.
+
+        :return: None
+        """
+        self.context.logger.info("tearing down HttpHandler")
```
DavidMinarsch, line 192: We avoid these now on the info level. It makes it too noisy. On the debug level there is already a log for each teardown call. So `pass` is probably best here.

packages/fetchai/skills/fetch_beacon/behaviours.py
```diff
@@ -0,0 +1,107 @@
...
+
+        # http request message
+        request_http_message, _ = http_dialogues.create(
+            counterparty=counterparty_address,
+            performative=HttpMessage.Performative.REQUEST,
+            method=method,
+            url=url,
+            headers="",
+            version="",
+            body=b"" if content is None else json.dumps(content).encode("utf-8"),
```
DavidMinarsch, line 72: content per your typing is never none; so ..

packages/fetchai/skills/coin_price/dialogues.py
```diff
@@ -0,0 +1,60 @@
...
+        def role_from_first_message(  # pylint: disable=unused-argument
+            message: Message, receiver_address: Address
+        ) -> BaseDialogue.Role:
+            """Infer the role of the agent from an incoming/outgoing first message
+
+            :param message: an incoming/outgoing first message
+            :param receiver_address: the address of the receiving agent
+            :return: The role of the agent
+            """
+            return BaseHttpDialogue.Role.CLIENT
```
DavidMinarsch, line 54: The role here depends on whether you're sending or receiving. If you're sending a HTTP request, you're a client, otherwise (receiving a HTTP request) a server. This is because in the skill you use both client and server

packages/fetchai/skills/fetch_beacon/handlers.py
```diff
@@ -0,0 +1,127 @@
...
+                    "entropy": keccak256(entropy.encode("utf-8")),
+                    "block_hash": bytes.fromhex(block_hash),
+                    "block_height": block_height,
+                }
+                self.context.logger.info(
+                    "Beacon info: "
+                    + str({"block_height": block_height, "entropy": entropy})
+                )
+                self.context.shared_state["oracle_data"] = beacon_data
+            return
```
DavidMinarsch, line 103: extract into helper method

packages/fetchai/skills/simple_oracle/behaviours.py
```diff
@@ -0,0 +1,232 @@
...
+        # Check for entropy value from fetch oracle skill
+        oracle_data = self.context.shared_state.get("oracle_data", None)
+        if oracle_data is None:
+            self.context.logger.info("No oracle value to publish")
+        else:
+            self.context.logger.info("Publishing oracle value")
+
+            # add expiration block
+            update_args = oracle_data.copy()
+            update_args["expiration_block"] = 1000000000000000
```
DavidMinarsch, line 97: should this be configurable maybe?

packages/fetchai/skills/simple_oracle/behaviours.py
```diff
@@ -0,0 +1,232 @@
...
+        contract_api_msg, contract_api_dialogue = contract_api_dialogues.create(
+            counterparty=str(LEDGER_API_ADDRESS),
+            performative=ContractApiMessage.Performative.GET_DEPLOY_TRANSACTION,
+            ledger_id=strategy.ledger_id,
+            contract_id=str(CONTRACT_PUBLIC_ID),
+            callable="get_deploy_transaction",
+            kwargs=ContractApiMessage.Kwargs(
+                {
+                    "deployer_address": self.context.agent_address,
+                    "ERC20Address": "0x0000000000000000000000000000000000000000",
```
DavidMinarsch, line 120: if meant to be hardcoded, declare as constant, otherwise make configurable.

packages/fetchai/skills/simple_oracle/behaviours.py
```diff
@@ -0,0 +1,232 @@
...
+                    "initialFee": strategy.initial_fee_deploy,
+                    "gas": 100000,
```
DavidMinarsch, line 122: same here

DavidMinarsch (approved):
Looks good, I have a few comments I'd like you to address before we can merge.

packages/fetchai/skills/simple_oracle/behaviours.py
jrriehl: Yes, good catch, I've meaning to add this and also the address below to the config.

marcofavorito (approved):
LGTM

packages/fetchai/contracts/oracle/contract.py
```diff
@@ -0,0 +1,130 @@
...
+class FetchOracleContract(Contract):
+    """The Fetch oracle contract."""
+
+    @classmethod
+    def get_grant_role_transaction(
+        cls,
+        ledger_api: LedgerApi,
+        contract_address: Address,
+        oracle_address: Address,
+        gas: int = 30000,
```
marcofavorito, line 49: might be good to put this in a top-module variable

packages/fetchai/contracts/oracle/contract.py
```diff
@@ -0,0 +1,130 @@
...
+
+    @classmethod
+    def get_update_transaction(
+        cls,
+        ledger_api: LedgerApi,
+        contract_address: Address,
+        oracle_address: Address,
+        update_function: str,
+        update_args: Dict[str, Any],
+        gas: int = 50000,
```
marcofavorito, line 84: see above

packages/fetchai/contracts/oracle/contract.py
```diff
@@ -0,0 +1,130 @@
...
+        try:
+            # try estimate the gas and update the transaction dict
+            gas_estimate = ledger_api.api.eth.estimateGas(transaction=tx)
+            tx["gas"] = gas_estimate
+        except Exception as e:  # pylint: disable=broad-except
+            _default_logger.debug(
+                "[OracleContract]: Error when trying to estimate gas: {}".format(e)
+            )
+
+        return tx
```
marcofavorito, line 130: @DavidMinarsch this is a code similar to other contracts, I think it's an opportunity for refactoring

packages/fetchai/skills/fetch_beacon/behaviours.py
```diff
@@ -0,0 +1,107 @@
...
+    """This class provides a simple beacon fetch behaviour."""
+
+    def __init__(self, **kwargs):
+        """Initialize the beacon fetch behaviour."""
+
+        super().__init__(**kwargs)
+        self.beacon_url = kwargs.pop("beacon_url", DEFAULT_URL)
+
+    def send_http_request_message(
+        self, method: str, url: str, content: Dict = None
```
marcofavorito, line 47: probably you meant `content: Optional[Dict]`

packages/fetchai/skills/simple_oracle/skill.yaml
```diff
@@ -0,0 +1,62 @@
...
+  signing_dialogues:
+    args: {}
+    class_name: SigningDialogues
+  strategy:
+    args:
+      contract_address: null
+      default_gas_deploy: 100000
+      default_gas_grant_role: 100000
+      default_gas_update: 150000
+      erc20_address: '0x0000000000000000000000000000000000000000'
```
marcofavorito, line 54: what is the meaning of this? Why `0x00...0`?

packages/fetchai/contracts/oracle/contract.py
jrriehl: Or maybe we should set the default limit to zero to make them fail if no gas limit is explicitly set? Thoughts @marcofavorito @DavidMinarsch ?

packages/fetchai/skills/simple_oracle/skill.yaml
jrriehl: Good question, maybe I can add a comment. The reason is that there is an explicit check for this in the contract. This value indicates that the payment for the oracle value will be in ETH, while a valid FET ERC20 contract address means the payment will be in FET.

marcofavorito: Thanks for the explanation, makes sense. 
> The reason is that there is an explicit check for this value in the contract.
We could have a check for `null`, `"null"`, `None` or equivalent, assuming Parity has an analogous type. That would make this part self-documenting. However, I don't think it's a priority, just a thought I had.

You could maybe add a comment just above that line, or at the end of the line, something like:
```suggestion
      # address of the ERC20 token contract to use for payment to the oracle.
      # Default: use the native currency ETH.
      erc20_address: '0x0000000000000000000000000000000000000000'
```

packages/fetchai/contracts/oracle/contract.py
marcofavorito: It's not a bad idea. One advantage is that things will not work "by chance" if you forget to pass the `gas` parameter.

jrriehl: Yes, that's what I was thinking too.

packages/fetchai/contracts/oracle/contract.py
DavidMinarsch: yes

packages/fetchai/skills/coin_price/dialogues.py
```diff
@@ -0,0 +1,64 @@
...
+            message: Message, receiver_address: Address
+        ) -> BaseDialogue.Role:
+            """Infer the role of the agent from an incoming/outgoing first message
+
+            :param message: an incoming/outgoing first message
+            :param receiver_address: the address of the receiving agent
+            :return: The role of the agent in this dialogue
+            """
+            if message.performative == HttpMessage.Performative.REQUEST:
+                return BaseHttpDialogue.Role.SERVER
```
DavidMinarsch, line 56: actually, we need a bit more. Because this is used for incoming and outgoing. 
comparing `message.sender`  and `receiver_address` will tell you what scenario you're in.

scripts/check_package_versions_in_docs.py
```diff
@@ -110,11 +110,25 @@ def get_public_id_from_yaml(configuration_file: Path):
 
     :param configuration_file: the path to the config yaml
     """
+
     data = unified_yaml_load(configuration_file)
-    author = data["author"]
+
+    # handle api spec files
+    if "openapi" in data:
+        return PublicId("author", "name", "0.0.0")
```
DavidMinarsch, line 118: We can do something better, we know there are only 5 valid config file names:
`aea-config.yaml`, `skill.yaml`,`connection.yaml`, `contract.yaml` and `protocol.yaml`. So let's check for inclusion and return otherwise. This should be done in the generator on line 126

scripts/check_package_dependencies.py
```diff
@@ -82,11 +82,25 @@ def get_public_id_from_yaml(configuration_file: Path):
 
     :param configuration_file: the path to the config yaml
     """
+
     data = unified_yaml_load(configuration_file)
-    author = data["author"]
+
+    # handle api spec files
```
DavidMinarsch, line 88: see below.

packages/fetchai/agents/coin_price_feed/aea-config.yaml
```diff
@@ -0,0 +1,37 @@
...
+skills:
+- fetchai/coin_price:0.1.0
+- fetchai/error:0.9.0
+default_connection: fetchai/http_server:latest
+default_ledger: fetchai
+default_routing:
+  fetchai/http:latest: fetchai/http_client:latest
+connection_private_key_paths: {}
+private_key_paths:
+  fetchai: fetchai_private_key.txt
```
DavidMinarsch, line 28: This needs to be:
```suggestion
private_key_paths: {}
```

packages/fetchai/agents/coin_price_feed/aea-config.yaml
```diff
@@ -0,0 +1,37 @@
+agent_name: coin_price_feed
+author: fetchai
+version: 0.1.0
+license: Apache-2.0
+description: An AEA providing a coin price feed.
+aea_version: 0.7.4
```
DavidMinarsch, line 6: This needs to be a range, see other agents for an example

packages/fetchai/agents/coin_price_oracle/aea-config.yaml
```diff
@@ -0,0 +1,53 @@
+agent_name: coin_price_oracle
+author: fetchai
+version: 0.1.0
+license: Apache-2.0
+description: An AEA providing a coin price oracle service.
+aea_version: 0.7.4
```
DavidMinarsch, line 6: see above

packages/fetchai/agents/coin_price_oracle/aea-config.yaml
```diff
@@ -0,0 +1,53 @@
...
+- fetchai/simple_oracle:0.1.0
+default_connection: fetchai/p2p_libp2p:latest
+default_ledger: ethereum
+default_routing:
+  fetchai/contract_api:latest: fetchai/ledger:latest
+  fetchai/http:latest: fetchai/http_client:latest
+  fetchai/ledger_api:latest: fetchai/ledger:latest
+connection_private_key_paths: {}
+private_key_paths:
+  ethereum: ethereum_private_key.txt
```
DavidMinarsch, line 35: see above

packages/fetchai/skills/coin_price/dialogues.py
```diff
@@ -0,0 +1,67 @@
...
+            """Infer the role of the agent from an incoming/outgoing first message
+
+            :param message: an incoming/outgoing first message
+            :param receiver_address: the address of the receiving agent
+            :return: The role of the agent in this dialogue
+            """
+            if (
+                message.performative == HttpMessage.Performative.REQUEST
+                and message.sender != receiver_address
+            ):
```
DavidMinarsch, line 58:
```suggestion
            if (
                message.performative == HttpMessage.Performative.REQUEST
                and message.sender != receiver_address
            ) or (message.performative == HttpMessage.Performative.RESPONSE and message.sender == receiver_address):
```

packages/fetchai/skills/coin_price/handlers.py
```diff
@@ -0,0 +1,202 @@
...
+from aea.protocols.base import Message
+from aea.skills.base import Handler
+
+from packages.fetchai.protocols.http.message import HttpMessage
+from packages.fetchai.skills.coin_price.dialogues import HttpDialogue, HttpDialogues
+from packages.fetchai.skills.coin_price.models import USE_HTTP_SERVER
+
+
+# Skill can be run with or without http_server
+if USE_HTTP_SERVER:
```
DavidMinarsch, line 35: This should be done dynamically in the model. Here it is now hard coded.

packages/fetchai/skills/fetch_beacon/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+# ------------------------------------------------------------------------------
+
+"""This module contains the implementation of the FetchBeacon skill."""
```
DavidMinarsch, line 20: Missing public id here (see other skills)

packages/fetchai/skills/coin_price/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+# ------------------------------------------------------------------------------
+
+"""This module contains the implementation of the CoinPrice skill."""
```
DavidMinarsch, line 20: Missing public id here, see other skills

packages/fetchai/skills/simple_oracle/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+# ------------------------------------------------------------------------------
+
+"""This module contains the implementation of the default skill."""
```
DavidMinarsch, line 20: Missing public id here

PR #2002, opened 2020-11-26
AEA-1092 add sequential tx processing to generic buyer

### LINKED ISSUE #1998: AW3 Transaction Fails with `Unknown error: unauthorized: signature verification failed; verify correct account sequence and chain-id`
**Describe the bug**
Seller agent on AW3 is registered and created using https://docs.fetch.ai/i_nets/quickstart-aw1/. No test fet is deposited into account.
Seller agent is deployed as a mobility agent with the following strategy configs in the config.yaml 
      ```
      service_data:
        key: seller_service
        value: mobility_data
      service_id: mobility_data
      shared_state_key: mobility_data
      unit_price: 5  
      ```
Agent does not receive transaction but agent explorer shows error when buyer tries to send transaction to mobility agent.

**To Reproduce**
1. Register agent using  https://docs.fetch.ai/i_nets/quickstart-aw1/
2. Change aea-config.yaml to 
`
public_id: dassy23/simple_seller:0.1.0
type: skill
models:
  strategy:
    args:
      location:
        latitude: 52.52
        longitude: 13.405
      service_data:
        key: seller_service
        value: mobility_data
      service_id: mobility_data
      shared_state_key: mobility_data
      unit_price: 5
`
3. Add mobility url to aea-config.yaml

`
public_id: dassy23/simple_data_request:0.1.0
type: skill
behaviours:
  http_request:
    args:
      request_interval: 20
      body: ''
      method: get
      url: 'YOUR URL
handlers:
  http:
    args:
      shared_state_key: mobility_data
`
4. Run and wait for transaction.

**Expected behavior**
Transaction should be received by agent

**Screenshots**
Screen Shot 2020-11-25 at 10 31 48 AM

**Additional context**
Screen Shot 2020-11-25 at 11 09 21 AM

DavidMinarsch:
## Proposed changes

This PR introduces a quick fix in the generic buyer to make tx submissions to the ledger sequential. Without this fix, tx after the first in a sequence of closely spaced txs can fail

## Fixes

#1998

## Types of changes

## Checklist

## Further comments

-


packages/fetchai/skills/generic_buyer/handlers.py
```diff
@@ -215,9 +216,11 @@ def _handle_match_accept(
             ledger_api_dialogue = cast(LedgerApiDialogue, ledger_api_dialogue)
             ledger_api_dialogue.associated_fipa_dialogue = fipa_dialogue
             fipa_dialogue.associated_ledger_api_dialogue = ledger_api_dialogue
-            self.context.outbox.put_message(message=ledger_api_msg)
-            self.context.logger.info(
```
DavidMinarsch, line 219: We now properly submit them in order; waiting till the previous one is finished or timed-out

packages/fetchai/skills/generic_buyer/behaviours.py
```diff
@@ -92,3 +95,64 @@ def teardown(self) -> None:
...
+            self.processing_time += self.tick_interval
+            return
+        if len(self.waiting) == 0:
+            # nothing to process
+            return
+        self.start_processing()
+
+    def start_processing(self) -> None:
+        """Process the next transaction."""
+        dialogue, message = self.waiting.pop(0)
```
marcofavorito, line 135: Minor: I suggest putting this as private method. The reason is, it only makes sense if called from act, because before the `pop` there isn't check of emptiness.
So either you move the check of emptiness inside `start_processing`, or put as private? Another is, put the check in both parts.

packages/fetchai/skills/generic_buyer/behaviours.py
```diff
@@ -92,3 +95,64 @@ def teardown(self) -> None:
...
+
+    def act(self) -> None:
+        """
+        Implement the act.
+
+        :return: None
+        """
+        if self.processing is not None and self.processing_time <= self.max_processing:
+            # already processing
+            self.processing_time += self.tick_interval
```
marcofavorito, line 126: is this safe? For our purposes I'd say it's good enough. A more resilient approach would be to take the current time and the last time `act` was called; the reason is that `tick_interval` might not be exactly the time elapsed between two acts, especially in case the AEA runtime is overloaded. 

Another point in favour of the current code is that `processing_time` would be an underestimation of the actual time elapsed, which means we would wait more than we did (which is safe to do in this case, but less responsive).

marcofavorito (approved):
LGTM

packages/fetchai/skills/generic_buyer/handlers.py
```diff
@@ -215,9 +216,11 @@ def _handle_match_accept(
...
-                "requesting transfer transaction from ledger api..."
+            tx_behaviour = cast(
+                GenericTransactionBehaviour, self.context.behaviours.transaction
+            )
+            tx_behaviour.waiting.append(
```
marcofavorito, line 222: A comment: `append` is fine because handlers and behaviours are processed sequentially or asynchronously by the scheduler.
It would also works in multithreading, because in CPython primitive operations over data structures are thread-safe.

packages/fetchai/skills/generic_buyer/behaviours.py
DavidMinarsch: I think it is safe in the sense that it will eventually terminate; it won't necessarily be exactly after max_processing seconds, but that;'s ok.

packages/fetchai/skills/generic_buyer/handlers.py
DavidMinarsch: Also, even if using multithreading; skills are always executed in the same thread atm.

marcofavorito: true, I had in mind an extension where many threads could write to it, e.g. one day we could have more threads for a skill

DavidMinarsch: ok but then we need to revisit other parts of the code too :)

packages/fetchai/skills/confirmation_aw3/skill.yaml
```diff
@@ -32,6 +32,11 @@ behaviours:
     args:
       search_interval: 1800
     class_name: SearchBehaviour
+  transaction:
+    args:
+      max_processing: 420
+      transaction_interval: 2
```
marcofavorito, line 38: not sure it's relevant; maybe a comment somewhere on what these parameters do? Don't know what's the best place; maybe in this YAML just above those lines, or in the skill README. We should update the other too though

packages/fetchai/skills/generic_buyer/strategy.py
```diff
@@ -192,6 +194,8 @@ def is_acceptable_proposal(self, proposal: Description) -> bool:
                 ]
             )
             and proposal.values["ledger_id"] == self.ledger_id
+            and proposal.values["price"] > 0
+            and proposal.values["quantity"] >= self._min_quantity
             and proposal.values["price"]
             <= proposal.values["quantity"] * self._max_unit_price
```
marcofavorito, line 200:
```suggestion
            and proposal.values["quantity"] >= self._min_quantity
            and 0 < proposal.values["price"] <= proposal.values["quantity"] * self._max_unit_price
```

DavidMinarsch: will do in separate pr later

marcofavorito (approved):
LGTM

Issue #1698, opened 2020-08-26
All custom types used by framework protocols should have a protobuf representation


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
Modules in `aea/helpers/transaction` and `aea/helpers/search` need a protobuf representation so we can phase out the use of pickle.

**Describe the solution you'd like**
See above

PR #2023, opened 2020-12-03
AEA-1100 Pickle --> Protobuf for OEF objects

### LINKED ISSUE #963: Serialisable Query and Description
Currently, to serialize Query and Description we use pickle.

Use protobuf instead: https://github.com/fetchai/oef-sdk-python/tree/master/oef-core-protocol

5A11:
This PR replaces the use of pickle for oef objects under `aea.helpers.search`

## Fixes

#963
#1698

## Types of changes

## Checklist


aea/helpers/search/models.py
```diff
@@ -918,32 +1379,30 @@ def __str__(self):
...
-
-        :param query_protobuf_object: the protocol buffer object whose type corresponds with this class.
-        :return: A new instance of this class that matches the protocol buffer object in the 'query_protobuf_object' argument.
+        :param query_protobuf_object: the protocol buffer object corresponding with this class.
+        :return: A new instance of this class matching the protocol buffer object
         """
-        query = pickle.loads(query_protobuf_object.query_bytes)  # nosec
-        return query
+        constraints = [ConstraintExpr._decode(c) for c in query_protobuf_object.constraints]
+        return cls(constraints, DataModel.decode(query_protobuf_object.model) if query_protobuf_object.HasField("model") else None)
```
DavidMinarsch, line 1405: simpler to read if the second arg is first declared in the line above

aea/helpers/search/models.py
```diff
@@ -840,6 +1249,58 @@ def __str__(self):
...
+        """
+        constraint_case = constraint_protobuf_object.WhichOneof("constraint")
+        constraint_type = None
+        if constraint_case == "relation":
+            constraint_type = ConstraintType.decode(constraint_protobuf_object.relation, "relation")
+        elif constraint_case == "set_":
+            constraint_type = ConstraintType.decode(constraint_protobuf_object.set_, "set")
+        elif constraint_case == "range_":
+            constraint_type = ConstraintType.decode(constraint_protobuf_object.range_, "range")
+        elif constraint_case == "distance":
```
DavidMinarsch, line 1299: add else for invalid case

aea/helpers/search/models.py
```diff
@@ -602,6 +907,44 @@ def check_validity(self) -> None:  # pylint: disable=no-self-use  # pragma: noco
...
+        :return: A new instance of this class matching the protocol buffer object
+        """
+        expression = constraint_expression_protobuf_object.WhichOneof("expression")
+        if expression == "and_":
+            return And.decode(constraint_expression_protobuf_object.and_)
+        elif expression == "or_":
+            return Or.decode(constraint_expression_protobuf_object.or_)
+        elif expression == "not_":
+            return Not.decode(constraint_expression_protobuf_object.not_)
+        elif expression == "constraint":
```
DavidMinarsch, line 945: else for non-matching type

aea/helpers/search/models.py
```diff
@@ -602,6 +907,44 @@ def check_validity(self) -> None:  # pylint: disable=no-self-use  # pragma: noco
...
+        """
+        constraint_expr_pb = models_pb2.Query.ConstraintExpr()
+        expression_pb = expression.encode()
+        if isinstance(expression, And):
+            constraint_expr_pb.and_.CopyFrom(expression_pb)
+        elif isinstance(expression, Or):
+            constraint_expr_pb.or_.CopyFrom(expression_pb)
+        elif isinstance(expression, Not):
+            constraint_expr_pb.not_.CopyFrom(expression_pb)
+        elif isinstance(expression, Constraint):
```
DavidMinarsch, line 925: else for non-matching type

aea/helpers/search/models.py
```diff
@@ -651,6 +994,28 @@ def __eq__(self, other):  # pragma: nocover
...
+
+    @classmethod
+    def decode(cls, and_protobuf_object) -> "And":
+        """
+        Decode a protocol buffer object that corresponds with this class into an instance of this class.
+
+        :param and_protobuf_object: the protocol buffer object corresponding with this class.
+        :return: A new instance of this class matching the protocol buffer object
+        """
+        expr = [ConstraintExpr._decode(c) for c in and_protobuf_object.expr]
```
DavidMinarsch, line 1016: you can do `cls._decode(c)`

aea/helpers/search/models.py
```diff
@@ -651,6 +994,28 @@ def __eq__(self, other):  # pragma: nocover
         """Compare with another object."""
         return isinstance(other, And) and self.constraints == other.constraints
 
+    def encode(self) -> models_pb2.Query.ConstraintExpr.And():
```
DavidMinarsch, line 997: If you make it a class method you can do `cls._encode(...)`

aea/helpers/search/models.py
```diff
@@ -568,6 +689,190 @@ def __str__(self):
...
+            models_pb2.Query.Relation.NOTEQ: ConstraintTypes.NOT_EQUAL,
+            models_pb2.Query.Relation.EQ: ConstraintTypes.EQUAL,
+            models_pb2.Query.Set.IN: ConstraintTypes.IN,
+            models_pb2.Query.Set.NOTIN: ConstraintTypes.NOT_IN
+        }
+
+        if category == "relation":
+            relation_enum = type_from_pb[constraint_type_protobuf_object.op]
+            value_case = constraint_type_protobuf_object.val.WhichOneof("value")
+            if value_case == "s":
```
DavidMinarsch, line 832: these magic values should be defined as constants somewhere

aea/helpers/search/models.py
```diff
@@ -292,36 +376,73 @@ def __str__(self):
...
+
+    @staticmethod
+    def _extract_value(value: models_pb2.Query.Value) -> ATTRIBUTE_TYPES:
+        """
+        From a Protobuf query value object to attribute type.
+        :param value: an instance of models_pb2.Query.Value.
+        :return: the associated attribute type.
+        """
+        value_case = value.WhichOneof("value")
+        if value_case == "s":
```
DavidMinarsch, line 424: see below; define constants

aea/helpers/search/models.proto
```diff
@@ -0,0 +1,150 @@
...
+            }
+            message Locations {
+                repeated Location vals = 1;
+            }
+            oneof values {
+                Strings s = 1;
+                Doubles d = 2;
+                Bools b = 3;
+                Ints i = 4;
+                Locations l = 5;
```
DavidMinarsch, line 107: I'd prefer if they were proper names; rather than single letter vars

5A11: I changed all abbreviations to proper names

aea/helpers/search/models.py
5A11: Done

aea/helpers/search/models.py
5A11: Done

aea/helpers/search/models.py
5A11: Done

aea/helpers/search/models.py
5A11: Done

aea/helpers/search/models.py
```diff
@@ -417,8 +561,8 @@ def check_validity(self):
                 )
             elif self.type == ConstraintTypes.WITHIN:
                 enforce(
-                    isinstance(self.value, (list, tuple)),
-                    f"Expected one of type in (list, tuple), got {self.value}",
+                    isinstance(self.value, tuple),
```
5A11, line 564: `value` can only be tuple. I removed the list type from being allowed. 

When `value` gets serialised into pb, it becomes `repeated int/double/...`. 
When deserialised, `repeated int/double/...` becomes `tuple[int/float/...]` in python. 

So for an original value to be identical to its serialised-deserialised version, it should be a `tuple`.

5A11: Similarly in the next couple of lines.

aea/helpers/search/models.py
```diff
@@ -613,6 +1052,7 @@ def __init__(self, constraints: List[ConstraintExpr]):
         :param constraints: the list of constraints expression (in conjunction).
         """
         self.constraints = constraints
+        self.check_validity()
```
5A11, line 1055: I noticed `check_validity` of `And`, `Or`, `Not` are not called at all, so I added it 
(probably an oversight as they are called in the other repo which seems to have been more developed).

aea/helpers/search/models.py
```diff
@@ -662,6 +1126,7 @@ def __init__(self, constraints: List[ConstraintExpr]):
         :param constraints: the list of constraints expressions (in disjunction).
         """
         self.constraints = constraints
+        self.check_validity()
```
5A11, line 1129: Same as above

tests/test_helpers/test_search/test_models.py
```diff
@@ -153,121 +160,217 @@ def test_constraint_type():
...
 
 
 def test_constraints_expressions():
     """Test constraint expressions: And, Or, Not."""
-    expression = And(
+    and_expression = And(
         [
-            ConstraintType(ConstraintTypes.EQUAL, 12),
-            ConstraintType(ConstraintTypes.EQUAL, 12),
+            Constraint("number", ConstraintType(ConstraintTypes.GREATER_THAN, 15)),
```
5A11, line 274: A side observation: The above is unintuitive. 

`Constraint("number", ConstraintType(ConstraintTypes.GREATER_THAN, 15))` intuitively (reading from left to right) means `number` must be greater than 15.

Currently it means 15 is greater than `number`.

All other constraint are read left to right (`within`, `in`, `not_in`, etc). But `>`, `>=`, `<`, and `<=` do not respect this. in the old `python-oef-sdk` repo, this is also respected.

To correct it, we need to modify lines 679-686 in `models.py` and just invert the operators.

DavidMinarsch: Good observation :). What's the effect of this proposed change on the rest of the code base?

aea/helpers/search/models.py
DavidMinarsch: Yes sounds good

aea/helpers/search/models.py
DavidMinarsch: ok, fine by me

aea/helpers/search/models.py
```diff
@@ -840,6 +1383,71 @@ def __str__(self):
...
+    @classmethod
+    def decode(cls, constraint_protobuf_object) -> "Constraint":
+        """
+        Decode a protocol buffer object that corresponds with this class into an instance of this class.
+
+        :param constraint_protobuf_object: the protocol buffer object corresponding with this class.
+        :return: A new instance of this class matching the protocol buffer object
+        """
+        constraint_case = constraint_protobuf_object.WhichOneof("constraint")
+        if constraint_case == "relation":
```
DavidMinarsch, line 1428: Magic values should be declared as constants. Here you can declare as class level constant `RELATION = "relation"` and then access as `self.RELATION`. That way we increase consistency across code and enable reuse.

Applies throughout the file

aea/helpers/search/models.py
5A11: Done

aea/helpers/search/models.py
5A11: I don't understand this comment. 

Do you mean, changing this `encode` method to be a class method (and by extension all other `encode` methods)? 
If so, what's the advantage of doing that? This method **always** needs an instance of this class to encode, so if we change it to class method, the instance has to always be passed as a parameter. With an instance method, we can just call `encode` off of the instance itself.

tests/test_helpers/test_search/test_models.py
5A11: I thought the effect might be a lot at first, hence commented first before doing it.

After I did a search in the framework, apparently we do not use these operators much ourselves (as in skills, etc). 

Here is a rough list of where they are used:
- `soef` connection -> `object_translator.py`
- `tac` (might be one of its failure reasons)
- bunch of tests

But if there are other developers who do use these, meaning they were using concepts under `helpers.search` (which they could but weren't recommended anywhere to do so afaik), the semantics would gets reversed in those codes. However, mu hunch is if there were any, they would probably complain/report this oddity somewhere.

All in all, I suggest we do it in a separate PR to see all its effects. It would make this one messy if we do it here.

aea/helpers/search/models.py
DavidMinarsch: Sorry, I got confused here.

tests/test_helpers/test_search/test_models.py
DavidMinarsch: Ok, let's do it but as you say in separate PR. We can make a note in upgrading guide re this breaking change.

aea/helpers/search/models.py
5A11: Done

DavidMinarsch (approved):
LGTM

Issue #2021, opened 2020-12-02
Add check for correct environment setup 


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
Most AEAs need `go` and `gcc` available.

**Describe the solution you'd like**
Add check at start `aea run`. Exit cleanly if not present. Use `shutil.which`

PR #2031, opened 2020-12-07
AEA-821 add 'build_entrypoint' field in all package configurations

### LINKED ISSUE #1710: `aea build` step in cli
Several packages contain go code which requires a build step.

Allow package configs to specify a build step requirement and entry point. On call of `aea build` the packages with these requirements are identified and the build performed.

This prevents a situation where building takes place during launch of the agent and we can also avoid rebuilding unnecessarily.

marcofavorito:
## Proposed changes

Add `entrypoint:` field in package configurations.
Add `aea build` command.

## Fixes

#1710 #2021 

## Types of changes

## Checklist

## Further comments

n/a


aea/configurations/base.py
```diff
@@ -994,6 +994,7 @@ def __init__(
         aea_version: str = "",
         fingerprint: Optional[Dict[str, str]] = None,
         fingerprint_ignore_patterns: Optional[Sequence[str]] = None,
+        entrypoint: Optional[str] = None,
```
DavidMinarsch, line 997: Probably should be more explicitly labelled `build_entrypoint` as there might be other entry points one day..

aea/crypto/registries/base.py
```diff
@@ -73,7 +73,7 @@ class EntryPoint(Generic[ItemType], RegexConstrainedString):
     )
 
     def __init__(self, seq):
-        """Initialize the entrypoint."""
+        """Initialize the build_entrypoint."""
```
DavidMinarsch, line 76: I don't think that's right

aea/cli/utils/loggers.py
```diff
@@ -78,7 +78,7 @@ def _set_level(ctx, param, value):  # pylint: disable=unused-argument
             level = logging.getLevelName(value)
             logger.setLevel(level)
             # save verbosity option so it can be
-            # read in the main CLI entrypoint
+            # read in the main CLI build_entrypoint
```
DavidMinarsch, line 81: please revert

aea/aea_builder.py
```diff
@@ -1273,6 +1323,34 @@ def _try_to_load_agent_configuration_file(aea_project_path: Path) -> None:
...
+            "Package has not a build entrypoint specified.",
+        )
+        build_entrypoint = cast(str, build_entrypoint)
+        script_path = Path(directory) / build_entrypoint
+        enforce(
+            script_path.exists(), f"File '{build_entrypoint}' does not exists.",
+        )
+        enforce(
+            script_path.is_file(), f"'{build_entrypoint}' is not a file.",
+        )
```
DavidMinarsch, line 1346: These are user defined values, so enforce seems not appropriate. We should probably raise an AEAException or similar.

aea/aea_builder.py
```diff
@@ -1475,6 +1508,7 @@ def from_aea_project(
         # load agent configuration file
         configuration_file = cls.get_configuration_file_path(aea_project_path)
         agent_configuration = cls.loader.load(configuration_file.open())
+        agent_configuration.directory = configuration_file
```
DavidMinarsch, line 1511: ?

DavidMinarsch (commented):
This looks mostly very good. It would be nice if it was already implemented for the `p2p_libp2p` package as this is the first important use case of this command.

aea/aea_builder.py
marcofavorito: not needed anymore

aea/aea_builder.py
marcofavorito: in fact, if the condition is not satisfied, `AEAException` is raised with the error message, am I right? How would you improve this code?

packages/fetchai/connections/p2p_libp2p/check_dependencies.py
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env python3
```
marcofavorito, line 1: relatively simple; we can iterate over it to add finer-grained checks, e.g. check the version is at least greater than some version etc.

solarw (approved):
LGTM

DavidMinarsch (approved):
LGTM

PR #2043, opened 2020-12-08
Fix/coverage

### LINKED ISSUE #2039: MAM: handle a case when "keys" folder wasn't cleaned
Currently if MAM wasn't correctly closed (by calling `stop_manager` method) it will fail on start with `FileExists` error because `keys` folder wasn't erased. That means that's impossible to restart MAM after force exit, only by manually deleting a `keys` folder.

Make MAM able to start with `keys` folder present (by default or with some flag) and handle that case.

solarw:
## Proposed changes

fix for manager. coverage fixes.

## Fixes

#2039 

## Types of changes

## Checklist


tests/test_cli/test_generate/test_generate.py
```diff
@@ -90,7 +87,7 @@ def test__generate_item_no_res_isort_missing(self, *_mocks):
         )
         self.assertEqual(cm.exception.message, expected_msg)
 
-    @pytest.mark.skip
+    # @pytest.mark.skip
```
DavidMinarsch, line 90: remove

aea/cli/fetch.py
```diff
@@ -137,6 +137,10 @@ def fetch_agent_locally(
         )
 
     # add dependencies
+    if not ctx.config.get("is_local") or not ctx.config.get("is_mixed"):
+        # used by add_item to download locally
+        ctx.config["is_local"] = True
```
DavidMinarsch, line 142: By definition we are in the local case here; so if anything this should be set at the start of the method. But preferably; this is correctly set outside this method and then only asserted at the start.

marcofavorito: yes. If not necessary, I'd put this outside and put the asserts at the start of the function

panasevychol (approved):
My approval to MAM fix

tests/test_cli/test_generate/test_generate.py
solarw: removed

aea/cli/fetch.py
solarw: updated, check is added

DavidMinarsch (approved):
LGTM

PR #2052, opened 2020-12-10
AEA-1017 Prometheus connection

jrriehl:
## Proposed changes

Prometheus connection allows agents to add and update metrics by sending messages. The metrics are exposed via a local http server using the package `aioprometheus`. From there they can be scraped by a prometheus server for remote monitoring.

## Types of changes

## Checklist


packages/fetchai/connections/prometheus/connection.py
```diff
@@ -0,0 +1,314 @@
...
+
+    async def _send(self, envelope: Envelope) -> None:
+        """Send a message.
+
+        :param envelope: the envelope
+        :return: None
+        """
+        await self.queue.put(envelope)
+
+    async def disconnect(self) -> None:
```
DavidMinarsch, line 230: what about prometheus_client?

packages/fetchai/connections/prometheus/connection.py
```diff
@@ -0,0 +1,314 @@
...
+                        f"{message.type} is not a recognized prometheus metric."
+                    )
+                else:
+                    self.metrics[message.title] = metric_type(
+                        message.title, message.description
+                    )
+                    response_code = 200
+                    response_msg = (
+                        f"New {message.type} successfully added: {message.title}."
+                    )
```
DavidMinarsch, line 187: extract in helper, similarly below

packages/fetchai/connections/prometheus/connection.py
```diff
@@ -0,0 +1,314 @@
...
+
+        :return: None
+        """
+        sender = envelope.sender
+        self.logger.debug("Processing message from {}: {}".format(sender, envelope))
+        if envelope.protocol_id != PrometheusMessage.protocol_id:
+            raise ValueError("This protocol is not valid for prometheus.")
+        await self.handle_prometheus_message(envelope)
+
+    async def handle_prometheus_message(self, envelope: Envelope) -> None:
```
DavidMinarsch, line 149: probably better indicated as private `_`

packages/fetchai/connections/prometheus/connection.py
```diff
@@ -0,0 +1,314 @@
...
+        Initialize a connection to a local prometheus server.
+
+        :param kwargs: the keyword arguments of the parent class.
+        """
+        super().__init__(**kwargs)
+
+        print(self.configuration)
+
+        self.port = cast(int, self.configuration.config.get("port", DEFAULT_PORT))
+        self.metrics = {}
```
DavidMinarsch, line 261: don't see why that needs instantiating here.

packages/fetchai/connections/prometheus/connection.py
```diff
@@ -0,0 +1,314 @@
...
+
+    def __init__(self, **kwargs):
+        """
+        Initialize a connection to a local prometheus server.
+
+        :param kwargs: the keyword arguments of the parent class.
+        """
+        super().__init__(**kwargs)
+
+        print(self.configuration)
```
DavidMinarsch, line 258: please use logger.debug

packages/fetchai/connections/prometheus/connection.yaml
```diff
@@ -0,0 +1,23 @@
...
+  __init__.py: QmWVrDiiePsr6vTnvbPTcDrayR89ji3hf25rs9V9TiJUPv
+  connection.py: QmWixZKE3wr7Ln3bTwxRWJ6dVqhtREVDc1xDHJjxxTdwXw
+fingerprint_ignore_patterns: []
+connections: []
+protocols: []
+class_name: PrometheusConnection
+config:
+  port: 9090
+excluded_protocols: []
+restricted_to_protocols: []
```
DavidMinarsch, line 19: `fetchai/prometheus:0.1.0`

packages/fetchai/connections/prometheus/connection.py
```diff
@@ -0,0 +1,314 @@
...
+        """
+        Start prometheus http server.
+
+        :return: None
+        """
+        if self._queue:  # pragma: nocover
+            return None
+        self._loop = asyncio.get_event_loop()
+        self._queue = asyncio.Queue()
+        prometheus_client.start_http_server(self._port)
```
solarw, line 135: I confused with this approach developers of this library provided.
you can start it, but no method to stop it, I just checked the code.
anyway the server will shutdown on main app exit, but it can be a problem for tests or several instances of the connection.


probably we can check some wrappers like https://github.com/claws/aioprometheus

packages/fetchai/connections/prometheus/connection.py
```diff
@@ -0,0 +1,314 @@
...
+
+        :param address: The address of the connection.
+        :param metrics: The prometheus metrics.
+        :param port: The port at which to expose the metrics.
+        """
+        self.address = address
+        self.metrics = metrics
+        self._loop: Optional[asyncio.AbstractEventLoop] = None
+        self._queue: Optional[asyncio.Queue] = None
+        self._threaded_pool: ThreadPoolExecutor = ThreadPoolExecutor(
```
solarw, line 97: is threaded pool used somewhere?

DavidMinarsch (commented):
LGTM, some comments

packages/fetchai/skills/coin_price/dialogues.py
```diff
@@ -68,3 +75,36 @@ def role_from_first_message(  # pylint: disable=unused-argument
...
+
+    def __init__(self, **kwargs) -> None:
+        """
+        Initialize dialogues.
+
+        :return: None
+        """
+
+        self.enabled = kwargs.pop("enabled", False)
+        self.metrics = kwargs.pop("metrics", [])
```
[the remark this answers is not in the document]
DavidMinarsch: On consideration, I think there is a benefit to have it here too. Let's keep it as an inspiration :).

packages/fetchai/protocols/prometheus/README.md
```diff
@@ -0,0 +1,43 @@
...
+...
+---
+initiation: [add_metric, update_metric]
+reply:
+  add_metric: [response]
+  update_metric: [response]
+  response: []
+termination: [response]
+roles: {agent, server}
+end_states: [successful]
```
DavidMinarsch, line 39:
```suggestion
end_states: [successful]
keep_terminal_state_dialogues: false
```

DavidMinarsch: Please regenerate after change.

solarw (approved):
LGTM
need to find the way to stop http server
metrics update methods are synchronous but I'm sure do not block for a long time so can use it.
also thread pool looks useless.

packages/fetchai/connections/prometheus/connection.py
jrriehl: Yes, exactly. I was also curious why I could not find any `stop_http_server()` function. Thanks for the suggestion!

packages/fetchai/skills/coin_price/handlers.py
```diff
@@ -204,3 +219,71 @@ def teardown(self) -> None:
...
+    def __init__(self, **kwargs):
+        """Initialize the handler."""
+        super().__init__(**kwargs)
+
+        self.handled_message = None
+
+    def setup(self) -> None:
+        """Set up the handler."""
+        if self.context.prometheus_dialogues.enabled:
+            self.context.logger.info("setting up PrometheusHandler")
```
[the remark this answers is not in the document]
jrriehl: Yes, this is true. I partly wanted to avoid doing a conditional import like we did for the `http_server`. It's a bit messy to give agents optional use of packages. Maybe something to keep in mind for the future, but for now we can just always log this setup info.

packages/fetchai/connections/prometheus/connection.py
jrriehl: Right, as @solarw also pointed out, this library somehow does not include stop server functionality. Hopefully the package he suggested can do this correctly.

packages/fetchai/connections/prometheus/connection.py
jrriehl: Are you suggesting that `metrics` be initialized to `None` or that it only needs to be defined within the channel? If the latter, couldn't there be some scenario where the channel gets disconnected but you don't want to lose the metrics that have already been created?

packages/fetchai/connections/prometheus/connection.py
```diff
@@ -0,0 +1,341 @@
...
+    ) -> Tuple[PrometheusMessage, Optional[PrometheusDialogue]]:
+        """
+        Get a message copy and dialogue related to this message.
+
+        :param envelope: incoming envelope
+
+        :return: Tuple[Message, Optional[Dialogue]]
+        """
+        message = cast(PrometheusMessage, envelope.message)
+        dialogue = cast(PrometheusDialogue, self._dialogues.update(message))
```
DavidMinarsch, line 112:
```suggestion
        dialogue = cast(Optional[PrometheusDialogue], self._dialogues.update(message))
```

packages/fetchai/connections/prometheus/connection.py
```diff
@@ -0,0 +1,341 @@
...
+    async def send(self, envelope: Envelope) -> None:
+        """
+        Process the envelopes to prometheus.
+
+        :return: None
+        """
+        sender = envelope.sender
+        self.logger.debug("Processing message from {}: {}".format(sender, envelope))
+        if envelope.protocol_id != PrometheusMessage.protocol_id:
+            raise ValueError("This protocol is not valid for prometheus.")
```
DavidMinarsch, line 143: string should contain the protocol id of the envelope for improved debugging

packages/fetchai/connections/prometheus/connection.py
```diff
@@ -0,0 +1,341 @@
...
+        else:
+            metric_type = getattr(aioprometheus, message.type, None)
+            if metric_type is None:
+                response_code = 404
+                response_msg = f"{message.type} is not a recognized prometheus metric."
+            else:
+                self.metrics[message.title] = metric_type(
+                    message.title, message.description, message.labels
+                )
+                self._service.register(self.metrics[message.title])
```
DavidMinarsch, line 209: is there no async equivalent?

DavidMinarsch: Can this fail?

packages/fetchai/connections/prometheus/connection.py
```diff
@@ -0,0 +1,341 @@
...
+                response_code = 400
+                response_msg = (
+                    f"Update function {message.callable} not found for metric {metric}."
+                )
+            else:
+                # Update the metric ("inc" and "dec" do not take "value" argument)
+                if message.callable in {"inc", "dec"}:
+                    update_func(message.labels)
+                else:
+                    update_func(message.labels, message.value)
```
DavidMinarsch, line 241: can we be sure this is a method? can it be a coroutine too?

DavidMinarsch: What if it fails?

packages/fetchai/connections/prometheus/connection.py
DavidMinarsch: I think the latter scenario cannot happen given how the connections are used. If the channel crashes the connection does to.

packages/fetchai/connections/prometheus/connection.py
```diff
@@ -0,0 +1,341 @@
...
+        """
+        super().__init__(**kwargs)
+
+        self.host = cast(int, self.configuration.config.get("host", DEFAULT_HOST))
+        self.port = cast(int, self.configuration.config.get("port", DEFAULT_PORT))
+        self.metrics = {}
+        self.channel = PrometheusChannel(
+            self.address, self.metrics, self.host, self.port
+        )
+        self._connection = None  # type: Optional[asyncio.Queue]
```
DavidMinarsch, line 290: Not used?

packages/fetchai/connections/prometheus/connection.py
```diff
@@ -0,0 +1,341 @@
...
+
+        :param kwargs: the keyword arguments of the parent class.
+        """
+        super().__init__(**kwargs)
+
+        self.host = cast(int, self.configuration.config.get("host", DEFAULT_HOST))
+        self.port = cast(int, self.configuration.config.get("port", DEFAULT_PORT))
+        self.metrics = {}
+        self.channel = PrometheusChannel(
+            self.address, self.metrics, self.host, self.port
```
DavidMinarsch, line 288: Feed through the logger (`self.logger` or something like that)

packages/fetchai/connections/prometheus/connection.py
```diff
@@ -0,0 +1,341 @@
...
+
+        :param address: The address of the connection.
+        :param metrics: The prometheus metrics.
+        :param port: The port at which to expose the metrics.
+        """
+        self.address = address
+        self.metrics = metrics
+        self._loop: Optional[asyncio.AbstractEventLoop] = None
+        self._queue: Optional[asyncio.Queue] = None
+        self.logger: Union[logging.Logger, logging.LoggerAdapter] = _default_logger
```
DavidMinarsch, line 95: logger should be an arg to and only default logger used if no logger passed through

DavidMinarsch (commented):
LGTM, I added some comments re missing logic and questions.

packages/fetchai/connections/prometheus/connection.py
jrriehl: There is now a check to make sure that is a callable and implemented function. There may be some other obscure failure cases (e.g. invalid strings in names or labels), but I think we cover the most important ones now.

packages/fetchai/connections/prometheus/connection.py
jrriehl: Added a check here to make sure this is a valid metric_type and not some other attribute.

According to the example here (https://github.com/claws/aioprometheus), I believe this is correct usage is async context. But maybe @solarw can confirm?

packages/fetchai/connections/prometheus/connection.py
jrriehl: Similar to above, this is the correct usage (not as a coroutine) based on my understanding.

DavidMinarsch (approved):
LGTM

packages/fetchai/connections/prometheus/connection.py
solarw: it's just a functions, not coroutines

PR #2056, opened 2020-12-14
Hotfix/libp2p node awx issues

lrahmani:
## Proposed changes

- Limit log spamming by reducing the number of log line to 2 maximum for each operation
- Add artificial delay for agent registration to limit spamming
- Add locks against race conditions

## Fixes

- race conditions
- logs spamming


lrahmani:
PR open to master branch by mistake

PR #2061, opened 2020-12-15
AEA-1158 Fix docs and protocol terminal states

DavidMinarsch:
## Proposed changes

This PR fixes an issue in docs and adds proper terminal states to all protocols.

## Fixes

-

## Types of changes

## Checklist

## Further comments

-


aea/protocols/generator/validate.py
```diff
@@ -685,6 +696,39 @@ def _validate_termination(
                 ),
             )
 
+    # check that there are no repetitive performatives in termination
+    duplicates_dict: Dict[str, int] = {}
+    for performative in termination:
+        duplicates_dict[performative] = duplicates_dict.get(performative, 0) + 1
+    duplicates = [k for k, v in duplicates_dict.items() if v > 1]
```
DavidMinarsch, line 703: just use `no_duplicates = len(termination) == len(list(set(termination)))`, this is very non-pythonic

aea/protocols/generator/validate.py
```diff
@@ -823,6 +871,7 @@ def _validate_dialogue_section(
         ) = _validate_termination(
             cast(List[str], protocol_specification.dialogue_config["termination"]),
             performatives_set,
+            cast(Set[str], terminal_performatives_from_reply),
```
DavidMinarsch, line 874: Don't cast here; just treat optional type properly in the method...

aea/protocols/generator/validate.py
5A11: Yea, I did something similar initially, but this gives a better error message, by printing what performatives are being repeated. 

If you think it is not that useful I simply the code.

aea/protocols/generator/validate.py
5A11: We know it won't be Optional (None). If it was, the previous `# Validate reply` block would return False and we won't reach this point. 

I moved cast at the beginning of the block for more clarity.

PR #2067, opened 2020-12-17
build command fix

solarw:
## Proposed changes

build command fix

## Types of changes

## Checklist


aea/aea_builder.py
```diff
@@ -780,7 +775,6 @@ def _set_component_build_directory(
             configuration.author,
             configuration.name,
         )
-        ensure_dir(configuration.build_directory)
```
DavidMinarsch, line 783: I think it should be kept here, but so:
```
if configuration.build_directory is not None:
    ensure_dir(configuration.build_directory)
```
This way, we don't require the package developer to do it!

solarw: no, package developer has to do it.
so it should be implemented in build script.
this will help to avoid empty directories creation

DavidMinarsch (approved):
LGTM

Issue #1991, opened 2020-11-24
MAM: no error handling when trying to add a different version of the same project


panasevychol:
**Describe the bug**
Throws `FileExistsError: [Errno 17] File exists: '/workdir/fetchai/my_first_aea/connections'` when trying to add project `fetchai/my_first_aea:0.8.0` when `fetchai/my_first_aea:0.15.0` is present.

**To Reproduce**
```python
from aea.configurations.base import PublicId
from aea.manager import MultiAgentManager

wd = "any_workdir"

m = MultiAgentManager(wd)
m.start_manager()
pid = PublicId.from_str("fetchai/my_first_aea:0.15.0")
m.add_project(pid)

pid = PublicId.from_str("fetchai/my_first_aea:0.8.0")
m.add_project(pid)
```

**Expected behavior**
Should throw an informative ValueError that tells that user is unable to add the same project twice (despite different versions) instead FileExistsError.

PR #2069, opened 2020-12-18
AEA-1138 multiagenmanager: same project add error handling

solarw:
## Proposed changes

MultiAgentManager add project fix
## Fixes

https://github.com/fetchai/agents-aea/issues/1991

## Types of changes

## Checklist

### DIFF
--- a/aea/manager.py
+++ b/aea/manager.py
@@ -27,7 +27,7 @@ from asyncio.tasks import FIRST_COMPLETED
 from pathlib import Path
 from shutil import rmtree
 from threading import Thread
-from typing import Any, Callable, Dict, List, Optional
+from typing import Any, Callable, Dict, List, Optional, Set
 
 from aea.aea import AEA
 from aea.aea_builder import AEABuilder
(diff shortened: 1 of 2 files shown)


aea/manager.py
```diff
@@ -310,8 +311,12 @@ def add_project(
         :param local: whether or not to fetch from local registry.
         :param restore: bool flag for restoring already fetched agent.
         """
-        if public_id in self._projects:
-            raise ValueError(f"Project {public_id} was already added!")
+        if public_id.to_latest() in self._projects_set:
```
DavidMinarsch, line 314: You should do a prefix check; this now checks whether `autor/name:latest` is in the set. What we should check is if `autor/name` is in the set!

aea/manager.py
```diff
@@ -327,13 +332,13 @@ def remove_project(
...
         if public_id not in self._projects:
             raise ValueError(f"Project {public_id} is not present!")
-
         if self._projects[public_id].agents:
             raise ValueError(
                 f"Can not remove projects with aliases exists: {self._projects[public_id].agents}"
             )
 
         project = self._projects.pop(public_id)
+        self._projects_set.remove(public_id.to_latest())
```
DavidMinarsch, line 341: I don't think this is a good idea. It is not necessarily the latest project which is being added!

DavidMinarsch (changes requested):

aea/manager.py
solarw: that correct, instead of forming prefix manually I use to_latest() to get rid of actual versions.
original version info is stored in self._projects

DavidMinarsch (approved):
LGTM

PR #2104, opened 2021-01-04
fix spelling errors in docs

DavidMinarsch:
## Proposed changes

This PR addresses spelling errors in the docs raised by executing `mdspell -n -a --en-gb '**/*.md' '!docker-images/*.md' '!docs/api/**/*.md'
` in root (https://www.npmjs.com/package/markdown-spellcheck)

## Fixes

-

## Types of changes

## Checklist

## Further comments

-


docs/.spelling
```diff
@@ -0,0 +1,143 @@
...
+faucet
+Fetch.ai.
+unregisters
+teardown
+unregister
+seller_AEA
+buyer_AEA
+Agentland
+generic_seller
+instantiable
```
marcofavorito, line 63: isn't a legal word?

docs/.spelling
```diff
@@ -0,0 +1,143 @@
...
+startup
```
marcofavorito, line 64: maybe `start-up` gets accepted? I don't think so.

docs/.spelling
```diff
@@ -0,0 +1,143 @@
...
+backend
```
marcofavorito, line 65: try `back-end` or `back end`.
The last one seems preferable, although less widespread.
https://en.wikipedia.org/wiki/Front_end_and_back_end

docs/.spelling
```diff
@@ -0,0 +1,143 @@
...
+SQLite
+backends
```
marcofavorito, line 67: see above

docs/.spelling
```diff
@@ -0,0 +1,143 @@
...
+permissionless
+libp2p
+ORMs
+standalone
+Multiaddresses
+namespaced
+prometheus
+Grafana
+toolset
+deserialised
```
marcofavorito, line 105: British vs American spelling?

marcofavorito (approved):
Very useful. I wonder if we could reduce the number of white-listed words by playing with hyphens or spaces in composite words, different types of spelling etc.

docs/.spelling
DavidMinarsch: https://en.wiktionary.org/wiki/instantiable

docs/.spelling
DavidMinarsch: its a valid noun, search `define: startup`

docs/.spelling
DavidMinarsch: its a valid noun, search `define: startup`

docs/.spelling
DavidMinarsch: see above

docs/.spelling
DavidMinarsch: Everything is set for British

DavidMinarsch:
> Very useful. I wonder if we could reduce the number of white-listed words by playing with hyphens or spaces in composite words, different types of spelling etc.

That only makes sense when the resulting words are more common. I think the white-listed words are valid additions (checked each one of them).

docs/.spelling
marcofavorito: Yes, this and other comments were to reduce at the minimum the size of `.spelling`. But as you said, it's OK.

PR #2125, opened 2021-01-06
fix validation of data pattern in config

DavidMinarsch:
## Proposed changes

This PR makes a quick fix noted by @5A11 when loading configurations with `null` values.

## Fixes

-

## Types of changes

## Checklist

## Further comments

-


aea/configurations/validation.py
```diff
@@ -304,7 +304,10 @@ def check_excludes(path):
             # one of the values is env variable: skip data type check
             continue
 
-        if not issubclass(type(new_value), type(pattern_value)):
+        if (
+            not issubclass(type(new_value), type(pattern_value))
+            and new_value is not None
```
DavidMinarsch, line 309: maybe `None` should also be excluded for `pattern_value`. Otherwise once setting `None` we cannot set anything else again...

DavidMinarsch: Ah no..I think `pattern_value` is taken from base config. So it must be non-null. Otherwise we don't know what's a valid value to set...

DavidMinarsch: Ah all good, line `297` has the answer!

5A11 (approved):
LGTM

Issue #2123, opened 2021-01-06
kwargs needed in MyScaffoldConnection constructor


marcofavorito:
**Describe the bug**

When a connection is scaffolded, it doesn't work as `kwargs` are needed in MyScaffoldConnection constructor.

**To Reproduce**

- Scaffold a connection;
- run the AEA

**Expected behavior**

The AEA runs correctly.

**Screenshots**
n/a

**Desktop (please complete the following information):**
 - Version: `0.8.0` and above

**Smartphone (please complete the following information):**
n/a

**Additional context**
n/a

PR #2134, opened 2021-01-10
AEA-1212 Fix for aea scaffold connection:

solarw:
Fix for aea scaffold connection: **kwargs for Connection.__init__ and public id fix for python files

## Fixes

https://github.com/fetchai/agents-aea/issues/2123

## Types of changes

## Checklist


aea/cli/scaffold.py
```diff
@@ -187,11 +186,13 @@ def scaffold_item(ctx: Context, item_type: str, item_name: str) -> None:
...
             loader.dump(config, fp)
 
-        # update 'PUBLIC_ID' variable with the right public id
-        init_module = Path(dest, "__init__.py")
-        init_module.write_text(
-            re.sub(SCAFFOLD_PUBLIC_ID, str(new_public_id), init_module.read_text())
-        )
+        # update 'PUBLIC_ID' variable with the right public id in all python files!
+
+        for file_path in glob.glob(str(Path(dest, "*.py"))):
```
DavidMinarsch, line 191: We're not looking for any file. We're looking for init file only..

aea/connections/scaffold/connection.py
```diff
@@ -33,22 +31,15 @@ class MyScaffoldConnection(Connection):
...
         Initialize a connection to an SDK or API.
 
         :param configuration: the connection configuration.
         :param crypto_store: object to access the connection crypto objects.
         :param identity: the identity object.
         """
-        super().__init__(  # pragma: no cover
-            configuration=configuration, crypto_store=crypto_store, identity=identity
-        )
+        super().__init__(**kwargs)  # pragma: no cover
```
DavidMinarsch, line 42: The reason we did miss this before is because we had no test. Pragma no cover needs removing and we need to test instantiation of an empty scaffolded class.

DavidMinarsch: I see test is there now, so remove pragma statement please

aea/connections/scaffold/connection.py
```diff
@@ -33,22 +31,15 @@ class MyScaffoldConnection(Connection):
...
-        identity: Optional[Identity] = None,
-        crypto_store: Optional[CryptoStore] = None,
-    ):
+    def __init__(self, **kwargs):
         """
         Initialize a connection to an SDK or API.
 
         :param configuration: the connection configuration.
         :param crypto_store: object to access the connection crypto objects.
         :param identity: the identity object.
```
DavidMinarsch, line 40: The docstring is no longer accurate.

aea/connections/scaffold/connection.py
solarw: the problem that file is not executed, but it's copy, this is why pragma: nocover is used

aea/connections/scaffold/connection.py
solarw: updated to the base connection definition

aea/cli/scaffold.py
solarw: now fixes only for `__init__.py` and `connection.py`

DavidMinarsch (approved):
LGTM

marcofavorito (approved):
LGTM

Issue #2127, opened 2021-01-07
AEA fetch


DavidMinarsch:
**Describe the bug**
Run `aea fetch PUBLIC_ID` when the to be fetched agent of `PUBLIC_ID` already exists.

**To Reproduce**
See above

**Expected behavior**
Should say: `Error: Item "{item_name}" already exists in target folder.`

**Actual**
Instead says: `Error: Not found in Registry.` which is the wrong error message!

PR #2135, opened 2021-01-11
AEA-1211 raise error if target folder exists on aea fetch

solarw:
## Proposed changes

raise error if target folder exists on aea fetch

## Fixes

https://github.com/fetchai/agents-aea/issues/2127

## Types of changes

## Checklist

### DIFF
--- a/aea/cli/core.py
+++ b/aea/cli/core.py
@@ -18,6 +18,7 @@
 #
 # ------------------------------------------------------------------------------
 """Core definitions for the AEA command-line tool."""
+from typing import Optional
 
 import click
 
@@ -58,6 +59,7 @@ from aea.cli.scaffold import scaffold
 from aea.cli.search import search
 from aea.cli.transfer import transfer
 from aea.cli.upgrade import upgrade
+from aea.cli.utils.click_utils import registry_path_option
 from aea.cli.utils.config import get_or_create_cli_config
 from aea.cli.utils.constants import AUTHOR_KEY
 from aea.cli.utils.context import Context
@@ -77,11 +79,15 @@ from aea.helpers.win32 import enable_ctrl_c_support
     default=False,
     help="Skip consistency checks of agent during command execution.",
 )
+@registry_path_option
 @click.pass_context
-def cli(click_context, skip_consistency_check: bool) -> None:
+def cli(
+    click_context, skip_consistency_check: bool, registry_path: Optional[str]
+) -> None:
     """Command-line tool for setting up an Autonomous Economic Agent (AEA)."""
     verbosity_option = click_context.meta.pop("verbosity")
     click_context.obj = Context(cwd=".", verbosity=verbosity_option)
+    click_context.obj.registry_path = registry_path
     click_context.obj.set_config("skip_consistency_check", skip_consistency_check)
 
     # enables CTRL+C support on windows!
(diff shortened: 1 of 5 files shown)


aea/cli/core.py
```diff
@@ -77,11 +79,15 @@
...
+@registry_path_option
 @click.pass_context
-def cli(click_context, skip_consistency_check: bool) -> None:
+def cli(
+    click_context, skip_consistency_check: bool, registry_path: Optional[str]
+) -> None:
     """Command-line tool for setting up an Autonomous Economic Agent (AEA)."""
     verbosity_option = click_context.meta.pop("verbosity")
     click_context.obj = Context(cwd=".", verbosity=verbosity_option)
+    click_context.obj.registry_path = registry_path
```
DavidMinarsch, line 90: By default, registry path is defined via aea-config. So now we're potentially overwriting this with `None`

aea/cli/utils/click_utils.py
```diff
@@ -148,6 +148,17 @@ def wrapper(f):
     return wrapper
 
 
+def registry_path_option(f):
+    """Add registry path aea option."""
+    return option(
+        "--registry-path",
+        "--alias",
```
DavidMinarsch, line 155: Not sure about `--alias`. That will be very confusing.

aea/cli/registry/fetch.py
```diff
@@ -55,16 +56,19 @@ def fetch_agent(
...
-    api_path = f"/agents/{author}/{name}/{version}"
-    resp = request_api("GET", api_path)
-    file_url = resp["file"]
-
-    filepath = download_file(file_url, ctx.cwd)
 
     folder_name = target_dir or (name if alias is None else alias)
     aea_folder = os.path.join(ctx.cwd, folder_name)
+    if os.path.exists(aea_folder):
+        raise ClickException(f'Item "{folder_name}" already exists in target folder.')
```
DavidMinarsch, line 63: What will be CLI output here? I.e. a full stack trace? We don't want a stack trace, just a single line echo error message.

aea/cli/utils/click_utils.py
solarw: sorry, forgot to remove. removed

aea/cli/registry/fetch.py
solarw: no, ClickException is not printed as a traceback. printed as a message to stderr and exit

aea/cli/core.py
solarw: for example in `aea add` it used as this. so it can be overridden by the option if needed, By default, it's None and should not cause a problem, cause ctx.agent_config.registry_path is used directly.


```
# check in registry
    registry_path = (
        os.path.join(ctx.cwd, ctx.agent_config.registry_path)
        if ctx.registry_path is None
        else ctx.registry_path
    )
```

aea/cli/utils/click_utils.py
```diff
@@ -148,6 +148,16 @@ def wrapper(f):
     return wrapper
 
 
+def registry_path_option(f):
+    """Add registry path aea option."""
+    return option(
+        "--registry-path",
+        type=str,
```
DavidMinarsch, line 155: Please use `click.Path` to have click validate the path (see other commands)

solarw: thanks! nice point

DavidMinarsch (approved):
LGTM

marcofavorito (approved):
LGTM

Issue #2133, opened 2021-01-09
AEA upgrade missing functionalities


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
A few bits are missing:
- [x] when using `aea upgrade` across `aea` version boundaries (as specified in `aea_version`) the upgrade command refuses to  work without `-s`
- [x] when `-s` is used to upgrade, the `aea_version` is not upgraded
- [x] when using a third party AEA which is not up-to-date, the upgrade command will not be sufficient as it will not be able to update the vendorised packages by the third party author (e.g. `aea fetch Kraken/BlockEnergy:0.4.0 && cd BlockEnergy` then `aea upgrade` will lead to an inconsistent project). Possible solution: prompt user in this scenario and say upgrade will eject packages and upgrade them (to the extend of eject command and local package upgrades).

**Describe the solution you'd like**
Fix the above

PR #2137, opened 2021-01-11
AEA-1208 Make 'aea upgrade' to don't check aea version (#2133)

marcofavorito:
## Proposed changes

Make 'aea upgrade' to don't check aea version (#2133).

## Fixes

Addresses #2133

## Types of changes

## Checklist

## Further comments

n/a

### DIFF
--- a/aea/cli/upgrade.py
+++ b/aea/cli/upgrade.py
@@ -50,7 +50,7 @@ from aea.exceptions import enforce
     help_remote="For fetching packages only from remote registry.",
 )
 @click.pass_context
-@check_aea_project
+@check_aea_project(check_aea_version=False)
 def upgrade(click_context, local, remote):  # pylint: disable=unused-argument
     """Upgrade the packages of the agent."""
     ctx = cast(Context, click_context.obj)
--- a/aea/cli/upgrade.py
+++ b/aea/cli/upgrade.py
@@ -50,8 +50,10 @@ from aea.exceptions import enforce
     help_remote="For fetching packages only from remote registry.",
 )
 @click.pass_context
-@check_aea_project(check_aea_version=False)
-def upgrade(click_context, local, remote):  # pylint: disable=unused-argument
+@check_aea_project(  # pylint: disable=unused-argument,no-value-for-parameter
+    check_aea_version=False
+)
+def upgrade(click_context, local, remote):
     """Upgrade the packages of the agent."""
     ctx = cast(Context, click_context.obj)
     ctx.set_config("is_local", local and not remote)
--- a/aea/cli/upgrade.py
+++ b/aea/cli/upgrade.py
@@ -22,7 +22,9 @@ from pathlib import Path
 from typing import Dict, Iterable, List, Set, Tuple, cast
 
 import click
+from packaging.version import Version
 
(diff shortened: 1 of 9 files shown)


aea/cli/utils/decorators.py
```diff
@@ -100,25 +106,27 @@ def _validate_config_consistency(ctx: Context):
...
         if not skip_consistency_check:
-            _validate_config_consistency(ctx)
+            _validate_config_consistency(ctx, check_aea_version=check_aea_version)
     except Exception as e:  # pylint: disable=broad-except
         raise click.ClickException(str(e))
 
 
-def check_aea_project(f):
+@decorator_with_optional_params
+def check_aea_project(f, check_aea_version: bool = True):
```
marcofavorito, line 129: now we can use either

```python
@check_aea_project
def cli_function
```

or


```python
@check_aea_project(*args, **kwargs)
def cli_function
```

marcofavorito: If the wrapper has some positional arguments other than the function being wrapped, we can only use the latter version.

aea/cli/upgrade.py
```diff
@@ -50,7 +50,7 @@
     help_remote="For fetching packages only from remote registry.",
 )
 @click.pass_context
-@check_aea_project
+@check_aea_project(check_aea_version=False)
```
marcofavorito, line 53: now when running `aea upgrade`, aea version is not checked (but we do check fingerprints!)

aea/cli/utils/decorators.py
```diff
@@ -40,20 +40,26 @@
...
 
     :param ctx: the context
+    :param check_aea_version: whether it should check also the AEA version.
     :raise ValueError: if there is a missing configuration file.
                        or if the configuration file is not valid.
                        or if the fingerprints do not match
     """
+
+    if check_aea_version:
+        _check_aea_version(ctx.agent_config)
```
marcofavorito, line 61: this was missing!

aea/cli/utils/decorators.py
marcofavorito: Basically, all of this was to avoid to update `@check_aea_project` with `@check_aea_project()` everywhere, just because of the added parameter `check_aea_version`.

To avoid the updates, it could have been done in another way, by using `partial` on `check_aea_version`, instead of this machinery. But this is a more general solution that we can reuse.

DavidMinarsch (approved):
LGTM, although I'd like to see a test which tests the aea upgrade command's skipping of `aea_version` inconsistencies.

marcofavorito:
> LGTM, although I'd like to see a test which tests the aea upgrade command's skipping of `aea_version` inconsistencies.

Yes, will do in this PR.

solarw (approved):
LGTM

marcofavorito:
@DavidMinarsch commit c95f10f tries to address the point in #2133 :

>  when -s is used to upgrade, the aea_version is not upgraded

Basically, we always check that the version of the installed `aea` is compatible with the project being upgraded. If not, we replace it using `compute_specifier_from_version` (version range varying the minor), as we do for `aea create`.

PR #2142, opened 2021-01-12
AEA-1208 Fix upgrade command to eject packages when needed (#2133)

marcofavorito:
## Proposed changes

Fix upgrade command to eject packages when needed (#2133)

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a


aea/cli/upgrade.py
```diff
@@ -103,23 +110,42 @@ def upgrade_project(ctx: Context) -> None:  # pylint: disable=unused-argument
...
+        old_aea_version_specifier = ctx.agent_config.aea_version
         new_aea_version_specifier = compute_specifier_from_version(version)
+        click.echo(
+            f"Upgrading version specifier from {old_aea_version_specifier} to {new_aea_version_specifier}."
+        )
         ctx.agent_config.aea_version = new_aea_version_specifier
         ctx.dump_agent_config()
 
+    eject_helper = InteractiveEjectHelper(ctx, agent_items, interactive=interactive)
+    eject_helper.get_latest_versions()
```
marcofavorito, line 135: we have to get the latest versions of all the packages, this makes the command a bit slow

marcofavorito: we can refine by filtering non-vendor packages, but not if package is in requirements of another package, as done before (or not?)

solarw (approved):
LGTM

aea/cli/upgrade.py
```diff
@@ -103,23 +110,42 @@ def upgrade_project(ctx: Context) -> None:  # pylint: disable=unused-argument
...
+        click.echo("Everything is already up to date!")
+        return
+    if not eject_helper.can_eject():
+        click.echo("Abort.")
+        return
+    eject_helper.eject()
+
     for package_id, deps in agent_items.items():
+        if package_id in eject_helper.to_eject:
+            continue
```
DavidMinarsch, line 146: can we isolate some code blocks here and give them proper names; for instance starting at 144, what is the code doing? would be much easier to read if this was better structured

aea/cli/upgrade.py
```diff
@@ -316,6 +357,115 @@ def add_item(self) -> None:
...
+        Ask the user permission for ejection of a package.
+
+        :param package_id: the package id.
+        :param dependencies_to_upgrade: the dependencies to upgrade.
+        :return: True or False, depending on the answer of the user.
+        """
+        package_type = str(package_id.package_type).capitalize()
+        message = (
+            f"{package_type} {package_id.public_id} prevents the upgrade of "
+            f"the following non-vendor packages:\n"
```
DavidMinarsch, line 460:
```suggestion
            f"the following vendor packages:\n"
```

aea/cli/upgrade.py
```diff
@@ -316,6 +357,115 @@ def add_item(self) -> None:
...
+            f"{pprint.pformat(dependencies_to_upgrade)}\n"
+            f"as there isn't an update available. "
```
DavidMinarsch, line 462:
```suggestion
            f"as there isn't a compatible version available on the AEA registry. "
```

DavidMinarsch (commented):
Looks mostly good to me. A few observations in comments and some more here:
- Run `aea fetch Aeolist/simple_weather_buyer:0.2.0` then `cd simple_weather_buyer` then `aea upgrade`. Command works as expected, however some issues remain: a) author is not updated in `aea-config`. Every time we upgrade we are changing the agent so the author needs to be changed if it doesn't match. Version needs to be reset if doesn't match (same as for eject). b) Eject command does not update the version ranges in the ejected packages.

aea/cli/upgrade.py
DavidMinarsch: I am ok with how it is. We don't care if the command is not that fast.

aea/configurations/constants.py
```diff
@@ -59,7 +59,6 @@
 DEFAULT_CONNECTION_CONFIG_FILE = "connection.yaml"
 DEFAULT_CONTRACT_CONFIG_FILE = "contract.yaml"
 DEFAULT_PROTOCOL_CONFIG_FILE = "protocol.yaml"
-DEFAULT_LICENSE = "Apache-2.0"
```
marcofavorito, line 62: defined above without usage

DavidMinarsch (approved):
LGTM

Issue #2128, opened 2021-01-08
Issues with TAC


5A11:
**List of issues with TAC**
Note: These issues may not specifically be in the TAC code but in the core aea code, but they appear in TAC.

**Issue 1**:
**Problem**:	Incorrect `dialogue.str()`
**Detail**: 	The function just interleaves messages, while the order may not be as such.
**Solution**:	It must order and represent messages based on their ids.

**Issue 2**:
**Problem**:	Simultaneous message sending
**Detail**:	When two agents simultaneously send a message to each other, the order becomes corrupted.
**Example**:	Consider the following:

(agent) (sends/receives) (message_id, message) 
X sends (1, a)
Y receives (1, a)
Y sends (2, b)
X receives (2, b)
X sends (3, c)
Y sends (3, d)

At this point:

X’s dialogue: (1, a) —> (2, b) —> (3, c)
Y’s dialogue: (1, a) —> (2, b) —> (3, d)

Now:

X receives (3, d)
Y receives (3, c)

And both those messages are dropped because their message ids are incorrect with respect to the local dialogue structure the two agents hold.
**Solution**:...

DavidMinarsch:
Message id's are only unique for the tuple `(sender, dialogue_ref)`. We need to account for this in the code base.

PR #2151, opened 2021-01-14
AEA-1210 dialogue resolution fix.

solarw:
## Proposed changes

dialogue resolution fix.

## Fixes

https://github.com/fetchai/agents-aea/issues/2128

## Types of changes

## Checklist


aea/protocols/dialogue/base.py
```diff
@@ -680,6 +680,12 @@ def _is_belonging_to_dialogue(self, message: Message) -> bool:
             result = other_initiated_dialogue_label in self.dialogue_labels
         return result
 
+    def get_next_message_id(self) -> int:
+        """Get next outgoinf message id."""
```
DavidMinarsch, line 684:
```suggestion
        """Get next outgoing message id."""
```

aea/protocols/dialogue/base.py
```diff
@@ -680,6 +680,12 @@ def _is_belonging_to_dialogue(self, message: Message) -> bool:
...
+        if self.last_outgoing_message:
+            return self.last_outgoing_message.message_id + 1
+        return Dialogue.STARTING_MESSAGE_ID
```
DavidMinarsch, line 687: Incorrect

tests/test_protocols/test_dialogue/test_msg_resolve.py
```diff
@@ -0,0 +1,104 @@
...
+    dialogue = part2.update(msg)
+    assert dialogue
+    game_data_msg = dialogue.reply(performative=TacMessage.Performative.GAME_DATA)
+
+    dialogue = part1.update(game_data_msg)
+    assert dialogue
+    transaction1_msg = dialogue.reply(performative=TacMessage.Performative.TRANSACTION)
+    transaction2_msg = dialogue.reply(performative=TacMessage.Performative.TRANSACTION)
+
+    print(transaction1_msg, transaction1_msg)
```
DavidMinarsch, line 81: remove print

aea/protocols/dialogue/base.py
```diff
@@ -857,30 +858,74 @@ def _basic_validation_non_initial_message(
...
             )
 
-        last_message_id = self.last_message.message_id  # type: ignore
-        if message_id != last_message_id + 1:
-            return (
-                False,
-                "Invalid message_id. Expected {}. Found {}.".format(
-                    last_message_id + 1, message_id
-                ),
-            )
```
DavidMinarsch, line 867: We should have a simple method which checks correct incrementation.

aea/protocols/dialogue/base.py
```diff
@@ -857,30 +858,74 @@ def _basic_validation_non_initial_message(
...
+                    False,
+                    "Invalid target. Expected a value less than or equal to {}. Found {}.".format(
+                        self.last_incoming_message.message_id, target
+                    ),
+                )
+            target_message = None
+            for msg in self._incoming_messages:
+                if msg.message_id == target:
+                    target_message = msg
+        else:
```
DavidMinarsch, line 892: A lot of duplication between both blocks. Opportunity to refactor and extract helper methods

aea/protocols/dialogue/base.py
```diff
@@ -892,32 +937,6 @@ def _basic_validation_non_initial_message(
 
         return True, "The non-initial message passes basic validation."
 
-    def _additional_validation(self, message: Message) -> Tuple[bool, str]:
```
DavidMinarsch, line 895: Why are we not keeping this. It was a good separation

aea/protocols/dialogue/base.py
```diff
@@ -720,7 +726,7 @@ def reply(
 
         reply = self._message_class(
             dialogue_reference=self.dialogue_label.dialogue_reference,
-            message_id=last_message.message_id + 1,
+            message_id=self.get_next_message_id(),
```
DavidMinarsch, line 729: A reply needs to use a `message_id` unused by and unknown  to `self`. So the `message_id` needs to be the max of all messages ids (incoming and outgoing) + 1.

aea/protocols/dialogue/base.py
```diff
@@ -699,7 +705,7 @@ def reply(
 
         :return: the reply message if it was successfully added as a reply, None otherwise.
         """
-        last_message = self.last_message
+        last_message = self.last_incoming_message
```
DavidMinarsch, line 708: This should not be changed. The last message != last_incoming_message in general.

aea/protocols/dialogue/base.py
```diff
@@ -749,13 +755,6 @@ def _validate_next_message(self, message: Message) -> Tuple[bool, str]:
         if not is_basic_validated:
             return False, msg_basic_validation
 
-        (
-            result_additional_validation,
-            msg_additional_validation,
-        ) = self._additional_validation(message)
-        if not result_additional_validation:
-            return False, msg_additional_validation
```
DavidMinarsch, line 757: I liked the previous split. Now we just have one massive method.

aea/protocols/dialogue/base.py
DavidMinarsch: See above, a valid message id increments the max of all seen message ids by 1

aea/protocols/dialogue/base.py
```diff
@@ -857,30 +858,74 @@ def _basic_validation_non_initial_message(
...
+                    False,
+                    "Invalid message_id. Expected  message id {} > {}.".format(
+                        message_id, self.get_next_message_id()
+                    ),
+                )
+
+            if not self.last_incoming_message:
+                return False, "Can not reply cause not incoming messages"
+
+            if target > self.last_incoming_message.message_id:
```
DavidMinarsch, line 881: Insufficient. Validity of targets can only be checked in conjunction with the dialogue reply structure (see GDoc).

aea/protocols/dialogue/base.py
DavidMinarsch: Message id must be evaluated as a tuple (message_id, sender), because only then do we have uniqueness.

aea/protocols/dialogue/base.py
```diff
@@ -857,30 +858,74 @@ def _basic_validation_non_initial_message(
...
                 ),
             )
-        if last_message_id < target:
+
+        if is_outgoing:
+            if message_id != self.get_next_message_id():
+                return (
+                    False,
+                    "Invalid message_id. Expected  message id {} > {}.".format(
+                        message_id, self.get_next_message_id()
```
DavidMinarsch, line 874:
```suggestion
                    "Invalid message_id. Expected  message id {} == {}.".format(
                        message_id, self.get_next_message_id()
```

aea/protocols/dialogue/base.py
```diff
@@ -857,30 +858,74 @@ def _basic_validation_non_initial_message(
...
+                    ),
+                )
+
+            if not self.last_incoming_message:
+                return False, "Can not reply cause not incoming messages"
```
DavidMinarsch, line 879: remove

packages/fetchai/protocols/default/message.py
```diff
@@ -250,13 +250,6 @@ def _is_consistent(self) -> bool:
                         self.target
                     ),
                 )
-            else:
-                enforce(
-                    0 < self.target < self.message_id,
-                    "Invalid 'target'. Expected an integer between 1 and {} inclusive. Found {}.".format(
-                        self.message_id - 1, self.target,
-                    ),
-                )
```
DavidMinarsch, line 259: We should probably extract the whole check (lines 245-259) into the message base class, use absolute value and then just call it here. That way we also avoid code duplication.

packages/fetchai/skills/simple_data_request/handlers.py
```diff
@@ -63,6 +63,7 @@ def handle(self, message: Message) -> None:
         # recover dialogue
         http_dialogues = cast(HttpDialogues, self.context.http_dialogues)
         http_dialogue = cast(HttpDialogue, http_dialogues.update(http_msg))
+
```
DavidMinarsch, line 66: please revert to avoid unnecessary bumping

tests/test_packages/test_protocols/test_default.py
```diff
@@ -129,38 +129,6 @@ def test_default_valid_performatives():
     )
 
 
-def test_light_protocol_rule_3_target_0():
```
DavidMinarsch, line 132: See above, we should still have this rule, but amended.

tests/test_packages/test_skills/test_tac_control/test_handlers.py
```diff
@@ -481,7 +481,7 @@ def test_on_transaction_valid(self):
             performative=TacMessage.Performative.TRANSACTION_CONFIRMATION,
             to=tac_participant_counterparty,
             sender=self.skill.skill_context.agent_address,
-            target=counterparty_dialogue.last_message.message_id - 1,
+            target=counterparty_dialogue.last_message.message_id + 1,
```
DavidMinarsch, line 484: Hm I think this was correct, this is supposed to be the target.

tests/test_packages/test_skills/test_tac_negotiation/test_handlers.py
```diff
@@ -1136,7 +1136,6 @@ def test_handle_signed_message_i(self):
             performative=FipaMessage.Performative.MATCH_ACCEPT_W_INFORM,
             to=fipa_dialogue.dialogue_label.dialogue_opponent_addr,
             # (line below) match-accept is already added to fipa_dialogue, hence "-1"
-            target=fipa_dialogue.last_message.message_id - 1,
```
DavidMinarsch, line 1139: target_message instead of target

tests/test_packages/test_skills/test_tac_participation/test_behaviours.py
```diff
@@ -241,7 +241,6 @@ def test_act_ii(self):
                 performative=TacMessage.Performative.TRANSACTION,
                 to=COUNTERPARTY_ADDRESS,
                 sender=self.skill.skill_context.agent_address,
-                target=tac_dialogue.last_message.message_id - no_tx + count,
```
DavidMinarsch, line 244: see above

aea/test_tools/test_skill.py
```diff
@@ -367,8 +364,8 @@ def prepare_skill_dialogue(
             message = self.build_incoming_message(
                 message_type=dialogues.message_class,
                 dialogue_reference=dialogue_reference,
-                message_id=message_id,
-                target=target,
+                message_id=Dialogue.STARTING_MESSAGE_ID,
```
DavidMinarsch, line 367: No, incoming message id must be whatever the id of the provided message.

aea/test_tools/test_skill.py
```diff
@@ -367,8 +364,8 @@ def prepare_skill_dialogue(
...
+                target=target or 0,
```
DavidMinarsch, line 368: target cannot be None, it should be a default value otherwise

aea/protocols/dialogue/base.py
```diff
@@ -857,66 +851,99 @@ def _basic_validation_non_initial_message(
...
+        if err:
+            return False, err
+
+        return True, "The non-initial message passes basic validation."
+
+    def _validate_message_target(self, message: Message) -> Optional[str]:
+        """Check message target corresponds to messages in the dialogue, if not return error string."""
+        target = message.target
+        performative = message.performative
+        if abs(target) < 1:
```
DavidMinarsch, line 872: use constant `STARTING_MESSAGE_ID`

aea/protocols/dialogue/base.py
```diff
@@ -857,66 +851,99 @@ def _basic_validation_non_initial_message(
...
             )
-        if last_message_id < target:
-            return (
-                False,
-                "Invalid target. Expected a value less than or equal to {}. Found {}.".format(
-                    last_message_id, target
-                ),
+
+        # quick target check.
+        if abs(target) > abs(cast(int, self._last_message_id)):
```
DavidMinarsch, line 878: ?

tests/test_packages/test_skills/test_tac_participation/test_behaviours.py
solarw: this quite hard to fix and not sure the target should be checked here, the test done for another check

tests/test_packages/test_skills/test_tac_negotiation/test_handlers.py
solarw: update

aea/test_tools/test_skill.py
solarw: done

tests/test_packages/test_skills/test_tac_control/test_handlers.py
solarw: done

packages/fetchai/skills/simple_data_request/handlers.py
solarw: done

aea/protocols/dialogue/base.py
```diff
@@ -803,7 +800,7 @@ def _basic_validation_initial_message(self, message: Message) -> Tuple[bool, str
                 ),
             )
 
-        if message_id != Dialogue.STARTING_MESSAGE_ID:
+        if abs(message_id) != Dialogue.STARTING_MESSAGE_ID:
```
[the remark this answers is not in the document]
solarw: done

aea/protocols/dialogue/base.py
solarw: done

aea/protocols/dialogue/base.py
solarw: done

aea/protocols/dialogue/base.py
```diff
@@ -857,66 +851,99 @@ def _basic_validation_non_initial_message(
...
-    def _additional_validation(self, message: Message) -> Tuple[bool, str]:
-        """
-        Check whether 'message' is a valid next message in the dialogue, according to additional rules.
+    def _validate_message_id(self, message: Message) -> Optional[str]:
+        """Check message id corresponds to message id sequences, if not return error string."""
+        is_outgoing = message.to != self.self_address
+        if is_outgoing:
+            next_message_id = self.get_outgoing_next_message_id()
+        else:
+            next_message_id = self.get_incoming_next_message_id()
```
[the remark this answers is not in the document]
solarw: done

aea/protocols/dialogue/base.py
```diff
@@ -1740,12 +1787,12 @@ def update(self, message: Message) -> Optional[Dialogue]:
...
             dialogue_reference[0] != Dialogue.UNASSIGNED_DIALOGUE_REFERENCE
             and dialogue_reference[1] == Dialogue.UNASSIGNED_DIALOGUE_REFERENCE
-            and message.message_id == 1
+            and message.message_id == Dialogue.STARTING_MESSAGE_ID
         )
         is_incomplete_label_and_non_initial_msg = (
             dialogue_reference[0] != Dialogue.UNASSIGNED_DIALOGUE_REFERENCE
             and dialogue_reference[1] == Dialogue.UNASSIGNED_DIALOGUE_REFERENCE
-            and message.message_id > 1
+            and message.message_id != Dialogue.STARTING_MESSAGE_ID
```
DavidMinarsch, line 1795:
```suggestion
            and message.message_id not in {Dialogue.STARTING_MESSAGE_ID, Dialogue.STARTING_TARGET}
```

aea/protocols/dialogue/base.py
```diff
@@ -857,66 +851,119 @@ def _basic_validation_non_initial_message(
...
-            )
-        if target < 1:
-            return (
-                False,
-                "Invalid target. Expected a value greater than or equal to 1. Found {}.".format(
-                    target
-                ),
+        # it's non initial check, assume there some messages in the dialogue
+        if self.is_empty:
+            return False, "Can not reply cause no messages in the dialogue."
```
5A11, line 856: No need for this check. This method is for non-initial case and is called when self.is_empty() fails.
Also the error message is incorrect.

aea/protocols/dialogue/base.py
```diff
@@ -857,66 +851,119 @@ def _basic_validation_non_initial_message(
...
+            message.message_id == self.STARTING_MESSAGE_ID
+            and target != self.STARTING_TARGET
+        ):
+            return "Invalid target. Expected 0. Found {}.".format(target)
+
+        if (
+            message.message_id != self.STARTING_MESSAGE_ID
+            and target == self.STARTING_TARGET
+        ):
+            return "Invalid target. Expected a absolute value greater or less than 0. Found {}.".format(
```
5A11, line 883: `Expected a absolute value greater or less than 0` --> `Expected a non-zero integer.`

aea/protocols/dialogue/base.py
```diff
@@ -614,7 +609,14 @@ def _has_message_id(self, message_id: int) -> bool:
...
 
-        return self.STARTING_MESSAGE_ID <= message_id <= self.last_message.message_id  # type: ignore
+        if abs(message_id) < self.STARTING_MESSAGE_ID:
+            return False
+
+        for msg in chain(self._incoming_messages, self._outgoing_messages):
+            if msg.message_id == message_id:
+                return True
+
+        return False
```
5A11, line 619: Instead of looping through all messages, you could simply check if it is between 0 and the last negative/positive message id, if message_id itself is negative/positive respectively.

aea/protocols/dialogue/base.py
```diff
@@ -857,66 +851,119 @@ def _basic_validation_non_initial_message(
...
+    def _validate_message_target(self, message: Message) -> Optional[str]:
+        """Check message target corresponds to messages in the dialogue, if not return error string."""
+        target = message.target
+        performative = message.performative
+
+        if (
+            message.message_id == self.STARTING_MESSAGE_ID
+            and target != self.STARTING_TARGET
+        ):
+            return "Invalid target. Expected 0. Found {}.".format(target)
```
5A11, line 877: Because of this if block, this method now covers initial messages as well. Either specialise it to the non-initial case, or use this method in the `_basic_validation_initial_message()` method as well.

aea/protocols/dialogue/base.py
```diff
@@ -857,66 +851,119 @@ def _basic_validation_non_initial_message(
...
+            return "Invalid target. Expected a value less than or equal to abs({}). Found abs({}).".format(
+                self._last_message_id, target
             )
 
-        target_message = self._get_message(target)
+        # detailed target check
+        target_message = self.get_message_by_id(target)
+
+        if not target_message:
+            return "Invalid target {}. target_message can not be found.".format(target)
```
5A11, line 905: Line 879 to 902 can be replaced with a simple call to `has_message_id(target)`.

aea/protocols/dialogue/base.py
DavidMinarsch: Here we can indeed use the fact that the message id on either side needs to be monotonically increasing/decreasing.

aea/protocols/dialogue/base.py
```diff
@@ -857,66 +851,119 @@ def _basic_validation_non_initial_message(
...
-        :return: Boolean result, and associated message.
-        """
-        if self.is_empty:
-            return True, "The message passes additional validation."
+        return None
 
-        last_target = self.last_message.target  # type: ignore
-        if message.target == last_target + 1:
-            return True, "The message passes additional validation."
+    def iter_all_messages(self) -> Iterable:
```
5A11, line 933: There are places where `chain(self._incoming_messages, self._outgoing_messages)` is used, yet this method encapsulates this. Why not use this method instead?

5A11: Although the place it is used it is not needed as per my comment. If removed there and this method isn't used, remove this method.

aea/protocols/dialogue/base.py
```diff
@@ -857,66 +851,119 @@ def _basic_validation_non_initial_message(
...
-            "Invalid target. Expected {}. Found {}.".format(
-                last_target + 1, message.target
-            ),
-        )
+    def get_message_by_id(self, message_id: int) -> Optional[Message]:
+        """Get message by id, if not presents return None."""
+        for message in self.iter_all_messages():
+            if message.message_id == message_id:
+                return message
+        return None
```
5A11, line 942: This isn't efficient. No need to loop through all messages. Just access the `message_id-1`th message of incoming/outgoing_messages respecting the positive or negative message_id. 

We know messages in say outgoing (or incoming_messages) list preserves their order with respect to other outgoing (incoming) messages.

5A11 (commented):
See comments

packages/fetchai/protocols/default/message.py
solarw: would be nice to solve it in another PR with generator/base message class refactoring

aea/protocols/dialogue/base.py
solarw: target message is used next, but updated get_message_by_id and has_mesasge_id to use without iterations

aea/protocols/dialogue/base.py
solarw: updated, will be used in initial message validation

aea/protocols/dialogue/base.py
solarw: updated

aea/protocols/dialogue/base.py
solarw: done

aea/protocols/dialogue/base.py
solarw: +

aea/protocols/dialogue/base.py
solarw: agree, this is non-public method and is_empty check done another place

aea/protocols/dialogue/base.py
solarw: +

aea/protocols/dialogue/base.py
solarw: removed, thanks

DavidMinarsch (approved):
LGTM

Issue #2136, opened 2021-01-11
Add protobuf linter


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
Currently, protobuf files are not linted.

**Describe the solution you'd like**
Add https://github.com/yoheimuta/protolint

Issue #1631, opened 2020-08-10
Windows line endings might affect development workflow


marcofavorito:
**Is your feature request related to a problem? Please describe.**

Windows line endings might affect development workflow on Windows systems. Developers should be aware of that.

**Describe the solution you'd like**

Better documentation of the needed actions to take when working with the framework on Windows systems; e.g. set the editor to use `LF` line endings instead of `CR LF`, use `dos2unix` tool etc.

**Describe alternatives you've considered**
n/a

**Additional context**

We already replace Windows line endings for text files with the common denominator of Linux line endings: https://github.com/fetchai/agents-aea/pull/1624/. However, the workaround is a bit naive and could be broken unexpectedly. 

Another problem is that the builtin function `open` uses universal newlines by default; hence line endings of text files in output will be replaced with the system's default ones. `ipfs` seems to care about line endings, so even if the framework replaces line endings for fingerprint purposes, it might diverge from `ipfs add` output. A more robust approach would be to always use `newline="\n"` when `open` is used in write text mode (or `pathlib.Path.write_text` variant)

PR #2171, opened 2021-01-20
AEA-1230 protolint check

solarw:
## Proposed changes

protolint check

## Fixes

https://github.com/fetchai/agents-aea/issues/2136

## Types of changes

## Checklist


protolint.yaml
```diff
@@ -0,0 +1,8 @@
+lint:
+  rules:
+    remove:
+      - MESSAGE_NAMES_UPPER_CAMEL_CASE
+      - ENUM_FIELD_NAMES_ZERO_VALUE_END_WITH
+      - PACKAGE_NAME_LOWER_CASE
+      - REPEATED_FIELD_NAMES_PLURALIZED
+      - FIELD_NAMES_LOWER_SNAKE_CASE
```
DavidMinarsch, line 8: We should try to fix `MESSAGE_NAMES_UPPER_CAMEL_CASE`, `REPEATED_FIELD_NAMES_PLURALIZED` and `FIELD_NAMES_LOWER_SNAKE_CASE`.

.github/workflows/workflow.yml
```diff
@@ -152,8 +153,27 @@ jobs:
...
+          go-version: '^1.14.0'
+      - name: Install protolint (ubuntu-latest)
+        run: |
+          sudo apt-get update --fix-missing
+          sudo apt-get autoremove
+          sudo apt-get autoclean
+          make protolint_install
+      - name: Protolint check
+        run: |
+          make protolint
```
marcofavorito, line 176: What do you think in adding a testing environment in `tox.ini`? So we can call `tox -e protolint` and have uniformity with other test environments.

aea/protocols/generator/base.py
```diff
@@ -1921,13 +1923,32 @@ def generate_protobuf_only_mode(self) -> None:
...
+        if sys.platform.startswith("win"):
+            protolint_base_cmd = "protolint"
+        else:
+            protolint_base_cmd = "PATH=${PATH}:${GOPATH}/bin/:~/go/bin protolint"
+
+        if call(f"{protolint_base_cmd} version", shell=True) != 0:  # nosec
+            raise ValueError("protolint is not installed!")
+
+        cmd = f'{protolint_base_cmd} lint -fix "{filepath}"'
+        call(cmd, shell=True)  # nosec
```
marcofavorito, line 1950: `shell=True` is not recommended for security reason, yet I think we have good control on the `cmd` variable. Maybe some checks on `filepath`? E.g. check that it is actually a path.

aea/protocols/generator/common.py
```diff
@@ -364,7 +364,7 @@ def _create_protocol_file(
     """
     pathname = os.path.join(path_to_protocol_package, file_name)
 
-    with open(pathname, "w") as file:
+    with open(pathname, "w", newline="\n") as file:
```
marcofavorito, line 367: why this? I guess related to #1631

marcofavorito (approved):
LGTM. I wouldn't mind to see a new line in `scripts/generate_all_protocols.py` to check `protolint` is installed:

https://github.com/fetchai/agents-aea/blob/feature/protolint/scripts/generate_all_protocols.py#L330-L332

marcofavorito:
Also, maybe some relevant parts of the docs should be updated to add protolint as necessary dependency.

.github/workflows/workflow.yml
DavidMinarsch: I agree!

DavidMinarsch (commented):
Screenshot 2021-01-25 at 13 23 02
some coverage gaps remain ^

aea/protocols/generator/common.py
solarw: the reason, that windows generation differs from unix one cause new lines. so decided to unify it

DavidMinarsch (approved):
LGTM

PR #2184, opened 2021-01-22
AEA-1238 filter config overrides

solarw:
## Proposed changes

filter data for config manager overrides, to apply only updated values

## Types of changes

## Checklist

### DIFF
--- a/aea/configurations/manager.py
+++ b/aea/configurations/manager.py
@@ -508,7 +508,7 @@ class AgentConfigManager:
 
         self.agent_config.update(overrides, env_vars_friendly=self.env_vars_friendly)
 
-    def _filter_overrides(self, overrides: Dict):
+    def _filter_overrides(self, overrides: Dict) -> Dict:
         """Stay only updated values for agent config."""
         agent_overridable, components_overridables = self.get_overridables()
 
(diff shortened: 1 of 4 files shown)


aea/configurations/manager.py
```diff
@@ -498,6 +508,15 @@ def update_config(self, overrides: Dict) -> None:
 
         self.agent_config.update(overrides, env_vars_friendly=self.env_vars_friendly)
 
+    def _filter_overrides(self, overrides: Dict):
```
DavidMinarsch, line 511: please add typing of return type

aea/configurations/validation.py
```diff
@@ -313,3 +313,31 @@ def check_excludes(path):
             )
 
     return errors
+
+
+SAME_MARK = object()
+
+
+def filter_data(base: Any, updates: Any) -> Any:
```
DavidMinarsch, line 321: Can we use type `JSONLike` instead (from `aea.common`)?

DavidMinarsch (approved):
LGTM

DavidMinarsch:
Replaces #2126

aea/configurations/validation.py
solarw: there is not only json friendly data, as i remember component_id also can be provided

aea/configurations/manager.py
solarw: done

PR #2197, opened 2021-01-26
Feature/acn relayed connection reconnect

lrahmani:
## Proposed changes

This pr allows a dht client to be notified when its relay peer is disconnected.
It'll loop for ever until it is connected back.
It won't necessarily re-register itself (it assumes that relay peer persist its agents records)


DavidMinarsch (approved):
LGTM

PR #2203, opened 2021-01-28
Experimental protocol id and spec mapping

DavidMinarsch:
## Proposed changes

This PR proposes an alternative implementation.

## Fixes

-

## Types of changes

## Checklist

## Further comments

-

### DIFF
--- a/aea/cli/interact.py
+++ b/aea/cli/interact.py
@@ -248,7 +248,11 @@ def _try_construct_envelope(
             counterparty=agent_name, performative=performative, content=message,
         )
         envelope = Envelope(
-            to=msg.to, sender=msg.sender, protocol_id=msg.protocol_id, message=msg,
+            to=msg.to,
+            sender=msg.sender,
+            protocol_id=msg.protocol_id,
+            message=msg,
+            is_framework_set=True,
         )
     except InterruptInputException:
         click.echo("Interrupting input, checking inbox ...")
--- a/aea/mail/base.py
+++ b/aea/mail/base.py
@@ -20,7 +20,7 @@
 
 import logging
 from abc import ABC, abstractmethod
-from typing import Optional, Tuple, Union, cast
+from typing import Optional, Tuple, Union
 from urllib.parse import urlparse
 
 from aea.common import Address
@@ -317,7 +317,7 @@ class Envelope:
     @staticmethod
     def _get_protocol_specification_id_by_the_protocol_id(
         protocol_id: PublicId,
-    ) -> PublicId:
+    ) -> Optional[PublicId]:
         """
         Get protocol specification id  from protocol id.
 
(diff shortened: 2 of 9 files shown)


aea/mail/base.py
```diff
@@ -327,31 +327,8 @@ def _get_protocol_specification_id_by_the_protocol_id(
...
-
-        :return: PublicId
-        """
-        return (
-            ProtocolSpecificationsRegistry.get_protocol_id_by_specification_id(
-                protocol_specification_id
-            )
-            or protocol_specification_id
+        return ProtocolSpecificationsRegistry.get_specification_id_by_protocol_id(
+            protocol_id
```
DavidMinarsch, line 331: If we cannot resolve, we return None. No magic resolution.

aea/mail/base.py
```diff
@@ -392,38 +369,42 @@ def __init__(
...
 
-        if not protocol_id:
-            protocol_id = self._get_protocol_id_by_protocol_specification_id(
-                cast(PublicId, protocol_specification_id)
-            )
-        if not protocol_specification_id:
-            # if no protocol_spcification_id, try to resolve by protocol_id
-            protocol_specification_id = self._get_protocol_specification_id_by_the_protocol_id(
-                protocol_id
+        if protocol_specification_id is None:
```
DavidMinarsch, line 402: An envelope must have a spec id. If it comes over the wire, we have it. If it comes from within AEA we can resolve it (protocol_id > spec_id works globally)

aea/protocols/base.py
```diff
@@ -434,16 +438,3 @@ def get_specification_id_by_protocol_id(
         :return: PublicId if protocol registered otherwise None
         """
         return cls.PROTOCOL_TO_SPECIFICATION.get(protocol_id, None)
-
-    @classmethod
```
DavidMinarsch, line 438: Not safe to do globally, must be scoped to agent instance.

aea/cli/interact.py
```diff
@@ -248,7 +248,11 @@ def _try_construct_envelope(
             counterparty=agent_name, performative=performative, content=message,
         )
         envelope = Envelope(
-            to=msg.to, sender=msg.sender, protocol_id=msg.protocol_id, message=msg,
+            to=msg.to,
+            sender=msg.sender,
+            protocol_id=msg.protocol_id,
```
DavidMinarsch, line 253: fix

aea/mail/base.py
```diff
@@ -392,38 +373,47 @@ def __init__(
...
-            )
-        if not protocol_specification_id:
-            # if no protocol_spcification_id, try to resolve by protocol_id
-            protocol_specification_id = self._get_protocol_specification_id_by_the_protocol_id(
-                protocol_id
-            )
+        if protocol_specification_id is None and not is_framework_set:
+            protocol_specification_id = protocol_id
+        if protocol_id is None and not is_framework_set:
+            protocol_id = protocol_specification_id
```
DavidMinarsch, line 409: We need to unfortunately introduce the flag and do this here. Otherwise every single use of the Envelope needs the registry. The alternative is to use a new type of Envelope framework side.

aea/mail/base.py
DavidMinarsch: (Problem is when it gets used standalone, see above)

aea/multiplexer.py
```diff
@@ -855,5 +855,6 @@ def put_message(
             protocol_id=message.protocol_id,
```
DavidMinarsch, line 855: not needed

DavidMinarsch:
Closed in favour of #2206

PR #2216, opened 2021-02-02
AEA-1268 Feature/misc

### LINKED ISSUE #2128: Issues with TAC

DavidMinarsch:
## Proposed changes

This PR plugs miscellaneous test gaps and addresses #2128 

## Fixes

#2128

## Types of changes

## Checklist

## Further comments

-

### DIFF
--- a/aea/protocols/dialogue/base.py
+++ b/aea/protocols/dialogue/base.py
@@ -25,7 +25,6 @@ This module contains the classes required for dialogue management.
 - Dialogues: The dialogues class keeps track of all dialogues.
 """
 import inspect
-import itertools
 import secrets
 import sys
 from collections import defaultdict, namedtuple
@@ -353,6 +352,7 @@ class Dialogue(metaclass=_DialogueMeta):
         self._message_class = message_class
         self._terminal_state_callbacks: Set[Callable[["Dialogue"], None]] = set()
         self._last_message_id: Optional[int] = None
+        self._ordered_message_ids: List[int] = []
 
     def add_terminal_state_callback(self, fn: Callable[["Dialogue"], None]) -> None:
         """
@@ -371,6 +371,7 @@ class Dialogue(metaclass=_DialogueMeta):
             and self.message_class == other.message_class
             and self._incoming_messages == other._incoming_messages
             and self._outgoing_messages == other._outgoing_messages
+            and self._ordered_message_ids == other._ordered_message_ids
             and self.role == other.role
             and self.self_address == other.self_address
         )
@@ -383,6 +384,8 @@ class Dialogue(metaclass=_DialogueMeta):
             "role": self._role.value,
             "incoming_messages": [i.json() for i in self._incoming_messages],
             "outgoing_messages": [i.json() for i in self._outgoing_messages],
+            "last_message_id": self._last_message_id,
+            "ordered_message_ids": self._ordered_message_ids,
         }
         return data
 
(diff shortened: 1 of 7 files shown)


aea/protocols/dialogue/base.py
```diff
@@ -353,6 +352,7 @@ def __init__(
         self._message_class = message_class
         self._terminal_state_callbacks: Set[Callable[["Dialogue"], None]] = set()
         self._last_message_id: Optional[int] = None
+        self._ordered_message_ids: List[int] = []
```
DavidMinarsch, line 355: This is useful to keep track of, otherwise we cannot debug the order in which messages where observed by this dialogue.

aea/protocols/dialogue/base.py
```diff
@@ -859,7 +868,7 @@ def _validate_message_target(self, message: Message) -> Optional[str]:
 
         if abs(target) > max(latest_ids):
             return "Invalid target. Expected a value less than or equal to abs({}). Found abs({}).".format(
-                self._last_message_id, target
+                max(latest_ids), abs(target)
```
DavidMinarsch, line 871: This seemed to have been a bug

5A11: I think it should be `target`. The text reads should be less than `abs(max(latest_id))`. Found `abs(X)`. Here X must be `target` and not `abs(target)`, no?

aea/protocols/dialogue/base.py
solarw: probably can reconstruct it from messages stored?
cause  have ids and targets

aea/protocols/dialogue/base.py
```diff
@@ -371,6 +371,7 @@ def __eq__(self, other) -> bool:
             and self.message_class == other.message_class
             and self._incoming_messages == other._incoming_messages
             and self._outgoing_messages == other._outgoing_messages
+            and self._ordered_message_ids == other._ordered_message_ids
```
solarw, line 374: probably redundant here, cause incoming and outgoing messages will check ids and targets that reflects order of the messages

aea/protocols/dialogue/base.py
solarw: it's general check, so if  messages in dialoge is like [1, -1] and target is 2 or -2 it's bad
but if messages are [1,-1, 2] and target is -2 this check will pass, but on get_message_by_id it will fail

solarw (approved):
LGTM, just a couple of comments

aea/protocols/dialogue/base.py
DavidMinarsch: But they don't tell us the full order.

aea/protocols/dialogue/base.py
DavidMinarsch: Disagree, we don't have total order from only incoming and outgoing messages lists.

PR #2217, opened 2021-02-02
AEA-1250 Add aggregation protocol, skill, and agent

jrriehl:
## Proposed changes

Add simple aggregation protocol, skill and agent.

## Checklist


packages/fetchai/agents/oracle_aggregator/aea-config.yaml
```diff
@@ -0,0 +1,30 @@
...
+fingerprint_ignore_patterns: []
+connections:
+- fetchai/p2p_libp2p:0.14.0
+- fetchai/soef:0.15.0
+contracts: []
+protocols:
+- fetchai/consensus:0.1.0
+- fetchai/default:0.11.0
+- fetchai/oef_search:0.12.0
+- fetchai/state_update:0.9.0
```
DavidMinarsch, line 17:
```suggestion
- fetchai/signing:0.9.0
- fetchai/state_update:0.9.0
```

DavidMinarsch: Without this protocol you won't be able to run the DM and hence AEA.

packages/fetchai/skills/oracle_aggregation/behaviours.py
```diff
@@ -0,0 +1,159 @@
...
+            float, kwargs.pop("search_interval", DEFAULT_SEARCH_INTERVAL)
+        )
+        super().__init__(tick_interval=search_interval, **kwargs)
+
+    def setup(self) -> None:
+        """Implement the setup for the behaviour."""
+        strategy = cast(GenericStrategy, self.context.strategy)
+        self._register_agent()
+        self._register_service_personality_classification()
+        strategy.make_observation()
```
DavidMinarsch, line 49: Unless it is actual setup, I would avoid doing it here. You'll probably want a separate behaviour which lets you configure this particular aspect independently of the search frequency.

packages/fetchai/skills/oracle_aggregation/handlers.py
```diff
@@ -0,0 +1,283 @@
...
+
+        strategy = cast(GenericStrategy, self.context.strategy)
+        obs = {
+            "value": obs_msg.value,
+            "time": obs_msg.time,
+            "source": obs_msg.source,
+            "signature": obs_msg.signature,
+        }
+
+        strategy.add_observation(obs_msg.sender, obs)
```
DavidMinarsch, line 126:
```suggestion
        obs = {
            "value": obs_msg.value,
            "time": obs_msg.time,
            "source": obs_msg.source,
            "signature": obs_msg.signature,
        }

        strategy.add_observation(obs_msg.sender, obs)
```
I can see it being useful having a helper method for extracting observation from message.

packages/fetchai/skills/oracle_aggregation/handlers.py
```diff
@@ -0,0 +1,283 @@
...
+        consensus_dialogues = cast(ConsensusDialogues, self.context.consensus_dialogues)
+        strategy.add_peers(oef_search_msg.agents)
+        for counterparty in strategy.peers:
+            obs_msg, _ = consensus_dialogues.create(
+                counterparty=counterparty,
+                performative=ConsensusMessage.Performative.OBSERVATION,
+                value=1,
+                time=1,
+                source="source",
+                signature="sign",
```
DavidMinarsch, line 262: I'd define a helper method somewhere which returns this. Then later you just need to update the method implementation.

packages/fetchai/skills/oracle_aggregation/skill.yaml
```diff
@@ -0,0 +1,56 @@
...
+    class_name: OefSearchDialogues
+  strategy:
+    args:
+      location:
+        latitude: 51.5194
+        longitude: 0.127
+      search_query:
+        constraint_type: ==
+        search_key: generic_service
+        search_value: generic_service
```
DavidMinarsch, line 50: It's a good idea to be more specific. sOEF is increasingly busy. This might interfere with other registered agents.

packages/fetchai/skills/oracle_aggregation/strategy.py
```diff
@@ -0,0 +1,316 @@
...
+
+    @property
+    def balance(self) -> int:
+        """Get the balance."""
+        return self._balance
+
+    @balance.setter
+    def balance(self, balance: int) -> None:
+        """Set the balance."""
+        self._balance = balance
```
DavidMinarsch, line 181: You have a lot of unused properties here as a result of copy & paste :).

jrriehl: Indeed, I got rid of a lot but missed a few. And I expect we might eventually want to bring back some of the ledger stuff, but for now it's probably better to keep things as light as possible.

packages/fetchai/protocols/consensus/README.md
```diff
@@ -0,0 +1,42 @@
...
+    source: pt:str
+    signature: pt:str
+  aggregation:
+    value: pt:int
+    time: pt:int
+    contributors: pt:list[pt:int]
+    signature: pt:str
+...
+---
+initiation: [observation, aggregation]
```
5A11, line 30: Is it just two independent messages?

packages/fetchai/protocols/consensus/README.md
```diff
@@ -0,0 +1,42 @@
...
+## Specification
+
+```yaml
+---
+name: consensus
+author: fetchai
+version: 0.1.0
+description: A protocol for agents to aggregate individual observations
+license: Apache-2.0
+aea_version: '>=0.9.0, <0.10.0'
```
5A11, line 16: I think this should now be different (include rc)

packages/fetchai/protocols/consensus/README.md
```diff
@@ -0,0 +1,42 @@
...
+speech_acts:
+  observation:
+    value: pt:int
+    time: pt:int
```
5A11, line 20: wouldn't `str` be easier to then convert to `datetime`?

packages/fetchai/protocols/consensus/README.md
```diff
@@ -0,0 +1,42 @@
...
+    source: pt:str
+    signature: pt:str
+  aggregation:
+    value: pt:int
+    time: pt:int
```
5A11, line 25: same as above

packages/fetchai/protocols/consensus/README.md
```diff
@@ -0,0 +1,42 @@
...
+    signature: pt:str
+...
+---
+initiation: [observation, aggregation]
+reply:
+  observation: []
+  aggregation: []
+termination: [observation, aggregation]
+roles: {agent}
+end_states: [successful, failed]
```
5A11, line 36: There seems to be only one state the dialogue ends in (only one message).

5A11: Actually, with the current setup, there are two outcomes, depending on which move is used (ending in observation or aggregation)

packages/fetchai/protocols/consensus/README.md
```diff
@@ -0,0 +1,42 @@
...
+    contributors: pt:list[pt:int]
+    signature: pt:str
+...
+---
+initiation: [observation, aggregation]
+reply:
+  observation: []
+  aggregation: []
+termination: [observation, aggregation]
+roles: {agent}
```
5A11, line 35: Do the sender/receiver have the same logical role? Just asking

packages/fetchai/skills/oracle_aggregation/behaviours.py
```diff
@@ -0,0 +1,157 @@
...
+
+from packages.fetchai.protocols.oef_search.message import OefSearchMessage
+from packages.fetchai.skills.oracle_aggregation.dialogues import OefSearchDialogues
+from packages.fetchai.skills.oracle_aggregation.strategy import GenericStrategy
+
+
+DEFAULT_SEARCH_INTERVAL = 5.0
+
+
+class GenericSearchBehaviour(TickerBehaviour):
```
5A11, line 34: Could this not be replaced with the generic search behaviour already implemented in other skills? Or at least subclass and only implement the diffs?

packages/fetchai/skills/oracle_aggregation/dialogues.py
```diff
@@ -0,0 +1,171 @@
...
+            return DefaultDialogue.Role.AGENT
+
+        BaseDefaultDialogues.__init__(
+            self,
+            self_address=self.context.agent_address,
+            role_from_first_message=role_from_first_message,
+        )
+
+
+class ConsensusDialogue(BaseConsensusDialogue):
```
5A11, line 83: You don't really need this as it doesn't add anything more to the `BaseConsensusDialogue`. You can just import it like the `DefaultDialogue` above.

packages/fetchai/skills/oracle_aggregation/behaviours.py
jrriehl: Thanks, this is really useful. I had thought that our approach to packages was to make them somewhat standalone, and that is why I see quite a bit of duplication across packages, but it seems to be much better practice to inherit as much base functionality as possible. I'm planning to do a bit of reorganization with this in mind.

packages/fetchai/protocols/consensus/README.md
jrriehl: Yes, I hope to add more logic here in the future, but for now the protocol is very light and these two message types can be sent independently.

packages/fetchai/protocols/consensus/README.md
jrriehl: Yes, the agents have effectively identical roles.

packages/fetchai/skills/oracle_aggregation/behaviours.py
jrriehl: In the end, I wasn't really able to inherit from another skill, the main reason being that the search here combined functionality from both the generic buyer and seller (each agent should register on the SOEF and also search for other AEAs who have registered). Also, it gets a bit messy to set the configuration for inherited skills because they use their own configuration.

packages/fetchai/skills/advanced_data_request/handlers.py
```diff
@@ -117,11 +126,12 @@ def _handle_response(self, http_msg: HttpMessage) -> None:
             json_path = output["json_path"]
 
             # find desired output data in msg_body
-            value = find(json_path, msg_body)
+            value = cast(SupportsFloat, find(json_path, msg_body))
 
             # if value is a numeric type, store it as fixed-point with number of decimals
-            if isinstance(value, (int, float)):
-                int_value = int(value * 10 ** model.decimals)
+            if is_number(value):
```
jrriehl, line 132: The `int_value` here with associated `decimals` is effectively a fixed-point representation of the value, which is kind of necessary when publishing to the oracle smart contract. However, we could consider defaulting to float here (or high precision decimal) and only do the conversion before updating the contract.

packages/fetchai/skills/simple_aggregation/handlers.py
```diff
@@ -0,0 +1,301 @@
...
+    AggregationDialogue,
+    AggregationDialogues,
+    DefaultDialogues,
+    OefSearchDialogue,
+    OefSearchDialogues,
+)
+from packages.fetchai.skills.simple_aggregation.strategy import AggregationStrategy
+
+
+def get_observation_from_message(obs_msg: AggregationMessage) -> Dict[str, Any]:
```
jrriehl, line 41: There is probably a case for defining `Observation` and `Aggregation` as custom types in the `aggregation` protocol.

packages/fetchai/protocols/aggregation/README.md
```diff
@@ -0,0 +1,42 @@
...
+    signature: pt:str
+...
+---
+initiation: [observation, aggregation]
+reply:
+  observation: []
+  aggregation: []
+termination: [observation, aggregation]
+roles: {agent}
+end_states: [successful, failed]
```
DavidMinarsch, line 37: I suppose only one end state possible (message received = successful).

packages/fetchai/protocols/aggregation/README.md
```diff
@@ -0,0 +1,42 @@
...
+name: aggregation
+author: fetchai
+version: 0.1.0
+description: A protocol for agents to aggregate individual observations
+license: Apache-2.0
+aea_version: '>=1.0.0, <2.0.0'
+protocol_specification_id: fetchai/aggregation:0.1.0
+speech_acts:
+  observation:
+    value: pt:int
```
DavidMinarsch, line 20: Would it not be better to make this more generic and offer a union of permitted types? Or have `value_int: pt:optional[int]` `value_bytes: optional[bytes]`? Otherwise this protocol is quite narrowly usable for only integer type observations.

DavidMinarsch (commented):
I see some test failures for the advanced data request skill.

DavidMinarsch (approved):
LGTM

PR #2243, opened 2021-02-11
AEA-1279 Coverage for Generic & AW skills

5A11:
This PR fill coverage gap for generic and AWX skills

## Types of changes

## Checklist


packages/fetchai/skills/generic_buyer/handlers.py
```diff
@@ -177,8 +177,10 @@ def _handle_decline(
         )
         target_message = fipa_dialogue.get_message_by_id(fipa_msg.target)
 
+        # to tell mypy target isn't none (it cannot be because this makes it
+        # invalid, therefore rejected by the dialogue when trying to add it)
```
DavidMinarsch, line 181: Please remove; that will confuse people more than help imo.

packages/fetchai/skills/tac_negotiation/handlers.py
```diff
@@ -235,8 +235,10 @@ def _on_decline(self, decline: FipaMessage, fipa_dialogue: FipaDialogue) -> None
 
         target_message = fipa_dialogue.get_message_by_id(decline.target)
 
+        # to tell mypy target isn't none (it cannot be because this makes it
+        # invalid, therefore rejected by the dialogue when trying to add it)
```
DavidMinarsch, line 239: This is ott; we don't explain mypy in our skills.

packages/fetchai/skills/generic_buyer/handlers.py
5A11: I'll remove these. 

I was very frustrated with this stupid check that is totally unnecessary because we went to all the trouble to design it in a way that all dialogue specific checks are on one place, and then you have to add this stupid check in your code and mess it up just because mypy can't understand the value cannot be none >:(

5A11: I'll remove these

DavidMinarsch: But the method doesn't know about the id you're passing to it. For all the method knows it is accepting an int and trying to find a message for it.

DavidMinarsch (approved):
LGTM

Issue #1830, opened 2020-10-16
Fix usage of setup/teardown in test classes


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
Currently, some test classes misuse `setup_class` and `teardown_class` and intend it to be test method specific setup and teardown.

**Describe the solution you'd like**
Consistently use `setup_class` for class wide setup and `setup` for test method specific setup. Same for `teardown*`

DavidMinarsch:
In case of using rerun functionality, it needs to be ensured that the setup / teardown is done on an instance not class basis.

PR #2248, opened 2021-02-11
Post-release fixes

DavidMinarsch:
## Proposed changes

This PR fixes some issues detected during release.

## Fixes

- #1830 

## Types of changes

## Checklist


tests/test_cli/test_freeze.py
```diff
@@ -45,39 +45,35 @@
...
         assert self.result.exit_code == 0
-
-    @pytest.mark.flaky(reruns=MAX_FLAKY_RERUNS)
-    def test_correct_output(self):
         """Test that the command has printed the correct output."""
         assert self.result.output == """protobuf\nvyper==0.1.0b12\n"""
 
-    @classmethod
-    def teardown_class(cls):
+    def teardown_class(self):
```
DavidMinarsch, line 73: fix

marcofavorito: should be called `teardown` thn

aea/test_tools/test_cases.py
```diff
@@ -933,6 +933,49 @@ def teardown_class(cls) -> None:
...
+            func
+            for func in dir(cls)
+            if callable(getattr(cls, func))
+            and not func.startswith("__")
+            and func.startswith("test_")
+        ]
+        if len(method_list) > 1:
+            raise ValueError(f"{cls.__name__} can only contain one test method!")
+        super(AEATestCaseEmptyFlaky, cls).setup_class()
+        cls.failure_count = 0
```
marcofavorito, line 960: this could be a boolean, as we don't care how many times a test failed (in fact, we only compare if `==0`).

aea/test_tools/test_cases.py
```diff
@@ -947,6 +990,47 @@ def teardown_class(cls) -> None:
...
+    def teardown(self) -> None:
+        """Tear down test case instance."""
+        self.failure_count += 1
+
+    @classmethod
+    def teardown_class(cls) -> None:
+        """Teardown the test class."""
+        super(AEATestCaseManyFlaky, cls).teardown_class()
+
+
```
marcofavorito, line 1033: this looks like redundant code. It could be solved by using a Mixin approach?

marcofavorito: I'm not that upset though. We have not violated the rule of three ;-)
https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)

DavidMinarsch: I was thinking to use a mixin but then there can be MRO issues when doing the super calls?

aea/test_tools/test_cases.py
DavidMinarsch: You're right. I'll keep it unless other issues need changing.

aea/test_tools/test_cases.py
marcofavorito: you can call specific super-methods by prepending the name of the class:

```
MixIn.setup_class(cls)
AEATestCaseMany.setup_class(cls)
```

MRO issues could happen only at definition-time of the class, not at runtime.

marcofavorito (approved):
LGTM

aea/test_tools/test_cases.py
DavidMinarsch: Yes but this is one more thing the users of this class need to get right.

marcofavorito: The class that inherits from `AEATestCaseManyFlaky`, in case needs to override `setup_class`, can use 
```
super(AEATestCaseManyFlaky, cls).setup_class()
```

marcofavorito: anyway, as I said, this is good enough.

aea/test_tools/test_cases.py
```diff
@@ -933,6 +941,31 @@ def teardown_class(cls) -> None:
         cls.agent_name = ""
 
 
+class AEATestCaseEmptyFlaky(AEATestCaseEmpty):
```
DavidMinarsch, line 944: Strictly speaking, one could also just use `AEATestCaseEmpty`. But the setup and teardown of class is called for each run. So it's better not to have multiple test methods to avoid unexpected behaviours.

tests/test_cli/test_upgrade.py
```diff
@@ -1185,7 +1185,12 @@ def test_upgrade(self, mock_confirm, confirm):
             self.current_agent_context, self.EXPECTED, ignore=ignore
         )
         _left_only, _right_only, diff = dircmp_recursive(dircmp)
-        assert diff == _left_only == _right_only == set()
+        if confirm:
+            assert diff == _right_only == _left_only == set()
+        else:
+            assert diff == _right_only == set()
+            # temporary: due to change in deps
+            assert _left_only == {"vendor/fetchai/skills/error"}
```
DavidMinarsch, line 1193: @marcofavorito  a hack, not sure you have a better idea.

tests/test_docs/test_skill_guide/test_skill_guide.py
```diff
@@ -62,13 +62,11 @@ def setup_class(cls):
...
         cls.code_blocks = extract_code_blocks(filepath=cls.doc_path, filter_="python")
 
-    def test_read_md_file(self):
-        """Teat that the md file is not empty."""
-        assert self.code_blocks != [], "File must not be empty."
-
     @pytest.mark.flaky(reruns=MAX_FLAKY_RERUNS_INTEGRATION)
     def test_update_skill_and_run(self):
         """Test that the resource folder contains scaffold handlers.py module."""
+        assert self.code_blocks != [], "File must not be empty."
```
DavidMinarsch, line 68: It makes sense to call it here as the method further in relies on the code_blocks being non-empty.

tests/test_test_tools/test_test_cases.py
```diff
@@ -369,3 +374,31 @@ def test_invoke(self):
...
+    @pytest.mark.flaky(reruns=1)
+    def test_fail_on_first_run(self):
+        """Test failure on first run leads to second run."""
+        file = os.path.join(self.t, "test_file")
+        if self.run_count == 1:
+            open(file, "a").close()
+            raise AssertionError("Expected error to trigger rerun!")
+        assert self.run_count == 2, "Should only be rerun once!"
+        assert not os.path.isfile(file), "File should not exist"
+
```
DavidMinarsch, line 391: This shows that both the setup_class and teardown_class are called on each re-run.

tests/test_cli/test_upgrade.py
marcofavorito: yeah it's OK

Issue #2189, opened 2021-01-23
CI script to check dependencies in setup.py correct


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
We don't have a script to check that the dependencies mentioned in setup.py are correctly specified.

**Describe the solution you'd like**
The script should check that all dependencies are listed and in the correct section.

PR #2249, opened 2021-02-12
[AEA-1286] setup.py dependencies check script

solarw:
## Proposed changes

setup.py dependencies check script

## Fixes

https://github.com/fetchai/agents-aea/issues/2189

## Types of changes

## Checklist


setup.py
```diff
@@ -44,10 +46,11 @@ def get_all_extras() -> Dict:
         "click",
         "pyyaml>=4.2b1",
         "jsonschema>=3.0.0",
+        "packaging>=20.3",
         *crypto_deps,
     ]
 
-    cli_gui = ["flask", "connexion[swagger-ui]>=2.4.0", "docker", *cli_deps]
```
DavidMinarsch, line 50: Are we sure docker is not needed @marcofavorito ?

DavidMinarsch: (I guess it was a requirement for oef node)

setup.py
```diff
@@ -76,10 +81,11 @@ def get_all_extras() -> Dict:
     "pyyaml>=4.2b1",
     "requests>=2.22.0",
     "python-dotenv>=0.14.0",
+    "urllib3==1.25.8",
```
DavidMinarsch, line 84: Can we do `>=`?

setup.py
marcofavorito: yes not needed for now.

marcofavorito: (In the future, we may want to support the Ganache Docker image in the test tools.)

setup.py
marcofavorito: I think so. Also, we only use `urllib3` for parsing URLs.

image

DavidMinarsch: can we use urllib instead? https://docs.python.org/3/library/urllib.parse.html

setup.py
```diff
@@ -34,8 +34,10 @@ def get_all_extras() -> Dict:
 
     ethereum_ledger_deps = [
         "web3==5.12.0",
-        "ipfshttpclient==0.6.1",
```
DavidMinarsch, line 37: we need to keep this here; it can otherwise lead to issues with our scripts

solarw: I thought that scripts mostly related to development

DavidMinarsch: Please revert (see slack)

DavidMinarsch (approved):
LGTM

scripts/check_imports_and_dependencies.py
```diff
@@ -0,0 +1,271 @@
...
+    @classmethod
+    def run(cls) -> None:
+        """Run dependency check."""
+        print("Dependencies check report:")
+        print("==========================")
+        files_and_modules = ImportsTool.list_all_pyfiles_with_3rdpart_imports(
+            AEA_ROOT_DIR, pattern="aea/**/*.py"
+        )
+
+        sections_dependencies = cls.get_section_dependecnies_from_setup()
```
marcofavorito, line 184: fix typo

```
get_section_dependecnies_from_setup
```

```
get_section_dependencies_from_setup
```

scripts/check_imports_and_dependencies.py
```diff
@@ -0,0 +1,271 @@
...
+        return grouped["IMPORT_NAME"]
+
+    @staticmethod
+    def get_module_file(module_name: str) -> str:
+        """Get module source file name."""
+        try:
+            mod = importlib.import_module(module_name)
+            return getattr(mod, "__file__", "")
+        except (AttributeError, ModuleNotFoundError):
+            return ""
```
marcofavorito, line 102: `__file__` is optional, so this might fail in some cases, right?
https://docs.python.org/3/reference/import.html#__file__

marcofavorito: maybe not, since `load_module` sets `__file__`:

https://docs.python.org/3/library/importlib.html#importlib.abc.Loader.load_module

scripts/check_imports_and_dependencies.py
solarw: fixed

scripts/check_imports_and_dependencies.py
solarw: anyway I need __file__ to find out what is the package.

scripts/check_imports_and_dependencies.py
```diff
@@ -0,0 +1,271 @@
...
+
+    @staticmethod
+    def get_package_files(package_name: str) -> List[str]:
+        """Get package files list."""
+        packages_info = list(search_packages_info([package_name]))
+        if len(packages_info) == 0:
+            raise Exception(f"package {package_name} not found")
+        files = packages_info[0]["files"]
+        location = packages_info[0]["location"]
+        return [str(Path(location) / i) for i in files]
```
marcofavorito, line 67: So basically we solved the problem `pypi-dependency-name != package-name` through `search_packages_info`

marcofavorito (approved):
LGTM

scripts/check_imports_and_dependencies.py
marcofavorito: yes; but we could find it by looking into `site-packages/python3.X` in case `__file__` is not present. Anyway, what I said is overkill, this should always work.

PR #2250, opened 2021-02-13
AEA-1299 Static type hints improved in scripts and benchmark folders

DavidMinarsch:
## Proposed changes

Minor typing fixes and enabling stronger checks.

## Fixes

-

## Types of changes

## Checklist

## Further comments

-


deploy-image/.aea/cli_config.yaml
```diff
@@ -1,2 +0,0 @@
-auth_token: <your-auth-token>
```
DavidMinarsch, line 1: Is it needed @marcofavorito ?

marcofavorito: I can see the `Dockerfile` copies it into the image, so at least it is used. However, I also think we might not need it, as the CLI tool will ask the user to authenticate if it can't find the configuration file.

Issue #2178, opened 2021-01-21
Investigate IPC issues in p2p lib with multiple agents


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
Supply chain project has issues with IPC

**Describe the solution you'd like**
Check if there is any clashes possible.

PR #2257, opened 2021-02-16
AEA-1302 Add data dir usage to p2p_libp2p and stub connections

DavidMinarsch:
## Proposed changes

This PR fixes some of the connections to use the correct directory for data.

## Fixes

- #2178 

## Types of changes

## Checklist

## Further comments

-


packages/fetchai/connections/p2p_libp2p/connection.py
```diff
@@ -202,11 +210,12 @@ def __init__(
 
         # log file
         self.log_file = log_file if log_file is not None else LIBP2P_NODE_LOG_FILE
-        self.log_file = os.path.join(os.path.abspath(os.getcwd()), self.log_file)
-
+        if not Path(self.log_file).is_absolute():
+            self.log_file = os.path.abspath(os.path.join(data_dir, self.log_file))
```
marcofavorito, line 214: @DavidMinarsch I noticed this:

```python
import os
p1 = "/some/abs/path
p2 = "/another/abs/path"
p3 = os.path.join(p1, p2)  # "/another/abs/path"
```
That is, concatenation of two absolute paths gives the latter one.

We could update this and other parts where we do a similar operation; would simplify the code (at the cost of worse readability!)

DavidMinarsch: I fixed it now I think (data_dir is absolute, self.log_file is not as per check)

marcofavorito: `data_dir` can be also a relative path wrt working directory. The concatenation would give the latter anyway if the latter is absolute

DavidMinarsch: The latter cannot be absolute as per check!

solarw (approved):
LGTM

aea/multiplexer.py
```diff
@@ -567,7 +567,11 @@ async def _send(self, envelope: Envelope) -> None:
         else:
             connection = self._id_to_connection[connection_id]
 
-        connection = cast(Connection, connection)
+        if connection is None:
```
DavidMinarsch, line 570: This was forgotten in the previous PR which removed the stub connection (and therefore makes it possible to have no connection specified)

Issue #2296, opened 2021-02-23
MAM: the agent re-created successfully from save but failed to start because component is outdated


panasevychol:
**Describe the bug**
MAM: the agent re-created successfully from save but failed to start because component is outdated

**Expected behavior**
If agent and/or it's components are outdated (doesn't support current AEA version) it should fail on building it.

**Screenshots**
![image](https://user-images.githubusercontent.com/6476789/108861169-9b6e4880-75f7-11eb-9a08-fa2111fc5873.png)

DavidMinarsch:
Building isn't the right term here. Reconstructing Alias from save file is what you mean.

DavidMinarsch:
In add_project under `Project.load` do this check. If it's not compatible, move the configuration to "deprecated_file".

Issue #2295, opened 2021-02-23
MAM: remove project and it's files when the last agent of that project deleted


panasevychol:
**Is your feature request related to a problem? Please describe.**
Currently if you remove all agents and restart the MAM the project files stays there. It shuld be removed so user can add it again.

**Describe the solution you'd like**
- remove the project without agents
- remove project files so it can be fetched again

**Additional context**
![image](https://user-images.githubusercontent.com/6476789/108860829-416d8300-75f7-11eb-9d37-7a251b91e246.png)

DavidMinarsch:
This functionality should be enabled via a flag on MAM initialisation and by default be disabled.

PR #2308, opened 2021-02-25
[AEA-1334] fixes and features for MAM: auto remove project, project load check.

solarw:
## Proposed changes

fixes and features for MAM: auto remove project, project load check.

## Fixes

https://github.com/fetchai/agents-aea/issues/2296
https://github.com/fetchai/agents-aea/issues/2295

## Types of changes

## Checklist


aea/manager/manager.py
```diff
@@ -249,20 +272,23 @@ def start_manager(
 
         :param local: whether or not to fetch from local registry.
         :param remote: whether or not to fetch from remote registry.
+
         :return: the MultiAgentManager instance.
         """
         if self._is_running:
-            return self
```
DavidMinarsch, line 255: That breaks compatibility. Why are we doing this? We could store this info in a property so user can request it.

aea/manager/project.py
```diff
@@ -126,6 +129,22 @@ def builder(self) -> AEABuilder:
...
+            default_ledger = builder.get_default_ledger()
+            create_private_key(default_ledger, key_file)
+            if not builder.private_key_paths:
+                # no keys, but create one for builder only
+                builder.add_private_key(default_ledger, key_file)
+            if not builder.connection_private_key_paths:
+                builder.add_private_key(
+                    default_ledger, key_file, is_connection=True,
+                )
+            builder.build()
```
DavidMinarsch, line 146: I think we should just do a version check. Not sure why we do a whole build..

solarw: the whole build is quite better, it checks many things, even python code load

DavidMinarsch: How about just this:
```
self.install_pypi_dependencies()
self.build()
```

aea/manager/project.py
```diff
@@ -126,6 +129,22 @@ def builder(self) -> AEABuilder:
         """Get builder instance."""
         return self._get_builder(self._get_agent_config(self.path), self.path)
 
+    def check(self) -> None:
+        """Check project is ok with builder.build."""
```
DavidMinarsch, line 133:
```suggestion
    def check(self) -> None:
        """Check we can still construct an AEA from the project with builder.build."""
```

aea/manager/manager.py
```diff
@@ -352,6 +379,14 @@ def add_project(
...
+            project.check()
+        except Exception as e:
+            project.remove()
+            raise ProjectCheckError(
+                f"Failed to load project: {public_id} Error: {str(e)}", e
+            )
+
         if not restore:
             project.install_pypi_dependencies()
             project.build()
```
DavidMinarsch, line 392: change order

DavidMinarsch (approved):
LGTM

PR #2286, opened 2021-02-22
Rough fix for beacon world and CosmWasm 0.10.1

MissingNO57:
These changes are necessary to make AEA working with FetchD v0.5.0 and BeaconWorld. 
agent-land is using an older version of FetchD (CosmWasm v0.9.1) - We would need to update FetchD on agent-land or have support for both versions.
NOTE! - All CosmWasm contracts need to be updated to work on CosmWasm v0.10.1

## Proposed changes

Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request.

## Fixes

If it fixes a bug or resolves a feature request, be sure to link to that issue.

## Types of changes

What types of changes does your code introduce to agents-aea?
_Put an `x` in the boxes that apply_

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

_Put an `x` in the boxes that apply._

- [ ] I have read the [CONTRIBUTING](../CONTRIBUTING.md) doc
- [ ] I am making a pull request against the `develop` branch (left side). Also you should start your branch off our `develop`.
- [ ] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have checked that code coverage does not decrease.
- [ ] I have checked that the documentation about the `aea cli` tool works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

## Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

DELETE INCLUSIVE THIS AND BELOW FOR STANDARD PR
------

## Release summary

Version number: [e.g. 1.0.1]

## Release details

Describe in short the main changes with the new release.

## Checklist

_Put an `x` in the boxes that apply._

- [ ] I have read the [CONTRIBUTING](../CONTRIBUTING.md) doc
- [ ] I am making a pull request against the `master` branch (left side), from `develop`
- [ ] Lint and unit tests pass locally
- [ ] I have checked the fingerprint hashes are correct by running (`scripts/generate_ipfs_hashes.py`)
- [ ] I have regenerated the latest API docs
- [ ] I built the documentation and updated it with the latest changes
- [ ] I have added an item in `HISTORY.md` for this release
- [ ] I bumped the version number in the `aea/__version__.py` file.
[...]

PR #2314, opened 2021-03-01
AEA-1349 Add support for new cosmwasm message format in crypto plugins

DavidMinarsch:
## Proposed changes

This PR introduces support for multiple message formats to the fetchai and cosmos crypto plugins.

It refactors the work proposed in #2286.

## Fixes

-

## Types of changes

## Checklist

## Further comments

-


plugins/aea-crypto-cosmos/aea_crypto_cosmos/cosmos.py
```diff
@@ -699,13 +723,17 @@ def get_handle_transaction(
         )
         if account_number is None or sequence is None:
             return None
+        if self.cosmwasm_versions != DEFAULT_COSMWASM_VERSIONS and amount == 0:
```
MissingNO57, line 726: I think it passing an empty list should work also for an older version of CosmWasm. But we would have to test it.

plugins/aea-crypto-cosmos/aea_crypto_cosmos/cosmos.py
```diff
@@ -62,6 +62,15 @@
 DEFAULT_CHAIN_ID = "cosmoshub-3"
 _BYTECODE = "wasm_byte_code"
 DEFAULT_CLI_COMMAND = "wasmcli"
+DEFAULT_COSMWASM_VERSIONS = "<0.10"
+MESSAGE_FORMAT_BY_VERSION = {
+    "store": {"<0.10": "wasm/store-code", ">=0.10": "wasm/MsgStoreCode"},
```
MissingNO57, line 67: This might be a little bit tricky because FetchD reports different versions than CosmWasm. For example, FetchD v0.5.0 is using CosmWasm 0.10.x. We just have to keep it in mind.

MissingNO57 (commented):
LGTM! I just have a few comments but I guess we will just leave it as it is.

plugins/aea-crypto-cosmos/aea_crypto_cosmos/cosmos.py
DavidMinarsch: Yes, this is purely on the Python side though, so no need for them to match strictly.

DavidMinarsch: Adding full version checks is a bit overkill here imo. Means another dependency to parse version etc.

PR #2315, opened 2021-03-01
AEA-1345 Update routing and envelope context

DavidMinarsch:
## Proposed changes

This PR:
- adds check for redundancy in routing declaration - > raise Exception
- auto set routing helper info when receiving envelopes multiple agent to agent connections are supported
- miscellaneous fixes
- adds slots in many places where it is safe to do so
- what to do about URI in envelope context? We keep it as is in the envelope context for now. But it is not exposed to the skill on purpose (transport only info!)

## Fixes

## Types of changes

## Checklist

## Further comments

-


aea/aea.py
```diff
@@ -297,11 +297,11 @@ def _handle_decoding(
             return None, []  # Tuple[Optional[Message], List[Handler]]
 
         handlers = self.filter.get_active_handlers(
-            protocol.public_id, envelope.to_as_public_id or envelope.skill_id
+            protocol.public_id, envelope.to_as_public_id
```
DavidMinarsch, line 300: For component to component we don't need the envelope context. For agent to agent, the other agent should not be able to target a specific skill.

aea/mail/base.py
```diff
@@ -54,171 +55,120 @@ def __init__(self, uri_raw: str) -> None:
         :param uri_raw: the raw form uri
         :raises ValueError: if uri_raw is not RFC3986 compliant
         """
-        self.uri_raw = uri_raw
```
DavidMinarsch, line 57: Storage vs compute.

aea/mail/base.py
```diff
@@ -54,171 +55,120 @@ def __init__(self, uri_raw: str) -> None:
...
 class EnvelopeContext:
-    """Extra information for the handling of an envelope."""
+    """Contains context information of an envelope."""
+
+    __slots__ = ("_connection_id", "_uri")
 
     def __init__(
-        self,
-        connection_id: Optional[PublicId] = None,
-        skill_id: Optional[PublicId] = None,
```
DavidMinarsch, line 147: Skill id removed, see above.

aea/mail/base.py
```diff
@@ -368,7 +320,12 @@ def __init__(
 
         self._protocol_specification_id: PublicId = protocol_specification_id
         self._message = message
-        self._context = context if context is not None else EnvelopeContext()
+        if self.is_component_to_component_message:
+            enforce(
+                context is None,
+                "EnvelopeContext must be None for component to component messages.",
```
DavidMinarsch, line 326: Component to component messages specify the target and source explicitly. Hence, there is no need for a context.

aea/multiplexer.py
```diff
@@ -506,12 +506,19 @@ async def _receiving_loop(self) -> None:
...
+                    connection = task_to_connection.pop(task)
                     envelope = task.result()
                     if envelope is not None:
+                        if not envelope.is_component_to_component_message:
+                            if envelope.context is None:
+                                envelope.context = EnvelopeContext(
+                                    connection_id=connection.public_id
+                                )
+                            else:
+                                envelope.context.connection_id = connection.public_id
```
DavidMinarsch, line 518: The framework sets the connection id so the skill knows where the envelope comes `via` in agent to agent communication.

aea/protocols/base.py
```diff
@@ -48,7 +49,7 @@ class Message:
     protocol_specification_id = None  # type: PublicId
     serializer = None  # type: Type["Serializer"]
 
-    __slots__ = ("_slots", "_to", "_sender")
+    __slots__ = ("_slots", "_to", "_sender", "_envelope_context")
```
DavidMinarsch, line 52: We need to expose the context on the message, otherwise the skill is unaware which route a message took.

aea/protocols/dialogue/base.py
```diff
@@ -716,6 +721,8 @@ def reply(
 
         self._update(reply)
 
+        reply.envelope_context = target_message.envelope_context
```
DavidMinarsch, line 724: Alternatively, we construct a new envelope context. But this is cheaper.

aea/protocols/base.py
solarw: not sure it's a good idea, it's obvious that `Handler.handle` accepts Message only instances.
but EnvelopeContext is out of the Messages/Protocols scope.

suppose would be nice to make Handler.handle to be able to accept some extra parameters with message related details. This will brake current api, but! it's possible to make it smart, and pass extra details if  handlers support it

aea/protocols/dialogue/base.py
solarw: I think it's ok to share it, most of the time it should stay immutable and used in reply as was provided.

aea/aea.py
```diff
@@ -297,11 +297,11 @@ def _handle_decoding(
...
         )
 
         if len(handlers) == 0:
-            handler.send_unsupported_skill(envelope, self.logger)
+            handler.send_unsupported_handler(envelope, self.logger)
```
solarw, line 304: probably unsupported protocol id/target?
what about checking target explicitly to provide an error like target skill does not present?
in case it's a check by the protocol id, we can inform that the protocol id is not supported by the agent.


also, not sure, do we have an agent2agent check for the target address match with current agent address?

aea/aea.py
```diff
@@ -338,6 +338,9 @@ def handle_envelope(self, envelope: Envelope) -> None:
         if msg is None:
             return
 
+        if envelope.context is not None:
```
solarw, line 341: I got the idea but don't like the way of the extending message class.

aea/mail/base.py
```diff
@@ -54,171 +55,120 @@ def __init__(self, uri_raw: str) -> None:
...
         """Get the host."""
-        return self._host
+        parsed = urlparse(self._uri_raw)
+        return parsed.hostname
 
     @property
     def port(self) -> Optional[int]:
         """Get the port."""
-        return self._port
+        parsed = urlparse(self._uri_raw)
```
solarw, line 117:
```
@property
def _parsed():
    return urlparse(self._uri_raw)
 
@property
def port(self) -> Optional[int]:
    return self._parsed.port
```

aea/mail/base.py
```diff
@@ -54,171 +55,120 @@ def __init__(self, uri_raw: str) -> None:
...
-            and self.port == other.port
-        )
+        return isinstance(other, URI) and str(self) == str(other)
 
 
 class EnvelopeContext:
-    """Extra information for the handling of an envelope."""
+    """Contains context information of an envelope."""
+
+    __slots__ = ("_connection_id", "_uri")
```
solarw, line 132: is it possible that uri will contain connection_id details?

solarw (commented):
Nice work, Slots are appreciated.
Not sure it's a good idea to extend base Message class with EnvelopContext property.

can be done in several ways:
* update all skills handlers to support extra data for handle method
* aea Handler.handle can do the smart check for arguments and dynamically set arguments required. not obvious but will help to stay skills code base the same and extend where needed.
* wrap a message in MessageProxy wrapper, so just provide extra property but mocks the message. so it acts like a message comlpetely, but has extra property outside of Message/Protocols

aea/protocols/base.py
DavidMinarsch: I thought about it. But it doesn't really help. Because then you have to in the skill write a lot of custom code to pass the envelope context around. The envelope context is really just a general message context. So we can simply rename it and use it here.

aea/configurations/data_types.py
```diff
@@ -223,6 +223,8 @@ class PublicId(JSONSerializable):
     True
     """
 
+    __slots__ = ("_author", "_name", "_package_version")
```
DavidMinarsch, line 226: Adding slots in heavily used framework-only classes is sensible. We only need to be careful with classes which we want the user to be able to extend.

aea/mail/base.py
```diff
@@ -42,187 +41,6 @@ class Empty(Exception):
     """Exception for when the inbox is empty."""
 
 
-class URI:
```
DavidMinarsch, line 45: Moved to avoid cyclic dependencies

aea/protocols/base.py
DavidMinarsch: Maybe we can also simplify the envelope context to be just a string? That would be less overhead.

aea/aea.py
DavidMinarsch: the address check happens automatically inside the skills' dialogues.

aea/aea.py
DavidMinarsch: Yeah there is a downside to it. However, any alternative approach ends up becoming quite verbose...

aea/aea.py
marcofavorito: +1 on `send_unsupported_protocol`.

aea/mail/common.py
```diff
@@ -0,0 +1,160 @@
+# -*- coding: utf-8 -*-
```
marcofavorito, line 1: just moved from the above? Or there are changes?

aea/multiplexer.py
```diff
@@ -699,6 +702,21 @@ def _setup(
         for c in connections:
             self.add_connection(c, c.public_id == default_connection)
 
+    @staticmethod
+    def _set_context_on_envelope(envelope: Envelope, connection: Connection) -> None:
+        """
+        Updat the envelope context.
```
marcofavorito, line 708: typo

aea/multiplexer.py
```diff
@@ -699,6 +702,21 @@ def _setup(
...
+
```
marcofavorito, line 709: also, if you don't mind, would be great to have the explanation in the docstring of the three cases below.

aea/protocols/base.py
marcofavorito: Another idea is to have a `MessageHeader`/`MessageContext` class, with all the attributes we want to expose (and nothing else!) to the user code. It also helps to avoid conflating two concepts (Envelope != Message), as @solarw pointed out. And could serve in the future if we want to put more information only in the message header and not in the envelope context.

marcofavorito (approved):
LGTM

aea/configurations/base.py
```diff
@@ -1594,6 +1675,13 @@ class ContractConfig(ComponentConfiguration):
 
     FIELDS_ALLOWED_TO_UPDATE: FrozenSet[str] = frozenset(["build_directory"])
 
+    __slots__ = (
+        "dependencies",
+        "description",
```
DavidMinarsch, line 1680: Can be extracted to base

aea/skills/base.py
```diff
@@ -767,7 +769,6 @@ def __init__(
         if kwargs is not None:
             pass
         super().__init__(configuration)
-        self.config = configuration
```
DavidMinarsch, line 770: didn't find the use of that ...

aea/connections/base.py
```diff
@@ -122,12 +122,8 @@ def _ensure_valid_envelope_for_external_comms(envelope: "Envelope") -> None:
         :param envelope: the envelope
         """
         enforce(
-            not envelope.is_sender_public_id,
-            f"Sender field of envelope is public id, needs to be address. Found={envelope.sender}",
-        )
-        enforce(
```
DavidMinarsch, line 128: Second one would not be reachable. So merged.

tests/test_aea.py
```diff
@@ -522,18 +523,20 @@ def test_no_handlers_registered():
...
             to=aea.identity.address, sender=aea.identity.address, message=msg,
         )
-        with patch.object(
-            aea.filter, "get_active_handlers", return_value=[]
-        ), patch.object(
-            aea.runtime.multiplexer, "put",
+        with patch(
+            "aea.registries.filter.Filter.get_active_handlers",
+            new_callable=PropertyMock,
         ):
```
DavidMinarsch, line 533: Patching of objects with slots is not straightforward

tests/test_connections/test_base.py
```diff
@@ -111,27 +111,15 @@ def test_ensure_valid_envelope_for_external_comms_negative_cases():
     """Test the staticmethod '_ensure_valid_envelope_for_external_comms', negative cases."""
     protocol_specification_id = PublicId("author", "name", "0.1.0")
     wrong_sender = wrong_to = "author/name:0.1.0"
-    envelope_wrong_sender = Envelope(
```
DavidMinarsch, line 114: Envelope only allows for either or, not a mix

aea/mail/base.py
DavidMinarsch: No; we don't allow URI to "target" anything in the AEA. This would be an obvious attack vector.

aea/protocols/base.py
DavidMinarsch: Resolved now on the multiplexer side to avoid message memory bloat and for KISS reasons.

aea/aea.py
DavidMinarsch: the correct cause is `no_active_handler`, updated now

aea/aea.py
DavidMinarsch: removed

aea/mail/base.py
DavidMinarsch: Just more work for me with no benefit?

aea/mail/common.py
DavidMinarsch: reverted

aea/multiplexer.py
DavidMinarsch: fixed

aea/multiplexer.py
```diff
@@ -586,6 +563,67 @@ async def _send(self, envelope: Envelope) -> None:
...
+        """
+        # component to component messages are routed by their component id
+        if envelope.is_component_to_component_message:
+            connection_id = envelope.to_as_public_id
+            self.logger.debug(
+                "Using envelope `to` field as connection_id: {}".format(connection_id)
+            )
+            enforce(
+                connection_id is not None, "Cannot be None by envelope construction"
+            )
```
marcofavorito, line 592:
```suggestion
            enforce(
                connection_id is not None, "Connection id cannot be None by envelope construction"
            )
```

docs/message-routing.md
```diff
@@ -24,12 +24,14 @@ It is important to keep in mind that <a href="../interaction-protocol">interacti
...
-- `Multiplexer` assigns messages to relevant `Connection` based on four rules:
-
-	1. checks if `to` field can be interpreted as `connection_id`, if so uses that else
-	2. checks if `EnvelopeContext` exists and specifies a `Connection`, if so uses that else
-	3. checks if default routing is specified for the `protocol_id` referenced in the `Envelope`, if so uses that else
-	4. sends to default `Connection`.
+- `Multiplexer` assigns messages to relevant `Connection` based on the following rules:
+
+	1. Component to component messages are routed by their `component_id`
+	2. Agent to agent messages, are routed following four rules:
```
marcofavorito, line 30:
```suggestion
	2. Agent to agent messages are routed following four rules:
```

marcofavorito (approved):
LGTM

aea/configurations/base.py
```diff
@@ -205,6 +207,19 @@ class PackageConfiguration(Configuration, ABC):
     - contracts
     """
 
+    __slots__ = (
```
solarw, line 210: why slots here?
I don't think it will save much memory, cause configuration instances amount is quite low during the agent life cycle in comparison to dialogues and messages

aea/mail/base.py
solarw: just don't repeat yourself

aea/configurations/base.py
DavidMinarsch: Yes, but there is no cost to adding it and a small benefit.

PR #2319, opened 2021-03-02
AEA-1346 Add 'file_path' to skill component configuration, and change loading policy

### LINKED ISSUE #2198: Splitting Handlers/Behaviours in different files.
I would like to be able to split the handlers.py, behaviours.py in different files. It will help to keep agents organised and also keep the lines of code inside the handlers.py to minimum.

marcofavorito:
## Proposed changes

Two changes:
- Add 'file_path' to skill component configuration;
- Change skill component loading policy

## Fixes

#2198

## Types of changes

## Checklist

## Further comments

n/a


aea/skills/base.py
```diff
@@ -427,70 +427,7 @@ def parse_module(  # pylint: disable=arguments-differ
...
-                    )
-                except Exception as e:  # pylint: disable=broad-except # pragma: nocover
-                    e_str = parse_exception(e)
-                    raise AEAInstantiationException(
-                        f"An error occured during instantiation of behaviour {skill_context.skill_id}/{behaviour_config.class_name}:\n{e_str}"
-                    )
-                behaviours[behaviour_id] = behaviour
-
-        return behaviours
+        return _parse_module(path, behaviour_configs, skill_context, "behaviour")
```
DavidMinarsch, line 430: What about the "models" case? can we use the same method (with modifications if necessary) please?

aea/skills/base.py
```diff
@@ -897,6 +779,74 @@ def from_config(
         return skill
 
 
+def _parse_module(
```
DavidMinarsch, line 782: Good that this is extracted! Let's have some tests +/- too :)

aea/skills/base.py
marcofavorito: yes. The plan is to make `Model.parse_module` to behave like the others. The loading from any Python module will be handled in another helper function.

aea/skills/base.py
marcofavorito: yes

aea/skills/base.py
```diff
@@ -912,3 +765,347 @@ def _print_warning_message_for_non_declared_skill_components(
...
+                            component_config,
+                            component_class,
+                            component_type,
+                        )
+                    )
+                else:
+                    # process the configuration at the end of the loop
+                    not_resolved_configurations[
+                        (component_type, component_id)
+                    ] = component_config
```
marcofavorito, line 1061: here we process skill component configurations that have the path field specified.

aea/skills/base.py
```diff
@@ -912,3 +765,348 @@ def _print_warning_message_for_non_declared_skill_components(
...
```
marcofavorito, line 1062: here we process skill component configurations that have the file_path field specified (!=None). The others will be processed in the loop below.

aea/skills/base.py
```diff
@@ -912,3 +765,348 @@ def _print_warning_message_for_non_declared_skill_components(
...
+                        None,
+                    )
+                    enforce(
+                        component_class_or_none is not None,
+                        self._get_error_message_prefix()
+                        + f"Cannot find class '{class_name}' for component '{component_id}' of type '{self._type_to_str(component_type)}' of skill '{self.configuration.public_id}' in module {path}",
+                    )
+                    component_class = cast(
+                        Type[SkillComponent], component_class_or_none
+                    )
```
marcofavorito, line 1048: TODO, need to check if the class found is of the right type.

aea/skills/base.py
```diff
@@ -912,3 +765,348 @@ def _print_warning_message_for_non_declared_skill_components(
...
+            enforce(
+                not_used_classes != 0,
+                f"Cannot find class of skill '{self.configuration.public_id}' for component configuration '{component_id}' of type '{self._type_to_str(component_type)}'.",
+            )
+            enforce(
+                len(not_used_classes) == 1,
+                self._get_error_message_ambiguous_classes(
+                    class_name, not_used_classes, component_type, component_id
+                ),
+            )
```
marcofavorito, line 1085: given a configuration:
1. we first check if there are classes with the class name specified (1069)
2. then we check if the class is of the right type (1069)

also, this time we need to:
- filter the classes that have been used by the configurations which specified the `file_path` field (1076)
- check that there is only one way to select such class, because we cannot resolve the ambiguity by path this time (1080)

aea/skills/base.py
```diff
@@ -912,3 +765,348 @@ def _print_warning_message_for_non_declared_skill_components(
...
+            not_used_class = list(not_used_classes)[0]
+            result.append(
+                _SkillComponentLoadingItem(
+                    component_id, component_config, not_used_class, component_type
+                )
+            )
+            used_classes.add(not_used_class)
```
marcofavorito, line 1092: the non-used class now becomes used. This information can be used later to print a warning message for skill components not used (not done yet).

aea/skills/base.py
marcofavorito: done in 035cecf

aea/skills/base.py
```diff
@@ -912,3 +765,356 @@ def _print_warning_message_for_non_declared_skill_components(
...
+                            component_config,
+                            component_class,
+                            component_type,
+                        )
+                    )
+                else:
+                    # process the configuration at the end of the loop
+                    not_resolved_configurations[
+                        (component_type, component_id)
+                    ] = component_config
```
marcofavorito, line 1070: here we process skill component configurations that have the file_path field specified (!=None). The others will be processed in the loop below.

aea/skills/base.py
```diff
@@ -912,3 +765,356 @@ def _print_warning_message_for_non_declared_skill_components(
...
+                if path is not None:
+                    classes_in_path = component_classes_by_path[path]
+                    component_class_or_none: Optional[Type[SkillComponent]] = next(
+                        (
+                            actual_class
+                            for actual_class_name, actual_class in classes_in_path
+                            if actual_class_name == class_name
+                        ),
+                        None,
+                    )
```
marcofavorito, line 1040: note: cannot have more than one classes with the same name in the same path (the last defined will overwrite the previous ones)

aea/skills/base.py
```diff
@@ -912,3 +765,356 @@ def _print_warning_message_for_non_declared_skill_components(
...
+
+    def _update_skill(self, components: _ComponentsHelperIndex) -> None:
+        self.skill.handlers.update(
+            cast(Dict[str, Handler], components.get(Handler, {}))
+        )
+        self.skill.behaviours.update(
+            cast(Dict[str, Behaviour], components.get(Behaviour, {}))
+        )
+        self.skill.models.update(cast(Dict[str, Model], components.get(Model, {})))
+        self.skill._set_models_on_context()  # pylint: disable=protected-access
```
DavidMinarsch, line 846: Previously we only did the during skill init. Why is it now needed here? Conceptually it makes sense but I don't understand why it worked before.

aea/skills/base.py
```diff
@@ -912,3 +765,356 @@ def _print_warning_message_for_non_declared_skill_components(
...
+        filtered_classes = filter(
+            lambda name_and_class: any(
+                re.match(component, name_and_class[0])
+                for component in self._all_component_names
+            )
+            and issubclass(name_and_class[1], SkillComponent)
+            and not str.startswith(name_and_class[1].__module__, "aea.")
+            and not str.startswith(
+                name_and_class[1].__module__, self.skill_dotted_path
+            ),
```
DavidMinarsch, line 892: Can you explain why we exclude this?

tests/data/dummy_skill/dummy_subpackage/foo.py
```diff
@@ -18,8 +18,32 @@
...
 """This module is in a skill sub-package (for testing purposes)."""
+from aea.skills.base import Behaviour
 
 
 def bar():
     """A bar function."""
     return 42
+
+
+class DummyBehaviour(Behaviour):
```
DavidMinarsch, line 29: WE should also add a handler class here and not give it a file path, just to have a richer test case.

tests/test_skills/test_base.py
```diff
@@ -468,20 +467,6 @@ def test_model_parse_module_missing_class():
         assert "dummy_model" in models_by_id
 
 
-def test_check_duplicate_classes():
```
DavidMinarsch, line 471: Why not keep such a test?

solarw (approved):
LGTM

DavidMinarsch (approved):
LGTM

Issue #1983, opened 2020-11-23
Module loading in tests


5A11:
**Describe the bug**
When running pytests on specific tests in isolation (not running all tests such as in CI) that cover code with references to something under `packages.fetchai....`, module loading error occurs. 

Note: this might be a broader issue than tests to do with the order of modules and dependencies being loaded.

Example: Running any of the test methods under `tests/test_packages/test_skills/test_tac_negotiation/test_dialogues.py::TestDialogues`, which test methods of the `dialogue.py` module of the `tac_negotiation` skill (note this skill has code in `/packages/fetchai/skills/tac_negotiation/strategy.py` line 45 that imports from `packages.fetchai.contracts.erc1155.contract`), the following is the result:

```
test setup failed
cls = <class 'tests.test_packages.test_skills.test_tac_negotiation.test_transactions.TestTransactions'>

    @classmethod
    def setup(cls):
        """Setup the test class."""
>       super().setup()

/Users/ali/Projects/agents-aea/tests/test_packages/test_skills/test_tac_negotiation/test_transactions.py:49: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/Users/ali/Projects/agents-aea/aea/test_tools/test_skill.py:449: in setup
    cls._skill = Skill.from_dir(str(cls.path_to_skill), agent_context)
/Users/ali/Projects/agents-aea/aea/skills/base.py:770: in from_dir
    return Skill.from_config(configuration, agent_context, **kwargs)
/Users/ali/Projects/agents-aea/aea/skills/base.py:819: in from_config
    str(directory / "handlers.py"), handlers_by_id, skill_context
/Users/ali/Projects/agents-aea/aea/skills/base.py:513: in parse_module
    handler_module = load_module("handlers", Path(path))
/Users/ali/Projects/agents-aea/aea/helpers/base.py:105: in load_module
    spec.loader.exec_module(module)  # type: ignore
<frozen importlib._bootstrap_external>:728: in exec_module
    ???
<frozen importlib._bootstrap>:219: in _call_with_frames_removed
    ???
/Users/ali/Projects/agents-aea/packages/fetchai/skills/tac_negotiation/handlers.py:53: in <module>
    from packages.fetchai.skills.tac_negotiation.strategy import Strategy
[...]

PR #2342, opened 2021-03-05
AEA-1348 remove dummy packages patch from the aea loader

solarw:
## Proposed changes

remove dummy packages patch from the aea loader

## Fixes

**https://github.com/fetchai/agents-aea/issues/1983**
## Types of changes

## Checklist


aea/components/base.py
```diff
@@ -131,9 +130,6 @@ def load_aea_package(configuration: ComponentConfiguration) -> None:
     prefix_author = prefix_root + f".{configuration.author}"
     prefix_pkg_type = prefix_author + f".{configuration.component_type.to_plural()}"
     prefix_pkg = prefix_pkg_type + f".{configuration.name}"
-    sys.modules[prefix_root] = types.ModuleType(prefix_root)
-    sys.modules[prefix_author] = types.ModuleType(prefix_author)
-    sys.modules[prefix_pkg_type] = types.ModuleType(prefix_pkg_type)
```
DavidMinarsch, line 136: @marcofavorito why was this added in the first place?

marcofavorito: to allow imports for prefixes of import paths. And to improve the error message in case of wrong prefix path. (e.g. existing author but missing package)

packages/fetchai/skills/ml_train/tasks.py
```diff
@@ -71,10 +68,11 @@ def _make_model() -> Any:
     def execute(self, *args: Any, **kwargs: Any) -> Any:
```
DavidMinarsch, line 68: return type could be specified more accurately

tests/conftest.py
```diff
@@ -367,6 +363,20 @@
 ]
 
 
+@contextmanager
```
DavidMinarsch, line 366: Is this used automatically for each test?

aea/components/base.py
DavidMinarsch: Where do we need `imports for prefixes of import paths.`?

Re error message, we then need to improve the error logging there to ensure this doesn't get worse as a result of this PR.

tests/conftest.py
solarw: no, need to set explicitly for every test.
probably will apply later to specific tests.

just want to complete this pr

aea/components/base.py
solarw: done

5A11 (approved):

DavidMinarsch (approved):
LGTM

aea/skills/tasks.py
```diff
@@ -64,7 +63,7 @@ def __call__(self, *args: Any, **kwargs: Any) -> "Task":
         finally:
             self._is_executed = True
             self.teardown()
-        return self
```
DavidMinarsch, line 67: This breaks what we have in the gym skill. Why not return self? This let's you check task status etc.

tests/test_task.py
```diff
@@ -107,25 +107,8 @@ def test_task_manager_task_object(self):
         task_result = self.task_manager.get_task_result(task_id)
         assert isinstance(task_result, AsyncResult)
 
-        expected_task = task_result.get(self.WAIT_TIMEOUT)
```
DavidMinarsch, line 110: See above!

Issue #2170, opened 2021-01-19
Plug-in mechanism for new CLI commands


marcofavorito:
**Is your feature request related to a problem? Please describe.**

Analogously to #1229 and #2112 , would be useful to let the user to define custom commands for their AEA project, or make it available for anyone via a plug-in mechanism.

**Describe the solution you'd like**

Something similar to Django admin custom commands:

https://docs.djangoproject.com/en/3.1/howto/custom-management-commands/

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

marcofavorito:
We could consider using `click-plugins` package: https://github.com/click-contrib/click-plugins.

It roughly breaks down in:
- [ ] refactor some code inside aea.cli and aea.cli_gui
- [ ] add another dependency: click-plugins
- [ ] add a decorator @with_plugins(iter_entry_points('aea.plugins.cli'))
- [ ] change CI workflow as needed (install CLI plugins manually)
- [ ] update docs

If we don't want to use `click-plugins`, we need to add a module similar to `crypto/plugin.py` (see #2112) that reads the registered CLI entry points and adds them to the main CLI command.

Related to #2090, the CLI GUI could be made a plugin.

PR #2348, opened 2021-03-08
Add cli plugin prototype and example

marcofavorito:
## Proposed changes

Add a CLI plug-in mechanism prototype, and an example.

## Fixes

addresses #2170 

## Types of changes

## Checklist

## Further comments

n/a


tests/test_cli/test_plugin.py
```diff
@@ -0,0 +1,218 @@
+# -*- coding: utf-8 -*-
```
marcofavorito, line 1: adapted from click-plugins

plugins/cli_plugin_ex/setup.py
```diff
@@ -0,0 +1,33 @@
...
+"""Setup script for the plug-in."""
+
+
+from setuptools import setup
+
+
+setup(
+    name="aea-cli-example",
+    version="0.1.0",
+    packages=["cli_plugin_example"],
```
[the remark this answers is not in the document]
marcofavorito: yes, it is not strictly necessary to depend on `aea`.

aea/cli/core.py
```diff
@@ -21,6 +21,7 @@
 from typing import Optional
 
 import click
+from pkg_resources import iter_entry_points
```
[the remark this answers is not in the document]
marcofavorito: it is not a new dependency, we used it for crypto plugin. It is from `setuptools`

DavidMinarsch (approved):
LGTM

scripts/check_copyright_notice.py
```diff
@@ -38,7 +38,7 @@
 )?# -\*- coding: utf-8 -\*-
 # ------------------------------------------------------------------------------
 #
-#   (Copyright 2018-(2019|2020) Fetch.AI Limited|Copyright [0-9]{4}(-[0-9]{4})? [a-zA-Z_]+)
+#   (Copyright 2018-(2019|2020|2021) Fetch.AI Limited|Copyright [0-9]{4}(-[0-9]{4})? [a-zA-Z_]+)
```
solarw, line 41: probably automatic substitution with list of valid years?

marcofavorito: good idea

marcofavorito: @DavidMinarsch although I'm thinking that maybe we should update everything to `2018-2021`. Thoughts?

DavidMinarsch: No! That's not a good idea. The diff will be absolutely massive and there is no gain from it.

solarw (approved):
LGTM

scripts/check_copyright_notice.py
marcofavorito: apart from the diff. I am wondering about the "right" thing to do. The copyright should be extended for early Python modules/files. It clearly has no practical gain, as often when dealing with such "legal" things.

DavidMinarsch: No the copyright does not need to be extended for the files we don't touch. We discussed this last year and came to the conclusion that this is ok.

marcofavorito: I think in 2020 we have touched some of those added in 2019, but we haven't done that. Anyway, not very important as I said.

PR #2359, opened 2021-03-16
AEA-173 Feature/password

### LINKED ISSUE #221: Investigate security issues on wallet private key handling.

Check security issues re. private keys generation/handling.

None in particular.  Open for discussion.

None.

**Additional context**

Take into account the current best practices.

DavidMinarsch:
## Proposed changes

This PR adds password support for private key encryption in the AEA.

Developers/users can run AEAs with or without private key encryption. Keys are only encrypted at rest. The password needs to be supplied for all interactions which utilise the wallet. It can be fed via CLI as an argument or for MAM/Builder directly. It is assumed that the same password is used for all keys (otherwise this would be impractical).

Crypto plugins are extended and refactored (breaking!) to support encryption/decryption of private keys when reading from/writing to file.

Missing:
- [x] unit tests
- [ ] end-to-end tests for AEA usage with password
- [x] multiprocess launcher support
- [x] password confirmation (?) for key creation
- [x] cosmos crypto needs encrypt/decrypt implementation

## Fixes

#221 

## Types of changes

## Checklist

## Further comments

-


aea/cli/add_key.py
```diff
@@ -50,22 +50,26 @@
 @click.argument(
     "file", metavar="FILE", type=key_file_argument, required=False,
 )
+@click.argument(
+    "password", metavar="PASSWORD", type=str, default=None, required=False,
```
DavidMinarsch, line 54: Did I do this right @marcofavorito ?

aea/cli/generate_key.py
```diff
@@ -64,7 +70,7 @@ def _generate_private_key(type_: str, file: Optional[str] = None) -> None:
             PRIVATE_KEY_PATH_SCHEMA.format(type__) if file is None else file
         )
         if _can_write(private_key_file):
-            create_private_key(type__, private_key_file)
+            create_private_key(type__, private_key_file, password)
```
DavidMinarsch, line 73: We might want to add a password confirmation step here!

DavidMinarsch: We might also want to disable autogeneration of keys outside the MAM context. As otherwise the password confirmation cannot be done. In the MAM context, we assume the password is correctly set.

aea/cli/utils/package_utils.py
```diff
@@ -648,16 +657,16 @@ def try_get_balance(  # pylint: disable=unused-argument
...
     """
     Get wallet from current click Context.
 
     :param ctx: click context
 
     :return: wallet
     """
-    verify_or_create_private_keys_ctx(ctx=ctx)
-    wallet = get_wallet_from_agent_config(ctx.agent_config)
+    verify_or_create_private_keys_ctx(ctx=ctx, password=password)
```
DavidMinarsch, line 668: Probably stop auto creating keys...

aea/crypto/base.py
```diff
@@ -128,14 +137,60 @@ def sign_transaction(self, transaction: JSONLike) -> JSONLike:
         :return: signed transaction
         """
 
-    @abstractmethod
-    def dump(self, fp: BinaryIO) -> None:
+    @classmethod
+    def load(cls, private_key_file: str, password: Optional[str] = None) -> str:
```
DavidMinarsch, line 141: For symmetry introduced this here, also means framework is responsible for IO and not plugin.

aea/launcher.py
```diff
@@ -253,16 +270,23 @@ def __init__(
...
+        self._password = password
         super().__init__(mode=mode, fail_policy=fail_policy)
 
     def _make_tasks(self) -> Sequence[AbstractExecutorTask]:
         """Make tasks to run with executor."""
         if self._mode == "multiprocess":
             return [
-                AEADirMultiprocessTask(agent_dir, log_level=self._log_level)
+                AEADirMultiprocessTask(
+                    agent_dir, log_level=self._log_level, password=self._password
```
DavidMinarsch, line 285: Not sure this gets propagated. Need to check...

aea/manager/manager.py
```diff
@@ -198,6 +200,7 @@ def __init__(
             )
         self._started_event = threading.Event()
         self._mode = mode
+        self._password = password
```
DavidMinarsch, line 203: In mam we store it in memory for the duration of mam usage. On AEA Manager app say, we can ask for password at login.

aea/manager/project.py
```diff
@@ -209,7 +218,7 @@ def _create_private_key(
         filepath = os.path.join(self._data_dir, file_name)
         if os.path.exists(filepath) and not replace:
             return filepath
-        create_private_key(ledger, filepath)
+        create_private_key(ledger, filepath, password=self._password)
```
DavidMinarsch, line 221: Note here, password confirmation step is too onerous. We would need to assume the supplied password is sufficient.

aea/cli/add_key.py
marcofavorito: I think that passing the password as an argument in plaintext is a bad practice. As a temporary (but *very* temporary) solution, it is OK. 

Apart from that. looks good. You may want to have a `click.option` instead of an argument: `--password/-p`.

marcofavorito: I would have used `getpass` instead:
https://docs.python.org/3/library/getpass.html#getpass.getpass

And an optional flag (default to false) to trigger the prompt (e.g. `--use-password`).

aea/cli/generate_wealth.py
```diff
@@ -38,21 +38,28 @@
     required=True,
 )
 @click.argument("url", metavar="URL", type=str, required=False, default=None)
+@click.argument(
+    "password", metavar="PASSWORD", type=str, default=None, required=False,
+)
```
marcofavorito, line 43: since it is used a lot, it could be implemented as "custom Click type". In the future, we may also want to restrict the type of passwords, e.g. see [here](https://en.wikipedia.org/wiki/Password_strength#Guidelines_for_strong_passwords).

marcofavorito: If my strategy above is applied, would be interesting to delegate the Click parameter type class to call getpass, so we get a pretty neat solution.

marcofavorito (approved):
LGTM. Instead of having password as parameter of every function, would be nice to refactor all this code a bit by introducing the `KeyLoader` abstraction. Then, we would have different cases, like `NotEncryptedKeyLoader`, `EncryptedKeyLoader` etc.

Another comment: instead of storing the password in plain-text, even if in-memory, we could hash it, to achieve greater security (depending on the context of usage of course)

aea/cli/add_key.py
DavidMinarsch: https://click.palletsprojects.com/en/7.x/options/

aea/cli/utils/package_utils.py
DavidMinarsch: done

solarw (approved):
LGTM

aea/cli/utils/click_utils.py
```diff
@@ -196,3 +196,37 @@ def handle_parse_result(
             )
 
         return super().handle_parse_result(ctx, opts, args)
+
+
+def password_option(confirmation_prompt: bool = False, **kwargs) -> Callable:  # type: ignore
```
marcofavorito, line 201: why this

plugins/aea-ledger-cosmos/aea_ledger_cosmos/cosmos.py
```diff
@@ -73,6 +76,136 @@
 }
 
 
+class DataEncrypt:
+    """Class to encrypt/decrypt data strings with password provided."""
+
+    hash_algo = "sha256"
+    hash_iterations = 20000
```
marcofavorito, line 83: isn't better to:

- add `hash_algo` and `hash_iterations` as parameters of the main APIs below
- make those parameters optional, with default values `_DEFAULT_HASH_ALGO` and `_HASH_ITERATIONS` respectively
- rename these class variables with `_DEFAULT_HASH_ALGO` and `_HASH_ITERATIONS`

marcofavorito: otherwise, say I wanted to use a different algorithm (either online or offline); I would have to modify these class variables which could break other code.

plugins/aea-ledger-ethereum/aea_ledger_ethereum/ethereum.py
```diff
@@ -328,7 +328,7 @@ def encrypt(self, password: str) -> str:
         :param password: the password to decrypt.
         :return: json string containing encrypted private key.
         """
-        encrypted = self.entity.encrypt(self.private_key, password)
+        encrypted = Account.encrypt(self.private_key, password)
```
marcofavorito: why this?

plugins/aea-ledger-fetchai/aea_ledger_fetchai/_cosmos.py
```diff
@@ -73,6 +76,136 @@
 }
 
 
+class DataEncrypt:
+    """Class to encrypt/decrypt data strings with password provided."""
+
+    hash_algo = "sha256"
+    hash_iterations = 20000
```
marcofavorito, line 83: see above

marcofavorito (approved):
LGTM

plugins/aea-ledger-cosmos/aea_ledger_cosmos/cosmos.py
DavidMinarsch: @solarw ^.

Issue #2300, opened 2021-02-23
Make `dependencies` in `aea-config.yaml` a mandatory field


marcofavorito:
**Is your feature request related to a problem? Please describe.**

Field `dependencies` in agent configuration files, (see #2289, #2291) has been made optional for backward compatibility; however, it should be a mandatory field.

**Describe the solution you'd like**

Make it mandatory, after `aea` release `0.11.0`

It would just be a matter of adding `dependencies` in the `required` field list of `aea-config_schema.json`.

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

marcofavorito:
also remove the "temporary" fix here: b72f757ac

Issue #1977, opened 2020-11-22
Configure agent goal setup


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
Currently, the agent's goals are implicitly defined by its skills and configurations thereof.

**Describe the solution you'd like**
We could provide further ways to configure it:
- expose DM and its configurations
- make preferences and ownership configurable
- see also https://github.com/fetchai/agents-aea/issues/658

PR #2362, opened 2021-03-16
AEA-1381 Enable config forwarding to error handler and decision maker handler

DavidMinarsch:
Make error handler instance rather than class based

## Proposed changes

The PR makes error handler and decision maker handler configurable. In particular, config object defined in yaml can be forwarded to it.

It also makes error handler instance based to avoid cross-agent effects for multiple agents within a given process.

## Fixes

- #1977, #2300

## Types of changes

## Checklist

## Further comments

-


aea/decision_maker/default.py
```diff
@@ -83,14 +83,13 @@ def role_from_first_message(  # pylint: disable=unused-argument
     signing_dialogue_class = SigningDialogue
     signing_msg_class = SigningMessage
 
-    def __init__(self, identity: Identity, wallet: Wallet) -> None:
```
DavidMinarsch, line 86: need to filter configs in base so they're not set on simple name space!

aea/aea_builder.py
```diff
@@ -1589,11 +1623,14 @@ def set_from_configuration(
...
             file_path = agent_configuration.decision_maker_handler["file_path"]
+            config = agent_configuration.error_handler["config"]
             self.set_decision_maker_handler_paths(dotted_path, file_path)
+            self.set_decision_maker_handler_config(config)
         if agent_configuration.error_handler != {}:
             dotted_path = agent_configuration.error_handler["dotted_path"]
             file_path = agent_configuration.error_handler["file_path"]
-            self.set_error_handler(dotted_path, file_path)
+            config = agent_configuration.error_handler["config"]
+            self.set_error_handler(dotted_path, file_path, config)
```
marcofavorito, line 1633: I wonder why we have `set_decision_maker_handler_paths` and `set_decision_maker_handler_config` but for handler we have only `set_error_handler` (maybe it is trivial, I haven't checked)

marcofavorito (approved):
LGTM

aea/aea_builder.py
DavidMinarsch: updated

aea/decision_maker/default.py
DavidMinarsch: updated

aea/aea_builder.py
marcofavorito: ok

DavidMinarsch: (the decision maker approach is the correct one as it guarantees loading happens **after** the protocols, contracts and connections are loaded, so error handler can depend on them (as it should be able to))

.pylintrc
```diff
@@ -34,7 +34,7 @@ max-locals=31
 max-statements=80
 max-parents=11
 max-branches=24
-max-attributes=34
+max-attributes=36
```
marcofavorito, line 37: instead of this, disable exception to "AEABuilder"?

solarw: actually AEABuilder is overbloated, lets keep max-attribtes low to get remind about attributes limit

solarw (approved):
LGTM

marcofavorito (approved):
LGTM

PR #2369, opened 2021-03-17
Feature/v0.11.2

DavidMinarsch:
## Proposed changes

This PR prepares develop for `v0.11.2`. It is mostly auto generated and includes a few bug fixes.

## Fixes

-

## Types of changes

## Checklist

## Further comments

-


HISTORY.md
```diff
@@ -1,5 +1,19 @@
 # Release History
 
+## 0.11.2 (2020-03-17)
+
+- Fixes a bug in testing tools
```
marcofavorito, line 5: I know we never did this, but now or in the future it might be good to reference the bug with its GitHub id, and also spend more words to describe the actual issue.

marcofavorito: I mean, if in this case it is a very minor bug, then it is fine (although in that case probably we should not mention it)

marcofavorito (approved):
LGTM

HISTORY.md
DavidMinarsch: I agree. We will adopt this approach from v1 onwards. You're welcome to do it for historic releases.

jrriehl (approved):
LGTM

PR #2373, opened 2021-03-18
Aea 1401/feature/golang based dialogues protocol

siddharthantier:
Adding initial work for creation mechanism for dialogue protocol - need review about the approach while moving ahead.


libs/go/aealite/protocols/dialogue.go
```diff
@@ -0,0 +1,142 @@
...
+// 	_role_from_first_message
+// 	_self_address
+// 	_outgoing_messages
+// 	_incoming_messages
+// }
+
+// TODO
+// Define RoleType and figure out how to find role
+
+type Label struct {
```
DavidMinarsch, line 32: LEt's call it DialogueLabel for symmetry. I'd implement the method `get_incomplete_version` for it too. As well as the getters for starter and responder reference.

libs/go/aealite/protocols/dialogue.go
```diff
@@ -0,0 +1,142 @@
...
+type Dialogue struct {
+	dialogueLabel Label
+	dialogueMessage *InitialMessage
+	selfAddress Address
+	// role RoleType
+}
+
+var dialogueStorage map[Address][]Label
+
+type InitialMessage struct {
```
DavidMinarsch, line 47: Not sure what's the point of this.

libs/go/aealite/protocols/dialogue.go
```diff
@@ -0,0 +1,142 @@
...
+// TODO
+// Define RoleType and figure out how to find role
+
+type Label struct {
+	dialogueReference [2]string
+	dialogueOpponentAddress Address
+	dialogueStarterAddress Address
+}
+
+type Dialogue struct {
```
DavidMinarsch, line 38: Missing: self_address, message_class which is associated with this dialogue, list of outgoing and incoming messages, list of ordered message ids, rules

libs/go/aealite/protocols/dialogue.go
```diff
@@ -0,0 +1,142 @@
...
+	dialogueReference [2]string
+	message_id int
+	target int
+	performative []string
+	message SomeMessageType
+	to Address
+	sender Address
+}
+
+func create(selfAddress Address, counterParty Address, performative Performative, message SomeMessageType) *Dialogue {
```
DavidMinarsch, line 57: first argument should be the Dialogues or DialogueCollection

DavidMinarsch: Drop the message and instead, we construct based on the message type assigned to the given dialogue.

libs/go/aealite/protocols/dialogue.go
```diff
@@ -0,0 +1,305 @@
...
+	getDialogueLabel() Label
+	getIncompleteDIalgueLabel() Label
+	getDialogueLabels() []Label // returns both dialogue and incomplete dialog labels Label
+	getSelfAddress() Address
+	getRole() RoleType
+	getRules() RuleType
+	getMessageCLass() InitialMessage
+	isSelfInitiated() bool
+	getLastIncomingMessage() InitialMessage
+	getLastOutgoingMessage() InitialMessage
```
DavidMinarsch, line 49: type..

DavidMinarsch: just call it `AbstractMessage` and define an interface for it according to what we do on python side

libs/go/aealite/protocols/dialogue.go
```diff
@@ -0,0 +1,305 @@
...
+	getLastMessage() InitialMessage
+	isEmpty() bool
+	counterPartyForMessage(InitialMessage) bool
+	isMessageBySelf(InitialMessage) bool
+	isMessageByOther(InitialMessage) bool
+	getMessage(MessageId) InitialMessage
+	hasMessageId(MessageId) bool
+	update(InitialMessage)
+	isBelongingToADialogue(InitialMessage) bool
+	reply(Performative, targetMessage, target) InitialMessage
```
DavidMinarsch, line 59: drop targetMessage and just keep `Target`

libs/go/aealite/protocols/dialogue.go
```diff
@@ -0,0 +1,300 @@
...
+	dialogueOpponentAddress Address
+	dialogueStarterAddress  Address
+}
+
+type Dialogue struct {
+	dialogueLabel    DialogueLabel
+	dialogueMessage  *AbstractMessage
+	selfAddress      Address
+	outGoingMessages []MessageId
+	GoingMessages    []MessageId
```
DavidMinarsch, line 131:
```suggestion
	outgoingMessageIds []MessageId
	incomingMessageIds    []MessageId
	orderedMessageIds []MessageId
	role string
```

libs/go/aealite/protocols/dialogue.go
```diff
@@ -0,0 +1,300 @@
...
+	tearDownDialogueStorage()
+}
+
+type AbstractMessageInterface interface {
+	initialize(diaglogReference [2]string, messagId int, target Target, performative Performative)
+	validPerformatives() []string
+	hasSender() bool
+	sender() Address
+	hasTo() bool
+	to() Address
```
DavidMinarsch, line 112:
```suggestion
	sender() Address
	to() Address
```

libs/go/aealite/protocols/dialogue.go
```diff
@@ -0,0 +1,300 @@
...
+	dialogueLabel    DialogueLabel
+	dialogueMessage  *AbstractMessage
+	selfAddress      Address
+	outGoingMessages []MessageId
+	GoingMessages    []MessageId
+	rules            RuleType
+}
+type AbstractMessage struct {
+	dialogueReference [2]string
+	message_id        int
```
DavidMinarsch:
```suggestion
	messageId        MessageId
```

libs/go/aealite/protocols/dialogue.go
```diff
@@ -0,0 +1,300 @@
...
+
+const (
+	Role1 Role = "role1"
+	Role2 Role = "role2"
+)
+
+type RuleType struct {
+	initialPerformative  Performative
+	terminalPerformative Performative
+	// validReplies
```
DavidMinarsch, line 41: valid replies is a map from a single performative to a set of performatives

libs/go/aealite/protocols/dialogue.go
```diff
@@ -0,0 +1,300 @@
...
+// 	_self_address
+// 	_outgoing_messages
+// 	_incoming_messages
+// }
+
+type Role string
+
+const (
+	Role1 Role = "role1"
+	Role2 Role = "role2"
```
DavidMinarsch: these are protocol specific, so not defined here

libs/go/aealite/protocols/dialogue.go
```diff
@@ -0,0 +1,300 @@
...
+	initialize(diaglogReference [2]string, messagId int, target Target, performative Performative)
+	validPerformatives() []string
+	hasSender() bool
+	sender() Address
+	hasTo() bool
+	to() Address
+	dialogueReference() [2]string
+	messageId() MessageId
+	performative() Performative
+	target() int
```
DavidMinarsch, line 116: Let's declare a specific type for target

libs/go/aealite/protocols/dialogue.go
```diff
@@ -0,0 +1,300 @@
...
+	equal(AbstractMessage) bool
+}
+
+type DialogueLabel struct {
+	dialogueReference       [2]string
```
DavidMinarsch, line 121: Let's prepare a specific type for `dialogueReference`

libs/go/aealite/protocols/dialogue.go
```diff
@@ -0,0 +1,314 @@
...
+const (
+	Role1             Role = "role1"
+	Role2             Role = "role2"
+	StartingMessageId      = 1
+	StartingTarget         = 0
+)
+
+type RuleType struct {
+	initialPerformative  Performative
+	terminalPerformative Performative
```
5A11, line 29: Not sure if these are singular or can be a collection later on, but initialPerformative and terminalPerformative are sets of performatives. So type-wise ensure that is possible.
Also then name-wise, initialPerformatives and terminalPerformatives (with s at the end).

libs/go/aealite/protocols/dialogue.go
```diff
@@ -0,0 +1,314 @@
...
+	// validReplies
+}
+
+type MessageId int
+type Target []int
```
5A11, line 34: target is a single int not an array of ints.

libs/go/aealite/protocols/dialogue.go
5A11: Why is that needed? 
Target is equivalent to messageID in type (it is essentially messageID of another message). So make sense for both to have the same type.

DavidMinarsch: Yes we can stick with int too.

libs/go/aealite/protocols/dialogue.go
siddharthantier: Okay, could I keep the type as a string that would act as the unique identifier for the specific protocol? Will this be passed with a message when initializing the dialogue?

libs/go/aealite/protocols/dialogue.go
```diff
@@ -0,0 +1,437 @@
...
+
+import (
+	"crypto/rand"
+	"encoding/hex"
+	"errors"
+	"fmt"
+)
+
+type Address string
+type Performative []string
```
marcofavorito, line 11: why a `Performative` is an array of strings?

In the Python version, we used an enumeration class to represent a "set of performatives (i.e. type identifiers)". However, a variable of type "PerformativeX" can be thought of as a specific value of that set, so I'd say `Performative` should be a string. Thoughts?

libs/go/aealite/protocols/dialogue.go
```diff
@@ -0,0 +1,437 @@
...
+	Role1             Role = "role1"
+	Role2             Role = "role2"
+	StartingMessageId      = 1
+	StartingTarget         = 0
+)
+
+type RuleType struct {
+	initialPerformatives  Performative
+	terminalPerformatives Performative
+	validReplies map[Performative][]Performative
```
marcofavorito, line 30: if the above is fixed, then this line makes sense to me.

libs/go/aealite/protocols/dialogue.go
```diff
@@ -0,0 +1,437 @@
...
+const (
+	Role1             Role = "role1"
+	Role2             Role = "role2"
+	StartingMessageId      = 1
+	StartingTarget         = 0
+)
+
+type RuleType struct {
+	initialPerformatives  Performative
+	terminalPerformatives Performative
```
marcofavorito, line 29: if the above comment is correct, the type of these fields should be changed to `Performative[]`

marcofavorito: However, since in Golang there is no built-in operator to check the membership of an element in an array, we could use the "map trick", i.e. use `map[Performative]bool`.

libs/go/aealite/protocols/dialogue.go
```diff
@@ -0,0 +1,437 @@
...
+	messageId() MessageId
+	performative() Performative
+	target() Target
+}
+
+func (message AbstractMessage) hasSender() (Address, error) {
+	if message.sender != "" {
+		return message.sender, nil
+	}
+	return "", errors.New("Sender address does not exist.")
```
marcofavorito, line 82: (disclaimer: very picky comment)

> Error strings should not be capitalized (unless beginning with proper nouns or acronyms) or end with punctuation, since they are usually printed following other context.
https://github.com/golang/go/wiki/CodeReviewComments#error-strings

marcofavorito: feel free to _not_ follow this advice :slightly_smiling_face:

libs/go/aealite/protocols/dialogue.go
```diff
@@ -0,0 +1,437 @@
...
+
+	// TODO
+	// validate next message
+	// check if dialogue message is valid
+	dialogue.validateNextMessage(message)
+	dialogue.updateIncomingAndOutgoingMessages(message)
+}
+
+func (dialogue Dialogue) validateNextMessage(message AbstractMessage) (bool, string) {
+	is_basic_validated, msg_basic_validation := dialogue.basicValidation(message)
```
marcofavorito, line 141: in Golang, identifiers should be camelCase:
```suggestion
	isBasicValidated, msgBasicValidation := dialogue.basicValidation(message)
```

libs/go/aealite/protocols/dialogue.go
```diff
@@ -0,0 +1,437 @@
...
+	if !is_basic_validated {
+		return false, msg_basic_validation
+	}
+	// TODO
```
marcofavorito, line 145: (this comment is just to keep track of that TODO)

libs/go/aealite/protocols/dialogue.go
```diff
@@ -0,0 +1,437 @@
...
+		return dialogue.basicValidationInitialMessage(message)
+	}
+	return dialogue.basicValidationNonInitialMessage(message)
+}
+
+func (dialogue Dialogue) basicValidationInitialMessage(message AbstractMessage) (bool, string) {
+	dialogue_reference := message.dialogueReference
+	message_id := message.messageId
+	// performative := message.performative
+	if dialogue_reference[0] != dialogue.dialogueLabel.dialogueReference[0] {
```
marcofavorito, line 161: see above on camelCase variables

libs/go/aealite/protocols/dialogue.go
```diff
@@ -0,0 +1,437 @@
...
+	// }
+	return true, "The initial message passes basic validation."
+}
+
+func (dialogue Dialogue) basicValidationNonInitialMessage(message AbstractMessage) (bool, string) {
+	dialogue_reference := message.dialogueReference
+	if dialogue_reference[0] != dialogue.dialogueLabel.dialogueReference[0] {
+		return false, "Invalid dialogue_reference[0]."
+	}
+	err := dialogue.validateMessageId(message)
```
marcofavorito, line 191: `validateMessageId` returns a `string`, but a nullable is expected (maybe `error` is a better return type?)

marcofavorito: line to change:

https://github.com/fetchai/agents-aea/pull/2373/files#diff-170ecc70a35a8331838e8a1c8489c844807589ac0a14491241ded20ff3c767deR251

libs/go/aealite/protocols/dialogue.go
```diff
@@ -0,0 +1,437 @@
...
+	if err != nil {
+		return false, err
```
marcofavorito, line 193: if `err` is changed to `error`, then you can use` err.Error()` (or return `error` to the caller as well). I would choose the latter option if there aren't other limitations

libs/go/aealite/protocols/dialogue.go
```diff
@@ -0,0 +1,437 @@
...
+	err = dialogue.validateMessageTarget(message)
+	if err != nil {
+		return false, err
+	}
+	return true, "The non-initial message passes basic validation."
+}
+
+func (dialogue Dialogue) validateMessageTarget(message AbstractMessage) string {
+	target = message.target
+    performative = message.performative
```
marcofavorito, line 204:
```suggestion
	target := message.target
    performative := message.performative
```

marcofavorito: beware of tabs vs spaces

libs/go/aealite/protocols/dialogue.go
```diff
@@ -0,0 +1,437 @@
...
+	incomingMessages  []AbstractMessage
+	lastMessageId     int
+	orderedMessageIds []int
+	rules             RuleType
+}
+type AbstractMessage struct {
+	dialogueReference [2]string
+	messageId         MessageId
+	target            Target
+	performative      []string
```
marcofavorito, line 112: Why a performative is a set of strings? Each message is associated with one and only one performative.
```suggestion
	performative      Performative
```

libs/go/aealite/protocols/dialogue.go
```diff
@@ -0,0 +1,437 @@
...
+
+func max(list []int) int {
+	max := list[0]
+    for i :=1; i < len(s); i++ {
+        if max < list[i] {
+            max = list[i]
+        }
+    }
+	return max
+}
```
marcofavorito, line 249: I propose the following implementation:
```suggestion
func max(list []int) int {
	max := list[0]
    for i :=1; i < len(s); i++ {
        if max < list[i] {
            max = list[i]
        }
    }
	return max
}
```

The above one has some minor issues: `s` not defined, raw for loop instead of range loop, `currentMax` variable name is different from the function name so to make it more readable)

libs/go/aealite/protocols/dialogue.go
```diff
@@ -0,0 +1,437 @@
...
+	"errors"
+	"fmt"
+)
+
+type Address string
+type Performative []string
+type SomeMessageType string
+
+const (
+	NONCE_BYTES_NB = 32
```
marcofavorito, line 15: my suggestion is to use camel case also for constants:

https://stackoverflow.com/a/22688926/7150172

marcofavorito: also, I find this comment is particularly meaningful:
https://stackoverflow.com/questions/22688906/go-naming-conventions-for-const#comment39214871_22688926

marcofavorito, line 15:
```suggestion
	NonceBytesNb = 32
```

libs/go/aealite/protocols/dialogue.go
```diff
@@ -0,0 +1,437 @@
...
+	isMessageBySelf(AbstractMessage) bool
+	isMessageByOther(AbstractMessage) bool
+	getMessage(MessageId) AbstractMessage
+	hasMessageId(MessageId) bool
+
+	update(AbstractMessage)
+	validateNextMessage(AbstractMessage) (bool, string)
+	basicValidations(AbstractMessage) (bool, string)
+	basicValidationInitialMessage(AbstractMessage) (bool, string)
+	basicValidationNonInitialMessage(AbstractMessage) (bool, string)
```
marcofavorito, line 54: why `(bool, string)` and not `error`? Then, if `error_obj == nil` means validation succeded, else `error_obj.Error()` returns the error message.

marcofavorito: Now I see; this comes from Python code. Let us leave this as it is, although we could think of a custom struct `ValidationResult` to better encapsulate the returned value of these methods.

DavidMinarsch: I think the python code is no good guide, I prefer the error to be returned as Marco suggests, that's more idiomatic. But can be done later of course.

libs/go/aealite/protocols/dialogue.go
```diff
@@ -0,0 +1,437 @@
...
+)
+
+type RuleType struct {
+	initialPerformatives  Performative
+	terminalPerformatives Performative
+	validReplies map[Performative][]Performative
+}
+
+type MessageId int
+type Target int
```
marcofavorito, line 34: I think `Target` should not be a type. A "target" of a message is, within the same dialogue, the message id of the (previous) message it is addressed to. So I suggest using `MessageId` wherever `Target` is used.

libs/go/aealite/protocols/dialogue.go
```diff
@@ -0,0 +1,437 @@
...
+	basicValidationNonInitialMessage(AbstractMessage) (bool, string)
+	isEmpty() bool
+	updateIncomingAndOutgoingMessages(AbstractMessage)
+	isBelongingToADialogue(AbstractMessage) bool
+
+	validateMessageTarget(AbstractMessage) string
+	validateMessageId(AbstractMessage) string
+	getMessageById(MessageId) AbstractMessage
+	getOutgoingNextMessageId() int
+	getIncomingNextMessageId() int
```
marcofavorito, line 63: use `MessageId` rather than `int`. Then, if needed, cast to `int`

libs/go/aealite/protocols/dialogue.go
```diff
@@ -0,0 +1,437 @@
...
+    if dialogue.lastIncomingMessage(){
+		next_message_id = dialogue.lastMessageId + 1
+	}
+    if dialogue.isSelfInitiated() {
+		next_message_id = 0 - next_message_id
+	}
+	return next_message_id
+}
+
+func (dialogue Dialogue)lastOutgoingMessage() (bool, int){
```
marcofavorito, line 286: I would use `*AbstractMessage`, the closest type to Python's `Optional[Message]`

libs/go/aealite/protocols/dialogue.go
```diff
@@ -0,0 +1,437 @@
...
+	if len(dialogue.outgoingMessages) > 0 {
+		return true, dialogue.outgoingMessages[-1]
+	} else {
+		false, 0
+	}
+}
+
+func (dialogue Dialogue)lastIncomingMessage() (bool, int){
```
marcofavorito, line 294: see above

libs/go/aealite/protocols/dialogue.go
```diff
@@ -0,0 +1,437 @@
...
+	basicValidations(AbstractMessage) (bool, string)
+	basicValidationInitialMessage(AbstractMessage) (bool, string)
+	basicValidationNonInitialMessage(AbstractMessage) (bool, string)
+	isEmpty() bool
+	updateIncomingAndOutgoingMessages(AbstractMessage)
+	isBelongingToADialogue(AbstractMessage) bool
+
+	validateMessageTarget(AbstractMessage) string
+	validateMessageId(AbstractMessage) string
+	getMessageById(MessageId) AbstractMessage
```
marcofavorito, line 61: To be more faithful with the Python implementation, this method should return a reference to the message:
```suggestion
	getMessageById(MessageId) *AbstractMessage
```

however, I would also accept returning by value (which implies a copy of the returned value)

libs/go/aealite/protocols/dialogue_test.go
```diff
@@ -0,0 +1,101 @@
...
+*   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+*   See the License for the specific language governing permissions and
+*   limitations under the License.
+*
+* ------------------------------------------------------------------------------
+ */
+
+package protocols
+
+//func TestDialogue(t *testing.T) {
```
marcofavorito, line 23: TODO

DavidMinarsch (approved):
LGTM

PR #2374, opened 2021-03-18
AEA-173 cli password promting improved

solarw:
## Proposed changes

reusable password_option decorator provided
it adds two options
-p flag to ask password interactively
--password to set password with command line
(can not combine it, cause one flag and another is option requires argument)

if -p and --password specified same time, password will be asked interactively
we can change priority and --password value will be used.

## Types of changes

## Checklist

### DIFF
--- a/aea/cli/utils/click_utils.py
+++ b/aea/cli/utils/click_utils.py
@@ -217,7 +217,7 @@ def password_option(confirmation_prompt: bool = False, **kwargs) -> Callable:  #
             type=bool,
             callback=callback,
             expose_value=False,
-            help="Ask for password interactivly",
+            help="Ask for password interactively",
         )(
             click.option(
                 "--password",
--- a/aea/cli/utils/click_utils.py
+++ b/aea/cli/utils/click_utils.py
@@ -204,7 +204,7 @@ def password_option(confirmation_prompt: bool = False, **kwargs) -> Callable:  #
     def callback(ctx, _, value: bool) -> bool:  # type: ignore
         if value is True:
             ctx.params["password"] = ctx.params.get("password") or click.prompt(
-                "Enter password please",
+                "Enter password",
                 hide_input=True,
                 confirmation_prompt=confirmation_prompt,
             )
--- a/aea/cli/utils/click_utils.py
+++ b/aea/cli/utils/click_utils.py
@@ -196,3 +196,37 @@ class MutuallyExclusiveOption(Option):
             )
 
         return super().handle_parse_result(ctx, opts, args)
+
+
+def password_option(confirmation_prompt: bool = False, **kwargs) -> Callable:  # type: ignore
+    """Decorator to ask for input if -p flag was provided or use --password to set password value in command line."""
+
+    def callback(ctx, _, value: bool) -> bool:  # type: ignore
+        if value is True:
+            ctx.params["password"] = ctx.params.get("password") or click.prompt(
+                "Enter password please",
+                hide_input=True,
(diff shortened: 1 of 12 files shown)


DavidMinarsch (approved):
LGTM

aea/cli/utils/click_utils.py
```diff
@@ -196,3 +196,37 @@ def handle_parse_result(
             )
 
         return super().handle_parse_result(ctx, opts, args)
+
+
+def password_option(confirmation_prompt: bool = False, **kwargs) -> Callable:  # type: ignore
+    """Decorator to ask for input if -p flag was provided or use --password to set password value in command line."""
```
marcofavorito, line 202: I think this is OK. Not important, but the only comment I have is that, ideally, `-p` could only be an abbreviation for `--password`, and not behave in a different way.

Some ideas to avoid the above:
1. `--password/-p` is an option that can have zero or one argument; if provided, then use that password, otherwise ask to the user. Basically, like `argparse` with `nargs='?'`. I'm not sure how to implement it with Click though.
2. Have `--password/-p` in combination with another flag, `--interactive/-i` (default: false). If only `--password/-p` is provided, then an argument must be provided (the option). Otherwise (`-i` provided), ask to user.

aea/cli/utils/click_utils.py
```diff
@@ -196,3 +196,37 @@ def handle_parse_result(
...
+        return value
+
+    def wrap(fn):  # type: ignore
+        return click.option(
+            "-p",
+            is_flag=True,
+            type=bool,
+            callback=callback,
+            expose_value=False,
+            help="Ask for password interactivly",
```
marcofavorito, line 220:
```suggestion
            help="Ask for password interactively",
```

aea/cli/utils/click_utils.py
```diff
@@ -196,3 +196,37 @@ def handle_parse_result(
...
         return super().handle_parse_result(ctx, opts, args)
+
+
+def password_option(confirmation_prompt: bool = False, **kwargs) -> Callable:  # type: ignore
+    """Decorator to ask for input if -p flag was provided or use --password to set password value in command line."""
+
+    def callback(ctx, _, value: bool) -> bool:  # type: ignore
+        if value is True:
+            ctx.params["password"] = ctx.params.get("password") or click.prompt(
+                "Enter password please",
```
marcofavorito, line 207:
```suggestion
                "Enter password",
```

(this is a very picky comment... feel free to ignore) I think `please` can be avoided here.

Consider adding a semicolon `:` if it is not already added by Click.

aea/cli/utils/click_utils.py
```diff
@@ -196,3 +196,37 @@ def handle_parse_result(
...
+            type=bool,
+            callback=callback,
+            expose_value=False,
+            help="Ask for password interactivly",
+        )(
+            click.option(
+                "--password",
+                **kwargs,
+                type=str,
+                is_eager=True,
```
marcofavorito, line 226: question: would this make `--password` to compete with other "eager" parameters, as `--version`?

solarw: should not be a problem,

aea/cli/utils/click_utils.py
solarw: 1. in short. can not.
i tried initially. but no success.
it can be flag or option, not both at the same time.
2. the idea was to probide self contain reusable solution, so -p and --password are togheter and i dont to want to introduce also interactive option cause it can be used for something else and be out of scope of the password

PR #2384, opened 2021-03-21
Minor fixes after removal of key auto-generation

marcofavorito:
## Proposed changes

- improve error message when the AEA is being built with no addresses
  - without this, we would have "Error: Wallet has no addresses".
- default_ledger set to None by default and checked that it has been set before build
- add a check that defaut_ledger is in wallet addresses
  - without this, since there isn't key auto-generation anymore, we would get an error.
- made `default_ledger` nullable in configurations. It is checked before building that it is not null.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a

### DIFF
--- a/aea/aea_builder.py
+++ b/aea/aea_builder.py
@@ -82,7 +82,13 @@ from aea.crypto.ledger_apis import DEFAULT_CURRENCY_DENOMINATIONS
 from aea.crypto.wallet import Wallet
 from aea.decision_maker.base import DecisionMakerHandler
 from aea.error_handler.base import AbstractErrorHandler
-from aea.exceptions import AEAException, AEAValidationError, enforce
+from aea.exceptions import (
+    AEAException,
+    AEANoDefaultLedgerException,
+    AEAValidationError,
+    AEAWalletNoAddressException,
+    enforce,
+)
 from aea.helpers.base import find_topological_order, load_env_file, load_module
 from aea.helpers.exception_policy import ExceptionPolicyEnum
 from aea.helpers.install_dependency import install_dependency
@@ -1186,20 +1192,26 @@ class AEABuilder(WithLogger):  # pylint: disable=too-many-public-methods
         if self._name is None:  # pragma: nocover
             raise ValueError("You must set the name of the agent.")
 
+        default_ledger = self.get_default_ledger_not_null()
         if not wallet.addresses:
-            raise ValueError("Wallet has no addresses.")
+            raise AEAWalletNoAddressException("Wallet has no addresses.")
+
+        if default_ledger not in wallet.addresses:
+            raise ValueError(
+                f"Specified default ledger '{default_ledger}' not found in available addresses of types: {'[' + ','.join(wallet.addresses.keys()) + ']'}"
+            )
 
         if len(wallet.addresses) > 1:
             identity = Identity(
                 self._name,
                 addresses=wallet.addresses,
-                default_address_key=self.get_default_ledger(),
+                default_address_key=default_ledger,
             )
         else:
             identity = Identity(
                 self._name,
-                address=wallet.addresses[self.get_default_ledger()],
-                default_address_key=self.get_default_ledger(),
+                address=wallet.addresses[default_ledger],
+                default_address_key=default_ledger,
             )
         return identity
 
@@ -1334,13 +1346,25 @@ class AEABuilder(WithLogger):  # pylint: disable=too-many-public-methods
         self._build_called = True
         return aea
 
-    def get_default_ledger(self) -> str:
+    def get_default_ledger(self) -> Optional[str]:
         """
         Return default ledger.
 
-        :return: the default ledger identifier.
+        :return: the default ledger identifier, or None.
         """
-        return self._default_ledger or self.DEFAULT_LEDGER
+        return self._default_ledger
+
+    def get_default_ledger_not_null(self) -> str:
+        """
+        Return default ledger if not None, else raise error.
+
+        :return: the default ledger identifier.
+        :raise: Value
+        """
+        default_ledger = self.get_default_ledger()
+        if default_ledger is None:
+            raise AEANoDefaultLedgerException("No default ledger specified.")
+        return default_ledger
 
     def _get_agent_act_period(self) -> float:
         """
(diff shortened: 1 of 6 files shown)


aea/aea_builder.py
```diff
@@ -1186,20 +1192,26 @@ def _build_identity_from_wallet(self, wallet: Wallet) -> Identity:
...
 
+        default_ledger = self.get_default_ledger_not_null()
         if not wallet.addresses:
-            raise ValueError("Wallet has no addresses.")
+            raise AEAWalletNoAddressException("Wallet has no addresses.")
+
+        if default_ledger not in wallet.addresses:
+            raise ValueError(
+                f"Specified default ledger '{default_ledger}' not found in available addresses of types: {'[' + ','.join(wallet.addresses.keys()) + ']'}"
+            )
```
marcofavorito, line 1202: without this, we would have a cryptic error message like
```
Error: 'fetchai'
```
typical of `ValueError` when a key is not present in a dictionary (see below `wallet.addresses[default_ledger]`)

aea/aea_builder.py
```diff
@@ -1334,13 +1346,25 @@ def build(
         self._build_called = True
         return aea
 
-    def get_default_ledger(self) -> str:
+    def get_default_ledger(self) -> Optional[str]:
```
marcofavorito, line 1349: this was missing.

aea/aea_builder.py
```diff
@@ -1334,13 +1346,25 @@ def build(
...
+        :return: the default ledger identifier, or None.
+        """
+        return self._default_ledger
+
+    def get_default_ledger_not_null(self) -> str:
+        """
+        Return default ledger if not None, else raise error.
+
         :return: the default ledger identifier.
+        :raise: Value
```
marcofavorito, line 1362:
```suggestion
        :raise: ValueError when the default ledger field is None.
```

aea/cli/create.py
```diff
@@ -180,7 +179,7 @@ def _create_agent_config(ctx: Context, agent_name: str, set_author: str) -> Agen
         license_=DEFAULT_LICENSE,
         registry_path=os.path.join("..", DEFAULT_REGISTRY_PATH),
         description="",
-        default_ledger=DEFAULT_LEDGER,
+        default_ledger=None,
```
marcofavorito, line 182: as the AEA has not the key generation anymore, I think it makes more sense to have it nullable (but mandatory).

aea/cli/run.py
```diff
@@ -195,5 +197,17 @@ def _build_aea(
...
+        raise click.ClickException(error_msg)
+    except AEAWalletNoAddressException:
+        error_msg = (
+            "You haven't specified any private key for the AEA project.\n"
+            "Please add one by using the following commands:\n"
+            "\n"
+            f"    aea generate-key {DEFAULT_LEDGER}\n"
+            f"    aea add-key fetchai {PRIVATE_KEY_PATH_SCHEMA.format(DEFAULT_LEDGER)}\n"
+        )
+        raise click.ClickException(error_msg)
```
marcofavorito, line 211: we need custom exceptions as we need to catch this special cases.
We couldn't use just `ValueError` because these error messages wouldn't make sense in `AEABuilder` for, say, programmatic usage.

aea/cli/create.py
marcofavorito: In theory, this can be reverted while keeping the rest.

DavidMinarsch: I don't agree. An AEA needs a ledger, so we might as well set the default one, as we do already. Then `aea install` installs the relevant plugin.

aea/aea_builder.py
DavidMinarsch: See below, nullable makes everything more complicated for no good reason.

aea/aea_builder.py
DavidMinarsch: Screenshot 2021-03-21 at 12 22 03
On current develop ^

DavidMinarsch: So this error message already exists somewhere.

DavidMinarsch: What needs fixing is to tell the user to run `aea install`!

DavidMinarsch (commented):
PR is too invasive for what it tries to fix. Please see my comments first. The key problem is that `aea install` needs to be run when ledger plugin is not yet installed. This should be made clear to the user. All the other changes are more problematic than helpful imo.

aea/aea_builder.py
marcofavorito: no, these are two different issues. Assume all the plug-ins are installed correctly.

1) `default_ledger` specified, NO addresses in the wallet:
you get `Wallet has no addresses`, which doesn't give suggestion on how to solve the problem (add key, generate key)
2) `default_ledger` specified, some addresses in the wallet but NONE of them of type `default_ledger`
you get `Error: 'fetchai'`, and the commented change fixes that by explaining what happened.

aea/cli/create.py
marcofavorito: that doesn't solve the issue of no key present in the AEA project, even if all the plugins are installed. Reason is that there isn't anymore the key generation.

marcofavorito:
I agree on removing nullable, but not on the other points, which are orthogonal to `aea install` and the plug-in mechanism.

aea/aea_builder.py
marcofavorito: if default_ledger would allow null values, we would have another case that would stop the build of the AEA.

aea/cli/create.py
DavidMinarsch: Yes, a simple error message informing the user of the missing key is needed.

aea/aea_builder.py
DavidMinarsch: revert

aea/aea_builder.py
DavidMinarsch: 1. Extend error message looks like the solution.
2. Yes a helpful error message is needed there.
Avoid null value of default ledger, not helpful.

marcofavorito:
closed this in favor of #2418

Issue #2360, opened 2021-03-16
CLI version vs package requirements blocking


eorituz:
**Describe the bug**
When adding a protocol (also works with agents etc.) that requires a different aea CLI version the installation succeeds but afterwards any further CLI command get's blocked due to version inconsistency. 
```
Error: The CLI version is 0.10.0, but package fetchai/http:0.13.0 requires version <0.12.0,>=0.11.0
```

**To Reproduce**
1. Install aea 0.10.0
2. Add a protocol/agent/package which requires another CLI version. e.g. `aea add protocol fetchai/http:0.13.0` (requires 0.11+)
3. Any other aea CLI command will result in an error due to the version inconsistency. 

**Expected behavior**
I would expect to at least be able to remove the "wrong" protocol:
```
(sensor_demo_aea) oliver@bbpc:~/Arbeit/aea/sensor_demo/buyer$ aea remove protocol fetchai/http:0.13.0
Error: The CLI version is 0.10.0, but package fetchai/http:0.13.0 requires version <0.12.0,>=0.11.0
```
This should not throw an error but work. Given the current state, I have no option to get rid of the mistakenly added protocol.
Sucks pretty hard because one can't simply delete some files to get rid of the protocol manaually (at least I wasn't able to come up with it in a reasonable amount of time) --> I had to hard reset to the last commit.

**Desktop (please complete the following information):**
 ```
(sensor_demo_aea) oliver@bbpc:~/Arbeit/aea/sensor_demo/buyer$ uname -a
Linux bbpc 5.8.0-44-generic #50~20.04.1-Ubuntu SMP Wed Feb 10 21:07:30 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
```
 - AEA Version: 0.10.0

**Additional context**

DavidMinarsch:
Thanks for the report. We'll fix it for v1.0.0rc next week.

PR #2390, opened 2021-03-23
[AEA-1420] [AEA-1424] [AEA-1415] [AEA-1414] Fix/misc coverage tests

solarw:
## Proposed changes

coverage fixes

## Fixes

fixes for cli remove command #2360
fix for mam plugins load after dependencies installed

## Types of changes

## Checklist

### DIFF
--- a/aea/crypto/plugin.py
+++ b/aea/crypto/plugin.py
@@ -22,7 +22,7 @@ import itertools
 import pprint
 from typing import Iterator, List, Set
 
-from pkg_resources import EntryPoint, iter_entry_points
+from pkg_resources import EntryPoint, WorkingSet
 
 from aea.configurations.constants import (
     ALLOWED_GROUPS,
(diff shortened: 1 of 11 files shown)


aea/multiplexer.py
```diff
@@ -342,9 +342,9 @@ async def disconnect(self) -> None:
                 self.connection_status.set(ConnectionStates.disconnecting)
                 await asyncio.wait_for(self._stop(), timeout=60)
                 self.logger.debug("Multiplexer disconnected.")
-            except (CancelledError) as e:
-                self.logger.exception("Exception on disconnect:")
-                raise AEAConnectionError("Failed to disconnect the multiplexer.") from e
+            except CancelledError:  # pragma: nocover
+                self.logger.debug("Multiplexer.disconnect cancellation!")
+                raise
```
DavidMinarsch, line 347: What problem does this relate to? Looks good, just wondering

DavidMinarsch (approved):
LGTM

aea/crypto/plugin.py
```diff
@@ -129,7 +129,8 @@ def _get_plugins(group: str) -> List[Plugin]:
     :param group: the plugin group.
     :return: a mapping from plugin name to Plugin objects.
     """
-    entry_points: List[EntryPoint] = list(iter_entry_points(group=group))
+
+    entry_points: List[EntryPoint] = list(WorkingSet().iter_entry_points(group=group))
```
marcofavorito, line 133: what is the difference?

solarw: `WorkingSet()` so looks like default workingset populated once, and iter_entry_points is the same during python process lifetime, even you added removed python packages.

aea/multiplexer.py
solarw: not a problem, just log that disconnect cancelled and exception propagated,
it's not assumed that disconnect should be cancelled, but it's possible

PR #2397, opened 2021-03-23
AEA-1410 Remove package registry_path from aea-config into cli config

DavidMinarsch:
## Proposed changes

This PR removes `registry_path` from the `aea-config.yaml` and object. The registry path is a cli side setting, not a project specific one. It makes no sense to have it in each project.

## Fixes

-

## Types of changes

## Checklist

## Further comments

-


aea/cli/search.py
```diff
@@ -121,29 +115,6 @@ def agents(ctx: Context, query: str, page: int) -> None:
     _output_search_results(item_type, *search_items(ctx, item_type, query, page), page)
 
 
-def setup_search_ctx(ctx: Context, local: bool) -> None:
```
DavidMinarsch, line 124: A lot of code for very little

aea/cli/utils/config.py
```diff
@@ -160,6 +165,24 @@ def set_cli_author(click_context: click.Context) -> None:
...
+    registry_path = Path(DEFAULT_REGISTRY_NAME)
+    if registry_path.absolute().is_dir():
+        return str(registry_path)
+    registry_path = Path("..", DEFAULT_REGISTRY_NAME)
+    if registry_path.absolute().is_dir():
+        return str(registry_path)
+    cli_logger.debug(
+        f"Registry path not provided and `packages` not found in current ({os.getcwd()}) and parent directory."
+    )
+    return ""
```
DavidMinarsch, line 183: I am not happy about the default. I'd prefer to raise an exception, but that is not ideal either as then commands unrelated to the registry stop working for no reason. So it's probably better to raise at the point of using the path. In `Context` getter of `registr_path` we could raise when the path is the empty string.

DavidMinarsch: Another question is when and where to set the registry path on cli config. One way to do it is during `aea init`. Otherwise we need another cli command which I would like to avoid..

aea/cli/utils/context.py
```diff
@@ -47,12 +47,7 @@ class Context:
 
     agent_config: AgentConfig
 
-    def __init__(
-        self,
-        cwd: str = ".",
-        verbosity: str = "INFO",
-        registry_path: Optional[str] = None,
-    ) -> None:
+    def __init__(self, cwd: str, verbosity: str, registry_path: str) -> None:
```
DavidMinarsch, line 50: The defaults did not help

aea/configurations/base.py
```diff
@@ -1134,7 +1133,6 @@ class AgentConfig(PackageConfiguration):
     FIELDS_ALLOWED_TO_UPDATE: FrozenSet[str] = frozenset(
         [
             "description",
-            "registry_path",
```
DavidMinarsch, line 1137: Kept in schema until v1 to avoid breaking everything.

tests/test_cli/test_create.py
```diff
@@ -429,8 +433,6 @@ def setup_class(cls):
 
         # calling 'aea create myagent' again within an AEA project - recursively.
         os.chdir(cls.agent_name)
-        os.mkdir("another_subdir")
-        os.chdir("another_subdir")
```
DavidMinarsch, line 433: test didn't require this...

tests/test_cli/test_search.py
```diff
@@ -53,7 +53,6 @@ class TestSearchProtocolsLocal:
     @classmethod
     def setup_class(cls):
         """Set the test up."""
-        cls.cwd = os.getcwd()
```
DavidMinarsch, line 56: pointless here

aea/cli/utils/context.py
```diff
@@ -47,12 +47,7 @@ class Context:
...
         """Init the context."""
         self.config = dict()  # type: Dict
         self.cwd = cwd
```
DavidMinarsch, line 53: raise on registry_path get if `registry_path == ""`

PR #2409, opened 2021-03-25
AEA-1375 aea.cli fingerprint for agent

solarw:
## Proposed changes

aea.cli fingerprint for agent

## Types of changes

## Checklist


aea/cli/utils/decorators.py
```diff
@@ -114,20 +114,34 @@ def _validate_config_consistency(ctx: Context, check_aea_version: bool = True) -
...
+    check_finger_prints: bool = False,
+) -> None:
     try:
         click_context = args[0]
         ctx = cast(Context, click_context.obj)
         try_to_load_agent_config(ctx)
         skip_consistency_check = ctx.config["skip_consistency_check"]
         if not skip_consistency_check:
             _validate_config_consistency(ctx, check_aea_version=check_aea_version)
+        if check_finger_prints:
```
DavidMinarsch, line 129: This is not used anywhere?

DavidMinarsch, line 129: It's false everywhere? Why was it added?

aea/cli/fingerprint.py
```diff
@@ -189,3 +189,17 @@ def fingerprint_package(
...
+    """Do a fingerprint for an agent."""
+    ctx = cast(Context, click_context.obj)
+    fingerprints_dict = _compute_fingerprint(
+        Path(ctx.cwd),
+        ignore_patterns=ctx.agent_config.fingerprint_ignore_patterns,
+        is_recursive=False,
+    )  # type: Dict[str, str]
+    ctx.agent_config.fingerprint = fingerprints_dict
+    ctx.dump_agent_config()
+    click.echo(f"Fingerprint for agent `{ctx.agent_config.name}` is calculated!")
```
DavidMinarsch, line 205: To be consistent with other commands, move to line 198 and print 
`Fingerprinting files in agent project 'agent_name' ...`

aea/cli/utils/decorators.py
solarw: it's used for tests at least, and probably will be used in future as a part of consistency check

DavidMinarsch (approved):
LGTM

marcofavorito:
Looks good, but I would like to share some thoughts.

I think this has the problem that some directories with project data will not be detected. That would also mean we need to manually exclude `vendor`, `protocols`, `connections`, `contracts` and `skills`, as its content is validated separately.

Moreover, files like private keys are fingerprinted as well... Why? I think they should be ignored by reading the content of the private key fields.

marcofavorito (approved):
LGTM

aea/configurations/base.py
```diff
@@ -1792,20 +1792,23 @@ def _create_or_update_from_json(
...
     hasher = IPFSHashOnly()
     fingerprints = {}  # type: Dict[str, str]
     # find all valid files of the package
     all_files = [
         x
-        for x in package_directory.glob("**/*")
+        for x in package_directory.glob("**/*" if is_recursive else "*")
         if x.is_file()
-        and (
-            x.match("*.py") or not any(x.match(pattern) for pattern in ignore_patterns)
```
DavidMinarsch, line 1807: Not sure why this was like that. ignore patterns were never evaluated for python files...

PR #2417, opened 2021-03-26
removing soef

devjsc:
Removed SOEF from aea docs, now a top level item in the main docs repo.


DavidMinarsch:
Rejecting this for two reasons:
- needs to be merging into develop and branching of develop as per repo policy
- we need to update various cross references.
I'll be happy to fix it in a PR from our end.

PR #2443, opened 2021-04-07
AEA-1435 AW5 AEA

5A11:
This PR adds an AW5 AEA, along the way modifying existing skills in a backwards compatible way. 
These changes to existing skills were required for the AW5 agent.

Here is the instructions for running these agents which worked on my machine locally:

```
1
aea fetch --local fetchai/confirmation_aea_aw5

2
aea fetch --local fetchai/simple_buyer_aw5

3
aea fetch --local fetchai/simple_seller_aw5

####################
#    In all
####################
aea install
aea build
aea generate-key fetchai
aea add-key fetchai
aea generate-key fetchai fetchai_connection_private_key.txt
aea add-key fetchai fetchai_connection_private_key.txt --connection
aea issue-certificates
aea get-address fetchai

####################
#    For 2
####################

Get funds from 
https://explore-agentworld.fetch.ai

Confirm with
aea get-wealth fetchai

####################
#    For 1
####################

Get funds from 
https://explore-agentworld.fetch.ai

Confirm with
aea get-wealth fetchai

(Ensure the wealth amount is more than token_dispense_amount in 1’s aea-config)

####################
#    Config 1
####################

awx_aeas: [ADDRESS OF 2]

####################
#    Config 2
####################

search_interval: APPROPRIATE AMOUNT (controls how often searches happen)

aw1_aea: ADDRESS OF 1
leaderboard_token: some_token
leaderboard_url: www.someurl.com

####################
#    Config 3
####################

developer_handle: fetchai
whitelist:
      - ADDRESS OF 1

value: fetchai

method: GET
url: https://api.tfl.gov.uk/Line/Mode/tube/Disruption

####################
#    In all
####################

aea run
```

## Types of changes

## Checklist


packages/fetchai/agents/aw5/aea-config.yaml
```diff
@@ -0,0 +1,160 @@
+agent_name: aw5
```
DavidMinarsch, line 1:
```suggestion
agent_name: simple_seller_aw5
```

packages/fetchai/agents/aw5/aea-config.yaml
```diff
@@ -0,0 +1,160 @@
...
+author: fetchai
+version: 0.1.0
+license: Apache-2.0
+description: An agent that participates in Agent World 5.
```
DavidMinarsch, line 5:
```suggestion
description: An agent that participates in Agent World 5 as a simple seller.
```

packages/fetchai/agents/aw5/aea-config.yaml
```diff
@@ -0,0 +1,160 @@
...
+public_id: fetchai/simple_service_registration:0.20.0
+type: skill
+models:
+  strategy:
+    args:
+      location:
+        latitude: 51.5194
+        longitude: 0.127
+      service_data:
+        key: agentworld-1
```
DavidMinarsch, line 95:
```suggestion
        key: agentworld-5
```

packages/fetchai/agents/aw5/aea-config.yaml
```diff
@@ -0,0 +1,160 @@
...
+      - PUT_WHITELIST_ADDRESSES_HERE
+---
+public_id: fetchai/simple_service_registration:0.20.0
+type: skill
+models:
+  strategy:
+    args:
+      location:
+        latitude: 51.5194
+        longitude: 0.127
```
DavidMinarsch, line 93: Hm this will not work..let's discuss

packages/fetchai/agents/aw5/aea-config.yaml
```diff
@@ -0,0 +1,160 @@
...
+models:
+  strategy:
+    args:
+      search_location:
+        latitude: 51.5194
+        longitude: 0.127
+      search_query:
+        constraint_type: ==
+        search_key: registration_service
+        search_value: aw1-registration
```
DavidMinarsch, line 109:
```suggestion
        search_value: aw5-registration
```

packages/fetchai/agents/aw5/aea-config.yaml
```diff
@@ -0,0 +1,160 @@
...
+      chain_id: agentworld-1
+---
+public_id: fetchai/simple_seller:0.11.0
+type: skill
+models:
+  strategy:
+    args:
+      location:
+        latitude: 51.5194
+        longitude: 0.127
```
DavidMinarsch, line 139: will conflict with above

packages/fetchai/skills/simple_data_request/behaviours.py
```diff
@@ -62,6 +63,13 @@ def act(self) -> None:
 
         :return: None
         """
+        if self.lookup_termination_key is not None:
```
DavidMinarsch, line 66: a skill can be deactivated via is_active flag on context, use it here

packages/fetchai/agents/aw5/aea-config.yaml
DavidMinarsch: Easiest option, first registration skill runs with location of registration agent, then when that's done we activate seller skill and register agent in new location. Alternatively, we have no registration and buyer buys from all agents (as we now have main net we no longer need eth address)

packages/fetchai/skills/simple_data_request/behaviours.py
5A11: The `self.lookup_termination_key` in `shared_state` may not be set at this point by the other skill yet (because it hasn't finished yet) but could be set later. So can't deactivate this skill if the condition fails.

packages/fetchai/skills/confirmation_aw1/handlers.py
```diff
@@ -129,24 +129,32 @@ def _handle_register(
...
+                contract_api_dialogue.terms = terms
+                contract_api_dialogue.associated_register_dialogue = register_dialogue
+                self.context.outbox.put_message(contract_api_msg)
+            else:
+                strategy.finalize_registration(register_msg.sender)
+                register_dialogue.terms = terms
+                tx_behaviour = cast(
+                    TransactionBehaviour, self.context.behaviours.transaction
+                )
+                tx_behaviour.waiting.append(register_dialogue)
```
DavidMinarsch, line 157: ? Not sure what this is for...

packages/fetchai/skills/confirmation_aw1/strategy.py
```diff
@@ -152,18 +162,50 @@ def valid_registration(
         :param sender: the sender
         :return: tuple of success, error code and error message
         """
-        if not all([key in registration_info for key in REQUIRED_KEYS]):
+        if not all([key in registration_info for key in DEVELOPER_ONLY_REQUIRED_KEYS]):
```
DavidMinarsch, line 165: the required keys should be a default arg to the function as you'll need to override it below.

packages/fetchai/skills/confirmation_aw1/strategy.py
```diff
@@ -152,18 +162,50 @@ def valid_registration(
...
+    ) -> Tuple[bool, int, str]:
+        """
+        Check if the registration info is valid.
+
+        :param registration_info: the registration info
+        :param sender: the sender
+        :return: tuple of success, error code and error message
+        """
+        is_valid, error_code, error_msg = self._valid_registration_developer_only(
+            registration_info, sender
```
DavidMinarsch, line 197: pass required keys here, as different depending on case

packages/fetchai/skills/confirmation_aw1/strategy.py
```diff
@@ -152,18 +162,50 @@ def valid_registration(
...
+        )
+
+        if self.developer_handle_only or not is_valid:
+            return (is_valid, error_code, error_msg)
+
+        if not all([key in registration_info for key in REQUIRED_KEYS]):
```
DavidMinarsch, line 203: See above

packages/fetchai/skills/registration_aw1/behaviours.py
```diff
@@ -43,24 +43,28 @@ def setup(self) -> None:
...
-                counterparty_address="",
-                amount_by_currency_id={},
-                quantities_by_good_id={},
-                nonce="",
-            ),
-        )
-        self.context.logger.info("sending signing_msg to decision maker...")
-        self.context.decision_maker_message_queue.put_nowait(msg)
+        if strategy.announce_termination_key is not None:
+            self.context.shared_state[strategy.announce_termination_key] = False
```
DavidMinarsch, line 47: What do we need this for?

packages/fetchai/skills/registration_aw1/strategy.py
```diff
@@ -43,38 +43,47 @@ def __init__(self, **kwargs: Any) -> None:
...
+                f"Not a valid signature_of_fetchai_address: {signature_of_fetchai_address}",
+            )
+            self._signature_of_fetchai_address = signature_of_fetchai_address
+            tweet = kwargs.pop("tweet", DEFAULT_TWEET)
+            enforce(isinstance(tweet, str), "Not a valid tweet link")
+            self._tweet = tweet
+            self._is_ready_to_register = False
+        else:
+            self._is_ready_to_register = True
+            self._ethereum_address = "some_dummy_address"
```
DavidMinarsch, line 83: ""/None is better imo

5A11: This messes up the property type hint for `ethereum_address` below and updating that with `Optional` leads to further mypy complains elsewhere in the handler.

packages/fetchai/skills/registration_aw1/behaviours.py
5A11: if `strategy.announce_termination_key` is not `None`, it means we want this skill to announce when it finishes its job (so another skill which needs to start strictly after this one knows when to start). 

The `announce_termination_key` is the key in the `shared_state` that both these skills try to write to/read from.

Here (one of the first codes executed by this skill) we are adding this key in the `shared_state` with `False` to announce this skill hasn't finished yet so when the other skill reads the value of this key from `shared_state`
it is there with value False. 
This value changes to True when the skill finishes (handler `_handle_success`).

packages/fetchai/skills/confirmation_aw1/strategy.py
5A11: The logic is slightly different here from what you have in mind. 

`_valid_registration_developer_only` contains all checks needed for the developer only mode. Then in `valid_registration` we call `_valid_registration_developer_only`. If we are in developer_only mode, we've done all checks so just return the results. If not, those checks are a prerequisite, so we continue with the other checks and return the overall result.

packages/fetchai/skills/confirmation_aw1/handlers.py
5A11: To register the terms of registration as a tx on the ledger.

DavidMinarsch (commented):
LGTM, just one gap `packages/fetchai/skills/registration_aw1/strategy.py                       73      4    95%   82-85`

tests/test_packages/test_skills/test_confirmation_aw1/test_strategy.py
```diff
@@ -297,16 +322,11 @@ def test_valid_registration_fails_v(self):
         }
 
         # operation
-        with patch.object(
```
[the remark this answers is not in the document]
5A11: We do. This covers another check failing. Previously every check was in the same place so the `_valid_signature` were called in all checks and so had to be mocked. This is one of those checks that now happen in `_valid_registration_developer_only` which does not include checks involving `_valid_signature `, hence no longer needing it to be mocked.

DavidMinarsch (approved):
LGTM

PR #2473, opened 2021-04-28
AEA-1494 Combine Genus and Classification registration

5A11:
This PR makes it possible to combine genus and classification registration steps.

[WIP]: Currently the PR only shows the solution in the generic_seller skill. Once approved, it will extend it to all other skills.

## Proposed changes

Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request.

## Fixes

## Types of changes

## Checklist

### DIFF
--- a/packages/fetchai/connections/soef/connection.py
+++ b/packages/fetchai/connections/soef/connection.py
@@ -879,7 +879,13 @@ class SOEFChannel:
         if not (isinstance(piece, str) and isinstance(value, str)):  # pragma: nocover
             raise SOEFException.debug("Personality piece bad values provided.")
 
-        await self._set_personality_piece(piece, value)
+        pieces = piece.split(",")
+        values = value.split(",")
+        if len(pieces) != len(values):
+            raise SOEFException.debug("Mismatched number of personality pieces and values.")
+
+        for idx in range(len(pieces)):
+            await self._set_personality_piece(pieces[idx], values[idx])
         await self._send_success_response(oef_message, oef_search_dialogue)
 
     async def _set_personality_piece(self, piece: str, value: str) -> None:
--- a/packages/fetchai/skills/generic_seller/behaviours.py
+++ b/packages/fetchai/skills/generic_seller/behaviours.py
@@ -148,8 +148,7 @@ class GenericServiceRegistrationBehaviour(TickerBehaviour):
         strategy = cast(GenericStrategy, self.context.strategy)
         descriptions = [
             strategy.get_register_service_description(),
-            strategy.get_register_personality_description(),
-            strategy.get_register_classification_description(),
+            strategy.get_registration_description(),
         ]
         oef_search_dialogues = cast(
             OefSearchDialogues, self.context.oef_search_dialogues
--- a/packages/fetchai/skills/generic_seller/strategy.py
+++ b/packages/fetchai/skills/generic_seller/strategy.py
@@ -78,6 +78,7 @@ class GenericStrategy(Model):
                 latitude=location["latitude"], longitude=location["longitude"]
             )
         }
+
         self._set_personality_data = kwargs.pop(
             "personality_data", DEFAULT_PERSONALITY_DATA
         )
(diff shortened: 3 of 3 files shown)


packages/fetchai/connections/soef/connection.py
```diff
@@ -879,7 +879,13 @@ async def _set_personality_piece_handler(
         if not (isinstance(piece, str) and isinstance(value, str)):  # pragma: nocover
             raise SOEFException.debug("Personality piece bad values provided.")
 
-        await self._set_personality_piece(piece, value)
+        pieces = piece.split(",")
+        values = value.split(",")
+        if len(pieces) != len(values):
+            raise SOEFException.debug("Mismatched number of personality pieces and values.")
+
+        for idx in range(len(pieces)):
```
DavidMinarsch, line 887: use enumerate instead

packages/fetchai/connections/soef/connection.py
```diff
@@ -879,7 +879,13 @@ async def _set_personality_piece_handler(
...
+            await self._set_personality_piece(pieces[idx], values[idx])
```
DavidMinarsch, line 888: Have you checked whether it is possible to set a whole (comma separated) list at once?

packages/fetchai/skills/generic_seller/behaviours.py
```diff
@@ -148,8 +148,7 @@ def register_service_personality_classification(self) -> None:
         strategy = cast(GenericStrategy, self.context.strategy)
         descriptions = [
             strategy.get_register_service_description(),
-            strategy.get_register_personality_description(),
-            strategy.get_register_classification_description(),
+            strategy.get_registration_description(),
         ]
         oef_search_dialogues = cast(
             OefSearchDialogues, self.context.oef_search_dialogues
```
DavidMinarsch, line 154: This loop below is a problem btw. It does not play well with the logic you submitted earlier this week. If both fail, then the first failed message can be overwritten with the second...

packages/fetchai/skills/generic_seller/behaviours.py
```diff
@@ -148,8 +148,7 @@ def register_service_personality_classification(self) -> None:
         strategy = cast(GenericStrategy, self.context.strategy)
         descriptions = [
             strategy.get_register_service_description(),
-            strategy.get_register_personality_description(),
-            strategy.get_register_classification_description(),
```
DavidMinarsch, line 152: I wasn't aware we already set these. So in some sense your PR isn't needed..it just offers a different way of doing it, which is still valuable in general.

packages/fetchai/skills/generic_seller/strategy.py
```diff
@@ -190,6 +200,17 @@ def get_register_classification_description(self) -> Description:
         )
         return description
 
+    def get_registration_description(self) -> Description:
```
DavidMinarsch, line 203: The other two methods are no longer used and can be removed

packages/fetchai/skills/generic_seller/behaviours.py
5A11: ok, the second bit can be broken down and moved into the SUCCESS and ERROR handlers

5A11: Will have to do that in all skills using the same pattern

packages/fetchai/skills/generic_seller/behaviours.py
5A11: We do in some skills, not all. 

So task 1 is to combine the two registration steps, and task 2 is to essentially copy paste this solution into all skills.

packages/fetchai/connections/soef/connection.py
5A11: I have to see the SOEF api for that. Is there such a thing accessible?

5A11: I had a look at a rough documentation.md file in the soef repo and the command doesn't seem to be getting multiple values (I may be mistaken as there was only some rough explanation there and nothing formal).

PR #2477, opened 2021-04-29
AEA-1534 Integrate aggregation with simple oracle skill

jrriehl:
## Proposed changes

Add an option to the simple oracle skill to publish an aggregated value rather than a single observation.

## Types of changes

## Checklist


packages/fetchai/skills/simple_aggregation/strategy.py
```diff
@@ -69,7 +70,7 @@ def __init__(self, **kwargs: Any) -> None:
         self._round = 0
         self._peers = set()  # type: Set[Address]
         self._observations = dict()  # type: Dict[Address, Dict[str, Any]]
-        self._aggregation = None  # type: Optional[Dict[str, Any]]
+        self._aggregation = None  # type: Optional[Any]
```
DavidMinarsch, line 73: Is it Any? or can it be more tightly defined?

jrriehl: Needed to make the typing compatible with the statistics package here, but probably we can do a cast with a more specific definition.

DavidMinarsch: ok I see

packages/fetchai/skills/simple_aggregation/strategy.py
```diff
@@ -179,7 +181,12 @@ def aggregate_observations(self) -> None:
             self.context.logger.info("No observations to aggregate")
             return
         self._aggregation = self._aggregate(values)
-        self.context.shared_state["aggregation"] = self._aggregation
+        self.context.shared_state["aggregation"] = {
```
DavidMinarsch, line 184: In the shared state the keys themselves should be configurable. Otherwise if two skills define the same key you have a problem if its hard coded.

packages/fetchai/skills/simple_oracle/behaviours.py
```diff
@@ -107,15 +107,19 @@ def act(self) -> None:
...
             return
 
-        # Check for observation from data collecting skill
-        observation = self.context.shared_state.get("observation", None)
-        if observation is None:
+        # Check for oracle from data collecting skill
+        if strategy.aggregation:  # pragma: nocover
+            oracle_data = self.context.shared_state.get("aggregation", None)
+        else:
+            oracle_data = self.context.shared_state.get("observation", None)
```
DavidMinarsch, line 114: see above for both keys

DavidMinarsch: Not sure why we need the if/else here. Difference is just the key, once that's configurable you can remove it.

5A11 (commented):
Agree with David's 'shared_state keys should be configurable comments', otherwise lgtm

PR #2506, opened 2021-05-11
AEA-1555 add --no-bump option to 'generate_all_protocols' script

marcofavorito:
## Proposed changes

Add `--no-bump` option to prevent automatic bump of protocol public id.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a


marcofavorito:
This error is likely due to the new Click release:

https://github.com/fetchai/agents-aea/pull/2506/checks?check_run_id=2560160487#step:7:91

https://pypi.org/project/click/

@DavidMinarsch @solarw

scripts/generate_all_protocols.py
```diff
@@ -427,18 +488,22 @@ def main() -> None:
         for package_path in all_protocols:
             log("=" * 100)
             log(f"Processing protocol at path {package_path}")
-            _bump_protocol_specification_id_if_needed(package_path)
-            _process_packages_protocol(package_path)
+            if not no_bump:
+                _bump_protocol_specification_id_if_needed(package_path)
+            # no_bump implies to ignore the docstring:
+            #  'It was created with protocol buffer compiler ... and aea version ...'
```
DavidMinarsch, line 494: Why are the two related?

DavidMinarsch (approved):
LGTM, one question

scripts/generate_all_protocols.py
marcofavorito: image

marcofavorito: if `--no-bump` is provided, then we don't expect the docstring to change due to just using an `aea` with a different version.
if `--no-bump` is not provided, then the script works as usual, taking the output of the generator as-is (after some post-processing that would happen anyway).

DavidMinarsch: ? still doesn't answer my question. The protocol spec is specified in README.md. So I don't understand why this would be bumped in this case...

solarw (approved):
LGTM

scripts/generate_all_protocols.py
DavidMinarsch: Might you be confusing protocol id and protocol spec id @marcofavorito ?

DavidMinarsch: Probably no harm to leave it as you have done, but I don't think the protocol spec id would be bumped anyway unless manually bumped.

marcofavorito: `protocol_id` is bumped via another script

PR #2523, opened 2021-05-23
AEA-1573 fix minor issues on MultiAgentManager

### LINKED ISSUE #2482: Examples of unhelpful error messages
This issue lists unhelpful error messages so they can be made more informative.

Make them informative and actionable

Screenshot 2021-04-29 at 14 15 51

marcofavorito:
## Proposed changes

- add check 'ensure_private_keys' at the end of AgentAlias.set_overrides, which in turn is called by 'MultiAgentManager.add_agent'.
- Print a warning message when an exception occurred but no error callback is registered.

## Fixes

Partially fixes #2482

## Types of changes

## Checklist

## Further comments

n/a


aea/manager/manager.py
```diff
@@ -263,8 +267,11 @@ async def _manager_loop(self) -> None:
                 agent_name = agents_run_tasks_futures[task]
                 self._agents_tasks.pop(agent_name)
                 if task.exception():
-                    for callback in self._error_callbacks:
-                        callback(agent_name, task.exception())
+                    if len(self._error_callbacks) == 0:
+                        self._print_exception_occurred_but_no_error_callback(agent_name)
```
solarw, line 271: make it default callback?

solarw (approved):
LGTM

aea/manager/manager.py
DavidMinarsch, line 271: why not print the exception too? I.e. pass task.exception() here?

aea/manager/manager.py
```diff
@@ -843,3 +850,20 @@ def _save_state(self) -> None:
...
+        """
+        Print a warning message when an exception occurred but no error callback is registered.
+
+        :param agent_name: the agent name.
+        :return: None
+        """
+        if self._warning_message_printed_for_agent.get(agent_name, False):
+            return
+        self._warning_message_printed_for_agent[agent_name] = True
+        self._print_exception_occurred_but_no_error_callback(agent_name)
```
DavidMinarsch, line 864: recursion?

aea/manager/project.py
```diff
@@ -301,7 +301,8 @@ def set_overrides(
                 )
 
         overrides["component_configurations"] = component_configurations
-        return self.agent_config_manager.update_config(overrides)
+        self.agent_config_manager.update_config(overrides)
+        self._ensure_private_keys()
```
DavidMinarsch, line 305: Not sure if this makes sense here; I'd prefer it to be only called if agent_overrides are not None; to make explicit fact that it depends on changes in agent overrides

aea/manager/manager.py
DavidMinarsch: Yes that's a good idea. Why don't we have a default callback?

marcofavorito: Then it means we want a default callback, not a warning message

marcofavorito: if someone adds a callback with `add_error_callback`, should the default callback be replaced?

DavidMinarsch: Yes

DavidMinarsch: Yeah I guess, it would be better for development

DavidMinarsch (approved):
LGTM

PR #2524, opened 2021-05-24
AEA-1524 [WIP] Dockerised SOEF

5A11:
[WIP] This PR adds SOEF docker image as a test fixture.

## Proposed changes

Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request.

## Fixes

## Types of changes

## Checklist


tests/conftest.py
```diff
@@ -730,6 +731,20 @@ def ganache(
...
 
+@pytest.mark.integration
+@pytest.fixture(scope="function")
+def soef(
+    soef_addr: str = "http://127.0.0.1",
+    soef_port: int = 9002,
+    timeout: float = 2.0,
+    max_attempts: int = 10,
+):
+    """Launch the Ganache image."""
```
DavidMinarsch, line 742:
```suggestion
    """Launch the SOEF image."""
```

tests/common/docker_image.py
```diff
@@ -337,6 +337,53 @@ def wait(self, max_attempts: int = 15, sleep_rate: float = 1.0) -> bool:
...
+        :param client: the Docker client.
+        :param addr: the address.
+        :param port: the port.
+        """
+        super().__init__(client)
+        self._addr = addr
+        self._port = port
+
+    @property
+    def tag(self) -> str:
```
5A11, line 358: One question is, where do we want to have the image? Upload to DockerHub and pull from there or have it built on every call?

tests/conftest.py
```diff
@@ -730,6 +731,20 @@ def ganache(
     yield from _launch_image(image, timeout=timeout, max_attempts=max_attempts)
 
 
+@pytest.mark.integration
+@pytest.fixture(scope="function")
```
5A11, line 735: Also, as @marcofavorito pointed out, this scope will be costly to setup and tear down on every test that uses this fixture, but on the other hand the whole point of having introduced this is to have clean separate image instances on different tests.

tests/common/docker_image.py
marcofavorito: The build would happen only once, and the next builds would use the cached Docker image layers, so it takes much less time than the first build.

The approach that pulls the image from DockerHub is easier to implement (basically the current code would already work), so if it is possible to publish the Docker image under `fetchai/` would be better in my opinion.

marcofavorito: The approach that builds the image manually would have to:

- clone the repository https://github.com/fetchai/soef in a temporary directory
- enter in the temporary directory, and call `docker build`

To save time, the clone should only happen if the image is not available. There should be some way to use the Python package `docker`'s API to check that.

tests/conftest.py
marcofavorito: true. Let's keep it with `function` for now; another approach could be to have a decorator for test functions that, in case of failure, would restart the SOEF and repeat the test.

tests/common/docker_image.py
```diff
@@ -337,6 +337,53 @@ def wait(self, max_attempts: int = 15, sleep_rate: float = 1.0) -> bool:
...
+        for i in range(max_attempts):
+            try:
+                response = requests.post(f"{self._addr}:{self._port}", json=request)
+                enforce(response.status_code == 200, "")
+                return True
+            except Exception:
+                logger.info(f"Attempt {i} failed. Retrying in {sleep_rate} seconds...")
+                time.sleep(sleep_rate)
+        return False
+
```
marcofavorito, line 385: here we should devise a health check mechanism for SOEF; that is, we should detect when the SOEF is ready to work.

tests/common/docker_image.py
```diff
@@ -337,6 +337,53 @@ def wait(self, max_attempts: int = 15, sleep_rate: float = 1.0) -> bool:
         return False
 
 
+class SOEFDockerImage(DockerImage):
+    """Wrapper to SOEF Docker image."""
+
+    def __init__(
+        self, client: DockerClient, addr: str, port: int = 9002,
```
marcofavorito, line 344: consider putting `9002` as a (constant) class field

tests/conftest.py
DavidMinarsch: I think scope to class would be preferable no?

tests/common/docker_image.py
DavidMinarsch: Good point. Should be simple to do. SOEF has an endpoint for this, just need to find out the details.

tests/conftest.py
5A11: We can do class and later change if need to.

Issue #2517, opened 2021-05-20
Docstring pass across the code base


5A11:
**Describe the bug**

There are various discrepancies and mismatches in docstrings. 
Need to do an complete pass over them across the codebase.

PR #2527, opened 2021-05-25
AEA-1597 feat: add darglint to CI

DavidMinarsch:
## Proposed changes

This PR adds darglint to CI (`aea` only for now).

## Fixes

- #2517 

## Types of changes

## Checklist

## Further comments

-


setup.cfg
```diff
@@ -215,3 +215,8 @@ ignore_missing_imports = True
 
 [mypy-packages/fetchai/protocols/yoti/yoti_pb2]
 ignore_errors = True
+
+[darglint]
+docstring_style=sphinx
+strictness=short
```
DavidMinarsch, line 221: This means we permit single line docstrings which we do want for properties. When the docstring is more than a single line the full checks are applied.

marcofavorito (approved):

setup.cfg
```diff
@@ -215,3 +215,9 @@ ignore_missing_imports = True
...
+ignore_regex=.*_pb2\.py
+ignore=DAR401
```
DavidMinarsch, line 223: For now we ignore unmentioned raised exceptions.

aea/aea.py
```diff
@@ -121,8 +125,6 @@ def __init__(
         :param storage_uri: optional uri to set generic storage
         :param task_manager_mode: task manager mode (threaded) to run tasks with.
         :param kwargs: keyword arguments to be attached in the agent context namespace.
-
-        :return: None
```
DavidMinarsch, line 125: Darglint does not accept this approach; so removing it.

aea/aea_builder.py
```diff
@@ -970,6 +965,7 @@ def add_component_instance(self, component: Component) -> "AEABuilder":
         You will have to `reset()` the builder before calling `build()` again.
 
         :param component: Component instance already initialized.
+        :return: self
```
marcofavorito, line 968: not consistent with `:return: the AEABuilder.` below

aea/aea_builder.py
```diff
@@ -970,6 +965,7 @@ def add_component_instance(self, component: Component) -> "AEABuilder":
...
         """
         self._to_reset = True
         self._check_can_add(component.configuration)
```
marcofavorito, line 971: below, `set_context_namespace` does not state the return value

aea/cli/remove.py
```diff
@@ -76,47 +76,31 @@ def remove(
 @click.argument("connection_id", type=PublicIdParameter(), required=True)
 @pass_ctx
 def connection(ctx: Context, connection_id: PublicId) -> None:
-    """
-    Remove a connection from the agent.
-
-    It expects the public id of the connection to remove from the local registry.
```
DavidMinarsch, line 82: For click it's better we have a single line string.

aea/context/base.py
```diff
@@ -118,14 +119,14 @@ def send_to_skill(
...
-        if message passed it will be wrapped into envelope with optional envelope context.
+        If message passed it will be wrapped into envelope with optional envelope context.
 
-        :return: None
+        :param message_or_envelope: envelope to send to another skill.
+        :param context: the optional envelope context
         """
         if self._send_to_skill is None:  # pragma: nocover
             raise ValueError("Send to skill feature is not supported")
-        return self._send_to_skill(message_or_envelope, context)
```
DavidMinarsch, line 128: inconsequential return

solarw (approved):
LGTM

aea/cli/local_registry_sync.py
```diff
@@ -111,8 +108,9 @@ def enlist_packages(
     Generate list of the packages in local repo directory.
 
     :param base_dir: path or str of the local repo.
+    :param skip_consistency_check: whether or not to skip consistency checks.
 
-    :return: generator with Tuple of package_id and package directory.
+    :yield: generator with Tuple of package_id and package directory.
```
marcofavorito, line 113: The function "yields" a tuple, not a generator; thoughts?

```suggestion
    :yield: a tuple of package_id and package directory.
```

aea/manager/manager.py
```diff
@@ -566,6 +561,7 @@ def set_agent_overrides(
             raise ValueError("Agent is running. stop it first!")
 
         self._agents[agent_name].set_overrides(agent_overides, components_overrides)
+        return self
```
DavidMinarsch, line 564: Altered return type in line with rest of manager. Should not break anything.

aea/cli/remove.py
```diff
@@ -188,7 +172,9 @@ def _get_item_requirements(
         """
         List all the requirements for item provided.
 
-        :return: generator with package ids: (type, public_id)
+        :param item: the item package configuration
+        :param ignore_non_vendor: whether or not to ignore vendor packages
+        :yield: generator with package ids: (type, public_id)
```
marcofavorito, line 177: see above re "yield" vs "return".

aea/cli/remove.py
```diff
@@ -281,6 +271,9 @@ def remove_unused_component_configurations(ctx: Context) -> Generator:
 
     Context manager!
     Clean all configurations on enter, restore actual configurations and dump agent config.
+
+    :param ctx: click context
+    :yield: generator
```
marcofavorito, line 276: This should be "yield None", although since we chose to ignore "return None" we should ignore this as well... what does Darglint thinks about this?
```suggestion
    :yield: None
```

aea/cli/remove.py
```diff
@@ -322,11 +315,9 @@ def __init__(
         :param ctx: click context.
         :param item_type: str, package type
         :param item_id: PublicId of the item to remove.
+        :param with_dependencies: whether or not to remove dependencies.
         :param force: bool. if True remove even required by another package.
-        :param ignore_non_vendor: bool. if True, ignore non-vendor packages when computing
-          inverse dependencies. The effect of this flag is ignored if force = True
-
-        :return: None
+        :param ignore_non_vendor: bool. if True, ignore non-vendor packages when computing inverse dependencies. The effect of this flag is ignored if force = True
```
marcofavorito, line 320: a bit long row? Maybe split if it is allowed

DavidMinarsch: Not trivial to split rows now for some reason. For now keeping it as one line.

aea/aea_builder.py
DavidMinarsch: Single line docstrings are permitted and not parsed. Without this we'd have to unnecessarily add an extensive docstring for each property.

aea/aea_builder.py
DavidMinarsch: (half the file uses one, half the other :D)...keeping as is now

marcofavorito (approved):

marcofavorito:
tests not passing

Issue #2508, opened 2021-05-12
ML Train Demo throws exception


5A11:
**Describe the bug**
The ML Train demo throws an exception after 1 round of training

**To Reproduce**
1. Follow the ml train demo
2. after running ml_train agent you see exception after the first round of training

**Expected behavior**
No exception; the agents must continue trading and training

**Screenshots**
<img width="1679" alt="Screenshot 2021-05-12 at 13 39 16" src="https://user-images.githubusercontent.com/38721653/117976301-874d0400-b327-11eb-9c10-fc9b89e0edcc.png">

**Desktop (please complete the following information):**
 - OS: macOS
 - AEA Version 11.3.1

PR #2529, opened 2021-05-26
ML skill fix

5A11:
This PR fixes a problem with the ML skill #2508. 

## Types of changes

## Checklist


packages/fetchai/skills/ml_train/handlers.py
```diff
@@ -129,36 +129,25 @@ def _handle_terms(
...
-            )
-            terms_ = strategy.terms_from_proposal(ml_trade_msg.terms)
-            ml_trade_dialogue.terms = terms_
-            _, ledger_api_dialogue = ledger_api_dialogues.create(
-                counterparty=LEDGER_API_ADDRESS,
-                performative=LedgerApiMessage.Performative.GET_RAW_TRANSACTION,
-                terms=terms_,
-            )
-            ledger_api_dialogue = cast(LedgerApiDialogue, ledger_api_dialogue)
-            ledger_api_dialogue.associated_ml_trade_dialogue = ml_trade_dialogue
```
DavidMinarsch, line 161: What happened to that? Why can it be removed?

packages/fetchai/skills/ml_train/behaviours.py
```diff
@@ -64,8 +64,7 @@ def act(self) -> None:
             if not result.successful():
                 return
 
-            ml_task = result.get()
-            strategy.weights = ml_task.result
```
DavidMinarsch, line 68: So the property result does not need to be used?

DavidMinarsch (changes requested):

packages/fetchai/skills/ml_train/behaviours.py
5A11: Yep

packages/fetchai/skills/ml_train/handlers.py
5A11: Not sure why this block was there. 

This logic (sending GET_RAW_TRANSACTION) is what happens when we add the ml_dialogue to the tx_behaviour for processing (two lines below).

DavidMinarsch: Well spotted!

DavidMinarsch (approved):
LGTM

PR #2562, opened 2021-06-18
AEA-1630, AEA-1634 SOEF Dockerised

5A11:
This PR adds a fixture that can be added to tests to run an isolated SOEF docker instance. It also uses the fixture on the tac integration test.

NOTE: Tests using the new fixture do NOT run on CI because of authentication issue.
Need to address authentication issue on google cloud on which the SOEF docker is hosted to be able to run the tests in CI and merge this PR.

## Types of changes

## Checklist

### DIFF
--- a/tests/common/docker_image.py
+++ b/tests/common/docker_image.py
@@ -340,7 +340,7 @@ class GanacheDockerImage(DockerImage):
 class SOEFDockerImage(DockerImage):
     """Wrapper to SOEF Docker image."""
 
-    PORT = 9002
+    PORT = 12002
     SOEF_MOUNT_PATH = os.path.abspath(os.path.join(os.sep, "etc", "soef"))
     SOEF_CONFIG_FILE_NAME = "soef.conf"
 
--- a/tests/common/docker_image.py
+++ b/tests/common/docker_image.py
@@ -396,7 +396,7 @@ class SOEFDockerImage(DockerImage):
 
     def _make_ports(self) -> Dict:
         """Make ports dictionary for Docker."""
-        return {f"{self.PORT}/tcp": ("0.0.0.0", self._port)}  # nosec
+        return {f"{self._port}/tcp": ("0.0.0.0", self._port)}  # nosec
 
     def create(self) -> Container:
         """Create the container."""
--- a/tests/common/docker_image.py
+++ b/tests/common/docker_image.py
@@ -373,7 +373,7 @@ class SOEFDockerImage(DockerImage):
             "# (Author Toby Simpson)",
             "#",
             "# Port we're listening on",
-            "port 9002",
+            f"port {self._port}",
             "#",
             "# Our declared location",
             "latitude 52.205278",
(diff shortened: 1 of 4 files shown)


DavidMinarsch (approved):
LGTM

tests/common/docker_image.py
```diff
@@ -337,6 +337,90 @@ def wait(self, max_attempts: int = 15, sleep_rate: float = 1.0) -> bool:
         return False
 
 
+class SOEFDockerImage(DockerImage):
+    """Wrapper to SOEF Docker image."""
+
+    PORT = 12002
```
5A11, line 343: I think we don't need anymore if we want it to be anything other than 9002 (the original value). If it works with any other port, we can just have the parameterised self._port assigned to all parts (marked below).

tests/common/docker_image.py
```diff
@@ -337,6 +337,90 @@ def wait(self, max_attempts: int = 15, sleep_rate: float = 1.0) -> bool:
...
+        """Make a temporary soef_config file to setup and run the an soef instance."""
+        soef_config_lines = [
+            "# SIMPLE OEF CONFIGURATION FILE",
+            "# Save as /etc/soef/soef.conf",
+            "#",
+            "# 27th May 2020",
+            "# (Author Toby Simpson)",
+            "#",
+            "# Port we're listening on",
+            f"port {self._port}",
```
5A11, line 376: here

tests/common/docker_image.py
```diff
@@ -337,6 +337,90 @@ def wait(self, max_attempts: int = 15, sleep_rate: float = 1.0) -> bool:
...
+            "# End.",
+        ]
+        soef_config_file = os.path.join(tmpdirname, self.SOEF_CONFIG_FILE_NAME)
+        with open(soef_config_file, "w") as file:
+            file.writelines(line + "\n" for line in soef_config_lines)
+        os.chmod(soef_config_file, 400)  # nosec
+
+    def _make_ports(self) -> Dict:
+        """Make ports dictionary for Docker."""
+        return {f"{self.PORT}/tcp": ("0.0.0.0", self._port)}  # nosec
```
5A11, line 399: here (both ports can be self._port)

tests/test_packages/test_skills_integration/test_tac.py
```diff
@@ -435,6 +439,14 @@ def test_tac(self):
             diff == []
         ), "Difference between created and fetched project for files={}".format(diff)
```
5A11, line 440: This assert in the newly changed test used to fail when I ran it locally (the agents are not the same because here we have the additional `is_https` in the config set to False. In the original package that is not present.)

tests/test_packages/test_skills_integration/test_tac.py
```diff
@@ -544,6 +556,15 @@ def test_tac(self):
             ), "Difference between created and fetched project for files={}".format(
                 diff
             )
```
5A11, line 558: Also here (see above).

tests/test_packages/test_skills_integration/test_tac.py
DavidMinarsch: The order matters!

tests/common/docker_image.py
DavidMinarsch: TBH there's a benefit from defining it here for clarity. There's no cost to keeping it imo.

tests/common/docker_image.py
DavidMinarsch:
```suggestion
        return {f"{self._port}/tcp": ("0.0.0.0", self._port)}  # nosec
```

tests/common/docker_image.py
DavidMinarsch: that's good like this

solarw (approved):
LGTM

PR #2570, opened 2021-06-30
Add option to return AEA logs in missing string check (AEA-1636)

jrriehl:
## Proposed changes

Currently it's hard to debug tests that fail only in CI because the AEA logs are not visible. This PR adds an option to return the AEA logs so they can be reported on test failures.

## Types of changes

## Checklist

### DIFF
--- a/aea/test_tools/test_cases.py
+++ b/aea/test_tools/test_cases.py
@@ -814,8 +814,7 @@ class BaseAEATestCase(ABC):  # pylint: disable=too-many-public-methods
         timeout: int = DEFAULT_PROCESS_TIMEOUT,
         period: int = 1,
         is_terminating: bool = True,
-        return_logs: bool = False,
-    ) -> Union[List[str], Tuple[List[str], str]]:
+    ) -> List[str]:
         """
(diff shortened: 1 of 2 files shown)


DavidMinarsch (changes requested):
@jrriehl this PR is backwards-incompatible as it changes the API of the test case. I suggest either logging the stdout directly or just making it accessible via a property (i.e. adding to rather than changing the API).

jrriehl:
> @jrriehl this PR is backwards-incompatible as it changes the API of the test case. I suggest either logging the stdout directly or just making it accessible via a property (i.e. adding to rather than changing the API).

@DavidMinarsch I added an optional argument `return_logs` that defaults to `False` in which case the function performs exactly as before. Only when set to `True`, the function returns a different type, so I think this is backwards compatible. Or am I missing something?

jrriehl:
Okay, I see now that `stdout` is already accessible via a property, which is indeed a simpler solution. I will using this approach.

DavidMinarsch (approved):
LGTM

PR #2557, opened 2021-06-15
Try to fix MAM flaky test

marcofavorito:
## Proposed changes

Try to fix MAM flaky test, e.g.:

https://github.com/fetchai/agents-aea/pull/2556/checks?check_run_id=2824321866#step:8:41817
https://github.com/fetchai/agents-aea/pull/2556/checks?check_run_id=2824321671#step:8:62118

## Fixes

n/a

## Types of changes

What types of changes does your code introduce to agents-aea?
_Put an `x` in the boxes that apply_

- [X] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

_Put an `x` in the boxes that apply._

- [ ] I have read the [CONTRIBUTING](../CONTRIBUTING.md) doc
- [ ] I am making a pull request against the `develop` branch (left side). Also you should start your branch off our `develop`.
- [ ] Lint and unit tests pass locally with my changes and CI passes too
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have checked that code coverage does not decrease.
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

## Further comments

n/a

PR #2571, opened 2021-07-01
MAM tests fixes.

solarw:
## Proposed changes

MAM tests fixes.

Takes inspiration and code from #2557.

## Types of changes

## Checklist


packages/fetchai/connections/p2p_libp2p/connection.py
```diff
@@ -173,7 +173,7 @@ def make_acn_envelope_message(envelope: Envelope) -> bytes:
         buf = acn_msg.SerializeToString()
         return buf
 
-    def make_agent_record(self) -> AcnMessage.AgentRecord:  # type: ignore
+    def make_agent_record(self) -> AcnMessage.AgentRecord:  # type: ignore  # pragma: nocover
```
DavidMinarsch, line 176: We should have a test for this. Why does this not get used in integration tests? That's worrying.

solarw: looks like this code is not used anywhere, so removed

DavidMinarsch (approved):
LGTM

Issue #2521, opened 2021-05-20
MAM: error installing an agent dependency if already present


panasevychol:
**Describe the bug**
MAM: error installing an agent dependency if already present.
If the dependency (`aiohttp`) already present installation return code `!=0` and that is recognized as a failure.

**To Reproduce**
1. Install `aiohttp<3.8>=3.7.4`
2. Try to fetch an agent `fetchai/simple_seller_aw5:0.2.0`
3. See error

**Expected behavior**
Agent is succesfully added.

**Screenshots**
From AEA Manager:
![image](https://user-images.githubusercontent.com/6476789/119001376-96215f80-b994-11eb-97d7-bda76fb6a044.png)

**Desktop (please complete the following information):**
 - OS: any
 - AEA Version 1.0.1

panasevychol:
Possible solutions:
- Use flag to reinstall the dependency (for the case if version doesn't match)
- Recognize the case if dependency is present

DavidMinarsch:
A dependency mismatch across agent projects cannot be engineered away. What we can do is provide more helpful error message. We could say: dependency x which you are trying to install with permitted versions y cannot be installed because it is already present with version z from project zz.

marcofavorito:
@panasevychol I can't reproduce the issue. Can you help me?

I am trying to reproduce it on a Ubuntu Docker container.

```
docker run -it ubuntu:20.04 /bin/bash
```

Then, copy-paste the following commands:
```
apt-get update
apt-get install python3-dev python3-pip
pip install "aiohttp<3.8>=3.7.4"
pip install "aea[all]"
aea fetch fetchai/simple_seller_aw5:0.2.0
cd simple_seller_aw5
aea install
```

marcofavorito:
I have also tried to add the project in a MAM. 
First, install `aiohttp`:

```
pip install "aiohttp<3.8>=3.7.4"
```

Then, run the following code in a Python console (with the same interpreter):
```python


import os
from pathlib import Path
from aea.manager import MultiAgentManager
from aea.configurations.base import PublicId

WORKING_DIR = "mam"

manager = MultiAgentManager(WORKING_DIR)
manager.start_manager()
public_id = PublicId.from_str("fetchai/simple_seller_aw5:0.2.0")
manager.add_project(public_id)
```

I couldn't get the error.

PR #2575, opened 2021-07-07
[AEA-1620] Fix/better error details on pip install

solarw:
## Proposed changes

better pip install error details. includes stderr from pip

## Fixes

https://github.com/fetchai/agents-aea/issues/2521

## Types of changes

## Checklist


aea/configurations/data_types.py
```diff
@@ -761,7 +761,7 @@ def to_json(self) -> Dict[str, Dict[str, str]]:
             result["ref"] = cast(str, self.ref)
         return {self.name: result}
 
-    def get_pip_install_args(self) -> List[str]:
+    def get_call_pip_args(self) -> List[str]:
```
DavidMinarsch, line 764: This breaks the API, not backwards compatible, please keep old name.

aea/helpers/install_dependency.py
```diff
@@ -20,54 +20,56 @@
...
+        pip_args = dependency.get_call_pip_args()
+        logger.debug("Calling 'pip install {}'".format(" ".join(pip_args)))
+        call_pip(["install", *pip_args], timeout=install_timeout, retry=True)
     except Exception as e:
         raise AEAException(
             f"An error occurred while installing {dependency_name}, {dependency}: {e}"
         )
 
 
-def run_install_subprocess(
```
DavidMinarsch, line 54: Removing the method means we break backwards compatibility. Please keep method and then just call the new one?

DavidMinarsch (changes requested):

aea/configurations/data_types.py
DavidMinarsch: @solarw

DavidMinarsch (approved):
LGTM

Issue #2460, opened 2021-04-20
MAM support for subprocesses


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
Currently, MAM only allows for modes `async` and `threaded`.

**Describe the solution you'd like**
Add support for subprocesses too.

PR #2596, opened 2021-08-17
mam support multiprocess backend

solarw:
## Proposed changes

mam support multiprocess backend

## Fixes

https://github.com/fetchai/agents-aea/issues/2460

## Types of changes

## Checklist


aea/manager/manager.py
```diff
@@ -85,13 +87,15 @@ def _build_error_message(self) -> str:
 class AgentRunAsyncTask:
     """Async task wrapper for agent."""
 
-    def __init__(self, agent: AEA, loop: asyncio.AbstractEventLoop) -> None:
-        """Init task with agent and loop."""
+    def __init__(
+        self, agent_alias: AgentAlias, loop: asyncio.AbstractEventLoop
```
DavidMinarsch, line 91: This breaks backwards compatibility.

aea/manager/manager.py
```diff
@@ -175,10 +181,102 @@ def stop(self,) -> None:
             self._thread.join()
 
 
+class AgentRunProcessTask(AgentRunAsyncTask):
```
DavidMinarsch, line 184: You're not using anything of `AgentRunAsyncTask` afaikt. So why not have a separate class for the process case and keep backwards compatibility above i.e. revert changes above.

aea/manager/manager.py
```diff
@@ -700,15 +798,18 @@ def start_agent(self, agent_name: str) -> "MultiAgentManager":
...
-            task = AgentRunAsyncTask(aea, self._loop)
+            task = AgentRunAsyncTask(agent_alias, self._loop)
         elif self._mode == "threaded":
-            task = AgentRunThreadTask(aea, self._loop)
+            task = AgentRunThreadTask(agent_alias, self._loop)
+        elif self._mode == "multiprocess":
+            task = AgentRunProcessTask(agent_alias, self._loop)
 
         task.start()
+        time.sleep(1)
```
DavidMinarsch, line 809: why is this needed?

aea/manager/manager.py
```diff
@@ -175,10 +181,102 @@ def stop(self,) -> None:
...
+                t.join(10)
+
+        result_queue.put(r)
+
+    def stop(self) -> None:
+        """Stop the task."""
+        if not self.process:
+            raise ValueError("Task not started!")
+        self._stop_event.set()
+        self.process.join(100)
```
DavidMinarsch, line 266: make join arg a kwarg to stop?

DavidMinarsch (changes requested):

aea/manager/manager.py
solarw: done

aea/manager/manager.py
solarw: updated. added common base class for agent run tasks

aea/manager/manager.py
solarw: removed, was added during investigating some issues

aea/manager/manager.py
solarw: added as a property of the task class

PR #2609, opened 2021-08-31
Update oracle contracts for cosmwasm v0.14

jrriehl:
## Proposed changes

Update oracle and oracle client contracts for cosmwasm 14.

## Types of changes

## Checklist


packages/fetchai/skills/simple_oracle/handlers.py
```diff
@@ -175,7 +175,7 @@ def _request_init_transaction(self, ledger_id: str, tx_receipt: JSONLike) -> Non
                         "label": "OracleContract",
                         "init_msg": {"fee": str(strategy.initial_fee_deploy)},
                         "gas": strategy.default_gas_deploy,
-                        "amount": 0,
+                        "amount": 1,
```
5A11, line 178: Why this change?

jrriehl: This was the easiest way to make sure some funds (along with the denomination) were sent to the contract. Otherwise, we would have needed to send the denomination as an extra string parameter. So this is actually what fixed the test!

tests/conftest.py
```diff
@@ -210,7 +210,7 @@
     "b6ef49c3078f300efe2d4480e179362bd39f20cbb2087e970c8f345473661aa5"
 )
 FUNDED_FETCHAI_PRIVATE_KEY_1 = (
-    "f848e125edb124d7752338fdd15825cd031b8c7f38627ec50ccccb7d75ecffdb"
+    "bbaef7511f275dc15f47436d14d6d3c92d4d01befea073d23d0c2750a46f6cb3"
```
5A11, line 213: So `FUNDED_FETCHAI_PRIVATE_KEY_1` and `FUNDED_FETCHAI_ADDRESS_ONE` now correspond with the validator account (corresponding with DEFAULT_FETCH_MNEMONIC) spawned in LocalFetchDockerImage right?

5A11: Don't we need to update `FUNDED_FETCHAI_PRIVATE_KEY_2` and `FUNDED_FETCHAI_ADDRESS_TWO` below?

tests/test_packages/test_skills_integration/test_simple_oracle.py
```diff
@@ -156,6 +156,10 @@ def test_oracle(
             f"{DEFAULT_FETCH_LEDGER_ADDR}:{DEFAULT_FETCH_LEDGER_REST_PORT}",
         )
 
+        # use alternate port for prometheus connection
+        setting_path = "vendor.fetchai.connections.prometheus.config.port"
+        self.set_config(setting_path, 9091, type_="int")
```
5A11, line 161: What was the port used before for prometheus connection?

tests/conftest.py
jrriehl: Yes, this key now corresponds to the validator on the local node running in docker. We send funds from this account to the other address, so no need to change the other key / address.

tests/test_packages/test_skills_integration/test_simple_oracle.py
jrriehl: It was `9090` which is exposed in the new 'fetchd` container, so this was just to remove that conflict.

MissingNO57 (approved):
LGTM

5A11 (approved):
LGTM

PR #2613, opened 2021-09-06
Stargate

5A11:
This PR adds Stargate support for fetchai ledger.

## Types of changes

## Checklist


panasevychol (approved):

packages/fetchai/agents/random_beacon_feed/aea-config.yaml
```diff
@@ -35,6 +35,6 @@ type: connection
 config:
   ledger_apis:
     fetchai:
-      address: https://rest-beaconworld.fetch.ai:443
```
jrriehl, line 38: I'll check on this, but I believe that Stargateworld does not have a random beacon.

jrriehl (approved):
Looks good

DavidMinarsch (changes requested):
The idea was that the switch to stargate finally resolves the test failures. However, I observe quite a few and it is harder to see now where they come from. I suggest @solarw has a look to address them so that we have a green test run before we merge.

5A11:
> The idea was that the switch to stargate finally resolves the test failures. However, I observe quite a few and it is harder to see now where they come from. I suggest @solarw has a look to address them so that we have a green test run before we merge.

We have addressed all Stargate related failures. There is none in this PR. 
The failures are either:
a) public DHT related: waiting for prod to be synced with latest ACN changes.
b) multi aea manager related: @solarw is looking into those.
c) odd failures here and there: also @solarw is looking into those.

PR #2623, opened 2021-09-10
WIP CW TAC

5A11:
This is a draft WIP PR that adds support for CW for Tac


aea/aea.py
```diff
@@ -220,6 +220,7 @@ def __init__(
             search_service_address,
             decision_maker_handler.self_address,
             data_dir,
+            public_keys=wallet.public_keys,
```
DavidMinarsch, line 223: It might be preferable to extend the identity object instead to contain the public keys. This would be consistent with its intended usage. You'll have to extend the identity object and also the constructor here: https://github.com/fetchai/agents-aea/blob/main/aea/aea_builder.py#L1221

aea/context/base.py
```diff
@@ -166,6 +170,11 @@ def addresses(self) -> Dict[str, Address]:
         """Get addresses."""
         return self.identity.addresses
 
+    @property
+    def public_keys(self) -> Optional[Dict[str, str]]:
+        """Get addresses."""
+        return self._public_keys
```
DavidMinarsch, line 176: If changed as proposed above this becomes `self.identity.public_keys`

aea/helpers/serializers.py
```diff
@@ -16,13 +16,12 @@
...
 # ------------------------------------------------------------------------------
-
 """This module contains Serializers that can be used for custom types."""
 
 import copy
-from typing import Any, Dict
+from typing import Any, Dict, Tuple
 
-from google.protobuf.struct_pb2 import Struct
+from google.protobuf.struct_pb2 import ListValue, Struct
```
DavidMinarsch, line 24: Changes like these in a big PR are not ideal. We should be careful not to break anything. Can you submit these changes as a standalone PR please?

aea/skills/base.py
```diff
@@ -173,6 +173,11 @@ def agent_address(self) -> str:
         """Get address."""
         return self._get_agent_context().address
 
+    @property
+    def public_keys(self) -> Optional[Dict[str, str]]:
```
DavidMinarsch, line 177: Remove the optional please, it is wrong

fund.py
```diff
@@ -0,0 +1,53 @@
+import time
```
DavidMinarsch, line 1: This file is in the wrong place

packages/fetchai/agents/tac_controller_contract/aea-config.yaml
```diff
@@ -25,7 +25,7 @@ skills:
 - fetchai/tac_control:0.23.0
 - fetchai/tac_control_contract:0.25.0
 default_connection: fetchai/p2p_libp2p:0.24.0
-default_ledger: ethereum
+default_ledger: fetchai
```
DavidMinarsch, line 28: Can we please keep the old one and add a new one?

packages/fetchai/protocols/cosm_trade/protocol.yaml
```diff
@@ -0,0 +1,21 @@
+name: cosm_trade
+author: fetchai
+version: 0.1.0
+protocol_specification_id: fetchai/cosm_trade:1.0.0
+type: protocol
+description: A protocol for preparing an atomic swap bilateral transaction for cosmos-based
```
DavidMinarsch, line 6: Can you explain its usage? Why is it needed?

packages/fetchai/skills/tac_control_contract/handlers.py
```diff
@@ -397,17 +453,29 @@ def _handle_transaction_receipt(
                 contract_api_dialogue.callable
                 == ContractApiDialogue.Callable.GET_DEPLOY_TRANSACTION
             ):
-                contract_address = cast(
-                    Optional[str],
-                    ledger_api_msg.transaction_receipt.receipt.get(
-                        "contractAddress", None
-                    ),
+                transaction_label = contract_api_dialogue.terms.kwargs.get(
+                    "label", "None"
```
DavidMinarsch, line 457:
```suggestion
                    "label", None
```

packages/fetchai/skills/tac_negotiation/handlers.py
```diff
@@ -267,32 +274,66 @@ def _on_accept(self, accept: FipaMessage, fipa_dialogue: FipaDialogue) -> None:
...
+                        "requesting batch transaction hash, sending {} to {}, message={}".format(
+                            contract_api_msg.performative,
+                            strategy.contract_id,
+                            contract_api_msg,
+                        )
+                    )
+                    self.context.outbox.put_message(message=contract_api_msg)
+                elif strategy.ledger_id == FetchAIApi.identifier:
+                    public_keys = self.context.public_keys
+                    if public_keys is None:
```
DavidMinarsch, line 311: Why would it ever be None? That makes no sense. An agent by design needs to have a wallet otherwise it won't run. See also above, the public_keys dictionary is not optional.

packages/fetchai/skills/tac_negotiation/handlers.py
```diff
@@ -267,32 +274,66 @@ def _on_accept(self, accept: FipaMessage, fipa_dialogue: FipaDialogue) -> None:
...
+                        self.context.logger.info(
+                            "The AEA does not have any public_keys!"
+                        )
+                        return
+                    public_key = public_keys.get(strategy.ledger_id, None)
+                    if public_key is None:
+                        self.context.logger.info(
+                            "Could not retrieve my own public key."
```
DavidMinarsch, line 319:
```suggestion
                            f"Agent has no public key for {strategy.ledger_id}."
```

tac_contract_local_ethereum.sh
```diff
@@ -0,0 +1,89 @@
+#!/bin/bash -e
+min=$1
```
DavidMinarsch, line 2: Move to tac folder in examples

tac_contract_local_fetchai.sh
```diff
@@ -0,0 +1,89 @@
+#!/bin/bash -e
+min=$1
+# 2
```
DavidMinarsch, line 3: see above

tac_local.sh
```diff
@@ -0,0 +1,89 @@
+#!/bin/bash -e
+min=$1
+# 2
+participants=$2
```
DavidMinarsch, line 4: see above

tests/test_helpers/test_serializers.py
```diff
@@ -17,38 +17,28 @@
...
 
 def test_encode_decode_i():
     """Test encode decode logic."""
     case = {
-        "key1": True,
-        "key2": 0.12,
-        "key3": 100,
-        "key4": "some string",
-        "key5": b"some bytes string",
-        "key6": Struct(),
```
DavidMinarsch, line 33: The struct case is no longer tested?

packages/fetchai/agents/tac_controller_contract/aea-config.yaml
5A11: I do not think it is justified to introduce another identical agent with only configuration changes. 
We can easily switch between Ethereum and Fetchai Tac by changing two config fields.

aea/helpers/serializers.py
5A11: There was a problem with `DictProtobufStructSerializer` (`_patch_dict` didn't cover nested dictionaries and `list` values) which the new Tac changes exposed. 

We had to make these changes to make Tac work and it was difficult to do this fix in isolation. This bug fix only involves this file and its associated test `tests/test_helpers/test_serializers.py` below. 

If it was a more involved bug fix, I agree that we should have done it separately to reduce confusion.

aea/identity/base.py
```diff
@@ -36,40 +37,79 @@ class Identity:
...
             address = addresses[default_address_key]
+            public_key = public_keys[default_address_key]
         self._address = address
+        self._public_key = public_key
+
         if addresses is None:
+            if public_keys is not None or public_key is None:
+                raise ValueError(
+                    "If you provide a single address, you must provide its corresponding single public key and not a dictionary of public keys."
+                )
```
DavidMinarsch, line 108: Have a separate error message for each case imo, easier to debug

DavidMinarsch:
A lot of new functionality has been added without any tests

aea/helpers/serializers.py
DavidMinarsch: Can we add a few more unit tests to make sure this is now well covered?

packages/fetchai/agents/tac_controller_contract/aea-config.yaml
DavidMinarsch: OK, please add the two test classes then

solarw:
https://github.com/fetchai/agents-aea/pull/2641

PR #2761, opened 2022-08-02
Update to reply-based oracle contracts

jrriehl:
## Proposed changes

Oracle contracts were recently updated so that when the client contract queries the oracle contract, it should handle the submessage reply rather than handing over control to the oracle contract to set the client's oracle value. This was made possible in newer versions of Cosmwasm. Since this only requires an update to the interaction between the contracts, only the contract binaries need to be updated.

## Types of changes

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/fetchai/agents-aea/blob/main/CONTRIBUTING.md) doc


solarw (approved):
lgtm

PR #2868, opened 2023-02-03
Update Dockerfile to match new docs

Archento:
## Proposed changes

The dockerfile and container are outdated and do not align with the current documentation.

## Issues

Links to any issues resolved.

## Types of changes

- [ ] Bugfix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to stop working as expected)
- [x] Something else (e.g. test, package, script, example, deployment, infrastructure, ...)

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/fetchai/agents-aea/blob/main/CONTRIBUTING.md) document.
- [x] I have based my branch, and I am making a pull request against, the `develop` branch.
- [x] Lint and unit tests pass locally with my changes.

### If applicable

- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] I have added/updated the documentations.
- [ ] Dependent changes have been merged and published in downstream modules.

## Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did, what alternatives you considered, etc.


5A11 (approved):

user-image/Dockerfile
```diff
@@ -1,8 +1,9 @@
-FROM ubuntu:18.04
+FROM ubuntu:22.10
+SHELL [ "/bin/bash", "-c" ]
 
 RUN apt-get update && apt-get upgrade -y
 
-RUN apt-get install sudo -y 
+RUN apt-get install sudo -y
```
jrriehl, line 6: I realize you didn't add this, but it's not immediately clear to me why `sudo` is needed since the docker container is typically entered as `root`. Maybe some of the guides use `sudo` somewhere? Anyway, I guess we can leave it in, just a very minor remark.

jrriehl (approved):
Looks good!

jrriehl:
Just one more comment: it's probably better to give the PR a somewhat more descriptive title. :slightly_smiling_face:

user-image/Dockerfile
Archento: You are right, typically the `root` user would be sufficient but I assume that some of the following instructions needed a user with sudo privileges so I haven't changed or removed it.
Currently it doesn't hurt to leave it in but if we need to add something later (like IPFS) I'll test if we can omit this.

PR #2875, opened 2023-02-20
Add n<-->n capability for carpark demo

Archento:
## Proposed changes

Add the capability of having multiple carpark data sellers and one carpark data buyer.

## Issues

Links to any issues resolved.

## Types of changes

- [ ] Bugfix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to stop working as expected)
- [ ] Something else (e.g. test, package, script, example, deployment, infrastructure, ...)

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/fetchai/agents-aea/blob/main/CONTRIBUTING.md) document.
- [x] I have based my branch, and I am making a pull request against, the `develop` branch.
- [x] Lint and unit tests pass locally with my changes.

### If applicable

- [ ] I have added tests that prove my fix is effective or that my feature works.
- [x] I have added/updated the documentations.
- [ ] Dependent changes have been merged and published in downstream modules.

## Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did, what alternatives you considered, etc.


packages/fetchai/skills/carpark_client/handlers.py
```diff
@@ -19,15 +19,145 @@
...
+            {
+                "sender": fipa_msg.sender,
+                "message": fipa_msg,
+                "decision": FipaMessage.Performative.DECLINE
+                if not strategy.is_acceptable_proposal(fipa_msg.proposal)
+                or not strategy.is_affordable_proposal(fipa_msg.proposal)
+                else None,
+            }
+        )
+        # self.context.logger.info(
```
jrriehl: `logger.debug(...)` maybe? Or delete.

packages/fetchai/skills/carpark_client/handlers.py
```diff
@@ -19,15 +19,145 @@
...
+        #     f"fipa_message_id={fipa_msg.message_id} with target={fipa_msg.target} added to carpark_data."
+        # )
+
+        # makes more sense to only wait for a certain time
```
jrriehl: Agree that there should probably be some timeout after which all received proposals are compared. Could also wait for some threshold (less than or equal to the number of agents) to respond. But not sure if that's better than a timeout.

packages/fetchai/skills/carpark_client/handlers.py
```diff
@@ -19,15 +19,145 @@
...
+                    if cheapest
+                    else FipaMessage.Performative.DECLINE
+                )
+            self._handle_propose_send(strategy)
+
+    def _handle_propose_send(self, strategy: GenericStrategy) -> None:
+        """
+        The actual sending of the accept/decline messages.
+
+        :param strategy: _description_
```
jrriehl: I guess `_description_` was meant to be replaced.

packages/fetchai/skills/carpark_client/strategy.py
```diff
@@ -19,7 +19,25 @@
...
+
+        e.g. agents = [
+                {'sender': 'agent_1', 'message': {'price': 10, ...}},
+                {'sender': 'agent_2', 'message': {'price': 5, ...}}
+            ]
+        """
+        return sorted(
+            agents,
+            key=lambda agent: agent["message"].proposal.values["price"],
+            reverse=False,
```
jrriehl, line 57: Isn't `reverse=False` the default?

jrriehl (commented):
Looks great! There are a few tests failing that probably should pass: `isort`, docs test, and hashes check, in particular. I think you just need to:
- run `make lint`
- run this script https://github.com/fetchai/agents-aea/blob/main/scripts/generate_ipfs_hashes.py
- update this file with your docs changes: https://github.com/fetchai/agents-aea/blob/main/tests/test_docs/test_bash_yaml/md_files/bash-car-park-skills.md.

packages/fetchai/skills/carpark_client/handlers.py
Archento: It's going to be thrown out eventually :)
I discovered a bug where the message_id of the fipa_messages is not correctly set (was always -1 in my case) and @5A11 wanted to have a look himself so I left it there for the time being.

packages/fetchai/skills/carpark_client/handlers.py
Archento: I haven't found any solution for non-blocking timeouts in skill handlers so far but if you've got any pointers on how to solve this I'd appreciate it :)

packages/fetchai/skills/carpark_client/handlers.py
Archento: fixed already, I simply forgot to add something

packages/fetchai/skills/carpark_client/strategy.py
Archento: Yes, but I'm planning to make the function reusable in the sense, that you can flip the order later to receive the proposal with e.g. the highest reward.
Also, I've encountered a case where a lib that we used in the project received an update and changed the necessary arguments without backward compatibility - so I tend to provide all arguments that I actually use

packages/fetchai/skills/carpark_client/handlers.py
jrriehl: Right, so I think this logic would have to be moved over to a new `TickerBehaviour`. Probably you would also need to store a timestamp. Should be doable, just not sure whether it's worth the effort for the demo.
@5A11 any thoughts on this?

Looneyestwolf (approved):

packages/fetchai/skills/carpark_client/handlers.py
Archento: Having a new tickerbehaviour did the trick :) thanks guys

packages/fetchai/skills/carpark_client/behaviours.py
```diff
@@ -19,11 +19,99 @@
...
+
+class ProposalCheckBehaviour(TickerBehaviour):
+    """
+    Controls the timer to check if proposals have been received.
+
+    :param TickerBehaviour: generic ticker behaviour to be inherited from
+    """
+
+    def __init__(self, **kwargs: Any) -> None:
+        """Initialize the search behaviour."""
```
jrriehl: Docstring needs updating.

packages/fetchai/skills/carpark_client/skill.yaml
```diff
@@ -25,6 +25,9 @@ protocols:
 skills:
 - fetchai/generic_buyer:0.27.6
 behaviours:
+  check:
+    args: {}
+    class_name: ProposalCheckBehaviour
```
jrriehl, line 30: It might make sense to expose the `ticker_interval` argument here to make that easier to configure.

jrriehl (approved):
LGTM

packages/fetchai/skills/carpark_client/skill.yaml
Archento: I think it's ok to have it exposed in the strategy as the `ticker_interval` itself is left default (1 sec) but the `proposal_check_interval` is more related to the overall behaviour of the agent.
I renamed it to `proposal_check_timeout` to make it more clear what I meant.

packages/fetchai/skills/carpark_client/behaviours.py
```diff
@@ -19,11 +19,99 @@
...
+        """Act according to the behaviour."""
+        strategy = cast(Strategy, self.context.strategy)
+        if not strategy.waiting_for_proposals:
+            return
+
+        # check if every agent has sent at least one proposal
+        if set(strategy.sent_proposals).issubset(
+            list(map(lambda x: x["sender"], strategy.received_proposals))
+        ):
+            self.context.logger.info("received all proposals, making decision...")
```
5A11, line 64: "received {number_of_receive_proposals} proposals, making decision..."

packages/fetchai/skills/carpark_client/behaviours.py
```diff
@@ -19,11 +19,99 @@
...
+        elif self.counter > strategy.proposal_check_timeout - 1:
+            self.context.logger.info(
+                "waiting for proposals timed out, making decision..."
```
5A11: "... making decision with {number_of_received_proposals} proposals..."

packages/fetchai/skills/carpark_client/behaviours.py
```diff
@@ -19,11 +19,99 @@
...
+        self._handle_all_proposals_received(strategy)
+
+    def teardown(self) -> None:
+        """Teardown the behaviour."""
+
+    def _handle_all_proposals_received(self, strategy: Strategy) -> None:
+        undecided_proposals = list(
+            filter(lambda x: x["decision"] is None, strategy.received_proposals)
+        )
+        """Prepare the accept/decline messages"""
```
5A11: Shouldn't this be before the function?

packages/fetchai/skills/carpark_client/behaviours.py
```diff
@@ -19,11 +19,99 @@
...
+        strategy = cast(Strategy, self.context.strategy)
+        if not strategy.waiting_for_proposals:
+            return
+
+        # check if every agent has sent at least one proposal
+        if set(strategy.sent_proposals).issubset(
+            list(map(lambda x: x["sender"], strategy.received_proposals))
+        ):
+            self.context.logger.info("received all proposals, making decision...")
+        elif self.counter > strategy.proposal_check_timeout - 1:
```
5A11: Why not just start counter at 1 and change this to counter >= proposal_check_timeout for better readability?

Archento: I had it like that before but changed it again to 0 but 1 is ok with me.

packages/fetchai/skills/carpark_client/behaviours.py
Archento: absolutely...

packages/fetchai/skills/carpark_client/handlers.py
```diff
@@ -27,7 +36,74 @@
...
+        query = strategy.get_service_query()
+        fipa_dialogues = cast(FipaDialogues, self.context.fipa_dialogues)
+        counterparties = strategy.get_acceptable_counterparties(oef_search_msg.agents)
+        for counterparty in counterparties:
+            cfp_msg, _ = fipa_dialogues.create(
+                counterparty=counterparty,
+                performative=FipaMessage.Performative.CFP,
+                query=query,
+            )
+            strategy.sent_proposals.append(counterparty)
```
5A11: `sent_proposals` should be renamed to `sent_cfps`

packages/fetchai/skills/carpark_client/behaviours.py
5A11: This just checks you have received _some_ proposals. You shouldn't necessarily trigger decision making here. Only when the timeout is reached.

packages/fetchai/skills/carpark_client/behaviours.py
```diff
@@ -19,11 +19,99 @@
...
+        """Set up the behaviour."""
+
+    def act(self) -> None:
+        """Act according to the behaviour."""
+        strategy = cast(Strategy, self.context.strategy)
+        if not strategy.waiting_for_proposals:
+            return
+
+        # check if every agent has sent at least one proposal
+        if set(strategy.sent_proposals).issubset(
```
5A11, line 61: This just checks you have received some proposals. You shouldn't necessarily trigger decision making here. Only when the timeout is reached.

packages/fetchai/skills/carpark_client/behaviours.py
```diff
@@ -19,11 +19,99 @@
...
+
+    def _handle_all_proposals_received(self, strategy: Strategy) -> None:
+        undecided_proposals = list(
+            filter(lambda x: x["decision"] is None, strategy.received_proposals)
+        )
+        """Prepare the accept/decline messages"""
+        if undecided_proposals:
+            cheapest_proposal = strategy.get_cheapest_proposal(undecided_proposals)
+            for carpark in undecided_proposals:
+                cheapest = cheapest_proposal["sender"] == carpark["sender"]
```
5A11: Some of the variables could have better names, such as `carpark["sender"]`, which could probably be something like `proposal["sender"]`.

packages/fetchai/skills/carpark_client/strategy.py
```diff
@@ -19,7 +19,90 @@
...
+    """This class defines a strategy for the agent."""
+
+    def __init__(self, **kwargs: Any) -> None:
+        """Initialize the strategy of the agent."""
+        self.proposal_check_timeout = kwargs.pop(
+            "proposal_check_timeout", DEFAULT_PROPOSAL_CHECK_TIMEOUT
+        )  # type: int
+        super().__init__(**kwargs)
+        self._received_proposals = cast(List, [])
+        self._sent_proposals = cast(List, [])
```
5A11: See above re. changing the name.

5A11:
Couple of comments, mostly on the code format. Also: 

- Some test failures (just the bash_yaml tests are relevant here).
- Noticed an issue for pulling soef on the CI. Could it be because the source branch is from your fork (it really shouldn't)? I think regardless, it is better to open PRs directly.

packages/fetchai/skills/carpark_client/behaviours.py
Archento: You are right, I was planning to rename the variables to be more generic and for them to ultimately end up in the generic buyer logic. Those fell under the radar and they will be changed.

